PR #88, opened 2025-04-07
`Athena`: Added mock tests for the basic approach of the module_text_llm

alekspetrov9e:
Add tests for text LLM module basic approach

This PR adds e tests for the text LLM module's basic approach functionality. The tests verify the feedback generation process for student submissions using extensive mocking to isolate the functionality being tested.

Key features of the test suite:
- Tests basic feedback generation with simple submissions
- Tests handling of empty submissions
- Tests processing of long input submissions
- Uses mocks for OpenAI model configuration, NLTK sentence tokenization, and module configuration
- Verifies proper feedback object creation and association with exercises/submissions

The tests ensure that the basic approach functionality works correctly across different scenarios while maintaining proper isolation from external dependencies.


athena/modules/text/module_text_llm/tests/utils/mock_llm.py
```diff
@@ -0,0 +1,96 @@
...
+    line_start: Optional[int] = None
+    line_end: Optional[int] = None
+    credits: float = 0.0
+    grading_instruction_id: Optional[int] = None
+
+class AssessmentModel(BaseModel):
+    """Mock assessment model containing feedback items."""
+    feedbacks: List[MockFeedbackModel] = Field(default_factory=lambda: [MockFeedbackModel()])
+
+class StructuredMockLanguageModel(BaseLanguageModel):
```
ahmetsenturk, line 24: These could also start with Mock, `MockAssessmentModel`, `MockStructuredLanguageModel`

athena/modules/text/module_text_llm/tests/utils/mock_llm.py
```diff
@@ -0,0 +1,96 @@
+"""Mock language model implementations for testing the text LLM module."""
+
```
ahmetsenturk, line 2: Not sure if this is pythonic - to explain what a file is there for. I think the name is already clear enough 😊

ahmetsenturk (commented):
Looks pretty nice 👍 Just added two nitpicks - other than that lgtm

maximiliansoelch:
> Currently, you're patching directly inside the test file, which is functional but clutters the test code and won’t scale well as the suite grows. I’d recommend moving these patches into fixtures in a `conftest.py` file. Since we’re likely to reuse much of this mocked functionality across different modules and tests, we should aim for a more reusable and modular setup.
> 
> There are two approaches I’d suggest considering:
> 
> 1. **Centralized `tests/` folder**: We could move all tests under a single root-level `tests/` directory. Inside that, we can organize shared fixtures, mocks, and test utilities in their own folders (e.g., `tests/utils/`, `tests/fixtures/`) and then have per-module test folders (like `tests/module_text_llm/`, etc.).
> 2. **Shared `test_helpers/` module**: Alternatively, we could create a shared `test_helpers/` package to store common fixtures, mocks, and utilities. Each module would keep its own `tests/` folder with local tests, but import shared components from `test_helpers/`.
> 
> In both setups, modules would import shared fixtures into their local `conftest.py` (or directly from the shared folder where appropriate) and can still keep their own specialised fixtures or mocks right next to their tests.
> 
> I personally lean toward option 1, but I think option 2 is equally valid depending on how modular we want to keep things.
> 
> What do you think @maximiliansoelch?

I am fine with both options. My only requirement is that the test files are clearly structured and identifiable to which module they belong, allowing for an easy modification/deletion of, for example, a deprecated module.

alekspetrov9e:
Thank you all for the reviews! I used the first approach that @LeonWehrhahn proposed and created a central tests folder and tests can be grouped by modules. I also resolved the two nitpicks from @ahmetsenturk :)

athena/tests/module_text_llm/test_basic_approach_mock.py
```diff
@@ -0,0 +1,165 @@
...
+
+@pytest.fixture
+def mock_config():
+    """Create a mock configuration for testing."""
+    return MockApproachConfig(
+        max_input_tokens=5000,
+        model=MockModelConfig(),
+        type="basic"
+    )
+
```
ahmetsenturk, line 77: the same snippet appears in `tests/module_text_llm/conftest.py`

athena/tests/module_text_llm/test_basic_approach_mock.py
```diff
@@ -0,0 +1,165 @@
...
+    model_name: str = "mock_model"
+    get_model: Callable[[], Any] = Field(default_factory=lambda: Mock())
+
+    class Config:
+        arbitrary_types_allowed = True
+
+mock_openai = Mock()
+mock_openai.OpenAIModelConfig = MockOpenAIModelConfig
+mock_openai.available_models = {'mock_model': Mock()}
+sys.modules['llm_core.models.openai'] = mock_openai
```
ahmetsenturk, line 27: You already have a dedicated file under utils that does the same thing, `tests/utils/mock_openai.py` 😊

athena/tests/module_text_llm/test_basic_approach_mock.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def __getitem__(self, key):
+        return self._data[key]
+
+    def read(self, *args, **kwargs):
+        pass
+
+# Start the patch context before other imports
+patcher = patch("athena.module_config.configparser.ConfigParser", return_value=MockConfigParser())
+patcher.start()
```
ahmetsenturk, line 43: this one as well - duplicate

athena/tests/module_text_llm/conftest.py
```diff
@@ -0,0 +1,31 @@
+# Import OpenAI mocks first to ensure they're in place before any other imports
+from tests.utils.mock_openai import mock_openai, mock_openai_client
```
ahmetsenturk, line 2: What do you mean here?

ahmetsenturk (changes requested):
You should make use of the utils you have implemented - looks like you are defining the mock models you have already separated from the test suites

athena/tests/module_text_llm/conftest.py
alekspetrov9e: this must happen before other code tries to import llm_core.models.openai

ahmetsenturk: which part, then why don't we import it there? because this is not being used in this file - might be confusing for people looking at this later

athena/tests/module_text_llm/test_basic_approach_mock.py
```diff
@@ -0,0 +1,119 @@
...
+        title="Test Exercise",
+        type=ExerciseType.text,
+        max_points=10,
+        bonus_points=2,
+        grading_instructions="Test grading instructions",
+        problem_statement="Test problem statement",
+        example_solution="Test example solution",
+        grading_criteria=[]
+    )
+
```
maximiliansoelch, line 22: Can you please adapt the PEP8 style to all files? Otherwise, viewing those files in my IDE is quite colorful:
image

athena/pyproject.toml
```diff
@@ -7,6 +7,8 @@ package-mode = true
 
 [tool.poetry.dependencies]
 python = "3.11.*"
+pytest = "^8.3.5"
+pytest-asyncio = "^0.26.0"
```
maximiliansoelch, line 11: These should only be dev dependencies, right?

maximiliansoelch (changes requested):
Minor thing in the test dependency declaration. See inline comment

athena/pyproject.toml
alekspetrov9e: yes

maximiliansoelch (changes requested):
Please also adapt the Github Action for testing to make use of your newly added test.
Additionally, please mark PR review suggestions that were resolved as such.

athena/scripts/test_modules.py
```diff
@@ -19,6 +19,46 @@ def main():
...
+            if result.returncode != 0:
+                print(f"Tests failed for {module}:")
+                print(result.stdout)
+                print(result.stderr)
+                success = False
+            else:
+                print(f"Tests passed for {module}")
+        except Exception as e:
+            print(f"Error running tests for {module}: {str(e)}")
+            success = False
```
maximiliansoelch, line 61: This looks way too complicated. Can we not just simply adapt the `Athena_test.yaml` GH action, and instead of calling this script, run `poetry run pylint`? Then all the existing tests should be run.

Then we also do not need additional pyproject.yaml files that also need to be maintained.

Or is there any reason why this proposed approach is better?

maximiliansoelch (changes requested):

maximiliansoelch (approved):

PR #113, opened 2025-05-04
`Athena`: Implement verification tests for basic approach of module_text_llm

alekspetrov9e:
### Motivation and Context

This PR improves the test coverage for the text feedback generation system by adding test cases that verify the LLM's ability to detect and provide feedback for simple text exercises. It is a stacked PR that extends the mock tests for the same module. Mock tests are executed in the pipeline and they pass. Real tests can be executed locally from module_text_llm dir and with using the venv of module_text_llm.

### Description

Added three test cases that verify feedback generation for different types of programming exercises:
- a recursive factorial function missing a base case and proper decrement
- a string reversal function incorrectly using built-in methods
- a list deduplication function that doesn't preserve order

### Steps for Testing


athena/tests/modules/text/module_text_llm/real/test_basic_approach_real.py
```diff
@@ -0,0 +1,118 @@
...
+    exercise = Exercise(
+        id=1,
+        title="Python Function Exercise",
+        type=ExerciseType.text,
+        max_points=10,
+        bonus_points=2,
+        grading_instructions="",
+        problem_statement="Implement a function that calculates the factorial of a given number n.",
+        example_solution="def factorial(n):\n    if n == 0:\n        return 1\n    return n * factorial(n-1)",
+        grading_criteria=[]
```
maximiliansoelch, line 18: The text module creates feedback for text exercises. It does not make sense and also is confusing if the task for the text exercise is to write code

athena/tests/modules/text/module_text_llm/real/conftest.py
```diff
@@ -0,0 +1,32 @@
...
+def real_config():
+    """Create a real configuration for testing with Azure OpenAI."""
+    return BasicApproachConfig(
+        max_input_tokens=5000,
+        model=OpenAIModelConfig(
+            model_name="azure_openai_gpt-4o",  
+            get_model=lambda: None  # This will be set by the module
+        ),
+        type="basic"
+    )
```
maximiliansoelch, line 32: Is the purpose of this to test against a real LLM instance?

maximiliansoelch (changes requested):
Running the tests locally fails. 
The GH test action only succeeds as it does not execute any tests.
Please update the GH action file accordingly. It would be also great to have the action log out the number of executed tests including passed and failed tests.

PR #463, opened 2026-03-24
`Iris`: Improve autonomous tutor prompt

toukhi:
## Summary

The autonomous tutor was responding to all student posts, including social comments and greetings. This PR adds a classification gate to the system prompt so Iris first decides whether a post warrants a tutoring response before taking any action. The pipeline now handles the `NO_RESPONSE_NEEDED` marker cleanly, signaling Artemis not to post with zero confidence. The prompt was also improved for thread context, conciseness, and clearer pedagogical guidance.

Closes IRIS-274

## Release Notes

* **New Features**
  * Tutoring assistant now intelligently determines when responses are unnecessary for certain student posts
  * Enhanced conversation awareness to prevent redundant or repetitive responses
  * Improved tool usage guidance for better information synthesis

### DIFF
--- a/iris/src/iris/pipeline/autonomous_tutor_pipeline.py
+++ b/iris/src/iris/pipeline/autonomous_tutor_pipeline.py
@@ -301,7 +301,7 @@ class AutonomousTutorPipeline(
                 variant_id="default",
                 name="Default",
                 description="Default autonomous tutor variant using the OpenAI GPT-OSS 120B model.",
-                cloud_agent_model="gpt-oss:120b",
+                cloud_agent_model="gpt-5-mini",
                 local_agent_model="gpt-oss:120b",
             ),
         ]


toukhi:
> Nice work! The prompt rewrite is noticeably tighter and the no-response filtering is a smart addition. Two things:
> 
> 1. **Substring vs strict equality for the marker check.** Right now the code uses `in` (substring match), which is lenient — it'll catch cases where the LLM doesn't output the marker perfectly cleanly, but it also means a response that _mentions_ `NO_RESPONSE_NEEDED` in a longer message would get suppressed. Worth thinking about which failure mode is worse here: false positives (suppressing a valid response) or false negatives (letting a non-response slip through as an actual reply). Up to you which trade-off fits better — just flagging it as something to consider deliberately.
> 2. **Not sure but don't we need a few test cases for the no-response path?** Something like one where `state.result` triggers the no-response branch and one with a normal response that proceeds as usual.

I’m leaning toward keeping the substring match. `NO_RESPONSE_NEEDED` is unlikely to appear organically in a channel message, so a false positive (suppressing a valid response) is the more acceptable failure mode here. A missed suppression would be more disruptive.

PR #472, opened 2026-03-30
`Athena`: Add scoring to module_programming_llm for preliminary feedback

simongundelwein:
Here’s a corrected version of the PR description summary:

## Summary
- add `credits` to non-graded feedback produced by `module_programming_llm` and pass those values through to Athena feedback objects
- include exercise grading instructions in the non-graded programming LLM prompt so feedback scores are grounded in the configured grading criteria
- restrict raw non-graded credits to `0`, `0.5`, or `1`, ignore bonus points for this flow, and scale the final feedback credits across the exercise `max_points`
- update the local-development request flow so repository authorization can fall back to the incoming `Authorization` header when no deployment secret is configured, and fix Azure/OpenAI-compatible model kwargs by remapping `max_tokens` to `max_completion_tokens` without leaving stale keys behind

## Summary by CodeRabbit

* **New Features**
  * Feedback now includes discrete credit values (0, 0.5, 1.0) with normalization and scaling across suggestions.
  * Exercise grading instructions are added to AI feedback prompts.

* **Improvements**
  * Better handling and normalization of model configuration parameters for provider compatibility.
  * Authorization credential forwarding to downstream module requests with safer fallback behavior.


athena/assessment_module_manager/assessment_module_manager/module/request_to_module.py
```diff
@@ -47,7 +55,15 @@ async def request_to_module(module: Module, headers: dict, path: str, lms_url: s
...
-        headers['X-Repository-Authorization-Secret'] = env.DEPLOYMENT_SECRETS.get(lms_url, "")
+        repository_secret = env.DEPLOYMENT_SECRETS.get(lms_url, "")
+        if not repository_secret and not env.PRODUCTION and lms_secret:
+            logger.warning(
+                "DEBUG MODE: No deployment secret configured for LMS URL %s. "
+                "Falling back to the incoming Authorization header for repository access.",
+                lms_url,
+            )
+            repository_secret = lms_secret
+        headers['X-Repository-Authorization-Secret'] = repository_secret
```
maximiliansoelch: I don't see why this relates to the scope of this PR of including scoring into preliminary feedback for programming.
As I understand, this change would simplify local development, but as it also concerns security, I would need to conduct a deeper review, which I currently do not have the time for.
Therefore, I suggest moving this change to a separate PR so I can review it once time allows.

athena/modules/programming/module_programming_llm/module_programming_llm/prompts/generate_non_graded_suggestions_by_file.py
```diff
@@ -23,6 +26,12 @@
 In git diff, lines marked with '-' were removed and with '+' were added by the student.
 
 # The student will be reading your response, use you instead of them
+
+# Credits
+For each feedback item, use exactly one of these `credits` values: `0`, `0.5`, or `1`.
+Use `0` for incorrect or missing work, `0.5` for partially correct work, and `1` for fully correct work.
+Never use negative credits and never use any other credit values.
+Do not invent grading criteria that are not supported by the provided grading instructions.
```
maximiliansoelch, line 34: Does the restriction to only these three possible scores improve the quality of scoring? 🤔
I understand that it should be restricted to positive values, but this restriction seems to me quite limiting regarding the score range of a feedback item.

athena/modules/programming/module_programming_llm/module_programming_llm/generate_non_graded_suggestions_by_file.py
```diff
@@ -65,6 +99,9 @@ async def generate_suggestions_by_file(
     # Feature extraction
     template_repo = exercise.get_template_repository()
     submission_repo = submission.get_repository()
+    grading_instructions = format_grading_instructions(
+        exercise.grading_instructions, exercise.grading_criteria
+    )
```
maximiliansoelch, line 105: This differs quite a bit from the logic used for generating graded feedback.
Ideally, this should be closely aligned so the functionality is the same regardless of the feedback mode.

But this can also be done in a follow-up.

maximiliansoelch (changes requested):

athena/assessment_module_manager/assessment_module_manager/module/request_to_module.py
simongundelwein: I needed this change to make my local deployment possible as it fails otherwise. I can remove it from the pr for now if it is just a problem on my machine

athena/modules/programming/module_programming_llm/module_programming_llm/prompts/generate_non_graded_suggestions_by_file.py
simongundelwein: Because the values get normalized to max-credits, having an arbitrary range of different numbers made the values seem random and confusing. Having only correct or half-correct made the credits have more meaning in my opinion.

maximiliansoelch (approved):
Thanks for implementing my change requests, Code looks good now 👍

Please align the non_graded and graded feedback generation workflows in a follow-up like we discussed today.

PR #523, opened 2026-04-27
`Iris`: Global Search intent routing with SetFit classifier and async webhook

Nayer-kotry:
### Summary
Adds intent-based routing to the global search pipeline. A lightweight on-device classifier decides per query whether to invoke the full LLM pipeline or return source results immediately. Also ships HyDE-based retrieval, an async webhook delivery pattern, and several pipeline performance and robustness improvements.

**Artemis counterpart:** ls1intum/Artemis#12612

### Motivation and Context
The previous lecture search pipeline ran the full HyDE + LLM flow for every query, including bare keyword lookups where an LLM answer adds no value. This wasted latency and tokens. Adding intent classification lets navigation and keyword queries return instantly while keeping AI-generated answers for conceptual questions.

### Description
**Intent classifier**
- SetFit model (paraphrase-multilingual-MiniLM-L12-v2, INT8-quantized ONNX) classifies queries as `trigger_ai` or `skip_ai` in ~2ms on CPU
- ONNX model (~113 MB) is downloaded from HuggingFace Hub (`NayerKotry/iris-intent-classifier`) during Docker build; tokenizer files are fetched at runtime and cached locally
- Pre-warmed on startup in a background thread so the first real request is not delayed

**Smart routing**
- `skip_ai` queries bypass HyDE and LLM entirely; sources are returned immediately with `answer: null`
- `trigger_ai` queries go through the full HyDE → embedding → retrieval → answer pipeline

**Async webhook pattern**
- `POST /api/v1/pipelines/global-search/run` returns 202 immediately
- Two status callbacks are sent to Artemis at `pipelines/global-search/runs/{runId}/status`: one when thinking starts, one when done

**Rename**
- All `LectureSearch*` pipeline classes, DTOs, and prompts renamed to `GlobalSearch*`

**Pipeline performance & robustness**
- Both HyDE and answer steps use `reasoning_effort=none` and pass it via `CompletionArguments`. This parameter is forwarded to the API only when `supports_reasoning_effort: true` is set on the model in `llm_config`. `reasoning_effort` is not available on models prior to GPT-5, and an effort level of `none` was introduced from GPT-5.2 and higher, which disables extended reasoning and minimises latency. Both roles (`hyde` and `answer`) should be mapped to `gpt-5.4-nano` in `llm_configuration.global_search_pipeline.default`.
- HyDE uses `max_tokens=150` (plain text output, no JSON risk); HyDE step reduced from ~3s to ~1s
- HyDE fallback: if the hypothetical-answer vector returns 0 hits, the pipeline retries with the raw query and `alpha=0.1`
- Hybrid search alpha tuned from 0.7 to 0.5
- Answer model uses `max_tokens=600`, `response_format=json_object`
- LaTeX JSON fix: inner retry that escapes bare backslashes (e.g. `\alpha`, `\sigma`) before `json.loads`, preventing parse failures on math-heavy answers
- Fixed missing `max_completion_tokens` mapping in `OpenAIChatModel`
- End-to-end latency: ~3–5s for typical queries (down from ~8s)

### Steps for Testing

Prerequisites:
- Iris running locally with the global search pipeline configured in `application.local.yml`
- At least one course with lecture content (PDF slides or videos) ingested in Artemis via the Iris ingestion pipeline

1. Open the global search in Artemis
2. Type a conceptual question about a topic covered in your ingested content
3. Verify the thinking indicator appears, then an AI answer with source citations renders correctly
4. If your content contains equations, verify KaTeX renders correctly in the Artemis UI with no raw JSON visible
5. Type a bare keyword from your course content
6. Verify no AI answer is returned, only source results appear
7. Type a bare verb like "explain"
8. Verify no AI answer and no sources are returned
9. Check the Iris server logs and verify `intent=SearchIntent.SKIP_AI` appears for steps 6-8 and `intent=SearchIntent.TRIGGER_AI` for step 2
10. (Optional) Trigger a query unlikely to match any ingested content and check logs for "retrying with keyword-heavy search" to verify the HyDE fallback fires

## Summary by CodeRabbit

* **New Features**
  * Global search pipeline with intent-aware behavior and authenticated POST /api/v1/pipelines/global-search/run.
  * Intent classifier with local model packaging and a training/export workflow.
  * Enhanced status callbacks for global search progress updates.

* **Changes**
  * Search results include source type, query parameters, and display metadata.
  * Answer field is now optional (null when not applicable).
  * GPT-5 model IDs and related LLM configs updated.

* **Removals**
  * Removed legacy search ask endpoint (/api/v1/search/ask).

* **Chores**
  * Container build now bundles intent model artifacts and downloads a pinned ONNX model.


iris/src/iris/pipeline/global_search_pipeline.py
```diff
@@ -0,0 +1,226 @@
...
+        # Parse structured response — strip markdown code fences if present
+        try:
+            cleaned = re.sub(r"```(?:json)?\s*|\s*```", "", raw).strip()
+            try:
+                parsed = json.loads(cleaned)
+            except json.JSONDecodeError:
+                # LaTeX backslashes (e.g. \alpha, \sum) are invalid JSON escape
+                # sequences. Escape any backslash not already part of a recognised
+                # JSON escape before retrying.
+                fixed = re.sub(r'\\(?!["\\/bfnrtu])', r"\\\\", cleaned)
```
bassner: **[HIGH] The LaTeX-JSON repair silently corrupts math.** This negative lookahead skips backslashes followed by `b f n r t u` — but those are exactly the first letters of common LaTeX commands. After this `re.sub`, `json.loads` interprets them as control-char escapes: `\beta`->`<BS>eta`, `\frac`->`<FF>rac`, `\tau`->`<TAB>au`, `\theta`->`<TAB>heta`, `\nu`->`<NL>u`; and `\u` not followed by 4 hex digits raises `JSONDecodeError`. The `except` at L193 then ships the **raw JSON envelope** to the student (`answer = raw`, L197). Since the answer prompt heavily uses `$$...$$` math, this is likely in practice, not theoretical. **Fix the root cause:** with `response_format=json_object` the model should already emit valid JSON (each LaTeX backslash doubled) — instruct it to and add examples; on parse failure return `answer=None` rather than the raw blob. Add parser tests for `\theta`, `\frac`, `\nabla`. _(flagged by both reviewers; reproduced empirically)_

bassner (commented):
## Automated cross-vendor code review — PR #523

Two independent review passes were run in parallel and merged, then the load-bearing findings were verified against the source:
- Claude `deep-code-reviewer` (read the full files), and
- Codex `gpt-5.5` at high reasoning effort (independent vendor).

**Both independently concluded: REQUEST CHANGES.** No critical/data-loss bugs. Inline comments cover the substantive findings; summary below.

### Merge-blockers (all low-effort)
- 🔴 **LaTeX-JSON repair corrupts math answers** — `global_search_pipeline.py:182`
- 🔴 **`skip_ai` uses `alpha=0.9` semantic-heavy retrieval**, defeating the keyword/navigation routing it exists for — `lecture_global_search_retrieval.py:54`
- 🟠 **Token usage collected but never sent** — `pipelines.py:386`

### Hardening / contract
- Unpinned, unverified HuggingFace model + tokenizer downloads (`Dockerfile:34`, `global_search_intent_classifier.py:62`)
- SSRF + 200s timeout on outbound status callbacks (`status_update.py:79`)
- SKIP_AI status-stage sequence is inconsistent and drops `initialStages` (`pipelines.py:367`)
- Removed synchronous `/ask` endpoint — confirm Artemis (#12612) is fully migrated (`search.py:21`)

### Nits
- Refusal-suppression regex is English-only while the classifier is multilingual.
- User queries logged at INFO (privacy).
- `GlobalSearchCallback.done` overrides the base and loses its "clear transient fields only on success" retry logic.
- Dead `if s.snippet` filter (snippet is a required field).
- `LectureGlobalSearchRetrieval` kept the `Lecture` name while everything else was renamed `Global`.

### Test gaps
No tests for the new pipeline at all — intent routing (SKIP vs TRIGGER), skip-path retrieval ranking, HyDE 0-hit fallback, JSON/LaTeX parsing, callback sequencing, `initialStages`, classifier warmup/failure, or token propagation. Only the `reasoning_effort` / `max_completion_tokens` mapping is covered.

### Strengths
Fail-open default to `TRIGGER_AI` when the model is unavailable; HyDE 0-hit → raw-query fallback; the `max_completion_tokens` mapping is a genuine missing-mapping fix; clean and tested `supports_reasoning_effort` gating; lazy load with success-only memoization + warmup; committed artifacts are small.

_Posted by Patrick Bassner; findings assisted and merged from the two review passes described above._

iris/src/iris/pipeline/global_search_pipeline.py
```diff
@@ -0,0 +1,226 @@
...
+        logger.info(
+            "Global search pipeline | mode=%s hyde_llm=%s answer_llm=%s embedding=%s",
+            "local" if local else "cloud",
+            hyde_model,
+            answer_model,
+            embedding_model,
+        )
+
+        hyde_completion_args = CompletionArguments(
+            reasoning_effort="none", max_tokens=150
```
bassner, line 70: **[LOW] `reasoning_effort="none"` is gated too coarsely.** The forwarding gate in `openai_chat.py:255` only checks the boolean `supports_reasoning_effort`, not whether the specific value `none` is valid for that model (`none` was introduced in GPT-5.2+). A model marked `supports_reasoning_effort: true` but older than 5.2 would get `none` rejected by the API. Consider modelling allowed effort values per model/role rather than hardcoding `none` here (also at L73). _(Codex)_

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -29,24 +34,21 @@
...
-        """
-        Search for lecture content based on a query.
-
-        :param query: The search query.
-        :param limit: The maximum number of results to return.
-        :return: Segments sorted by relevance.
-        """
         query_embedding = self.llm_embedding.embed(query)
         return self._run_hybrid_search(
             query=query, vector=query_embedding, alpha=0.9, limit=limit
```
bassner: **[HIGH] `skip_ai` queries get vector-heavy retrieval.** `search()` hardcodes `alpha=0.9` (near-semantic), but the intent classifier deliberately routes keyword/navigation queries (e.g. "Lecture 4") to this path — exactly the queries that need keyword-heavy matching. The HyDE fallback already uses `alpha=0.1` for keyword-style lookup. Parameterize `alpha` (or add a dedicated keyword-heavy lookup) for the `skip_ai` path, and add tests for lecture-number/title queries. _(Codex)_

iris/src/iris/web/routers/pipelines.py
```diff
@@ -332,6 +340,75 @@ def run_autonomous_tutor_pipeline(dto: AutonomousTutorPipelineExecutionDTO):
...
+            "[global-search] query=%r  intent=%s  → %s",
+            dto.query[:120],
+            intent,
+            (
+                "calling LLM (HyDE + answer)"
+                if intent == SearchIntent.TRIGGER_AI
+                else "skipping LLM (sources only)"
+            ),
+        )
+        if intent == SearchIntent.TRIGGER_AI:
```
bassner: **[MED] SKIP_AI status sequence is inconsistent and `initialStages` is dropped.** `thinking()` is only called for `TRIGGER_AI`, so on skip the "Thinking" stage jumps `NOT_STARTED`->`DONE` (the base class enforces `NOT_STARTED`->`IN_PROGRESS`->`DONE` at `status_update.py:114`), and Artemis receives only the final callback. Also `GlobalSearchCallback` builds a fresh 2-stage list and ignores `dto.initial_stages`, unlike every other callback. Mark Thinking `SKIPPED` (not silently `DONE`), thread `initial_stages` through, and confirm the single-callback-on-skip contract with Artemis (#12612). _(both reviewers)_

iris/src/iris/web/routers/pipelines.py
```diff
@@ -332,6 +340,75 @@ def run_autonomous_tutor_pipeline(dto: AutonomousTutorPipelineExecutionDTO):
...
+            callback.thinking()
+        client = VectorDatabase().get_client()
+        result = GlobalSearchPipeline(client, local=dto.settings.is_local())(
```
bassner: **[MED] `skip_ai` still constructs the full HyDE/answer pipeline.** `GlobalSearchPipeline.__init__` resolves the HyDE + answer models (`global_search_pipeline.py:56-57`) before `__call__` can short-circuit on intent. A sources-only request therefore pays the setup cost and **can fail on unused LLM config**. Short-circuit `SKIP_AI` in the worker (call the retriever directly), or lazily initialize the LLM clients inside the trigger branch. _(Codex)_

iris/src/iris/web/routers/pipelines.py
```diff
@@ -332,6 +340,75 @@ def run_autonomous_tutor_pipeline(dto: AutonomousTutorPipelineExecutionDTO):
...
+                "[global-search] LLM produced an answer (%d chars, %d sources)",
+                len(result.answer),
+                len(result.sources),
+            )
+        else:
+            logger.info(
+                "[global-search] answer=null  sources=%d  (LLM returned null or was skipped)",
+                len(result.sources),
+            )
+        callback.done(answer=result.answer, sources=result.sources)
```
bassner: **[MED] Token usage is collected but never sent.** The pipeline accumulates HyDE/answer tokens via `_append_tokens` (`global_search_pipeline.py:121,222`), but the instance is discarded here and `callback.done()` is called without `tokens=`, so `GlobalSearchCallback.done` always emits `tokens=[]` — `IRIS_GLOBAL_SEARCH_PIPELINE` cost is never reported. Keep a reference: `pipeline = GlobalSearchPipeline(...); result = pipeline(query=..., ...); callback.done(answer=result.answer, sources=result.sources, tokens=pipeline.tokens)`. _(both reviewers)_

iris/src/iris/web/routers/pipelines.py
```diff
@@ -332,6 +340,75 @@ def run_autonomous_tutor_pipeline(dto: AutonomousTutorPipelineExecutionDTO):
...
+)
+def run_global_search_pipeline(dto: GlobalSearchRequestDTO):
+    """
+    Answer a student's question using course content retrieved via HyDE.
+
+    Returns 202 immediately. Sends webhook callbacks to Artemis:
+      - TRIGGER_AI (LLM path, ~5-8s): thinking callback first, then result
+      - SKIP_AI (sources only, ~200ms): result callback only (no loading state needed)
+    """
+    thread = Thread(
```
bassner: **[LOW] Unbounded thread-per-request.** Each accepted request starts a raw `Thread` with no backpressure. This matches existing routes, but global search is likely higher-volume and can exhaust worker resources under load. Consider a bounded executor / task queue with observable failures. _(Codex)_

iris/src/iris/web/status/status_update.py
```diff
@@ -469,6 +472,53 @@ def __init__(
         )
 
 
+class GlobalSearchCallback(StatusCallback):
+    """Status callback for the global search pipeline."""
+
+    def __init__(self, run_id: str, base_url: str):
+        url = f"{base_url}/{self.api_url}/global-search/runs/{run_id}/status"
```
bassner, line 486: **[HIGH/MED] Outbound status callback: SSRF + 200s hang.** `on_status_update` (`status_update.py:79-87`) does `requests.post(self.url, timeout=200)`, where `url` is built here from request-body `artemis_base_url` with no allowlist/scheme validation. An authenticated caller can point Iris at internal addresses (cloud metadata, internal services) or a slow endpoint and hold the worker thread for up to 200s per callback. This is a pre-existing pattern across all callbacks, but worth: validating `artemis_base_url` against a configured host/allowlist, rejecting private-IP/non-https hosts, and cutting the timeout to ~10-30s. _(deep-review)_

iris/Dockerfile
```diff
@@ -26,6 +27,16 @@ COPY iris/src/iris ./iris
 # Copy logging configuration
 COPY iris/log_conf.yml ./log_conf.yml
 
+# Copy intent-classifier model config files
+COPY iris/models ./models
+
+# Download the quantized ONNX model from Hugging Face
+RUN curl -fsSL \
```
bassner, line 34: **[MED] Unpinned, unverified model download.** Pulls `model_quantized.onnx` from HuggingFace `.../resolve/main/...` with no revision pin and no checksum — builds are not reproducible and a changed/compromised upstream silently alters routing behavior. Pin an immutable commit SHA in the URL and verify a `sha256` after download. _(both reviewers)_

iris/src/iris/pipeline/shared/global_search_intent_classifier.py
```diff
@@ -0,0 +1,159 @@
...
+
+class _IntentClassifier:
+    """Holds all loaded artifacts and runs inference."""
+
+    def __init__(self, model_dir: Path) -> None:
+        logger.info("Loading intent classifier from %s", model_dir)
+
+        onnx_path = _find_onnx_file(model_dir)
+        meta = json.loads((model_dir / "meta.json").read_text())
+        self._tokenizer = AutoTokenizer.from_pretrained(
```
bassner, line 67: **[MED] Runtime tokenizer fetch is unpinned + pickle deserialization.** `AutoTokenizer.from_pretrained(meta["base_model"])` fetches from HF Hub at runtime with no `revision=` (the `# nosec B615` suppresses exactly this Bandit warning), and `joblib.load(...)` at L65 deserializes a pickle. Pin `revision="<sha>"`, prefer vendoring the tokenizer into the image and loading with `local_files_only=True`, and treat `model_head.joblib` as trusted checked-in build input. _(both reviewers)_

iris/src/iris/web/routers/search.py
```diff
@@ -22,26 +19,6 @@
     "/lectures", dependencies=[Depends(TokenValidator())], response_model_by_alias=True
 )
 def lecture_search(dto: LectureSearchRequestDTO) -> list[LectureSearchResultDTO]:
```
bassner, line 21: **[contract] Breaking API change.** The old synchronous answer pipeline (`lecture_search_answer_pipeline.py`) was deleted and this router now exposes only `POST /api/v1/search/lectures`. Confirm no client still calls the removed synchronous `/ask` path and that Artemis (#12612) is fully migrated to the new async `POST /api/v1/pipelines/global-search/run`. _(deep-review)_

Claudia-Anthropica (changes requested):
@Nayer-kotry I’m requesting changes without adding duplicate inline comments because the unresolved review threads on this head already cover the blockers. Please address those existing high/medium items before approval; I don’t see a safe path to approve while they remain. The failing Docker jobs look like runner shutdowns during image pull, so I’m not treating them as a PR-specific code failure.

iris/src/iris/pipeline/shared/global_search_intent_classifier.py
Nayer-kotry: > One option to address this properly would be creating a dedicated HuggingFace organization for Pyris/Artemis and hosting the model artifacts there

iris/Dockerfile
Nayer-kotry: One option to address this properly would be creating a dedicated HuggingFace organization for Pyris/Artemis and hosting the model artifacts there

iris/src/iris/web/routers/pipelines.py
Nayer-kotry: In the Artemis side, the single-callback flow for SKIP_AI is intentional. The handler only looks at whether stages[0] is IN_PROGRESS (thinking) or all stages are terminal (done), so the state transition doesn't matter.

iris/Dockerfile
Claudia-Anthropica: @Nayer-kotry Keeping this open. A project-owned HuggingFace org would help ownership, but the Docker build is currently failing at this exact curl step with exit 22, and the artifact still needs an immutable revision plus checksum verification so builds are reproducible.

iris/src/iris/pipeline/shared/global_search_intent_classifier.py
Claudia-Anthropica: @Nayer-kotry Keeping this open too. Moving the model under a project-owned repo would help ownership, but the runtime tokenizer load still needs to be pinned or vendored with `revision=` / `local_files_only=True` so every deployment uses the same tokenizer files.

Claudia-Anthropica (changes requested):
@Nayer-kotry I’m still requesting changes. The current head has PR-specific CI failures: lint fails because Black reformats `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, and both Docker builds fail at the new HuggingFace model download step. I also kept the two existing model-artifact/tokenizer threads open because the downloads are still unpinned/unverified.

iris/Dockerfile
Claudia-Anthropica: @Nayer-kotry Still open on the current head: line 35 still uses `resolve/main` and there is still no `sha256` check after the download, so the image can change across builds. Please pin an immutable HF revision and verify the downloaded ONNX file before the build continues.

iris/src/iris/pipeline/shared/global_search_intent_classifier.py
Claudia-Anthropica: @Nayer-kotry Still open on the current head: `AutoTokenizer.from_pretrained(meta["base_model"])` still has no `revision=` and is not loading vendored local files. Please pin the tokenizer revision or copy the tokenizer files into the image and load them with `local_files_only=True`.

Claudia-Anthropica (changes requested):
@Nayer-kotry I’m still requesting changes. The model download/tokenizer threads are still unresolved, and CI lint is still failing because Black reformats `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`. Tests pass and the amd64 image now builds, but these need to be fixed before approval.

iris/Dockerfile
Nayer-kotry: will be done later on

iris/src/iris/pipeline/shared/global_search_intent_classifier.py
Nayer-kotry: will be done later on

iris/Dockerfile
Claudia-Anthropica: @Nayer-kotry Thanks for clarifying. This still needs to be done before this PR can merge: the Docker build currently downloads from `resolve/main` without a checksum, so production images are not reproducible and can silently change.

iris/src/iris/pipeline/shared/global_search_intent_classifier.py
Claudia-Anthropica: @Nayer-kotry Same here: this still needs to be fixed in this PR. The runtime still fetches tokenizer files from HuggingFace without a pinned `revision=` or vendored `local_files_only=True` path, so deployments can load different tokenizer files over time.

Claudia-Anthropica (changes requested):
@Nayer-kotry Still requesting changes. The two model provenance items are still unresolved on the current head: the ONNX download is mutable and unchecked, and the runtime tokenizer fetch is still unpinned. CI lint and tests are passing now; please fix those two reopened threads before approval.

iris/Dockerfile
Claudia-Anthropica: @Nayer-kotry Looks good now, thanks. The ONNX download is pinned to an immutable revision and verified with sha256.

iris/src/iris/pipeline/shared/global_search_intent_classifier.py
Claudia-Anthropica: @Nayer-kotry Looks good now, thanks. The tokenizer load now uses an immutable revision, so deployments will not silently drift.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -29,24 +34,33 @@
...
         self.collection = init_lecture_unit_segment_schema(client)
         self.lecture_unit_collection = init_lecture_unit_schema(client)
+        self.transcription_collection = init_lecture_transcription_schema(client)
 
     def search(
-        self, query: str, limit: int, course_ids: list[int] | None = None
+        self,
+        query: str,
+        limit: int,
+        alpha: float = 0.5,
```
Claudia-Anthropica, line 55: @Nayer-kotry [high] Adding `alpha` before `course_ids` changes the positional API. `lecture_search()` still calls `search(dto.query, dto.limit, dto.course_ids)`, so normal lecture searches now pass `None` or a course-id list as the hybrid alpha and never apply the course filter. Move `alpha` after `course_ids` or make it keyword-only, then update the router call to use keyword args.

In `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, the new `alpha` parameter was inserted before `course_ids`, but `iris/src/iris/web/routers/search.py` still passes `dto.course_ids` as the third positional argument. Fix this by preserving the old positional contract or making `alpha` keyword-only, and update the lecture search router to pass `course_ids=dto.course_ids` explicitly.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The async global-search path mostly hangs together, but the shared retrieval signature change currently breaks the existing lecture search endpoint. I left the specific integration issue inline.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Claudia-Anthropica: @Nayer-kotry Looks good now, thanks. The lecture search router now passes `course_ids` by keyword, so the old positional call no longer routes course IDs into `alpha`.

Claudia-Anthropica (dismissed):
@Nayer-kotry All previously requested feedback is addressed and CI is green now. Approving.

iris/src/iris/pipeline/global_search_pipeline.py
```diff
@@ -0,0 +1,226 @@
...
+            except json.JSONDecodeError:
+                # LaTeX backslashes (e.g. \alpha, \sum) are invalid JSON escape
+                # sequences. Escape any backslash not already part of a recognised
+                # JSON escape before retrying.
+                fixed = re.sub(r'\\(?!["\\/])', r"\\\\", cleaned)
+                parsed = json.loads(fixed)
+            answer = parsed.get("answer") or None  # treat null and "" as no answer
+            used_indices = {
+                i - 1
+                for i in parsed.get("used_sources", [])
```
Claudia-Anthropica, line 187: @Nayer-kotry [medium] This can return a non-null answer with `sources=[]` whenever the model omits `used_sources` or emits indices as strings, because only integer entries survive the set comprehension. The answer prompt makes source attribution part of the contract, so the UI can end up showing an answer with no citations even though sources were retrieved. Normalize digit strings and, if `answer` is present but no valid source index remains, either fall back to `grounded_sources` or suppress the answer instead of returning it uncited.

In `iris/src/iris/pipeline/global_search_pipeline.py`, parsed answers can keep a non-null answer while `used_sources` becomes empty when the model omits the field or emits numeric strings. Fix this by normalizing valid integer/string source indices and ensuring a non-null answer never returns with an empty `sources` list; fall back to `grounded_sources` or set `answer` to `None`, and add parser tests for missing/string/null `used_sources`.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The new global search flow is mostly in good shape and CI is green. I found one parser contract issue where answers can be returned without any source citations; see inline.

bassner (approved):

PR #548, opened 2026-05-09
`Iris`: Detect and save slidepages

vivienxr:
FIXES IRIS-321

This PR need to have [#12682](https://github.com/ls1intum/Artemis/pull/12682) applied in Artemis to work.

### Summary

This branch introduces automated display page number extraction for lecture slides during ingestion and uses these numbers consistently across ingestion, segment construction, retrieval context, and Artemis callback payloads.
In addition to extracting page numbers, the ingestion now also produces an academic page description per PDF page and integrates it into the stored page content.

#### Vision-based slide parsing per PDF page (number + description)

During slide ingestion, Pyris processes every PDF page with a vision call:
- It renders the page as an image.
- Sends it to the vision model.
- It extracts display_page_number (the number visible on the slide) and academic_description (page-level academic interpretation).
- If no visible number is found, display_page_number is set to -1.
- The description is generated per page and merged into the page content that gets indexed (not only in image-only edge cases as before).

#### Dual page numbering is stored

For slide chunks, Pyris now stores both:
- Index page number (page_number): technical PDF order (1..n)
- Display page number (display_page_number): visible number on the slide (-1 if missing)
This enables stable internal indexing while still maintaining the displayed context (which might not be unique and therethrough not stable). 

#### Segment building aligns slides and transcript using display page semantics

Segment generation still iterates index pages, but when slides are present, the segment reads the slide’s display_page_number and transcript matching is then performed using that display number. For slides with no displayed number (display is -1), no transcript is matched. This keeps slide/transcript pairing aligned with transcript-side slide detection (which also uses visible/display slide numbers).

#### Retrieval context vs citation identity

For answer-generation context: slide and segment context is surfaced with display page numbers.
For citation identity: citations keep using index page numbers (page_number), which remain stable even when display numbers are missing or non-unique. To the user the page numbers will be shown as the display page numbers but the index page numbers are necessary internally for the deep linking due to their uniqueness. 

#### Artemis callback payload

After ingestion, Pyris returns slide display numbers to Artemis in JSON via:
{
  "slidePageNumbers": [1, 2, 3, -1, 5, 6]
}
This data is included in the final_result field of the ingestion callback response, allowing Artemis to maintain a mapping between displayed and indexed page numbers. 

### Test

Use the instruction of Artemis since it directly includes this PR.

## Summary by CodeRabbit

* **New Features**
  * Added extraction and persistent tracking of slide page numbers from lecture presentations
  * Added automated academic description generation from lecture slide imagery
* **Improvements**
  * Enhanced page number attribution and display throughout lecture content retrieval and segmentation
  * Improved backward compatibility handling for legacy lecture materials without page tracking data
  * Better organization of lecture chunks and segments with accurate page number metadata


vivienxr:
@coderabbitai review

thynguyentumde (dismissed):
Tested locally along with Artemis PR#12682 and it works as expected. Approving.

toukhi (dismissed):
Tested locally and works as described

bassner (dismissed):

iris/src/iris/retrieval/lecture/lecture_unit_segment_retrieval.py
```diff
@@ -175,6 +175,9 @@ def generate_retrieval_dtos(self, lecture_unit_segment, uuid: str):
             page_number=lecture_unit_segment[
                 LectureUnitSegmentSchema.PAGE_NUMBER.value
             ],
+            display_page_number=lecture_unit_segment.get(
```
Claudia-Anthropica, line 178: @vivienxr [high] Legacy segment objects will not have `display_page_number`, but this default turns a missing field into `-1`. `get_lecture_transcription_of_lecture_unit()` treats `-1` as "unknown slide" and returns no transcript, so already-ingested lectures lose transcript context after this deploy. Fall back to the stored `page_number` when the property is missing; keep `-1` only when it was explicitly stored.

In `iris/src/iris/retrieval/lecture/lecture_unit_segment_retrieval.py`, missing `display_page_number` values are defaulted to `-1`, which breaks transcript lookup for legacy lecture segments. Fix this by defaulting missing segment and page chunk display numbers to their stored index `page_number` in all DTO creation paths, while preserving an explicitly stored `-1`.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The new display/index page split mostly hangs together, and CI is green. I am requesting changes because legacy segment/page chunk objects currently fall into the unknown-display-page path, which drops transcript context for already-ingested lectures. See the inline comment.

iris/src/iris/retrieval/lecture/lecture_unit_segment_retrieval.py
Claudia-Anthropica: @vivienxr This is still open. The DTO builders still use `.get(..., -1)` for missing `display_page_number`, so `get_lecture_transcription_of_lecture_unit()` never sees `None` and still returns no transcript for legacy segment objects. Please default the missing field to the stored `page_number` there and in the page-chunk DTO paths, while preserving an explicitly stored `-1`.

Claudia-Anthropica (changes requested):
@vivienxr The previous change request is still open, so I am keeping this as changes requested. Missing `display_page_number` still defaults to `-1` in the DTO builders, which means legacy lectures still lose transcript context; I replied on the existing thread with the exact remaining issue.

iris/src/iris/retrieval/lecture/lecture_unit_segment_retrieval.py
Claudia-Anthropica: @vivienxr Still seeing this on the current head: the segment DTO builder and both page-chunk DTO paths still use `.get(..., -1)` for a missing `display_page_number`, so the `None` fallback in `get_lecture_transcription_of_lecture_unit()` never runs for legacy objects. Keeping this thread open.

Claudia-Anthropica (changes requested):
@vivienxr The previous change request is still open on the current head, so I am keeping this as changes requested. CI is green, but the legacy fallback still needs to default missing display page numbers to the stored index page number rather than `-1`.

iris/src/iris/retrieval/lecture/lecture_unit_segment_retrieval.py
Claudia-Anthropica: @vivienxr Nice fix. The helper now falls back to `page_number` when `display_page_number` is missing, preserves an explicitly stored `-1`, and both the segment plus page-chunk DTO paths use it. Looks good now, thanks.

Claudia-Anthropica (dismissed):
@vivienxr All feedback addressed, including the legacy display-page fallback. CI is green, approving.

iris/src/iris/pipeline/lecture_ingestion_update_pipeline.py
```diff
@@ -432,6 +432,7 @@ def _run_ingestion(self, callback: IngestionStatusCallback) -> None:
         )
         callback.done(
             "Ingested lecture unit summary into vector database",
+            display_page_numbers=self.dto.lecture_unit.display_page_numbers,
```
Claudia-Anthropica, line 435: @vivienxr [medium] `display_page_numbers` is still `None` when `LectureUnitPageIngestionPipeline` skips slide ingestion because the attachment version already exists in Weaviate. On a retry after the chunks were written but before the terminal callback succeeded, this final callback sends no mapping, and the matching Artemis handler stores `null` instead of the slide numbers. Populate the DTO from the existing page chunks in the skip path, ordered by stored `page_number`, so retries return the same mapping as a fresh ingestion.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `iris/src/iris/pipeline/lecture_ingestion_update_pipeline.py`, the final ingestion callback sends `self.dto.lecture_unit.display_page_numbers`, but the slide-ingestion skip path leaves it as `None` when page chunks already exist. Fix this by making `LectureUnitPageIngestionPipeline` populate `lecture_unit.display_page_numbers` from existing `LectureUnitPageChunkSchema.DISPLAY_PAGE_NUMBER` values, ordered by `PAGE_NUMBER`, whenever it skips re-ingesting an unchanged attachment; preserve explicitly stored `-1` values and use the stored index page number only as the legacy fallback when the display field is missing.\n\n</details>

Claudia-Anthropica (changes requested):
@vivienxr The display/index page split looks good overall and CI is green. I found one retry-path issue where an unchanged PDF can complete with no display page-number mapping, so Artemis never gets the mapping from already-indexed chunks.

iris/src/iris/pipeline/lecture_ingestion_update_pipeline.py
vivienxr: Artemis now does not overwrite with null anymore

Claudia-Anthropica: @vivienxr I do not think that fully covers this retry case. If the first run writes Weaviate chunks but the final callback fails before Artemis stores any mapping, the retry still takes the unchanged-attachment skip path and sends no `displayPageNumbers`, so Artemis has nothing to preserve. Iris should reconstruct `self.dto.lecture_unit.display_page_numbers` from the existing chunks in that skip path.

Claudia-Anthropica (changes requested):
@vivienxr The retry-path issue is still open. The current Iris skip path still sends no display page-number mapping when chunks already exist, so a retry after a failed terminal callback can leave Artemis without any mapping to preserve.

iris/src/iris/pipeline/lecture_ingestion_update_pipeline.py
vivienxr: @Claudia-Anthropica I see. This is an edge case, but valid. Can you suggest code to resolve this?

Claudia-Anthropica: @vivienxr I would keep this inside `LectureUnitPageIngestionPipeline`, so the update pipeline can keep passing `self.dto.lecture_unit.display_page_numbers` unchanged. In the unchanged-attachment branch, fetch the existing page chunks with the same base URL/course/lecture/unit filter as `check_if_attachment_needs_update()`, deduplicate by stored `PAGE_NUMBER`, then sort by that page number and assign the DTO field before returning.

The core of the helper can look like this:

```python
chunks = self.collection.query.fetch_objects(filters=page_chunk_filter, limit=10000).objects
by_page: dict[int, int] = {}
for chunk in chunks:
    props = chunk.properties
    page_number = int(props[LectureUnitPageChunkSchema.PAGE_NUMBER.value])
    display_page_number = props.get(LectureUnitPageChunkSchema.DISPLAY_PAGE_NUMBER.value)
    if display_page_number is None:
        display_page_number = page_number
    by_page.setdefault(page_number, int(display_page_number))

self.dto.lecture_unit.display_page_numbers = [by_page[page] for page in sorted(by_page)]
```

Then call that helper in the `if not self.check_if_attachment_needs_update():` path before the skip callbacks finish. This preserves an explicitly stored `-1`, but legacy chunks that do not have the new property still return their index page number.

Claudia-Anthropica (approved):
@vivienxr The retry-path display-page mapping is now restored from existing chunks, and the dedicated `displayPageNumbers` callback contract matches the paired Artemis DTO. CI is green, approving.

thynguyentumde (approved):
Approved.

bassner (approved):

PR #549, opened 2026-05-11
`Iris`: Add support for reasoning effort configuration per model

Nayer-kotry:
## Summary

- Adds `reasoning_effort` parameter to `CompletionArguments`, typed directly from the OpenAI SDK's `ReasoningEffort` type (`none`, `minimal`, `low`, `medium`, `high`, `xhigh`)
- Adds `supports_reasoning_effort: bool = False` field to `OpenAIChatModel` so each model declares this capability in its YAML config, defaulting to `false`
- Adds `supports_temperature: bool = True` field to `OpenAIChatModel`. The previous string-match heuristic that skipped `temperature` for GPT-5 / o-series models has been removed; each model now declares explicitly whether it accepts `temperature`
- In `chat()`, `reasoning_effort` is only forwarded when the model declares `supports_reasoning_effort: true` and the pipeline passes a value; `temperature` is dropped when the model declares `supports_temperature: false`. A debug log fires when `reasoning_effort` is passed to a non-supporting model, including the model id and a concrete fix hint
- Adds GPT-5.4 as an example entry in `llm_config.example.yml` with `supports_temperature: false` and `supports_reasoning_effort: true`
- Adds unit tests for the gating logic in `tests/test_openai_chat_reasoning_effort.py`

## ⚠️ Migration required (breaking change for deployments)

The name-based heuristic that used to detect reasoning models and skip `temperature` (matching `gpt-5*`, `o1*`, `o3*`, `o4*`, `gpt-o*`) has been **removed**. Each model now declares its API surface explicitly via two booleans:

- `supports_temperature: bool = True` — set `false` on reasoning models that reject the `temperature` parameter
- `supports_reasoning_effort: bool = False` — set `true` on models that accept the `reasoning_effort` parameter

**Every reasoning model in every deployed `llm_config.yml` must declare `supports_temperature: false`.** If a deployment is missed, `temperature` is forwarded to the model and the API rejects the request.

### Deployment status

- [x] `/opt/pyris/llm_config.yml` on **pyris-prod.artemis.cit.tum.de** — flags applied, backup at `/opt/pyris/llm_config.yml.bak.<timestamp>`, not yet restarted (will be picked up by the post-merge image deploy)
- [x] `/opt/pyris/llm_config.yml` on **pyris-test.artemis.cit.tum.de** — flags applied, backup retained, `pyris-app` restarted; LangFuse, Uvicorn, Weaviate all came up cleanly
- [ ] **Local developer `llm_config.yml` files** — each developer must add `supports_temperature: false` to their reasoning-model entries

Applied to both servers on the three GPT-5 chat entries (`gpt-5-mini`, `gpt-5-nano`, `gpt-5.4`):

```yaml
supports_temperature: false
supports_reasoning_effort: true
```

Non-reasoning chat models (`gpt-4.1-mini`, `gpt-oss-120b`, `gemma-3-4b-it`) keep the defaults — no flags needed.

See the in-tree migration note at the top of [`iris/llm_config.example.yml`](iris/llm_config.example.yml) for the canonical reference.

## How to use

In any pipeline, pass `reasoning_effort` via `CompletionArguments`:

```python
# Disable reasoning entirely (GPT-5.4+)
CompletionArguments(reasoning_effort="none")

# Use medium reasoning
CompletionArguments(reasoning_effort="medium")

# Don't touch it (default, all existing pipelines unchanged)
CompletionArguments(temperature=0.3)
```

In the model's YAML config, opt in:

```yaml
- id: gpt-5.4
  type: azure_chat
  model: gpt-5.4
  supports_temperature: false
  supports_reasoning_effort: true
  ...
```

## Test plan

- [x] Unit tests cover: `reasoning_effort` dropped when model doesn't support it; not forwarded when unset even if model supports it; forwarded when both true; `temperature` dropped when `supports_temperature: false`; `temperature` forwarded by default; debug log fires with the helpful hint
- [x] pyris-test config-only restart confirms the existing (pre-merge) code silently ignores the new fields — no startup regression
- [ ] Post-merge: verify an Iris chat against `gpt-5-mini` or `gpt-5.4` on pyris-test still returns 200 (no 400 from temperature)
- [ ] Post-merge: verify all valid effort levels (`none`, `minimal`, `low`, `medium`, `high`, `xhigh`) are accepted by `CompletionArguments` against a live Azure GPT-5.4 deployment


vivienxr (dismissed):
Code looks good.

thynguyentumde (dismissed):
Approved.

Senan04 (dismissed):
Code LGTM

bassner (dismissed):

PR #561, opened 2026-05-18
`Iris`: Do not stop new ingestion job for lecture units

vivienxr:
## Problem
If Pyris received a second ingestion job for the same lecture unit while the first one was still running, it aborted the second job, while Artemis had already switched to expecting the second job's callback. As a result, Artemis could miss the successful completion entirely, because the first job finished but Artemis was no longer waiting for that token.

## Description 
The fix is to cancel the first (older) job when a newer one arrives, because Artemis explicitly expects the newer job and because it may legitimately contain newer content. Whenever a new ingestion request comes in for a lecture unit that is already being processed, the handler now signals the running job to stop and starts the new one immediately, instead of dropping the incoming request.

To make this possible, each ingestion job receives a cancellation event that is passed down through all pipeline stages — from the webhook worker into the update pipeline and on into the page-ingestion, transcription-ingestion, lecture-unit-summary, and the heavy/light transcription pipelines, down to the slide-turn detector and the Whisper client. The pipelines regularly check this event during AI operations such as generating embeddings, summarizing chunks, detecting slide changes, and transcribing audio, allowing them to stop cleanly between individual units of work (e.g. between chunks, between embeddings, or between Whisper chunks).                                                                     
                                                                                                                                                                                     
Weaviate operations like deleting old data and inserting new chunks are grouped into atomic blocks that never get interrupted mid-operation, preventing half-finished database states. To achieve this, the pipeline flow was refactored to separate AI operations from database operations: embeddings are now generated upfront for all chunks (cancellable between each embedding), and only then are the old chunks deleted and all new chunks batch-inserted into Weaviate together inside a single lock, as one atomic operation. No cancellation is checked once that delete-and-insert block has started.                                                                                                             
                                                                                                                                                                                     
When the cancellation event is detected, the affected pipeline raises an IngestionCancelledException. The old thread treats this as a controlled cancellation rather than an error: it notifies Artemis with a dedicated INGESTION_CANCELLED status, logs the cancellation, and exits cleanly, while the new thread starts fresh with the latest content.              
                                                                                                                                                                                     
Alongside this, the change closes PDF and temporary-file resources reliably in finally blocks during chunking, so cancellations and errors no longer leak file handles.

## Test
Ingest a lecture and while the ingestion is still running, edit the attached file. Check that the Pyris logs show the first job being cancelled and the second job completing successfully.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * “Most recent request wins”: submitting a new ingestion for the same lecture cancels any in-progress run and starts the new one.
  * End-to-end cooperative cancellation: transcription and ingestion pipelines (including webhook workers) now support graceful, progress-aware cancellation and report cancelled runs cleanly.

* **Tests**
  * Updated ingestion tests to construct pipelines normally and verify transcription/branching behavior.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


iris/src/iris/pipeline/lecture_ingestion_update_pipeline.py
```diff
@@ -124,14 +126,32 @@ def __init__(
...
             self.dto.settings and self.dto.settings.artemis_llm_selection == "LOCAL_AI"
         )
 
+    def _check_cancellation(self, stage_name: str = None):
+        """Check if job has been cancelled.
+
+        Raises:
+            IngestionCancelledException: If cancellation event is set.
+        """
+        if self.cancel_event is not None and self.cancel_event.is_set():
```
Claudia-Anthropica, line 145: @vivienxr [medium] This is what is breaking the current Iris test job: the existing YouTube wiring tests construct this pipeline with `__new__`, so `cancel_event` is missing and `_check_cancellation()` raises `AttributeError` before `_run_full_transcription()` or `_run_slide_detection_only()` reaches the code under test. Make this helper tolerate the absent attribute, or update those tests to initialize it, so the existing tests keep exercising the video-source branches.

In `iris/src/iris/pipeline/lecture_ingestion_update_pipeline.py`, `_check_cancellation()` assumes `self.cancel_event` always exists, but existing tests instantiate `LectureIngestionUpdatePipeline` via `__new__` and now fail before the transcription code runs. Fix this by reading the event with `getattr(self, "cancel_event", None)` inside `_check_cancellation()` and keep the existing YouTube wiring tests passing.

</details>

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
```diff
@@ -199,6 +211,10 @@ def __call__(self) -> (str, []):
                 self.callback.in_progress("skipping slide ingestion")
                 self.callback.done()
                 return self.course_language, self.tokens
+
+            self._check_cancellation()
+
+            # Atomic DELETE - no cancellation during DB operation
```
Claudia-Anthropica: @vivienxr [high] This delete is not atomic with the later insert. If job A has already passed the last cancellation check and is inside `batch_update()`, job B can run this delete while job A continues adding old slide chunks afterward, leaving stale chunks mixed with the newer ingestion. Move the delete into the same lock-protected write section as the batch insert, after embedding generation and a final cancellation check.

In `iris/src/iris/pipeline/lecture_ingestion_pipeline.py`, the page-chunk delete runs before chunking and outside the `batch_update_lock`, so an older cancelled job can still insert stale chunks after a newer job deletes. Fix this by generating chunks and embeddings first, checking cancellation, then acquiring the same write lock and performing delete plus batch insert together with no cancellation point between them.

</details>

iris/src/iris/pipeline/transcription_ingestion_pipeline.py
```diff
@@ -77,29 +81,59 @@ def __init__(
...
+                self.dto.lecture_unit.lecture_unit_id if self.dto else 0,
+                "Cancelled during transcription ingestion",
+            )
+
     @observe(name="Transcription Ingestion Pipeline")
     def __call__(self) -> (str, []):
         try:
+            self._check_cancellation()
+
+            # Atomic DELETE - no cancellation during DB operation
```
Claudia-Anthropica: @vivienxr [high] Same race here: the newer job can delete transcription chunks while the cancelled older job is already in `batch_insert()`, and the older job can then keep adding stale objects after that delete. The delete and insert need to be one serialized write section for this lecture unit, with embeddings/summaries prepared before the lock and no cancellation point between delete and insert.

In `iris/src/iris/pipeline/transcription_ingestion_pipeline.py`, transcription deletion is separated from the later locked batch insert, which lets overlapping ingestion jobs interleave stale writes. Fix this by preparing chunks, summaries, and embeddings before the write lock, then performing `delete_existing_transcription_data()` and `batch_insert()` inside the same lock-protected section without a cancellation check between them.

</details>

iris/src/iris/pipeline/lecture_unit_pipeline.py
```diff
@@ -57,7 +77,12 @@ def __call__(self, lecture_unit: LectureUnitDTO):
             )()
         )
 
-        # Delete existing lecture unit
+        self._check_cancellation()
+        embedding = self.llm_embedding.embed(lecture_unit.lecture_unit_summary)
+
+        self._check_cancellation()
+
+        # Atomic DELETE+INSERT - no cancellation during DB operations
```
Claudia-Anthropica: @vivienxr [high] The comment says DELETE+INSERT is atomic, but the delete happens before `batch_update_lock` is acquired. With the new overlapping-job behavior, a newer job can delete while an older job is about to insert its stale summary, so the final collection can still contain the old lecture-unit object. Put the delete and insert under the same lock after a final cancellation check.

In `iris/src/iris/pipeline/lecture_unit_pipeline.py`, the lecture-unit summary delete runs outside `batch_update_lock` while only the insert is locked, so overlapping ingestion jobs can interleave stale summary writes. Fix this by acquiring `batch_update_lock` before the delete and keeping both delete and insert in that same locked block after the final cancellation check.

</details>

Claudia-Anthropica (dismissed):
@vivienxr The overall direction makes sense, but this SHA still has a failing Iris test job and the vector write path can race stale data back into Weaviate when old and new ingestions overlap. Please fix the `cancel_event` test regression and make the delete+insert sections actually serialized before merge.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
```diff
@@ -211,14 +223,10 @@ def __call__(self) -> (str, []):
...
-                self.dto.lecture_unit.course_id,
-                self.dto.lecture_unit.lecture_id,
-                self.dto.lecture_unit.lecture_unit_id,
-                self.dto.settings.artemis_base_url,
-            )
-            self.callback.done("Old slides removed")
+
+            self._check_cancellation()
+
+            # Chunking - cancellable between pages
```
Claudia-Anthropica: @vivienxr [medium] This moved chunking ahead of the delete, but the callback is still on the `Old slides removal` stage here. The `done("Lecture Chunking and interpretation Finished")` below will mark old-slide removal as complete before the delete actually runs inside `batch_update()`, and the real delete then gets reported as slide interpretation. Please reorder the ingestion stages or callback sequence so each operation completes the matching Artemis stage.

In `iris/src/iris/pipeline/lecture_ingestion_pipeline.py`, page chunking now runs before old slide deletion while `IngestionStatusCallback` still starts with the `Old slides removal` stage. Fix this by reordering the slide-related status stages or callback sequence so slide interpretation/chunking, old slide removal, and slide ingestion are reported on the stages where they actually happen.

</details>

iris/src/iris/pipeline/transcription_ingestion_pipeline.py
```diff
@@ -77,29 +81,55 @@ def __init__(
...
+
     @observe(name="Transcription Ingestion Pipeline")
     def __call__(self) -> (str, []):
         try:
-            self.callback.in_progress("Deleting existing transcription data")
-            self.delete_existing_transcription_data(self.dto.lecture_unit)
-            self.callback.done("Old slides deleted")
+            self._check_cancellation()
 
+            # Chunking - cancellable
```
Claudia-Anthropica, line 97: @vivienxr [medium] Same stage drift happens in the transcription path. `Chunking transcription` still runs while the current stage is `Old transcriptions removal`, then summarization and the delayed delete are shifted into the next stage names, so Artemis will show the wrong stage as done or failed. Align the transcription stage order with the new chunk/summarize/delete/insert order.

In `iris/src/iris/pipeline/transcription_ingestion_pipeline.py`, transcription chunking and summarization now happen before deleting old transcription data, but the callback stages still start with `Old transcriptions removal`. Fix the status stage ordering or callback sequence so chunking, summarization, removal, and ingestion update the matching stages.

</details>

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
```diff
@@ -352,6 +401,8 @@ def chunk_data(
         old_page_text = ""
         display_page_numbers: list[int] = []
         for page_num in range(doc.page_count):
+            self._check_cancellation()  # Cancellable between pages
```
Claudia-Anthropica: @vivienxr [medium] This new cancellation point can raise after `__call__()` has created `pdf_path`, but before it reaches `cleanup_temporary_file(pdf_path)`. A superseded PDF ingestion will then leak the temp PDF, and `doc` is never closed on that path either. Wrap the `chunk_data()` call and the `fitz.open()` body in `finally` cleanup so cancellation releases both resources.

In `iris/src/iris/pipeline/lecture_ingestion_pipeline.py`, cancellation inside `chunk_data()` can raise before the caller cleans up the temp PDF and before the opened fitz document is closed. Fix this by putting `cleanup_temporary_file(pdf_path)` in a `finally` around the `chunk_data()` call and closing the `fitz` document in a `finally` or context manager inside `chunk_data()`.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The atomic write refactor is heading in the right direction and CI is green, but the callback stage order now no longer matches the actual work, and cancellation during PDF chunking can leak the temporary file. See the inline comments.

iris/src/iris/pipeline/lecture_ingestion_update_pipeline.py
Claudia-Anthropica: @vivienxr Looks good now, thanks.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
Claudia-Anthropica: @vivienxr Nice, the slide delete and insert are now under the same lock.

iris/src/iris/pipeline/transcription_ingestion_pipeline.py
Claudia-Anthropica: @vivienxr Nice, the transcription delete and insert are now under the same lock.

iris/src/iris/pipeline/lecture_unit_pipeline.py
Claudia-Anthropica: @vivienxr Looks good now, thanks.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
Claudia-Anthropica: @vivienxr This is still shifted one stage too early for the embedding generation step: after `Lecture Chunking and interpretation Finished`, the callback advances to `Old slides removal`, but `generate_embeddings()` reports progress on that stage before the delete actually starts. Keep the slide-interpretation stage open until embeddings are generated, or add/reorder a matching stage before old-slide removal.

iris/src/iris/pipeline/transcription_ingestion_pipeline.py
Claudia-Anthropica: @vivienxr Same remaining issue here: after `Summarized transcription`, the callback advances to `Old transcriptions removal`, but `generate_embeddings()` still reports progress there before removal begins. Keep summarization/embedding on the matching stage, or add a separate embedding stage before old-transcription removal.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
Claudia-Anthropica: @vivienxr Looks good now, thanks.

Claudia-Anthropica (changes requested):
@vivienxr Most of the prior issues are addressed and CI is green. I’m still requesting changes because the status callbacks now report embedding progress under the old-slide/old-transcription removal stages; I replied on the two open threads with the remaining adjustment.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
Claudia-Anthropica: @vivienxr Looks good now, thanks.

Claudia-Anthropica (changes requested):
@vivienxr The slide callback thread is fixed now and I resolved it. I’m still requesting changes because the transcription embedding progress is still reported under the old-transcription removal stage before deletion actually starts; the existing open thread has the detail.

iris/src/iris/pipeline/transcription_ingestion_pipeline.py
Claudia-Anthropica: @vivienxr Looks good now, thanks.

Claudia-Anthropica (dismissed):
@vivienxr All previously requested feedback is addressed now; I resolved the remaining thread. Approving based on the code review, with lint and image builds still pending at the time of review.

iris/src/iris/pipeline/lecture_unit_pipeline.py
```diff
@@ -48,6 +65,9 @@ def __call__(self, lecture_unit: LectureUnitDTO):
                 callback=self.callback,
             )()
         )
+
+        self._check_cancellation()
```
Claudia-Anthropica, line 70: @vivienxr [high] This cancellation check happens after `LectureUnitSegmentSummaryPipeline` has already generated and upserted every slide-level segment summary. That subpipeline does not receive the cancel event and writes `LectureUnitSegmentSchema` objects per slide, so a superseded job can keep overwriting the newer ingestion's segment summaries even though the final lecture-unit summary write is skipped. Wire cancellation into the segment-summary pipeline and keep its Weaviate writes behind a final cancellation check and the same serialized write path.

In `iris/src/iris/pipeline/lecture_unit_pipeline.py`, cancellation is checked only after `LectureUnitSegmentSummaryPipeline()` returns, but that subpipeline writes lecture-unit segment summaries while it runs and does not receive `cancel_event`. Fix this by adding `cancel_event` support to `LectureUnitSegmentSummaryPipeline`, checking it before each slide summary and before any Weaviate write, and moving the segment-summary writes behind a final cancellation check under the existing ingestion write lock so superseded jobs cannot update `LectureUnitSegmentSchema`.

</details>

iris/src/iris/pipeline/shared/transcription/slide_turn_detector.py
```diff
@@ -186,10 +199,12 @@ def detect(self) -> List[Tuple[float, int]]:
             logger.debug("[Lecture %s] Anchors: %s", self.job_id, anchor_indices)
 
             for idx in anchor_indices:
+                self._check_cancellation()
                 if self.labels[idx] is None:
                     self.labels[idx] = self._query_label(idx)
 
             for left, right in zip(anchor_indices, anchor_indices[1:]):
+                self._check_cancellation()
```
Claudia-Anthropica, line 207: @vivienxr [medium] This check only runs before entering each top-level interval. Once `_resolve_interval()` starts, it can recursively issue many GPT Vision calls before returning here, so a cancelled job can keep spending vision requests long after the newer ingestion has superseded it. Add cancellation checks inside `_resolve_interval()`, especially before the midpoint `_query_label()` call.

In `iris/src/iris/pipeline/shared/transcription/slide_turn_detector.py`, cancellation is checked before top-level interval resolution but not inside `_resolve_interval()`, which can recursively make many GPT Vision calls after cancellation. Fix this by calling `_check_cancellation()` at the start of `_resolve_interval()` and immediately before `_query_label(mid)` so slide detection stops between recursive probes.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The core delete+insert refactor is mostly in place, but cancellation still leaves one stale write path and one long recursive AI path uncovered. Please fix the segment-summary writes and recursive slide detection cancellation; CI also still has lint and image builds pending at the time of review.

iris/src/iris/pipeline/lecture_unit_pipeline.py
Claudia-Anthropica: @vivienxr Still seeing this in the current head: `LectureUnitPipeline` calls `LectureUnitSegmentSummaryPipeline(...)` without passing `cancel_event`, and that subpipeline still calls `_upsert_lecture_object()` per slide with no cancellation or write-lock boundary. This still lets a superseded job write stale segment summaries.

iris/src/iris/pipeline/shared/transcription/slide_turn_detector.py
Claudia-Anthropica: @vivienxr This is still unchanged in the current head. `_resolve_interval()` still jumps straight to `_query_label(mid)` and the recursive calls without checking `cancel_event`, so cancellation is only observed after the whole top-level interval returns.

Claudia-Anthropica (changes requested):
@vivienxr The two remaining follow-up issues are still present in the current head, so I'm keeping this at changes requested. The segment-summary path still has no cancellation/write-lock protection, and recursive slide detection still only observes cancellation after a top-level interval returns.

Nayer-kotry (dismissed):
tested on pyris test server, works as described

Claudia-Anthropica (changes requested):
@vivienxr The two existing open threads are still valid in the current head, so I’m keeping this at changes requested. The segment-summary path still writes without cancellation/write-lock protection, and recursive slide detection still only observes cancellation after a top-level interval returns.

iris/src/iris/pipeline/lecture_unit_pipeline.py
Claudia-Anthropica: @vivienxr Looks good now, thanks.

iris/src/iris/pipeline/shared/transcription/slide_turn_detector.py
Claudia-Anthropica: @vivienxr Looks good now, thanks.

Claudia-Anthropica (approved):
@vivienxr All previously requested feedback is addressed now. I resolved the remaining threads; approving based on the code review, with CI still pending and local tests not run because `poetry` is unavailable in this environment.

Nayer-kotry (approved):
reapprove

bassner (approved):
Validated exact head ff6ad67f4a50e636e9d73206abf3f14cf0bc278c in an isolated paired Artemis/Pyris stack. A live two-thread supersession test showed the old event was signaled and observed, the new job was not cancelled, completed, and remained registered. Cancellation checks raised correctly in the page, transcription, summary, heavy/light transcription, and slide-turn stages. Paired Browser lecture search also completed successfully.

vivienxr:
Too many merge conflicts. New PR for same issue: https://github.com/ls1intum/edutelligence/pull/715

PR #567, opened 2026-05-21
`Athena`: Implement med api for text and modeling llm module

maximiliansoelch:
### Steps for Testing

## Summary by CodeRabbit

* **New Features**
  * Added submission evaluation API (POST /evaluate) and a health/status endpoint with standardized request/response schemas, capability reporting, and payload translation to/from downstream modules.
  * Published new public models for evaluate requests/responses and translation helpers.

* **Bug Fixes / Behavior Changes**
  * Authentication now resolves LMS identity from the Authorization header; missing/invalid secrets return 401/403.
  * Improved error mapping and propagation, stricter proxy validation (e.g., reject GET with body), always inject resolved LMS server URL into proxied requests, and enhanced connection/error logging.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ls1intum/edutelligence/pull/567?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


athena/assessment_module_manager/assessment_module_manager/authenticate.py
```diff
@@ -1,27 +1,90 @@
...
+            "" if configured_secret is None else str(configured_secret),
+            provided_secret_str,
+        )
+    ]
+
+    if len(matching_lms_urls) == 1:
+        logger.debug("Resolved LMS URL from Authorization header: %s", matching_lms_urls[0])
+        return matching_lms_urls[0]
+
+    if len(matching_lms_urls) > 1:
```
Claudia-Anthropica, line 51: @maximiliansoelch [high] This makes the bundled non-production config reject every authenticated request. `env.py` assigns the same fallback secret (`abcdef12345`) to both `local` and `playground` when env vars are missing, so resolving by secret alone always hits `len(matching_lms_urls) > 1` and returns 401 in the default dev setup. Either make the default secrets unique or keep a validated way to disambiguate the LMS URL when secrets collide.

In `athena/assessment_module_manager/assessment_module_manager/authenticate.py`, resolving the LMS only from `Authorization` makes the default non-production deployments ambiguous because `env.py` assigns the same fallback secret to every deployment. Fix this by making default deployment secrets unique or by adding a safe disambiguation path that still validates the secret against the selected LMS. Ensure the bundled local config can authenticate with its documented defaults.

</details>

athena/assessment_module_manager/assessment_module_manager/endpoints/evaluate_endpoint.py
```diff
@@ -0,0 +1,549 @@
...
+
+    logger.info(
+        "Dispatching /evaluate request request_id=%s to module=%s type=%s graded=%s lms_url=%s",
+        x_request_id,
+        module.name,
+        module.type.value,
+        bool(feedback_request["isGraded"]),
+        resolved_lms_url,
+    )
+    module_response = await request_to_module(
```
Claudia-Anthropica: @maximiliansoelch [medium] `request_to_module()` can raise `HTTPException` when the target module is unavailable, so this path bypasses `_map_module_error()` and `_error_response()`. A down module returns FastAPI's default `{'detail': ...}` body without the MED error shape or echoed request/version headers, even though this endpoint declares `ErrorResponse` for 503. Catch the exception around this call and map it to the same 503 response format.

In `athena/assessment_module_manager/assessment_module_manager/endpoints/evaluate_endpoint.py`, `/evaluate` does not catch `HTTPException` raised by `request_to_module()` when a module cannot be reached. Wrap the module call in a try/except and return `_error_response(status_code=exc.status_code, ...)` with the existing MED headers and module details.

</details>

Claudia-Anthropica (changes requested):
@maximiliansoelch The new MED evaluate surface is mostly wired cleanly, but I found two runtime issues that should be fixed before merging. The auth resolver breaks the default local deployment secrets, and `/evaluate` still leaks FastAPI's default error body on module connection failures; details inline.

athena/assessment_module_manager/assessment_module_manager/authenticate.py
Claudia-Anthropica: @maximiliansoelch Looks good now, the per-deployment dev fallback secrets remove the default ambiguity. Thanks.

athena/assessment_module_manager/assessment_module_manager/endpoints/evaluate_endpoint.py
Claudia-Anthropica: @maximiliansoelch Looks good now, the dispatch exception is mapped back into the MED error shape.

Claudia-Anthropica (dismissed):
@maximiliansoelch All previous feedback is addressed, thanks. The current log_viewer image matrix job failed before logs were available, but the same Athena image build step passed in other matrix jobs on this SHA, so I am not blocking on it.

athena/assessment_module_manager/assessment_module_manager/endpoints/evaluate_endpoint.py
```diff
@@ -0,0 +1,575 @@
...
+    try:
+        resolved_lms_url = resolve_lms_url_from_secret(authorization)
+    except HTTPException as exc:
+        logger.warning(
+            "Rejected /evaluate request request_id=%s during authentication: %s",
+            x_request_id,
+            exc.detail,
+        )
+        return _error_response(
+            status_code=status.HTTP_403_FORBIDDEN,
```
Claudia-Anthropica: @maximiliansoelch [medium] `resolve_lms_url_from_secret()` raises 401 for missing or invalid `Authorization`, but `/evaluate` turns every auth failure into 403 here. That makes this endpoint disagree with the existing authenticated proxy and with clients that distinguish missing credentials from a forbidden request. Preserve `exc.status_code` and set the MED error title/code from that status.

In `athena/assessment_module_manager/assessment_module_manager/endpoints/evaluate_endpoint.py`, the `/evaluate` auth exception handler always returns 403 even when `resolve_lms_url_from_secret()` raised 401. Fix it by preserving `exc.status_code`, returning an `UNAUTHORIZED`/`Unauthorized` MED error for 401, and updating the documented responses if needed.

</details>

athena/assessment_module_manager/assessment_module_manager/med_api/translation.py
```diff
@@ -0,0 +1,368 @@
...
+
+    raise UnsupportedEvaluateRequestError(
+        f"µEd feedback conversion is currently not implemented for {artefact_type.value}.",
+    )
+
+
+def _build_text_request(evaluate_request: EvaluateRequest) -> dict[str, Any]:
+    exercise_id = _stable_int_id(
+        "exercise",
+        _first_non_empty(
```
Claudia-Anthropica: @maximiliansoelch [medium] When none of these optional identifiers are present, `_first_non_empty()` returns `""`, so every anonymous MED request hashes to the same Athena exercise id. That can mix unrelated tasks in module storage and, for modeling, collide with the one-structured-grading-cache-row-per-exercise constraint. Add a deterministic fallback from the task/submission context or reject requests without a task identifier before building the Athena payload.

In `athena/assessment_module_manager/assessment_module_manager/med_api/translation.py`, `_build_text_request()` and `_build_model_request()` derive `exercise_id` only from optional task IDs/title, so requests without those fields all hash the empty string. Fix this by adding a deterministic fallback based on task/submission context or by rejecting such requests with a MED error before translation, and apply the fix to both request builders.

</details>

Claudia-Anthropica (changes requested):
@maximiliansoelch The MED endpoint is close, and the current CI is green. I found two API/runtime edge cases that need fixing before merge: auth failures lose their 401 status, and requests without task identifiers can collide in Athena storage.

athena/assessment_module_manager/assessment_module_manager/med_api/translation.py
Claudia-Anthropica: @maximiliansoelch This is still present in the current head: both text and modeling builders still call _stable_int_id("exercise", _first_non_empty(...)) with only optional task identifiers, so the empty-source collision remains possible.

athena/assessment_module_manager/assessment_module_manager/endpoints/evaluate_endpoint.py
Claudia-Anthropica: @maximiliansoelch Still seeing this in the current head: the handler still returns 403 unconditionally at this catch block instead of preserving the 401 from resolve_lms_url_from_secret().

Claudia-Anthropica (changes requested):
@maximiliansoelch The two current change requests are still open on this head: /evaluate still flattens auth failures to 403, and anonymous MED tasks can still collide on the generated Athena exercise id. I replied on the existing threads with the exact lines.

athena/assessment_module_manager/assessment_module_manager/endpoints/evaluate_endpoint.py
Claudia-Anthropica: @maximiliansoelch Looks good now, this preserves the auth status and maps 401 correctly.

athena/assessment_module_manager/assessment_module_manager/med_api/translation.py
Claudia-Anthropica: @maximiliansoelch Nice fix, the fallback context prevents anonymous requests from collapsing onto the same exercise id.

Claudia-Anthropica (approved):
@maximiliansoelch All previous feedback is addressed, and CI is green. Approving this follow-up.

PR #581, opened 2026-05-29
`Logos`: Enhance Logos UI

alex7sz:
## Summary by CodeRabbit

* New Features
  * Connect/manage multiple models per provider; configure per-model endpoints and API keys; edit or disconnect connections
  * Searchable Model Picker for selecting models
  * Team and API-key provider permission management, including a new Providers tab in Team Management
  * API Key modal: toggle custom permissions with granular provider/model overrides
  * Interactive Billing with time-range presets and budget history charts (admin-only)
  * Models tab now respects provider permissions when listing available models
* Changes
  * Provider type options narrowed (removed Azure)
* Tests
  * Added unit tests for provider-permissions pruning operations

### DIFF
--- a/logos/db/migrations/034_merge_model_provider_tables.sql
+++ b/logos/db/migrations/034_merge_model_provider_tables.sql
@@ -1,0 +1,46 @@
+ALTER TABLE model_provider
+    ADD COLUMN IF NOT EXISTS api_key TEXT DEFAULT NULL,
+    ADD COLUMN IF NOT EXISTS endpoint TEXT DEFAULT NULL;
+
+DO $$
+BEGIN
+    IF EXISTS (
+        SELECT 1
+        FROM information_schema.tables
+        WHERE table_schema = 'public' AND table_name = 'model_api_keys'
+    ) THEN
+        EXECUTE '
+            INSERT INTO model_provider (model_id, provider_id)
+            SELECT mak.model_id, mak.provider_id
+            FROM model_api_keys mak
+            WHERE NOT EXISTS (
+                SELECT 1
+                FROM model_provider mp
+                WHERE mak.model_id = mp.model_id
+                AND mak.provider_id = mp.provider_id
+            );
+        ';
+
+        EXECUTE '
+            UPDATE model_provider mp
+            SET api_key = mak.api_key,
+                endpoint = mak.endpoint
+            FROM model_api_keys mak
+            WHERE mak.model_id = mp.model_id
+                AND mak.provider_id = mp.provider_id;
+        ';
+    END IF;
+END $$;
+
+DO $$
+BEGIN
+    IF NOT EXISTS (
+        SELECT 1 FROM pg_constraint
+        WHERE conname = 'uq_model_provider_mapping'
+    ) THEN
+        ALTER TABLE model_provider
+        ADD CONSTRAINT uq_model_provider_mapping UNIQUE(model_id, provider_id);
+    END IF;
+END $$;
+
+DROP TABLE IF EXISTS model_api_keys CASCADE;
--- a/logos/logos-ui/components/modals/api-key-modal.tsx
+++ b/logos/logos-ui/components/modals/api-key-modal.tsx
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from "react";
-import { View, Pressable, ScrollView } from "react-native";
+import { View, Pressable, ScrollView, Switch } from "react-native";
 import { BaseModal } from "./base-modal";
 import { Text } from "@/components/ui/text";
 import { VStack } from "@/components/ui/vstack";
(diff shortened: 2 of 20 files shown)


logos/db/migrations/034_merge_model_provider_tables.sql
```diff
@@ -0,0 +1,46 @@
...
+END $$;
+
+DO $$
+BEGIN
+    IF NOT EXISTS (
+        SELECT 1 FROM pg_constraint
+        WHERE conname = 'uq_model_provider_mapping'
+    ) THEN
+        ALTER TABLE model_provider
+        ADD CONSTRAINT uq_model_provider_mapping UNIQUE(model_id, provider_id);
```
Claudia-Anthropica, line 42: @alex7sz [high] Existing `model_provider` rows can still contain duplicate `(model_id, provider_id)` pairs because the old schema had no unique constraint and `connect_model_provider()` inserted blindly. This migration adds `uq_model_provider_mapping` without deleting duplicates first, so upgraded databases with repeated connections will fail here. Deduplicate `model_provider` before adding the constraint, keeping the row with populated `api_key`/`endpoint` or otherwise the lowest id.

In `logos/db/migrations/034_merge_model_provider_tables.sql`, the migration adds `uq_model_provider_mapping` but existing `model_provider` rows may already duplicate `(model_id, provider_id)`. Before the `ALTER TABLE ... ADD CONSTRAINT`, delete duplicate rows while preserving one row per pair, preferably the row with non-null credentials/endpoints or otherwise the lowest id.

</details>

logos/src/logos/dbutils/dbmanager.py
```diff
@@ -3791,6 +3824,30 @@ def get_model(self, model_id: int):
...
+                   ORDER BY m.name ASC
+                   """
+        )
+        rows = self.session.execute(sql, {"pid": int(provider_id)}).fetchall()
+        return [
+            {
+                "model_id": r.model_id,
+                "model_name": r.model_name,
+                "endpoint": r.endpoint or "",
+                "api_key": r.api_key or "",
```
Claudia-Anthropica, line 3869: @alex7sz [high] Returning `mp.api_key` from this read helper exposes the model-provider secret to the UI; `openEditConn()` seeds it into the edit form. Even after the missing auth check on `get_provider_models` is fixed, these credentials should stay write-only. Return a masked flag or empty value instead and only update the stored key when the user enters a replacement.

In `logos/src/logos/dbutils/dbmanager.py`, `get_connections_for_provider()` returns `mp.api_key` for every provider model connection. Stop returning the stored secret; return an empty string or a boolean/masked indicator, and update the provider connection edit flow to keep the API key field blank and only send `api_key` when a replacement is entered.

</details>

logos/logos-ui/components/modals/api-key-modal.tsx
```diff
@@ -242,19 +256,34 @@ export function ApiKeyModal({
...
+            }
+          );
         }
-      );
 
-      if (!permsRes.ok) {
-        const err = await permsRes.json().catch(() => null);
-        alert(err?.detail || "Failed to save model permissions.");
-        return;
+        await fetch(
```
Claudia-Anthropica, line 276: @alex7sz [medium] These permission updates are not checked at all. If either `PUT` returns 403/500, the modal still calls `onSaved()` and closes, so the user sees a successful save while custom provider/model permissions remain stale. Capture each response, check `ok`, show the backend error, and return before closing when a permission write fails.

In `logos/logos-ui/components/modals/api-key-modal.tsx`, `handleSave()` awaits the provider/model permission `fetch()` calls but never checks their responses. Store each response, check `.ok`, surface the parsed error/detail to the user, and do not call `onSaved()` or `onClose()` when a permission update fails.

</details>

Claudia-Anthropica (changes requested):
@alex7sz The UI direction looks useful, but I found a migration blocker and two permission/secret handling issues that need fixing before merge. I avoided repeating the existing CodeRabbit comments; the new items are inline.

logos/db/migrations/034_merge_model_provider_tables.sql
Claudia-Anthropica: @alex7sz Still unresolved: migration 034 still goes straight from the `model_api_keys` merge to adding `uq_model_provider_mapping`; there is no cleanup of duplicate `(model_id, provider_id)` rows before line 42. Upgraded databases with old duplicate rows can still fail here.

logos/src/logos/dbutils/dbmanager.py
Claudia-Anthropica: @alex7sz Still unresolved: `get_connections_for_provider()` still selects and returns `mp.api_key`, and `openEditConn()` still seeds `conn.api_key` into the edit field. The stored secret is still exposed to the UI instead of staying write-only.

logos/logos-ui/components/modals/api-key-modal.tsx
Claudia-Anthropica: @alex7sz Still unresolved: both permission `fetch()` calls are still awaited without checking `.ok`, so a rejected provider/model permission update still falls through to `onSaved()` and `onClose()`.

Claudia-Anthropica (changes requested):
@alex7sz The same three blockers from the previous review are still present: migration 034 can still fail on duplicate model-provider rows, model-provider API keys are still returned to the UI, and permission writes in the API-key modal still ignore failed responses. I replied on the existing threads with the current code references.

PR #588, opened 2026-06-03
`Athena`: Allow overriding base LLM model via env var

maximiliansoelch:
Read `LLM_DEFAULT_MODEL` before the configured `base_model` when materializing the LLM config.

### Steps for Testing

[](https://athena-test1.aet.cit.tum.de)
[](https://athena-test2.aet.cit.tum.de)

## Release Notes

* **New Features**
  * The default LLM model can now be configured using the `LLM_DEFAULT_MODEL` environment variable, allowing flexible model selection without modifying configuration files.


Claudia-Anthropica (approved):
@maximiliansoelch This looks good: the loader now gives `LLM_DEFAULT_MODEL` the intended precedence while preserving the YAML fallback. The only failing check is a `module_text_cofee` image push ending in GHCR 502, which looks unrelated to this change.

PR #591, opened 2026-06-04
`Logos`: Add GSM8K performance benchmarking suite

wasnertobias:
## Summary
- Adds a standalone GSM8K benchmarking suite under `logos/benchmarks/` that measures **TTFT** (Time to First Token), **TTLT** (Time to Last Token), and **GPU energy per request** against Logos / vLLM / Ollama backends.
- Includes workload preparation script (`prepare_benchmark.py`), the runner (`benchmark_logos.py`), and scenario config (`benchmark_config.py`).
- Workloads (`benchmarks/workloads/`) and per-run result artifacts (`benchmarks/benchmark_results/`) are gitignored — only tooling is tracked.

## More changes coming
This branch will collect additional improvements before merge, starting with an admin UI button to trigger one-shot calibration for all uncalibrated models on a specific node.

## Test plan
- [ ] Run `prepare_benchmark.py` to regenerate a workload CSV from GSM8K
- [ ] Run `benchmark_logos.py` end-to-end against a Logos test deployment with `--logos-key`
- [ ] Verify `run_meta.json`, detailed/summary CSVs, and charts are produced under `benchmark_results/<timestamp>_*/`
- [ ] Confirm `.gitignore` keeps workloads + results out of the tree

## Summary by CodeRabbit

* **New Features**
  * Full benchmarking suite: workload generation, runner CLI, analysis/plots, CSV outputs, and local/remote GPU energy measurement
  * Admin endpoint to batch-calibrate uncalibrated models and UI calibration action (with API-key gated button)
  * Node-wide "disable sleep mode" toggle to force sleep off across lanes

* **Bug Fixes**
  * Fixed crash when skipping KV-cache calibration path

* **Tests**
  * Added regression and endpoint tests for calibration and scheduler offline handling

* **Documentation**
  * Added comprehensive benchmark README

* **Chores**
  * Updated ignore rules, pre-commit hook config, and benchmark deps list

### DIFF
--- a/logos/benchmarks/Readme.md
+++ b/logos/benchmarks/Readme.md
@@ -1,0 +1,419 @@
+# Logos Benchmark
+
+Misst **TTFT** (Time to First Token), **TTLT** (Time to Last Token) und **GPU-Energieverbrauch** pro Request über drei Szenarien:
+
+| Szenario | Beschreibung |
+|---|---|
+| `logos-sleep` | Logos mit aktiviertem Sleep Mode (Kapazitätsplaner darf Modelle entladen) |
+| `logos-nosleep` | Logos ohne Sleep Mode (Modelle bleiben dauerhaft geladen) |
+| `ollama` | Direkt gegen Ollama ohne Logos-Layer |
+
+Jedes Szenario wird in zwei Konfigurationen getestet:
+- **2-LLM-Config**: Requests verteilen sich auf 2 Modelle
+- **5-LLM-Config**: Requests verteilen sich auf 5 Modelle
+
+---
+
+## Architektur
+
+```
+┌──────────────────────────────────────────┐
+│  Logos-Host  (benchmark läuft hier)      │
+│  ┌──────────────┐   ┌──────────────────┐ │
+│  │ benchmark_   │   │   Logos-Server   │ │
+│  │ logos.py     │   └────────┬─────────┘ │
+│  └──────────────┘            │ HTTP      │
+└─────────────────────────────┼────────────┘
+                               │ vLLM-API
+          ┌────────────────────┼────────────────────┐
+          │  GPU-Node A        │       GPU-Node B   │
+          │  ┌──────────┐  ┌──┴──────┐ ┌─────────┐ │
+          │  │NVML-Poll │  │  vLLM   │ │  vLLM   │ │
+          │  │(via SSH) │  │RTX Ada  │ │RTX Ada  │ │
+          │  └──────────┘  └─────────┘ └─────────┘ │
+          └─────────────────────────────────────────┘
+```
+
+**Wichtig:** Die GPU-Energie wird auf den **GPU-Nodes** gemessen, nicht auf dem Logos-Host. Das Benchmark-Script verbindet sich per SSH zu den GPU-Nodes und startet dort einen NVML-Poller, der kontinuierlich Leistungsaufnahme und Energiezähler streamt.
+
+---
+
+## Voraussetzungen
+
+### Logos-Host (wo das Benchmark-Script läuft)
+
+```bash
+# Alte pynvml entfernen falls vorhanden
+pip uninstall pynvml -y
+
+pip install httpx paramiko numpy matplotlib
+```
+
+### GPU-Nodes (einmalig mit Root-Zugriff einrichten)
+
+```bash
+pip install nvidia-ml-py
+```
+
+Der Remote-Poller (`_REMOTE_POLLER` im Script eingebettet) nutzt dann `nvidia-ml-py` auf den Nodes. Falls `nvidia-ml-py` nicht installiert ist, fällt der Poller automatisch auf `nvidia-smi` zurück — allerdings ohne Hardware-Energiezähler (nur Power-Polling).
+
+### Vorbereitung der Workloads
+
+```bash
+pip install datasets  # nur für prepare_benchmark.py benötigt
+```
+
+---
+
+## Konfiguration
+
+### `benchmark_config.py`
+
+Zentrales Config-File für beide LLM-Konfigurationen und die Ollama-Namens-Übersetzung:
+
+```python
+# Zwei-LLM-Konfiguration
+MODELS_2 = [
+    "Qwen3-30B-A3B",
+    "Llama-3.3-70B",
+]
+
+# Fünf-LLM-Konfiguration
+MODELS_5 = [
+    "Qwen3-30B-A3B",
+    "Llama-3.3-70B",
+    "Gemma3-4B",
+    "microsoft/Phi-4-reasoning",
+    "Gemma4-26b",
+]
+
+# Übersetzung Logos-Modellname → Ollama-Tag
+OLLAMA_MODEL_MAP = {
+    "Qwen3-30B-A3B":             "qwen3:30b-a3b",
+    ...
+}
+```
+
+**Vor dem ersten Lauf die Ollama-Tags überprüfen:**
+```bash
+# Auf dem Ollama-Server:
+ollama list
+```
+
+---
+
+## Schritt 1 — Workloads vorbereiten
+
+`prepare_benchmark.py` lädt den **openai/gsm8k**-Datensatz von HuggingFace herunter und erzeugt pro LLM-Konfiguration eine Workload-CSV-Datei. Die Modelle werden **gleichmäßig per Round-Robin** über alle Requests verteilt.
+
+```bash
+python prepare_benchmark.py [OPTIONEN]
+```
+
+### Optionen
+
+| Option | Standard | Beschreibung |
+|---|---|---|
+| `--split` | `test` | GSM8K-Split: `test` (1 319 Fragen) oder `train` (7 473 Fragen) |
+| `--num-samples N` | alle | Nur die ersten N Beispiele verwenden |
+| `--rps RATE` | `1.0` | Ankunftsrate in Requests/Sekunde (bestimmt `arrival_offset`) |
+| `--max-tokens N` | `512` | `max_tokens` je Request |
+| `--output-dir DIR` | `workloads/` | Zielverzeichnis |
+
+### Beispiele
+
+```bash
+# 100 Test-Fragen, 0.5 req/s Ankunftsrate
+python prepare_benchmark.py --num-samples 100 --rps 0.5
+
+# Alle Test-Fragen, kein Timing (immer mit --sequential verwenden)
+python prepare_benchmark.py --rps 0
+```
+
+### Ausgabe
+
+```
+workloads/
+├── workload_gsm8k_2llm.csv    ← 2-Modell-Konfiguration
+└── workload_gsm8k_5llm.csv    ← 5-Modell-Konfiguration
+```
+
+Jede Zeile enthält:
+
+| Spalte | Bedeutung |
+|---|---|
+| `request_id` | Eindeutige ID (z. B. `gsm8k-0001`) |
+| `arrival_offset` | Geplante Ankunftszeit in ms ab Benchmark-Start |
+| `mode` | `interactive` (Logos-Scheduling-Parameter) |
+| `priority` | `mid` (Logos-Scheduling-Parameter) |
+| `body_json` | Vollständiger OpenAI-Chat-Payload als JSON-String |
+| `question` | Original-GSM8K-Frage (Referenz/Auswertung) |
+| `answer` | Original-GSM8K-Antwort inkl. Lösungsweg (Referenz/Auswertung) |
+| `model_assigned` | Welches Modell diesem Request zugewiesen wurde |
+
+---
+
+## Schritt 2 — Benchmark ausführen
+
+```bash
+python benchmark_logos.py --scenario SZENARIO [OPTIONEN] \
+    --workload PFAD/ZUR/workload.csv
+```
+
+### Optionen
+
+| Option | Standard | Beschreibung |
+|---|---|---|
+| `--scenario` | `logos-sleep` | Szenario: `logos-sleep`, `logos-nosleep`, `ollama` |
+| `--workload CSV` | — | Workload-CSV von `prepare_benchmark.py` |
+| `--logos-url URL` | `http://localhost:8080` | Ziel-URL (Logos oder Ollama) |
+| `--logos-key KEY` | — | Logos API-Key (erforderlich für `logos-*`-Szenarien) |
+| `--sequential` | aus | **Sequentieller Modus** (siehe unten) |
+| `--max-concurrent N` | `64` | Max. parallele Requests (ohne `--sequential`) |
+| `--request-timeout-s S` | `600` | HTTP-Timeout pro Request |
+| `--output-dir DIR` | `benchmark_results/` | Ausgabeverzeichnis |
+
+### GPU-Energie-Optionen
+
+| Option | Standard | Beschreibung |
+|---|---|---|
+| `--gpu-host HOST [HOST...]` | — | Hostnames der GPU-Nodes (SSH). Typischer Fall: Logos-Szenarien |
+| `--gpu-ssh-user USER` | aktueller User | SSH-Benutzername auf den GPU-Nodes |
+| `--gpu-ssh-key PATH` | — | Pfad zum SSH-Private-Key |
+| `--gpu-ssh-port PORT` | `22` | SSH-Port |
+| `--gpu-device-index IDX` | `0` | NVML-GPU-Index auf jedem Remote-Node |
+| `--gpu-indices IDX [IDX...]` | — | Lokale GPU-Indices (nur wenn GPU auf diesem Server liegt) |
+| `--poll-interval-ms MS` | `100` | GPU-Polling-Intervall (ms) |
+
+`--gpu-host` und `--gpu-indices` schließen sich gegenseitig aus.
+
+---
+
+## `--sequential` — was es bewirkt
+
+### Ohne `--sequential` (Concurrent-Modus, Standard)
+
+Der Benchmark liest die `arrival_offset`-Spalte aus der CSV und sendet Requests zu den darin hinterlegten Zeitpunkten. Wenn mehrere Requests gleichzeitig aktiv sind, kann die GPU parallel für sie arbeiten — genau wie in realistischer Produktionslast.
+
+```
+t=0s    Request 1 → [===== vLLM berechnet =====]
+t=1s    Request 2 →      [===== vLLM berechnet ======]
+t=2s    Request 3 →           [===== vLLM berechnet =====]
+                  ↑ Requests überlappen sich auf der GPU
+```
+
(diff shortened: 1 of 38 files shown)


logos/benchmarks/benchmark_logos.py
```diff
@@ -0,0 +1,1495 @@
...
+        if not _PARAMIKO:
+            print("  [gpu] paramiko not installed — remote GPU measurement disabled.")
+            print("        Install with: pip install paramiko")
+            return
+
+        self._mono_start = time.monotonic()
+        self._wall_start = time.time()
+
+        # Base64-encode the poller script to avoid all shell quoting issues
+        script = _REMOTE_POLLER.format(gpu_index=self._gpu_index, interval=self._poll_s)
```
Claudia-Anthropica: @wasnertobias [high] This will crash before any SSH connection is attempted. `_REMOTE_POLLER` is formatted with `str.format()`, but the embedded script still contains f-string placeholders like `{GPU_INDEX}` and `{t:.6f}`, so `format()` tries to resolve those names and raises `KeyError`. Escape the embedded braces or replace only explicit sentinel placeholders.

In `logos/benchmarks/benchmark_logos.py`, `RemoteGPUTracker.start()` formats `_REMOTE_POLLER` with `str.format()` while the embedded poller contains unescaped f-string braces. Fix this by escaping every brace that belongs to the remote script's own f-strings, or by replacing `{gpu_index}` and `{interval}` with a safer placeholder mechanism that leaves the embedded f-strings intact.

</details>

logos/benchmarks/benchmark_logos.py
```diff
@@ -0,0 +1,1495 @@
...
+    def energy_from_samples(self, t_start: float, t_end: float) -> Optional[float]:
+        """Trapezoidal integration of total power across all hosts → Joules."""
+        # Build a merged (local_monotonic_t, total_mw) series
+        combined: list[tuple[float, float]] = []
+        for idx, (samples, lock) in enumerate(zip(self._all_samples, self._locks)):
+            with lock:
+                snap = list(samples)
+            for s in snap:
+                lt = self._to_local_mono(s.remote_wall_t, idx)
+                combined.append((lt, s.power_mw))
```
Claudia-Anthropica: @wasnertobias [medium] In the polling fallback this does not sum power across remote hosts. It appends each host's samples into one time series and integrates adjacent points, so two 100 W hosts sampled at the same second still integrate as roughly 100 W instead of 200 W. Build a per-host integral and sum those Joules, or construct an actual total-power time series before integrating and charting.

In `logos/benchmarks/benchmark_logos.py`, `RemoteGPUTracker.energy_from_samples()` merges per-host power samples without aggregating simultaneous host power, which undercounts energy whenever multiple remote hosts are measured in polling mode. Fix this by integrating each host's samples over the request window separately and summing the resulting Joules, or by constructing an actual total-power time series before integration.

</details>

logos/benchmarks/Readme.md
```diff
@@ -0,0 +1,419 @@
...
+# Logos-Server vorher mit aktiviertem Sleep Mode konfigurieren/starten
+
+python benchmark_logos.py \
+    --scenario logos-sleep \
+    --logos-url http://logos.ase.cit.tum.de \
+    --logos-key YOUR_KEY \
+    --workload workloads/workload_gsm8k_2llm.csv \
+    --gpu-host gpu-node-a gpu-node-b \
+    --gpu-ssh-user ubuntu --gpu-ssh-key ~/.ssh/id_rsa \
+    --sequential --output-dir results
```
Claudia-Anthropica, line 253: @wasnertobias [low] These examples write to `results`, but `.gitignore` only ignores `benchmarks/benchmark_results/`. Following the README from `logos/benchmarks` will leave generated CSV/JSON/PNG outputs as untracked files, which is exactly what the PR summary says should be avoided. Either use the default `benchmark_results` in the examples or ignore `benchmarks/results/` too.

In `logos/benchmarks/Readme.md`, the example commands use `--output-dir results` while `logos/.gitignore` only ignores `benchmarks/benchmark_results/`. Fix this by changing the examples/output section to use `benchmark_results`, or by adding the documented `results` path to `logos/.gitignore`.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The benchmark suite is a useful addition, but the remote GPU measurement path needs fixes before this can be relied on. The remote poller currently crashes during setup, and the polling fallback undercounts multi-host energy; I also left a small docs/gitignore mismatch inline.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias This is still unresolved. `energy_from_samples()` still merges all host samples into one curve at lines 562-585, so multi-host polling still integrates individual host power points instead of summed host power.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still seeing this in current head: `_REMOTE_POLLER.format(...)` still hits `KeyError: 'GPU_INDEX'` because the embedded f-strings are unescaped. This still needs the sentinel replacement or brace escaping fix.

logos/benchmarks/Readme.md
Claudia-Anthropica: @wasnertobias This is still present too. The README examples still use `--output-dir results`, but the ignored benchmark output path is `benchmarks/benchmark_results/`.

Claudia-Anthropica (changes requested):
@wasnertobias The previous benchmark feedback is still open: the remote poller still crashes on `_REMOTE_POLLER.format(...)`, polling-mode multi-host energy is still undercounted, and the README output path mismatch remains. CI also fails because Black reformats `logos/src/logos/main.py`, so this still needs another pass before approval.

Claudia-Anthropica (changes requested):
@wasnertobias The same blockers are still present: the remote poller still fails during `_REMOTE_POLLER.format(...)`, multi-host polling still integrates individual host samples instead of total power, and the README examples still write to an unignored `results` directory. CI is also still failing because Black reformats `logos/src/logos/main.py`, so I’m keeping this as changes requested.

Claudia-Anthropica (changes requested):
@wasnertobias The previous benchmark blockers are still open on this SHA: the remote poller formatting crash, multi-host polling energy undercount, and README output path mismatch remain. Keeping this as changes requested.

Claudia-Anthropica (changes requested):
@wasnertobias The new commit only updates root pre-commit delegation, so the three existing benchmark blockers are still open: `_REMOTE_POLLER.format(...)` still crashes, multi-host polling energy is still integrated incorrectly, and the README still writes to `results`. Keeping this as changes requested.

Claudia-Anthropica (changes requested):
@wasnertobias I checked the current SHA and the unresolved benchmark threads from my previous review still apply, so I’m keeping this as changes requested.

Claudia-Anthropica (changes requested):
@wasnertobias I checked this SHA and the unresolved benchmark threads from my previous review still apply: the remote poller formatting crash, the multi-host polling energy integration bug, and the README output path mismatch are all still present. Keeping this as changes requested.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still not fixed on current head: `_REMOTE_POLLER.format(...)` still sees `{GPU_INDEX}` and `{t:.6f}` from the embedded f-strings as format fields, so it raises before any SSH connection. The poller needs escaped braces or sentinel replacement.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still unresolved on current head: `energy_from_samples()` still merges every host sample into one curve and integrates adjacent individual points. That does not represent summed cluster power for multiple hosts; please integrate per host and sum the resulting Joules, or build a real total-power series before integrating.

logos/benchmarks/Readme.md
Claudia-Anthropica: @wasnertobias This README mismatch is still present too: the examples and output tree still use `--output-dir results`, but `.gitignore` ignores `benchmarks/benchmark_results/` instead. Following the docs from `logos/benchmarks` will still leave generated benchmark artifacts untracked.

Claudia-Anthropica (changes requested):
@wasnertobias The three earlier benchmark blockers are still present on this SHA: the remote poller string formatting crash, the multi-host polling energy integration bug, and the README output path mismatch. Keeping this as changes requested.

logos/logos-workernode/logos_worker_node/runtime.py
```diff
@@ -176,4 +176,5 @@ async def build_runtime_status(app: FastAPI) -> WorkerRuntimeStatus:
         model_profiles=model_profiles if model_profiles else None,
         max_lanes=cfg.worker.max_lanes,
         gpu_performance_score=cfg.worker.gpu_performance_score,
+        sleep_mode_disabled=cfg.engines.vllm.disable_sleep_mode,
```
Claudia-Anthropica, line 235: @wasnertobias [medium] This new field breaks the worker-node CI path where `build_runtime_status()` is exercised with a minimal config object: `cfg.engines` is missing, so the status builder raises `AttributeError` before returning the degraded Ollama runtime status. Either make this read fall back to `False` when `engines.vllm` is absent, or update the runtime test helper to provide an `AppConfig`-compatible `engines.vllm.disable_sleep_mode` value.

In `logos/logos-workernode/logos_worker_node/runtime.py`, the new `sleep_mode_disabled=cfg.engines.vllm.disable_sleep_mode` access breaks worker-node tests when `app.state.config` has no `engines` attribute. Fix this by safely deriving the flag with a `False` fallback when `engines` or `engines.vllm` is missing, or by updating `logos/logos-workernode/tests/test_runtime.py` so its `_make_app` helper provides an AppConfig-compatible engines object. Ensure the worker-node test suite passes.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias CI is currently red in the worker-node tests because `build_runtime_status()` now assumes `cfg.engines` exists. The three earlier benchmark blockers are also still unresolved in the existing threads, so this still needs changes before approval.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Looks good now, thanks.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still open in the current SSH tracker: `energy_from_samples()` collects each host samples into one `combined` list and integrates adjacent points, so hosts sampled around the same time are not summed into cluster power. This still needs per-host integration and summing, or an actual total-power series.

logos/benchmarks/Readme.md
Claudia-Anthropica: @wasnertobias Still open: the examples still pass `--output-dir results` while only `benchmarks/benchmark_results/` is ignored. Please align the documented output directory with the ignored path, or ignore the documented path too.

logos/logos-workernode/logos_worker_node/runtime.py
Claudia-Anthropica: @wasnertobias Still open, and CI confirms it: `build_runtime_status()` still reads `cfg.engines.vllm.disable_sleep_mode` directly, so the worker-node runtime test fails when its minimal config has no `engines`.

Claudia-Anthropica (changes requested):
@wasnertobias The remote poller crash is fixed, so I resolved that thread. The multi-host energy aggregation, README output path mismatch, and runtime-status config access are still open; CI is also red from the worker-node runtime test and pre-commit on `logos/benchmarks/benchmark_logos.py`.

logos/benchmarks/Readme.md
Claudia-Anthropica: @wasnertobias Still open on `1a10000`: the README examples still use `--output-dir results`, while `.gitignore` only ignores `benchmarks/benchmark_results/`.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still open on `1a10000`: `energy_from_samples()` at lines 390-403 still concatenates all host samples and integrates adjacent points, so multi-host polling still does not sum per-host power/Joules.

logos/logos-workernode/logos_worker_node/runtime.py
Claudia-Anthropica: @wasnertobias Still open on `1a10000`: CI still fails because `build_runtime_status()` reads `cfg.engines.vllm.disable_sleep_mode` directly when the worker-node runtime test config has no `engines`.

Claudia-Anthropica (changes requested):
@wasnertobias The previous blockers are still open on this head: remote multi-host energy is still integrated from concatenated host samples, the README output dir still does not match the ignored path, and worker-node CI still fails on the runtime status config access. Pre-commit is also red on `logos/benchmarks/benchmark_logos.py` formatting/F541, so this still needs another pass.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still open on `7a9815d`: `energy_from_samples()` still builds one combined list from all host samples and integrates adjacent points. That still treats interleaved host readings as one GPU curve instead of summing per-host Joules or total cluster power.

logos/benchmarks/Readme.md
Claudia-Anthropica: @wasnertobias Still open on `7a9815d`: the examples still pass `--output-dir results`, but `.gitignore` only ignores `benchmarks/benchmark_results/`. Following the README still leaves generated benchmark artifacts untracked.

logos/logos-workernode/logos_worker_node/runtime.py
Claudia-Anthropica: @wasnertobias Still open on `7a9815d`: `build_runtime_status()` still reads `cfg.engines.vllm.disable_sleep_mode` directly, so the minimal worker-node runtime test config without `engines` can still raise `AttributeError`.

Claudia-Anthropica (changes requested):
@wasnertobias Three previous blockers are still open on this head: multi-host energy integration, the README output directory mismatch, and the runtime-status config access. I replied on the existing threads, so I’m keeping this as changes requested.

Claudia-Anthropica (changes requested):
@wasnertobias I’m keeping this as changes requested because there are still unresolved current-head review threads covering functional issues, and I don’t want to duplicate those inline. CI is also still partly pending, so this head is not ready for approval yet.

logos/src/logos/main.py
```diff
@@ -3364,6 +3371,129 @@ async def logosnode_reconfigure_lane(data: LogosNodeReconfigureLaneRequest):
...
+    if not candidates:
+        candidates = _logosnode_facade.get_worker_capabilities(provider_id)
+    try:
+        profiles = _logosnode_facade.get_model_profiles(provider_id)
+    except Exception:
+        profiles = {}
+    uncalibrated: list[str] = []
+    for model_name in candidates:
+        profile = profiles.get(model_name)
+        if (
```
Claudia-Anthropica, line 3395: @wasnertobias [medium] This helper no longer actually mirrors `CalibrationOrchestrator._find_uncalibrated_model()`: it does not skip profiles marked `calibration_unsupported`, and it does not treat `sleep_mode_disabled` profiles as having sleep fields intentionally unavailable. The admin button can therefore report and enqueue models the nightly orchestrator correctly avoids, so users see calibration started for models the worker will only reject. Apply the same skip and sleep-NA logic here.

In `logos/src/logos/main.py`, `_find_uncalibrated_models_on_provider()` omits the `calibration_unsupported` and `sleep_mode_disabled` checks from `CalibrationOrchestrator._find_uncalibrated_model()`. Fix it by applying the same profile skip and sleep-field exemption logic before appending a model.

</details>

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -641,6 +741,23 @@ def _run_calibration() -> None:
...
+                    # When calibrate_model identified a permanent
+                    # model-identity-level failure (bad repo id, gated
+                    # repo, unsupported architecture), persist the flag in
+                    # the profile so the master's orchestrator stops
+                    # scheduling this model. Also persisted is the reason
+                    # code, for ops visibility. Recoverable failures
+                    # (kv-cache too high, transient I/O, etc.) carry no
+                    # ``unsupported_reason`` and remain eligible for
+                    # retry on the next maintenance window.
+                    if getattr(result, "unsupported_reason", None):
```
Claudia-Anthropica, line 940: @wasnertobias [medium] `calibrate_model()` can now return `node_unhealthy_reason`, but this bridge drops it on the floor. A node-level failure detected from the vLLM log tail is logged as a generic calibration failure, so the master never gets a health/profile signal to stop scheduling more calibrations for that degraded node. Wire this result into the node-unhealthy path before returning.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, `_handle_start_calibration()` ignores `CalibrationResult.node_unhealthy_reason` after `calibrate_with_tp_escalation()` returns. Fix this by checking that field in the failure branch and surfacing a node-unhealthy result to the master or runtime status so the orchestrator stops scheduling that worker until recovery.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The branch is in much better shape, but I found two more calibration-flow mismatches inline. Both are about making the manual/admin path and worker failure handling follow the same signals as the orchestrator, so they do not keep scheduling calibrations that cannot succeed.

logos/logos-workernode/logos_worker_node/calibration.py
```diff
@@ -1522,6 +1984,150 @@ def _try_calibrate(
...
+    if not result.success and not _fatal and tp > original_tp:
+        logger.info(
+            "  %s failed at max tp=%d — falling back to configured tp=%d",
+            model_name,
+            tp,
+            original_tp,
+        )
+        tp = original_tp
+        current_plan = {**plan, "tensor_parallel_size": tp}
+        result = _try_calibrate(current_plan, **cal_kwargs)
```
Claudia-Anthropica, line 2091: @wasnertobias [medium] The trust-remote-code retry only runs before the fallback to the configured TP. If the max-TP probe fails for a tensor-parallel compatibility reason and the configured-TP probe is the one that reports `trust_remote_code=True`/`contains custom code`, this returns the failure instead of retrying with `--trust-remote-code`. Apply the same retry helper after the fallback probe too.

In `logos/logos-workernode/logos_worker_node/calibration.py`, `calibrate_with_tp_escalation()` only retries with `--trust-remote-code` after the initial max-TP probe, so the configured-TP fallback can fail permanently when that attempt is the first one to emit the custom-code error. Fix this by applying the same trust-remote-code retry logic to each failed calibration probe, including the configured-TP fallback, and add a regression test for max-TP failure followed by configured-TP custom-code retry.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias I found one more calibration edge case in the trust-remote-code fallback path; see inline. Lint and tests are green now, with Docker image builds still pending.

Claudia-Anthropica (changes requested):
@wasnertobias Keeping this as changes requested because there are still unresolved current-head review threads, and several Docker build checks are still pending. I did not add new inline comments to avoid duplicating feedback that is already on the PR.

PR #593, opened 2026-06-05
`Logos`: Reboot host automatically when the GPU enters an unrecoverable wedge

wasnertobias:
## Summary

- New `logos_worker_node.gpu_watchdog`: monitors `node_health` for `gpu-*` failures and calls `reboot(2)` directly via `libc` when the wedge persists for 3 consecutive ticks. Closes the loop on RTX 6000 Ada / Quadro RTX 5000 GSP RPC failures where `nvidia-smi` flips Pwr/Fan to `ERR!` and every subsequent CUDA context allocation returns `cudaErrorDevicesUnavailable` — until now the only fix was a manual operator reboot, leaving nodes "up but useless" for hours.
- `node_health._check_gpu` extended to also flag `ERR!` / `[Error]` on `power.draw`, `fan.speed`, and `temperature.gpu` (the hochbruegge signature where memory was still readable).
- Container gains `CAP_SYS_BOOT` (for `reboot(2)`) and `CAP_SYSLOG` (for `dmesg` capture in the diagnostics dump) in both prod and dev compose.

## Why

- **deimama 2026-06-05**: GPU 1 wedged with `NV_ERR_RESET_REQUIRED` in `dmesg`; every `add_lane` failed with `cudaErrorDevicesUnavailable`. Node was "healthy" from Logos's view but couldn't run inference.
- **hochbruegge 2026-06-05**: `nvidia-smi` showed `ERR!` on Pwr/Fan with memory still readable — the existing memory-only sensor missed it entirely.

## Safety rails

- 5-min startup grace (no reboot in first 5 min after container start — avoids boot-loop).
- 30-min rate limit, persisted via `.gpu_watchdog_last_reboot` mtime in the state dir. **Survives the reboot itself**, so if the GPU comes back wedged immediately, the watchdog logs LOUDLY but doesn't reboot again — that's an operator-needed condition.
- 3 consecutive unhealthy ticks required (~3 min at default poll interval); a single noisy probe never reboots.
- Only `gpu-*` reason codes trigger the watchdog. Storage / Ceph EIO does not — rebooting would just amplify the outage.
- Diagnostics dump written to `gpu_watchdog_diagnostics/<utc-timestamp>.log` before triggering: `nvidia-smi`, `dmesg | grep -iE 'nvidia|nvrm|xid'`, full `node_health` snapshot.
- Missing `CAP_SYS_BOOT` (EPERM) is logged but doesn't propagate out of the poll loop.

## Branch note

This branch (`logos/benchmarking`) already contained 14 unrelated commits ahead of `main` from prior `Logos` work (calibration handover, RAM cache, blacklist hardening, …). Per the user's instruction the watchdog commit was added on top rather than rebasing onto a fresh branch.

## Test plan

- [x] `pytest logos-workernode` — 326 passed (8 new tests + updates to test_node_health to match the wider CSV)
- [x] `pre-commit run` clean on all changed files
- [ ] Deploy to a worker, simulate a wedge (`fuser -k /dev/nvidia*` won't quite reproduce, but on a real wedged host the watchdog should now self-heal). Verify diagnostics dump lands in `data/gpu_watchdog_diagnostics/`.
- [ ] Verify `CAP_SYS_BOOT` is actually granted on production hosts (run `capsh --print` inside the container after deploy).

## Summary by CodeRabbit

* **New Features**
  * GPU watchdog for automatic host reboot on unhealthy GPU conditions
  * Node health monitoring with GPU and storage sensors
  * KV cache envelope discovery and calibration for memory optimization

* **Chores**
  * Updated vLLM dependency to version 0.22.1

* **Tests**
  * Added comprehensive test suites for GPU watchdog, node health evaluation, KV cache envelope selection, and calibration persistence

### DIFF
--- a/logos/benchmarks/benchmark_logos.py
+++ b/logos/benchmarks/benchmark_logos.py
@@ -36,10 +36,12 @@ Supports three scenarios via --scenario:
                  tags. No logos_key header is sent.
 
 Requirements (on this machine):
-    pip install httpx paramiko numpy matplotlib
+    pip install httpx numpy matplotlib
+    # SSH uses the system 'ssh' binary — no extra library needed
 
-Requirements (on each GPU node, installed once with root access):
-    pip install nvidia-ml-py
+Requirements (on each GPU node):
+    nvidia-smi is used automatically (always available on NVIDIA systems).
+    Optionally: pip install nvidia-ml-py  (enables hardware energy counter)
 
 Usage — remote GPU nodes (typical setup):
     python benchmark_logos.py \\
@@ -69,24 +71,25 @@ from __future__ import annotations
 
 import argparse
 import asyncio
-import base64
 import csv
 import importlib.util
 import json
 import math
+import shlex
+import subprocess
 import sys
 import threading
 import time
-import warnings as _warnings
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Optional
+from typing import List, Optional
 
 import httpx
 
 # ── Optional deps ─────────────────────────────────────────────────────────
 
+import warnings as _warnings
 
 try:
     with _warnings.catch_warnings():
@@ -99,88 +102,20 @@ try:
 except ImportError:
     _NVML = False
 
-try:
-    import paramiko as _paramiko
-
-    _PARAMIKO = True
-except ImportError:
-    _PARAMIKO = False
 
 try:
-    import matplotlib
     import numpy as np
-
+    import matplotlib
     matplotlib.use("Agg")
     import matplotlib.pyplot as plt
-
     _PLOT = True
 except ImportError:
     _PLOT = False
 
 
-# ── Remote poller script (embedded, sent to GPU nodes via SSH) ────────────
-#
-# Runs on each GPU node. Continuously streams one line per poll interval:
-#   <unix_timestamp> <power_mw> <energy_mj>
-# energy_mj = -1 when the hardware counter is unavailable (falls back to
-# power-integration only). Uses nvidia-smi as a last resort if pynvml is
-# not installed on the node.
-
-_REMOTE_POLLER = r"""
-import sys, time, warnings, subprocess
-warnings.filterwarnings("ignore", category=FutureWarning)
-
-GPU_INDEX = {gpu_index}
-INTERVAL  = {interval:.3f}
-
-use_nvml   = False
-has_energy = False
-
-try:
-    import pynvml
-    pynvml.nvmlInit()
-    _h = pynvml.nvmlDeviceGetHandleByIndex(GPU_INDEX)
-    use_nvml = True
-    try:
-        pynvml.nvmlDeviceGetTotalEnergyConsumption(_h)
-        has_energy = True
-    except Exception:
-        pass
-except Exception:
-    pass
-
-def _read_nvml():
-    try:
-        p = pynvml.nvmlDeviceGetPowerUsage(_h)
-        e = pynvml.nvmlDeviceGetTotalEnergyConsumption(_h) if has_energy else -1
-        return p, e
-    except Exception:
-        return 0, -1
-
-def _read_smi():
-    try:
-        out = subprocess.check_output(
-            ["nvidia-smi", "--query-gpu=power.draw",
-             "--format=csv,noheader,nounits", f"--id={GPU_INDEX}"],
-            text=True, timeout=2.0,
-        )
-        return int(float(out.strip()) * 1000), -1
-    except Exception:
-        return 0, -1
-
-_read = _read_nvml if use_nvml else _read_smi
-
-while True:
-    t = time.time()
-    p_mw, e_mj = _read()
-    sys.stdout.write(f"{t:.6f} {p_mw} {e_mj}\n")
-    sys.stdout.flush()
-    time.sleep(INTERVAL)
-"""
 
 
 # ── Load benchmark_config (optional sibling file) ─────────────────────────
-
 
 def _load_config_attr(attr: str, default):
     config_path = Path(__file__).parent / "benchmark_config.py"
@@ -196,7 +131,6 @@ def _load_config_attr(attr: str, default):
 
 
 # ── Local GPU Tracker (NVML on this machine) ──────────────────────────────
-
 
 class GPUTracker:
     """
@@ -267,7 +201,10 @@ class GPUTracker:
     def _poll_loop(self) -> None:
         while not self._stop.is_set():
             t = time.monotonic()
-            mw = sum(_pynvml.nvmlDeviceGetPowerUsage(h) for h in self._handles if True)  # errors silently ignored below
+            mw = sum(
+                _pynvml.nvmlDeviceGetPowerUsage(h) for h in self._handles
+                if True  # errors silently ignored below
+            )
             with self._lock:
                 self._samples.append((t, mw))
             time.sleep(self._poll_s)
@@ -304,7 +241,7 @@ class GPUTracker:
         window = [(t, p) for t, p in samples if t_start <= t <= t_end]
         if len(window) < 2:
             before = [s for s in samples if s[0] < t_start]
-            after = [s for s in samples if s[0] > t_end]
+            after  = [s for s in samples if s[0] > t_end]
             if before and after and not window:
                 avg_mw = (before[-1][1] + after[0][1]) / 2.0
                 return avg_mw / 1000.0 * (t_end - t_start)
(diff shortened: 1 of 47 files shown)


logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -528,51 +558,224 @@ async def _execute_command(self, action: str, params: dict[str, Any]) -> dict[st
...
+            from logos_worker_node.node_health import evaluate_node_health  # noqa: PLC0415
+
+            _health = evaluate_node_health()
+            if not _health.healthy:
+                logger.error(
+                    "[Calibration] refusing start_calibration_session: node unhealthy (reason=%s) — %s",
+                    _health.reason_code,
+                    _health.reason_detail,
+                )
+                return {
```
Claudia-Anthropica, line 602: @wasnertobias [high] This refusal still goes back over the websocket as a successful command, because `_execute_command_and_respond()` sets `success: true` whenever the handler returns normally and `LogosNodeRuntimeRegistry.send_command()` ignores the inner `ok`. If this branch fires for an unhealthy node, the orchestrator leaves `_active_provider_id` set and waits for a terminal event that will never come, while the admin endpoint also reports “session started”. Raise an exception for refused starts, or make both callers treat `{"ok": false}` as `LogosNodeCommandError`.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, refused `start_calibration_session` requests return `{"ok": false}` but the server command layer treats normal returns as successful commands. Fix this by raising a command exception from the worker refusal paths or by checking the returned `ok` flag in the orchestrator and admin endpoint and clearing/reporting failure accordingly.

</details>

logos/benchmarks/benchmark_logos.py
```diff
@@ -0,0 +1,1305 @@
...
+        return None  # nvidia-smi has no cumulative energy counter
+
+    def energy_from_counter(self, start_mj: float, end_mj: float) -> float:
+        return (end_mj - start_mj) / 1000.0
+
+    def energy_from_samples(self, t_start: float, t_end: float) -> Optional[float]:
+        combined: list[tuple[float, float]] = []
+        for samples, lock in zip(self._host_samples, self._locks):
+            with lock:
+                combined.extend((t, p) for t, p in samples if t_start <= t <= t_end)
```
Claudia-Anthropica, line 395: @wasnertobias [medium] This concatenates samples from different GPU hosts and integrates across alternating host timestamps. With two nodes, the trapezoids connect host A's power sample to host B's next timestamp, so per-request Joules are not the sum of node energy. Integrate each host's request window separately and sum the per-host Joules.

In `logos/benchmarks/benchmark_logos.py`, `SshGpuTracker.energy_from_samples()` concatenates power samples from all remote hosts before integration. Fix this by computing trapezoidal energy independently for each host's samples in the request window and returning the sum across hosts.

</details>

logos/logos-workernode/logos_worker_node/node_health.py
```diff
@@ -0,0 +1,291 @@
...
+    memory_fields = ("total", "used", "free")
+    telemetry_fields = ("power", "fan", "temp")
+    bad_gpus: list[str] = []
+    for line in raw.strip().splitlines():
+        parts = [p.strip() for p in line.split(",")]
+        if len(parts) < 7:
+            continue
+        idx = parts[0]
+        for name, value in zip(memory_fields, parts[1:4]):
+            if _is_gpu_error_token(value) or value.upper() == "N/A":
```
Claudia-Anthropica, line 156: @wasnertobias [medium] `nvidia-smi --query-gpu` commonly returns bracketed `[N/A]`, not just bare `N/A`. A wedged memory query that reports `[N/A]` slips through this check as healthy, so the watchdog never reaches the GPU failure threshold. Normalize unavailable tokens for memory fields to catch both `N/A` and `[N/A]`, and add a regression test.

In `logos/logos-workernode/logos_worker_node/node_health.py`, memory-field GPU health checks only treat bare `N/A` as unavailable. Fix this by normalizing unavailable tokens so both `N/A` and `[N/A]` mark memory fields unhealthy, and add a test covering bracketed `[N/A]` on memory fields.

</details>

logos/src/logos/main.py
```diff
@@ -3364,6 +3371,83 @@ async def logosnode_reconfigure_lane(data: LogosNodeReconfigureLaneRequest):
...
+    if not candidates:
+        candidates = _logosnode_facade.get_worker_capabilities(provider_id)
+    try:
+        profiles = _logosnode_facade.get_model_profiles(provider_id)
+    except Exception:
+        profiles = {}
+    uncalibrated: list[str] = []
+    for model_name in candidates:
+        profile = profiles.get(model_name)
+        if (
```
Claudia-Anthropica, line 3395: @wasnertobias [medium] This helper no longer matches the worker/orchestrator selection rules. Profiles marked `calibration_unsupported` or `sleep_mode_disabled` with base residency already measured are returned as “uncalibrated”, so the admin endpoint can start a no-op session and tell the UI it is calibrating models the worker will skip. Mirror `_provider_has_uncalibrated_models()` / `_list_uncalibrated_models()` here before returning candidates.

In `logos/src/logos/main.py`, `_find_uncalibrated_models_on_provider()` ignores `calibration_unsupported` and `sleep_mode_disabled` profile flags. Fix this helper so it uses the same skip and missing-field rules as the calibration orchestrator and worker-side `_list_uncalibrated_models()`.

</details>

logos/logos-workernode/logos_worker_node/model_cache.py
```diff
@@ -372,13 +384,161 @@ async def cache_models_by_priority(self, models: list[str]) -> dict[str, str]:
...
+        if self._caching_now == model_name:
+            return event
+        # Already queued: only act if we need to bump priority.
+        if model_name in self._cache_queue:
+            if priority:
+                self._cache_queue.remove(model_name)
+                self._cache_queue.appendleft(model_name)
+                logger.info("RAM cache: bumped %s to front of queue", model_name)
+            return event
+        # Fresh enqueue.
```
Claudia-Anthropica, line 458: @wasnertobias [medium] A model's completion event stays set after a previous background copy completes or fails. If the model is later evicted or re-enqueued after a failed attempt, `wait_for_cached()` gets the already-set event back and returns immediately instead of waiting for the new copy. Clear the event whenever an uncached model is freshly enqueued or moved back onto the queue.

In `logos/logos-workernode/logos_worker_node/model_cache.py`, `_enqueue()` reuses already-set completion events for uncached models. Fix this by clearing the model's completion event whenever the model is not cached and is newly enqueued or reprioritized for another cache attempt, and add a regression test for re-enqueue after failure or eviction.

</details>

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -627,53 +914,71 @@ def _run_calibration() -> None:
...
+                        try:
+                            lane_manager._mark_status_dirty()  # noqa: SLF001
+                        except Exception:  # noqa: BLE001
+                            logger.debug("[Calibration] _mark_status_dirty failed", exc_info=True)
                 else:
-                    logger.warning(
-                        "[Calibration] Failed: model=%s error=%s",
-                        model_name,
-                        result.error,
+                    logger.warning("[Calibration] Failed model=%s error=%s", model_name, result.error)
```
Claudia-Anthropica, line 939: @wasnertobias [medium] `CalibrationResult.node_unhealthy_reason` is dropped here. When calibration detects a node-level EIO/read-only failure after the initial health check, the session records a normal model failure and keeps walking the remaining models, and the server never gets a node-unhealthy signal from this result. Handle this field explicitly: emit a node-health/degraded event or status update, stop the session, and avoid continuing with more models on the broken node.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, failed calibration results with `node_unhealthy_reason` are handled like ordinary model failures. Fix this by detecting `result.node_unhealthy_reason`, surfacing the node-degraded state to the server, terminating the calibration session, and preventing the loop from calibrating additional models on the degraded node.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The watchdog direction makes sense, but a few pieces need tightening before this can land. The biggest one is that refused calibration starts are still treated as successful bridge commands, which can wedge the orchestrator's active-session state; there are also correctness issues in the benchmark energy accounting, GPU health parsing, calibration candidate selection, and RAM-cache waiting.

logos/logos-workernode/logos_worker_node/logos_bridge.py
Claudia-Anthropica: @wasnertobias Still seeing this one. `_handle_start_calibration_session()` still returns `{"ok": False}` for refused starts, and `send_command()` still only raises on the outer websocket `success`, so both orchestrator/admin paths can still treat the refusal as a started session.

logos/logos-workernode/logos_worker_node/node_health.py
Claudia-Anthropica: @wasnertobias This still needs the `[N/A]` normalization. The memory check is still `value.upper() == "N/A"`, so `[N/A]` falls through as healthy, and I do not see a regression test for bracketed memory N/A.

logos/logos-workernode/logos_worker_node/model_cache.py
Claudia-Anthropica: @wasnertobias Still open. `_enqueue()` still reuses the old completion event for a non-cached model and never clears it before a fresh queue attempt, so `wait_for_cached()` can return immediately after a previous failed/completed attempt.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still open. `SshGpuTracker.energy_from_samples()` still builds one combined list from all `_host_samples` and integrates across the sorted aggregate; it needs to integrate each host window separately and sum the results.

logos/src/logos/main.py
Claudia-Anthropica: @wasnertobias Still open. `_find_uncalibrated_models_on_provider()` still ignores `calibration_unsupported` and still requires the sleep fields unconditionally, so it can disagree with the orchestrator/worker selection rules.

logos/logos-workernode/logos_worker_node/logos_bridge.py
Claudia-Anthropica: @wasnertobias Still open. Failed calibration results still go through the generic `calibration_model_failed` path without checking `result.node_unhealthy_reason`, so a node-level failure can still be treated like a per-model failure and the loop can keep going.

Claudia-Anthropica (changes requested):
@wasnertobias I re-checked the previous change-request threads and they are still open on this head. The main blockers are unchanged: refused calibration starts still look successful to the server, benchmark GPU energy is still aggregated across hosts incorrectly, GPU `[N/A]` memory values still pass as healthy, and the calibration/model-cache selection paths still disagree with the worker behavior.

logos/logos-workernode/logos_worker_node/node_health.py
Claudia-Anthropica: @wasnertobias Still open on this head. The memory check still compares `value.upper() == "N/A"`, so bracketed `[N/A]` memory values still bypass the watchdog; please normalize that token and add the regression test.

logos/logos-workernode/logos_worker_node/model_cache.py
Claudia-Anthropica: @wasnertobias This file is no longer part of the current PR diff after the branch update, so this thread no longer blocks this PR. Resolving.

logos/logos-workernode/logos_worker_node/logos_bridge.py
Claudia-Anthropica: @wasnertobias This file is no longer part of the current PR diff after the branch update, so this thread no longer blocks this PR. Resolving.

logos/src/logos/main.py
Claudia-Anthropica: @wasnertobias This file is no longer part of the current PR diff after the branch update, so this thread no longer blocks this PR. Resolving.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias This file is no longer part of the current PR diff after the branch update, so this thread no longer blocks this PR. Resolving.

logos/logos-workernode/logos_worker_node/logos_bridge.py
Claudia-Anthropica: @wasnertobias This file is no longer part of the current PR diff after the branch update, so this thread no longer blocks this PR. Resolving.

Claudia-Anthropica (changes requested):
@wasnertobias Most older blocker threads are no longer part of this PR diff, so I resolved those. The current watchdog change still misses bracketed `[N/A]` memory values in `node_health`, which means that wedge signature can still bypass the reboot watchdog. Requesting changes until that thread is fixed.

logos/logos-workernode/logos_worker_node/node_health.py
Claudia-Anthropica: @wasnertobias Still open on the current head. Memory fields still use `value.upper() == "N/A"`, so bracketed `[N/A]` is treated as healthy, and the tests still do not cover bracketed memory N/A.

Claudia-Anthropica (changes requested):
@wasnertobias The previous blocker is still open: bracketed `[N/A]` memory values still pass the GPU health check as healthy, so the watchdog can miss that wedge signature. Requesting changes until the existing node_health thread is fixed.

PR #594, opened 2026-06-06
`Logos`: true request count per model in charts breakdown (bugfix)

JakubJakubczyk:
## Logos: fix duplicated models in per-model stats
  
  **Problem:** Models showed up doubled/multiplied in the "By Model" volume chart and pie. The breakdown query grouped by `(model_id, provider)`, so a model served by N
  providers came back as N rows with the same name, and the UI (keyed by model name, stacked bars) summed them.

  **Fix:** Group the breakdown by model only, at the SQL stage. Dropped the unused `providerName` from the payload and the two TS types.

  ### Changes (3 files, +7/−8)
  - `dbmanager.py` — `GROUP BY re.model_id, model_name` (removed provider from SELECT/GROUP BY and the `providers` join); dropped `providerName` from the result dict.
  - `types.ts` / `plotly-request-volume-chart.web.tsx` — removed the now-dead `providerName` field.

  ### Notes
  - No frontend logic change needed: one row per model makes the names unique on their own.
  - `providerName` was verified unused by any consumer.
  - Aggregates (counts/averages) recompute correctly across providers.

## Summary by CodeRabbit

* **New Features**
  * Added VRAM provider selector dropdown on the Statistics screen
  * Added VRAM remaining column to the Request volume section

* **Bug Fixes**
  * Fixed transient empty charts after clearing zoom
  * Improved chart zoom and pan behavior with automatic range reset

* **Style**
  * Restructured Statistics page layout for improved organization
  * Updated footer credits with additional contributor


logos/logos-ui/app/statistics.tsx
```diff
@@ -2066,29 +2083,42 @@ export default function Statistics() {
...
-        value: stats.totals.cloudRequests,
-        color: CHART_PALETTE.cloud,
-        text: "Cloud",
-      },
-      {
-        value: stats.totals.localRequests,
-        color: CHART_PALETTE.local,
-        text: "Local",
-      },
+      { value: cloudSum, color: CHART_PALETTE.cloud, text: "Cloud" },
```
Claudia-Anthropica, line 2092: @JakubJakubczyk [medium] The slices now come from `cloudLineData`/`localLineData`, but the Request type center label still renders `cloudPct` from `stats.totals`. Right after a zoom, and if the range refresh fails, the donut can show windowed slices with a full-period percentage in the middle. Derive that percentage from the same `cloudSum + localSum` values.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `logos/logos-ui/app/statistics.tsx`, `providerPieData` uses the visible cloud/local series while the Request type chart center still renders `cloudPct` from `stats.totals`. Fix this by deriving a windowed cloud percentage from the same `cloudSum` and `localSum` values used for `providerPieData`, then use that value in the Request type `centerText`.\n\n</details>

Claudia-Anthropica (changes requested):
@JakubJakubczyk The per-model SQL aggregation looks good, but the request type pie can still show inconsistent range data after zooming. Please fix the center percentage issue inline.

PR #595, opened 2026-06-06
`Logos`: Dedupe per-model stats breakdown at SQL stage (bugfix)

JakubJakubczyk:
### Problem
The statistics dashboard's per-model request breakdown showed **duplicated / doubled entries** for the same model. The breakdown query grouped by `(model, provider)`, so
any model served by more than one provider produced one row per provider. The stats UI keys rows by model **name**, so those rows rendered as duplicate/stacked entries
with split counts.

### Fix
Aggregate to a **true per-model** breakdown at the SQL stage — group by model only, not by provider.

  - `dbmanager.py`: drop `provider_id` / `provider_name` from the breakdown query's `SELECT`, `GROUP BY`, and result mapping; now `GROUP BY re.model_id, model_name`. A
  single model served by N providers collapses into one row with the summed count. Added a comment explaining provider is an orthogonal axis surfaced separately via the
  VRAM/lane views.
  - `types.ts` + `plotly-request-volume-chart.web.tsx`: remove the now-unused `providerName` field from `modelBreakdown`.

### Impact
 - One row per model in the per-model breakdown; counts are no longer split/doubled across providers.
 - Provider-level detail remains available via the VRAM/lane (scheduler signals) panels.

### DIFF
--- a/logos/logos-ui/components/statistics/plotly-request-volume-chart.web.tsx
+++ b/logos/logos-ui/components/statistics/plotly-request-volume-chart.web.tsx
@@ -9,7 +9,6 @@ import { useDarkMode } from "@/components/statistics/use-dark-mode";
 type ModelBreakdownItem = {
   modelId: number;
   modelName: string;
-  providerName: string;
   requestCount: number;
 };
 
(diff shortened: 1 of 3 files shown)


Claudia-Anthropica (approved):
@JakubJakubczyk Looks good, the backend aggregation now matches the UI's per-model breakdown contract and the removed provider field is cleaned up consistently. Approving as-is.

PR #596, opened 2026-06-07
`Logos`: Benchmark preperations

wasnertobias:
## Summary by CodeRabbit

* **New Features**
  * Optional benchmark warmup with CLI controls (enable/skip, timeout) and abort-on-warmup-failure.
  * Benchmarks updated to reference newer model identifiers.

* **Refactor**
  * Improved GPU energy tracking and clearer connected-host warnings.
  * Explicit HTTP error handling for request dispatching.

* **Bug Fix**
  * Calibration now records correct KV-cache minima, detects collapsed envelopes, and auto-retries using suggested model-length hints; workers re-announce newly calibrated models.

* **Tests**
  * Added regression and unit tests for calibration, envelope detection, profile persistence, and process command building.


logos/benchmarks/benchmark_logos.py
```diff
@@ -1198,30 +1250,27 @@ async def _async_main(args: argparse.Namespace) -> None:
...
+    if not args.skip_warmup:
+        await _warmup(
+            args.logos_url, args.logos_key, workload,
+            args.scenario, model_map,
+            timeout_s=args.warmup_timeout,
+        )
+
     # ── Run ───────────────────────────────────────────────────────────────
-    print("\nRunning...")
+    print(f"\nRunning...")
```
Claudia-Anthropica: @wasnertobias [medium] CI is red here: the Logos pre-commit job reformats this file with autoflake/isort/black, then flake8 still fails on the two new bare f-strings (`print(f"  [warmup] Done.")` and this line). Run the Logos pre-commit hooks and replace those with plain string prints so the required check can pass.

In `logos/benchmarks/benchmark_logos.py`, CI pre-commit fails because the file is not formatted and contains two f-strings without placeholders. Run `pre-commit run --config logos/.pre-commit-config.yaml --all-files` or apply the equivalent autoflake/isort/black changes, and replace `print(f"  [warmup] Done.")` and `print(f"\nRunning...")` with plain string prints.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The benchmark warmup direction looks reasonable, but CI is currently red because Logos pre-commit reformats `benchmark_logos.py` and flake8 still finds two bare f-strings. Please run the hooks and fix that before merge.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still seeing this on the current head: `benchmark_logos.py` still has the bare f-strings at lines 785 and 1262, and the Logos pre-commit job is still failing because the hooks want to reformat this file. Please run the Logos pre-commit hooks and replace those two literal f-strings before this can merge.

Claudia-Anthropica (changes requested):
@wasnertobias The previous CI/pre-commit feedback is still open on this head: `benchmark_logos.py` still needs the Logos pre-commit formatting changes and the two bare f-strings replaced. Keeping changes requested until that required check is green.

Claudia-Anthropica (changes requested):
@wasnertobias Keeping changes requested on this head. The required Logos pre-commit check is still red because the hooks rewrite `benchmark_logos.py`, and the two literal f-strings at 785/1262 are still present. The current calibration review thread also remains unresolved, so please address that before merge.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Looks good now, thanks.

Claudia-Anthropica (changes requested):
@wasnertobias The CI/pre-commit issue is fixed now, but I’m keeping changes requested because the calibration issue CodeRabbit flagged is still present. The current code still records `_KV_CACHE_MIN_STEP_MB` as `min_kv_observed_mb` even after the floor probe failed, so the planner can still use a KV size that never succeeded.

PR #598, opened 2026-06-07
`Iris`: Add context awareness

vivienxr:
FIXES IRIS-38

This PR needs https://github.com/ls1intum/Artemis/pull/12879 to be applied in Artemis. 

## Summary

This PR implements context-aware lecture content retrieval in Iris. When students send messages from the combined view, Iris receives information about their current position in the lecture materials (PDF page and/or video timestamp) and prioritizes this content in its responses, ensuring highly relevant and contextual assistance.

## Description

When a student sends a message in LECTURE mode, Artemis can include an optional context array in the request payload containing structured context objects. These are defined in src/iris/domain/data/lecture_context_dto.py:
- SlidesContextDTO — lecture_unit_id + page for slide viewers
- VideoContextDTO — lecture_unit_id + timestamp (seconds) for viewers of transcribed videos
- CombinedViewContextDTO — indicates the message was sent from the combined/fullscreen view; wrapps an optional SlidesContextDTO and/or VideoContextDTO. Exposes a lecture_unit_id property that derives the unit id from slides (falling back to video).
- LectureContextDTO — a Union of the three, so future context types can be added easily.

Workflow (all in src/iris/pipeline/chat/chat_pipeline.py)

1. Flatten positions — _build_current_view (line 527) calls _collect_context_positions (line 473), which flattens standalone slides/video entries and the slides/video nested inside a combinedView entry into two lists: context_pages ({lecture_unit_id, page}) and context_timestamps ({lecture_unit_id, timestamp}).
2. Fetch the viewed content from Weaviate — _build_current_view calls LectureRetrieval.fetch_context_content (src/iris/retrieval/lecture/lecture_retrieval.py), which does a direct, non-RAG lookup:
    - _fetch_page_chunks_by_page — filters page chunks by course_id + lecture_unit_id + page_number (+ base_url).
    - _fetch_transcriptions_by_timestamp — filters transcription segments where segment_start_time <= timestamp < segment_end_time (+ course_id, lecture_unit_id, base_url).
    - If a viewed position has no ingested material, it's simply skipped — Iris can't see it.
3. Group multi-chunk pages — in _build_current_view, page chunks are grouped by (lecture_unit_id, page_number) so all chunks of one slide page are bundled into a single block. Each position renders one block: a position description (page/timestamp + lecture unit name and id) followed by the corresponding content.
4. Add combined-view info to the prompt — build_system_message (line 372) computes current_view_is_combined and passes it to the template. src/iris/pipeline/prompts/templates/chat_system_prompt.j2 renders the blocks and, for combined view, notes that all is shown side by side.
5. Retrieval stays independent — the viewed content is stored under a dedicated key state.lecture_content_storage["current_view"], kept separate from the lecture retrieval tool's "content". The tool (provide_lecture_retrieval in src/iris/tools/chat_tool_providers.py) is unchanged in behavior. It now just reuses the per-request state.lecture_retriever (also cached via _get_lecture_retriever, line 512) instead of instantiating a second LectureRetrieval (and its models).
6. Merge for citations — in _add_citations (line 634), _merge_lecture_content (line 68) merges current_view with the tool's retrieved content, de-duplicating items by uuid (_dedup_by_uuid, line 57) so the same paragraph isn't cited twice. This means answers about the current position get lecture citations even when the agent never calls the retrieval tool.

Tests:
- tests/test_lecture_context_dto.py — DTO parsing/validation, including the combinedView lecture_unit_id derivation.
- tests/test_lecture_retrieval_context_content.py — fetch_context_content and the page/timestamp filter lookups.

## Testing

See Artemis PR: https://github.com/ls1intum/Artemis/pull/12879

## Summary by CodeRabbit

* **New Features**
  * System recognizes and accepts the user's current position in videos and slide decks to contextualize replies.
  * Chat messages include a "Current Position" block showing lecture unit, slide page, and video timestamp when available.
  * Retrieval prioritizes and boosts content matching current slide pages or video timestamps, returning context-first results.
  * Viewing context is pre-parsed and forwarded so retrieval tools automatically scope results to the user's position.
  * Execution payloads may include an optional viewing-context list to enable this behavior.


iris/src/iris/pipeline/chat/chat_pipeline.py
```diff
@@ -339,6 +343,13 @@ def build_system_message(
         """
         dto = state.dto
 
+        # Parse lecture context from messages if present
+        # (Artemis only sends context blocks in LECTURE_CHAT mode)
+        lecture_context = self._parse_lecture_context(dto)
+
+        # Store context in state for later use (e.g., by tools)
+        state.lecture_context = lecture_context
```
Claudia-Anthropica: @vivienxr [medium] `_parse_lecture_context()` strips the messages only after `prepare_state()` and the early `user_query` capture have already read the latest user text, so the retrieval tool and Memiris memory creation can still receive the raw `[context:...]` block. Make the latest-user-text path return the context-stripped text, or move parsing/cleanup before those cached query values are computed.

In `iris/src/iris/pipeline/chat/chat_pipeline.py`, context blocks are removed after `state.query_text` and the early `user_query` value have already been computed. Fix this by stripping context blocks before the latest user text is cached, for example by overriding `get_text_of_latest_user_message()` in `ChatPipeline` to return `remove_context_block(...)` for text messages, while still keeping `state.lecture_context` populated for retrieval.

</details>

iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2
```diff
@@ -68,6 +68,20 @@ You are currently helping with the lecture "{{ lecture_name }}".
...
 
+{% if lecture_context %}
+# Current Position
+{% if lecture_unit_name %}
+The student is currently viewing the lecture unit "{{ lecture_unit_name }}".
+{% endif %}
+{% if lecture_context.page %}
+The student is currently viewing page {{ lecture_context.page }} of the lecture slides.
+{% endif %}
+{% if lecture_context.timestamp %}
```
Claudia-Anthropica: @vivienxr [low] A timestamp of `0` is valid at the start of a video, but this `if` treats it as false and omits the current-position line from the prompt. Use an explicit `is not none` check so zero-valued context still renders.

In `iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2`, the timestamp condition treats `0` as false, so the start of a video is omitted from the current-position prompt. Change the Jinja condition to `lecture_context.timestamp is not none`; apply the same explicit `is not none` style to `lecture_context.page` if zero can be valid.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The context-aware retrieval path is close, but the raw context marker can still leak into the retrieval query because it is cached before cleanup. I’m requesting changes for that; the timestamp prompt edge case is minor but worth fixing at the same time.

iris/src/iris/pipeline/chat/chat_pipeline.py
Claudia-Anthropica: @vivienxr This is still present in the current HEAD. `prepare_state()` still sets `state.query_text` from the raw latest user message before `_parse_lecture_context()` mutates the chat history in `build_system_message()`, so retrieval and MCQ detection can still see the `[context:...]` marker.

iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2
Claudia-Anthropica: @vivienxr This one is also still unchanged. `lecture_context.timestamp` is checked as a truthy value, so a valid timestamp of `0.0` will not be rendered in the prompt.

Claudia-Anthropica (dismissed):
@vivienxr The two open follow-up items are still unresolved on this SHA: the context marker can still be cached into `state.query_text`, and timestamp `0.0` is still skipped in the prompt. Keeping this as changes requested; details are in the existing threads.

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

vivienxr:
@coderabbitai review

iris/src/iris/retrieval/lecture/lecture_retrieval.py
```diff
@@ -200,12 +211,251 @@ def __call__(
...
+        transcription_filter &= Filter.by_property(
+            LectureTranscriptionSchema.LECTURE_UNIT_ID.value
+        ).equal(lecture_unit_id)
+        if base_url is not None:
+            transcription_filter &= Filter.by_property(
+                LectureTranscriptionSchema.BASE_URL.value
+            ).equal(base_url)
+
+        lecture_transcriptions = (
+            self.lecture_transcription_collection.query.fetch_objects(
```
Claudia-Anthropica, line 329: @vivienxr [high] This fallback fetches only Weaviate's default page of transcription objects, then filters by timestamp in Python. On a normal video with more than the default number of chunks, any timestamp outside that first batch will never match, so the current-position transcript will not be boosted. Fetch the matching timestamp range directly or page/fetch with an explicit limit before filtering.

In `iris/src/iris/retrieval/lecture/lecture_retrieval.py`, `_fetch_transcriptions_by_lecture_unit()` fetches transcription objects without an explicit limit and `_boost_transcriptions_by_timestamps()` filters that partial result by timestamp. Fix this by querying for the timestamp range directly, or by using explicit pagination/limit handling so all candidate transcript chunks for the lecture unit are available before filtering. Add a test covering a timestamp that is not in the first fetched batch.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The structured DTO path and current-position prompt look good, and CI is green. One remaining video-context issue needs fixing: timestamp fallback can miss later transcript chunks because it filters after a default-limited fetch.

iris/src/iris/pipeline/chat/chat_pipeline.py
Claudia-Anthropica: @vivienxr Looks good now, this moved to structured DTO context so there is no raw context marker on the query path anymore.

iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2
Claudia-Anthropica: @vivienxr Looks good now, the template renders video contexts directly and `0.0` is no longer filtered out by a truthiness check.

iris/src/iris/retrieval/lecture/lecture_retrieval.py
Claudia-Anthropica: @vivienxr Nice fix, the fallback now queries by timestamp range directly and the test covers the late-segment case.

Claudia-Anthropica (dismissed):
@vivienxr All feedback addressed, nice work. CI is green, approving.

iris/src/iris/retrieval/lecture/lecture_retrieval.py
```diff
@@ -132,7 +132,9 @@ def __call__(
         lecture_unit_id: int = None,
         base_url: str = None,
     ) -> LectureRetrievalDTO:
-        lecture_unit = self.get_lecture_unit(course_id, lecture_id, lecture_unit_id)
+        lecture_unit = self.get_lecture_unit(
+            course_id, lecture_id, lecture_unit_id, base_url
```
Claudia-Anthropica: @vivienxr [high] Passing `base_url` into `get_lecture_unit()` makes the returned `LectureUnitRetrievalDTO.base_url` non-null, and the three downstream search methods currently replace their course/lecture filters with a `BASE_URL` filter whenever that field is set. Normal lecture retrieval can then search every lecture on the same Artemis instance instead of the selected course or lecture unit. Combine the base-url filter with the existing course/lecture/unit filters in the page chunk, transcription, and segment retrieval pipelines before enabling this parameter.

In `iris/src/iris/retrieval/lecture/lecture_retrieval.py`, passing `base_url` into `get_lecture_unit()` causes downstream retrieval to use `LectureUnitRetrievalDTO.base_url`, but the retrieval subpipelines overwrite their course/lecture filters with a base-url-only filter. Fix this by combining the base-url filter with the existing course/lecture/unit filters in `lecture_page_chunk_retrieval.py`, `lecture_transcription_retrieval.py`, and `lecture_unit_segment_retrieval.py`. Add a focused test with two lectures on the same base URL to prove retrieval stays scoped.

</details>

iris/src/iris/pipeline/chat/chat_pipeline.py
```diff
@@ -405,6 +416,128 @@ def is_memiris_memory_creation_enabled(
...
+        context_pages, context_timestamps = self._collect_context_positions(
+            lecture_contexts
+        )
+        if not context_pages and not context_timestamps:
+            return None
+
+        base_url = state.dto.settings.artemis_base_url if state.dto.settings else None
+        try:
+            retriever = self._get_lecture_retriever(state)
+            page_chunks, transcriptions = retriever.fetch_context_content(
```
Claudia-Anthropica: @vivienxr [medium] This fetch result only gets formatted into the system prompt, while `_add_citations()` later cites lecture content only from `state.lecture_content_storage["content"]`. Because the prompt explicitly tells the model it does not need to call the lecture retrieval tool, answers based on the current slide/video can now skip citations entirely. Store these fetched DTOs in the same citation storage, for example as a `LectureRetrievalDTO` with empty segments, before returning the formatted prompt content.

In `iris/src/iris/pipeline/chat/chat_pipeline.py`, `_build_current_view_content()` injects fetched page chunks/transcriptions into the prompt but never stores them in `state.lecture_content_storage`. Fix this by storing a `LectureRetrievalDTO` containing the fetched page chunks and transcriptions, with an empty `lecture_unit_segments` list, in `state.lecture_content_storage["content"]` before returning the formatted current-view text. Add a test that current-view content can be cited without invoking the retrieval tool.

</details>

iris/src/iris/domain/chat/chat_pipeline_execution_dto.py
```diff
@@ -34,7 +38,27 @@ class ChatPipelineExecutionDTO(PipelineExecutionDTO):
...
+        """Extract lectureUnitId from a combinedView context if present.
+
+        This populates the lecture_unit_id field from the combinedView context's
+        nested slides/video objects, which is used for RAG filtering to scope
+        retrieval to the specific lecture unit.
+        """
+        # Only extract if lecture_unit_id is not already set
+        if self.lecture_unit_id is None and self.context:
+            for ctx in self.context:
+                if isinstance(ctx, CombinedViewContextDTO):
```
Claudia-Anthropica, line 61: @vivienxr [medium] The DTO supports standalone `slides` and `video` context entries, but this validator only copies the unit id for `combinedView`. If Artemis sends one of the standalone context types without a top-level `lectureUnitId`, the retrieval tool is not scoped to the unit the student is viewing even though the context carries that id. Extract the first available `lecture_unit_id` from any context type, including nested combined-view objects.

In `iris/src/iris/domain/chat/chat_pipeline_execution_dto.py`, `extract_lecture_unit_id_from_combined_view_context()` ignores standalone `SlidesContextDTO` and `VideoContextDTO` entries even though they have `lecture_unit_id`. Fix the validator to populate `self.lecture_unit_id` from the first available context lecture unit id across standalone and combined-view contexts when the top-level field is unset. Add DTO tests for standalone slides and video contexts populating `lecture_unit_id`.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The structured context flow is much cleaner now, but I found a retrieval scoping regression plus two gaps in the new direct current-view path. Details are inline.

iris/src/iris/domain/chat/chat_pipeline_execution_dto.py
vivienxr: When not being in the combined view then several units can be viewed at the same time. So the unit id should only be set if its in the combined view.

iris/src/iris/retrieval/lecture/lecture_retrieval.py
Claudia-Anthropica: @vivienxr Looks good now. `__call__()` no longer passes `base_url` into `get_lecture_unit()`, so the normal lecture retrieval path keeps the existing course/lecture/unit scoping.

iris/src/iris/pipeline/chat/chat_pipeline.py
Claudia-Anthropica: @vivienxr Looks good now, the current-view chunks are stored in `lecture_content_storage` and the merge path preserves them when the tool runs.

iris/src/iris/domain/chat/chat_pipeline_execution_dto.py
Claudia-Anthropica: @vivienxr Good point, that makes sense for standalone contexts where multiple units can be open. I withdraw this one and will resolve the thread.

Claudia-Anthropica (dismissed):
@vivienxr All feedback addressed, nice work.

thynguyentumde (dismissed):
Tested locally with the Artemis PR. Worked as expected.

Nayer-kotry (dismissed):
tested locally, works as expected

iris/src/iris/retrieval/lecture/lecture_retrieval.py
```diff
@@ -206,6 +206,138 @@ def __call__(
...
+        page_chunk_filter &= Filter.by_property(
+            LectureUnitPageChunkSchema.PAGE_NUMBER.value
+        ).equal(page_number)
+        if base_url is not None:
+            page_chunk_filter &= Filter.by_property(
+                LectureUnitPageChunkSchema.BASE_URL.value
+            ).equal(base_url)
+
+        lecture_page_chunks = (
+            self.lecture_unit_page_chunk_collection.query.fetch_objects(
```
Claudia-Anthropica, line 288: @vivienxr [medium] `fetch_objects()` uses Weaviate's default limit here, but one PDF page is split into multiple 512-character chunks during ingestion. For text-heavy slides, this can return only the first batch, while `_build_current_view()` expects all chunks of that page for the prompt and citation storage. Add explicit pagination or a sufficiently high limit for this exact page lookup.

In `iris/src/iris/retrieval/lecture/lecture_retrieval.py`, `_fetch_page_chunks_by_page()` fetches page chunks without an explicit limit, so Weaviate can return only its default page even though ingestion splits one PDF page into multiple chunks. Fix this by adding explicit pagination or a sufficiently high limit to fetch all chunks for the filtered course, lecture unit, page, and base URL. Add a test with more chunks than the Weaviate default.

</details>

Claudia-Anthropica (dismissed):
@vivienxr The structured context path is in good shape, but the new exact slide lookup can still truncate current-page content because it relies on Weaviate’s default fetch limit. I left one inline comment; CI is green otherwise.

Claudia-Anthropica (approved):
@vivienxr The structured lecture-context flow looks good on this pass, and the previous current-view retrieval concerns are addressed. CI is green, approving.

DominikRemo (approved):
Tested in testing session on Artemis ts1. LGTM

laadvo (approved):
Tested on ts1 during the testing session, lgtm

bassner (approved):

PR #597, opened 2026-06-07
`Logos`: Spring Migration

alex7sz:

PR #603, opened 2026-06-11
`Logos`: Stop stats DB queries from freezing streaming responses

### LINKED ISSUE #601: `Logos`: Loading the statistic pages has influence on streaming LLM responses
This is a critical performance issue.

The loading of the statistic pages takes a few seconds, in this time I noticed that all streaming responses are NO LONGER forwarded! It is temporarily paused in fact.

JakubJakubczyk:
### Problem

  While the statistics page was loading (which can take a few seconds), **all in-flight streaming responses stopped delivering tokens** until the stats finished
  loading.  

  The cause: the orchestrator handles everything on a single thread. Database queries were executed directly on that thread, so one slow stats query paused chunk
  forwarding for every active stream. The same (smaller) hiccup happened on every request when usage logging was written to the database.

  ### Solution
  
  Blocking database calls are now executed in a background worker thread pool via a small `with_db()` helper, so the main thread stays free to forward stream
  chunks. Applied to:

  - the VRAM statistics endpoint (the heaviest remaining Python stats path)
  - the per-request logging writes in all streaming and non-streaming response paths (time-to-first-token is now written in the background so it doesn't delay
  the stream itself)
  - worker status snapshot writes

  A concurrency cap ensures dashboard queries can't hog the database connection pool, and TTFT timestamps are captured at chunk arrival so the metric stays
  accurate.

  ### Testing

  - New regression test (`test_event_loop_blocking.py`) measures whether the main loop stalls during a slow query — it fails on `main` (detects a ~1–1.5 s
  freeze) and passes with this change.
  - Full unit suite passes (388 tests).
  - Added two scripts under `tests/performance/` to measure stream smoothness against a live deployment (`stream_gap_probe.py` + a mock streaming backend).
  - ⚠️  Not verified end-to-end on the test or production environments due to current limitations (no GPU worker attached to the local dev stack). The probe
  scripts above are ready for that once a real deployment is available.

  Note: most of the other stats endpoints were already moved off the Python service by the Spring Migration (#597), which is why this PR only touches the paths
  that remained.

  Closes #601

## Release Notes

* **New Features**
  * Added performance monitoring tools for streaming response latency analysis.

* **Tests**
  * Added regression tests to detect event-loop blocking issues.
  * Added performance probe utilities for streaming behavior analysis.

* **Chores**
  * Enhanced database operations with asynchronous support to prevent blocking the request handler.
  * Updated project dependencies.

### DIFF
--- a/logos/logos-orchestrator/pyproject.toml
+++ b/logos/logos-orchestrator/pyproject.toml
@@ -30,6 +30,7 @@ sentence-transformers = "^5.0.0"
 matplotlib = "^3.9.2"
 aiohttp = "^3.9.0"
 prometheus-client = "^0.21.0"
+anyio = "^4.5"
 
 [tool.poetry]
 packages = [{ include = "logos", from = "src" }]
--- a/logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
+++ b/logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
@@ -11,6 +11,8 @@ import secrets
 import threading
 from typing import Any, Dict, List, Optional, Tuple, cast
 
+import anyio
+import anyio.to_thread
 import sqlalchemy.exc
 import yaml
 from dateutil.parser import isoparse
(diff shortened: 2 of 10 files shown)


wasnertobias:
no longer relevant as fixed with spring migration

PR #608, opened 2026-06-16
`Logos`: Benchmark per-node warmup pre-fetch and config hardening

wasnertobias:
## Summary

- **Per-node warmup pre-fetch**: Before each benchmark scenario, each GPU node is started in isolation so the Logos scheduler is forced to route all warmup requests there — guaranteeing every benchmark model is downloaded to that node's local cache before traffic starts.
- **Benchmark config patching**: Filters `capabilities_models` to benchmark-only models and disables the 400 GB tmpfs RAM cache (`TMPFS_SIZE=0`) for the duration of the run; config is restored from backup on completion or abort.
- **Sleep mode fix**: Rewrote `_set_logos_sleep_mode_via_ssh` to use pyyaml, injecting `enable_sleep_mode` into `model_overrides` for all models — the old sed-based approach silently skipped models with no pre-existing override entry.
- **Local model cache**: New `--benchmark-local-cache` arg redirects `OLLAMA_MODELS_MOUNT` to a local NVMe path; vLLM downloads models there on first run and reuses them on subsequent runs.
- **Warmup timeout**: Default raised from 1800s → 3600s to accommodate per-node cycling (`num_nodes × download_time`).

## Test plan

- [ ] Run `--run-all-scenarios` with two GPU nodes and verify pre-fetch log shows each node cycling in isolation
- [ ] Verify all 5 benchmark models appear on both nodes after warmup
- [ ] Verify `config.yml` and `.env` are restored after benchmark completes
- [ ] Verify `config.yml` and `.env` are restored on Ctrl-C abort
- [ ] Confirm models are not sleeping during `logos-nosleep` scenario

## Summary by CodeRabbit

* **New Features**
  * Added SSH relay routing for remote benchmark orchestration and Ollama/workernode control.
  * Enhanced energy/power tracking with per-source GPU vs wall metrics and expanded Shelly transport support (UDP/TCP/HTTP).
  * Improved worker start-up with background prefetch for missing capability models and more robust calibration reset/warmup flows.
  * Added vLLM `--max-num-seqs` calibration handling for hybrid models.
  * Introduced new benchmark CLI flags for relay execution, local caching, and calibration control.
* **Documentation**
  * Updated benchmark “Setup & Run” guide with full examples and clarified `results_detailed.csv` schema; added Shelly monitoring guidance/templates.
* **Tests**
  * Added coverage for `max-num-seqs` calibration retry/precedence behavior.
* **Chores**
  * Added environment-driven `run_bench.sh` and expanded benchmark `.gitignore`.


logos/benchmarks/benchmark_logos.py
```diff
@@ -2081,6 +2643,8 @@ def _open_ssh_tunnel(
         "-L",
         f"{local_port}:localhost:{remote_port}",
     ]
+    if relay_host:
+        parts += ["-J", f"{relay_user}@{relay_host}"]
```
Claudia-Anthropica, line 2647: @wasnertobias [high] `ProxyJump` still authenticates the final `logos-server@<gpu>` hop from the local machine, not from the relay. That contradicts the relay mode described above where the developer machine only has access to `--logos-ssh-host` and the relay owns the GPU-node key, so the remote commands work but the Ollama tunnel fails and the whole Ollama scenario gets skipped. Build this tunnel through a command running on the relay, or explicitly require and validate local GPU-node SSH credentials before using `-J`.

In `logos/benchmarks/benchmark_logos.py`, `_open_ssh_tunnel()` uses `ssh -J` for relay mode, which still requires the local machine to authenticate to the GPU node. Fix relay mode so the tunnel is established using credentials available on the relay host, or fail fast with a clear error when only local `ProxyJump` authentication is supported.

</details>

logos/benchmarks/benchmark_logos.py
```diff
@@ -1749,12 +2127,183 @@ def _stop_logos_workernodes_if_running_via_ssh(
...
+            )
+            if read_res.returncode != 0:
+                raise RuntimeError(f"  [config] {host}: Cannot read config.yml: {read_res.stderr.strip()}")
+
+            subprocess.run(
+                _build_ssh_cmd(
+                    host,
+                    ssh_user,
+                    ssh_key,
+                    f"{sudo}cp {shlex.quote(config_path)} {shlex.quote(config_bak)}",
```
Claudia-Anthropica, line 2184: @wasnertobias [medium] This overwrites `config.yml.benchmark_bak` every time the benchmark starts. If a previous run aborted after patching but before restore, the next run copies the already-filtered benchmark config over the only backup, so the original model list and cache settings are unrecoverable. Preserve an existing backup or restore it before creating a new one, and apply the same guard to `.env.benchmark_bak` below.

In `logos/benchmarks/benchmark_logos.py`, `_apply_benchmark_workernode_config_via_ssh()` unconditionally overwrites benchmark backup files. Fix this by preserving an existing `config.yml.benchmark_bak` and `.env.benchmark_bak`, or by restoring them before creating fresh backups, so an aborted previous run cannot lose the original configuration.

</details>

logos/benchmarks/benchmark_logos.py
```diff
@@ -2746,10 +3747,26 @@ async def _async_run_all(args: argparse.Namespace) -> None:
...
+    _benchmark_config_applied = [False]  # mutable flag so _cleanup can restore
+
+    # Wall power over HTTPS/443: a Traefik-routed ingest sidecar this run manages.
+    want_shelly_http = getattr(args, "shelly", False) and getattr(args, "shelly_transport", "udp") == "http"
+    args._shelly_ingest_file = None  # set when the sidecar starts; read by _benchmark_scenario
+
+    def _ensure_shelly_sidecar() -> None:
+        if not want_shelly_http or args._shelly_ingest_file:
+            return
+        args._shelly_ingest_file = _start_shelly_ingest_sidecar(
```
Claudia-Anthropica, line 3767: @wasnertobias [medium] In relay mode this still starts the Shelly HTTP ingest sidecar with local Docker, but the Traefik instance receiving `https://<logos>/shelly-ingest` is on `--logos-ssh-host`. From a developer machine this either disables wall-power collection because no local Traefik exists, or tails a local file that the Pi never writes to. Make the HTTP ingest path relay-aware, including how the benchmark process reads the relay-side NDJSON file, or reject `--shelly-transport http` when `--logos-ssh-host` is set.

In `logos/benchmarks/benchmark_logos.py`, `_ensure_shelly_sidecar()` starts the HTTP Shelly ingest sidecar locally even when `--logos-ssh-host` is used. Fix this by making the sidecar startup, cleanup, and NDJSON reading path work through the relay host, or by failing early with a clear unsupported-combination error for `--shelly-transport http` plus `--logos-ssh-host`.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The benchmark hardening is useful, but the new relay mode still has a couple of paths that do not actually run through the relay. The Ollama tunnel is the blocker, and the config backup logic also needs to avoid overwriting the only restore point.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still open on the current head: both benchmark backup copies are still unconditional `cp ... .benchmark_bak`, so a rerun after an aborted patch can overwrite the original restore point.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still open on the current head: `_open_ssh_tunnel()` still uses `ssh -J` from the local machine, so relay mode still depends on local GPU-node authentication.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still open on the current head: `_ensure_shelly_sidecar()` still starts the ingest sidecar locally when `--logos-ssh-host` is set, so the HTTP Shelly path is not relay-aware and the combination is not rejected either.

Claudia-Anthropica (changes requested):
@wasnertobias The new commit only touches worker calibration/profile code, but the benchmark relay/config feedback from the previous review is still open. I replied on those threads and am keeping this at requested changes until they are addressed.

PR #610, opened 2026-06-18
`Logos`: Benchmark improvements

wasnertobias:


.github/workflows/logos_deploy-prod.yml
```diff
@@ -29,17 +29,17 @@ jobs:
...
-      github.event.workflow_run.conclusion == 'success' }}
-    uses: ls1intum/.github/.github/workflows/deploy-docker-compose.yml@main
-    with:
-      environment: "Logos Worker - Test - hochbruegge"
-      docker-compose-file: "./logos/logos-workernode/docker-compose.yml"
-      main-image-name: ls1intum/edutelligence/logos-workernode-vllm
-      image-tag: ${{ inputs.image-tag || 'latest' }}
-      deployment-base-path: "/opt/logos-workernode"
-    secrets: inherit
+  #  deploy-worker-hochbruegge:
```
Claudia-Anthropica: @wasnertobias [medium] This comments out the entire production deploy job for `hochbruegge`, so future Logos prod deployments will stop updating that worker while `deioma` still deploys. That looks unrelated to the benchmark changes and leaves one prod node stale; either keep the job active with the corrected prod environment or move the intentional decommissioning into a separate PR with context.

In `.github/workflows/logos_deploy-prod.yml`, the `deploy-worker-hochbruegge` job is commented out, removing that worker from production deploys. Fix this by re-enabling the job with the intended production environment, or remove it only in a dedicated deployment change with an explicit replacement/decommissioning note.

</details>

logos/benchmarks/benchmark_logos.py
```diff
@@ -2081,6 +2682,8 @@ def _open_ssh_tunnel(
         "-L",
         f"{local_port}:localhost:{remote_port}",
     ]
+    if relay_host:
+        parts += ["-J", f"{relay_user}@{relay_host}"]
```
Claudia-Anthropica, line 2686: @wasnertobias [medium] `ProxyJump` does not use the relay's SSH key for the final GPU hop; OpenSSH still authenticates `logos-server@host` from the machine running the benchmark. That breaks the new `--logos-ssh-host` mode described above, where the developer machine only has access to the relay and the relay has the GPU-node key. Build the tunnel from the relay side, or fail clearly when that topology is unsupported.

In `logos/benchmarks/benchmark_logos.py`, `_open_ssh_tunnel` uses `ssh -J` when `relay_host` is set, which still requires the local client to authenticate to the GPU node. Fix relay mode by creating the Ollama port tunnel through the relay using the relay's access to the GPU node, or reject that mode with a clear error before starting the scenario.

</details>

logos/benchmarks/benchmark_logos.py
```diff
@@ -2713,6 +3365,388 @@ async def _run_all_traffic_patterns(
...
+            print(f"  [{i + 1}/{n}] {host}: pre-fetch done — stopping before next node ...")
+            _stop_workernode_via_ssh([host], ssh_user, ssh_key, workernode_dir, use_sudo, relay_host, relay_user)
+
+    # The last node is still running.  Start all remaining nodes so the benchmark
+    # has the full cluster.  Models are already cached — startup is fast.
+    if n > 1:
+        remaining = hosts[:-1]
+        print(f"\n[Warmup pre-fetch] Starting remaining nodes to restore full cluster: {remaining} ...")
+        _start_workernode_via_ssh(remaining, ssh_user, ssh_key, workernode_dir, use_sudo, relay_host, relay_user)
+        if not await _wait_for_logos(logos_url, timeout_s=120.0, logos_key=logos_key):
```
Claudia-Anthropica, line 3742: @wasnertobias [medium] This readiness check can return immediately because the last warmed node is still connected, so it does not prove that the `remaining` workers have rejoined before the benchmark starts. On multi-node runs the benchmark can start measuring with only the last node active. Wait for the expected worker/provider count or verify each host connected before returning success.

In `logos/benchmarks/benchmark_logos.py`, `_warmup_workernodes_sequentially` starts the remaining workers and then calls `_wait_for_logos`, which can pass using the already-connected last worker. Fix this by waiting for all expected workers/providers to reconnect, or by otherwise verifying every host before the benchmark traffic starts.

</details>

logos/benchmarks/benchmark_logos.py
```diff
@@ -2746,10 +3786,26 @@ async def _async_run_all(args: argparse.Namespace) -> None:
...
+    _benchmark_config_applied = [False]  # mutable flag so _cleanup can restore
+
+    # Wall power over HTTPS/443: a Traefik-routed ingest sidecar this run manages.
+    want_shelly_http = getattr(args, "shelly", False) and getattr(args, "shelly_transport", "udp") == "http"
+    args._shelly_ingest_file = None  # set when the sidecar starts; read by _benchmark_scenario
+
+    def _ensure_shelly_sidecar() -> None:
+        if not want_shelly_http or args._shelly_ingest_file:
+            return
+        args._shelly_ingest_file = _start_shelly_ingest_sidecar(
```
Claudia-Anthropica, line 3806: @wasnertobias [medium] In relay mode this starts the Shelly ingest sidecar on the machine running the benchmark, not on `--logos-ssh-host` where Traefik is running. A developer-machine run with `--logos-ssh-host ... --shelly-transport http` will never register the `/shelly-ingest` route on logos-test and wall-power collection gets disabled even though the CLI says that mode is supported. Start and stop the sidecar on the relay host when `relay_host` is set, or reject that combination.

In `logos/benchmarks/benchmark_logos.py`, `_ensure_shelly_sidecar` always starts the HTTP ingest sidecar locally even when `--logos-ssh-host` points at the Traefik host. Fix this by starting/stopping the sidecar on the relay host in relay mode, or by rejecting `--shelly-transport http` with `--logos-ssh-host` before the benchmark starts.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The worker-node `max_num_seqs` propagation looks solid, but the benchmark orchestration has a few correctness issues and the prod workflow currently drops a worker deploy. Please fix the inline items before merging.

PR #611, opened 2026-06-18
`Logos`: Benchmark adaptions

wasnertobias:
## Summary by CodeRabbit

* **New Features**
  * Benchmarks now support an `all` option to combine GSM8K train and test splits, and the default was updated to run with the full dataset.
  * Added support for storing and using per–KV cache context calibration curves to drive better model sizing decisions during runs.
* **Bug Fixes**
  * Improved “needs calibration” detection so calibrated profiles without KV calibration curves are re-calibrated.
* **Tests**
  * Expanded unit coverage for KV curve selection, profile serialization, persistence, and uncalibrated-model detection.

### DIFF
--- a/logos/benchmarks/prepare_benchmark.py
+++ b/logos/benchmarks/prepare_benchmark.py
@@ -148,8 +148,9 @@ def main() -> None:
     parser.add_argument(
         "--split",
         default="test",
-        choices=["train", "test"],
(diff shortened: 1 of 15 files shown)


Claudia-Anthropica (dismissed):
@wasnertobias Looks good from my side; the all-split path is straightforward and CI is green.

PR #612, opened 2026-06-18
`Logos`: Hotfix — prune unused tagged images on prod deploy

wasnertobias:
## Problem

The prod deploy workflow's **Clean up old images** step ran:

```
docker image prune -f --filter "until=168h"
```

Without `-a`, `docker image prune` only removes **dangling** (untagged) images. Superseded *tagged* builds (old `IMAGE_TAG` versions) are unused but never collected, so they accumulate on the prod VMs and consume disk. This is a regression from the pre-Harbor-migration behavior.

## Fix

Add the `-a` flag so the step also removes unused tagged images older than 168h that aren't used by a running container:

```
docker image prune -a -f --filter "until=168h"
```

## Notes

- Manually verified on `logos-test` (`prune -a -f --filter "until=48h"` reclaimed ~4.4GB of stale tagged images that the dangling-only prune left behind).
- The 168h window + "not used by a running container" guard means in-use images (incl. running worker/vLLM containers) are never touched.
- `logos_deploy-test.yml` has the identical step and the same issue — not changed here to keep the prod hotfix scoped; can follow up if desired.

### DIFF
--- a/.github/workflows/logos_deploy-prod.yml
+++ b/.github/workflows/logos_deploy-prod.yml
@@ -149,4 +149,4 @@ jobs:
           proxy_key: ${{ secrets.DEPLOYMENT_GATEWAY_SSH_KEY }}
           proxy_port: ${{ vars.DEPLOYMENT_GATEWAY_PORT }}
           script: |
-            docker image prune -f --filter "until=168h"
+            docker image prune -a -f --filter "until=168h"


wasnertobias:
Superseded — fix pushed directly to main in 51261b80, and extended to the test deploy workflow as well.

PR #614, opened 2026-06-19
`Logos`: Calibration — compute KV/max_model_len curve from vLLM's rep…

wasnertobias:
…orted rate

The KV sweep crawled the rising edge 1 GiB at a time (a full vLLM start per step), which is very slow for models whose full context needs more KV than fits (no plateau → walks the whole window, ~20+ loads/model/node).

vLLM's "KV too small" error already reports the model's full context (M) and the KV needed to serve it ("<X> GiB KV cache is needed") — the KV→context rate. Use it:
- Binary-search the startable-window ceiling (records real anchor points), ~log2 probes instead of a per-GiB walk.
- Derive per-token KV from the reported rate and COMPUTE max_model_len at each KV step (capped at M), validated against the probed anchors; fall back to anchors-only if the rate is unavailable or off by >10%.

Cuts a ~20-probe rising crawl to a handful. Adds a test asserting the computed curve is produced with <12 probes for an unreachable-plateau model.

Please go to the `Preview` tab and select the appropriate sub-template:

* [Athena](?expand=1&template=athena.md)
* [Atlas](?expand=1&template=atlas.md)

## Release Notes

* **Improvements**
  * Enhanced KV-cache calibration with validated KV→context curve generation and improved plateau handling for more reliable results.
  * Benchmark calibration now more reliably finishes pending calibrations (without full resets), shows live progress during runs, and always applies configured calibration provider IDs.
  * Safer capacity planning for calibrated KV via improved envelope selection and corrected KV headroom estimation.
  * vLLM serving enables prefix caching by default; increased inference/stream timeouts help reduce load-time stream/protocol errors.
* **Tests**
  * Added coverage for calibration edge cases, orchestration behavior, and calibrated KV envelope/headroom logic.

### DIFF
--- a/logos/benchmarks/benchmark_logos.py
+++ b/logos/benchmarks/benchmark_logos.py
@@ -3543,6 +3543,42 @@ async def _benchmark_scenario(
 _TRAFFIC_PATTERNS = ["burst", "poisson", "sequential", "mixed"]
 
 
+def _resolve_patterns(raw: Optional[str]) -> list[str]:
+    """Resolve the --patterns selection (comma-separated) to canonical order.
+
+    Empty/None → all four. Unknown names raise so a typo fails fast rather than
+    silently running everything.
+    """
+    if not raw or not str(raw).strip():
+        return list(_TRAFFIC_PATTERNS)
+    wanted = [p.strip().lower() for p in str(raw).split(",") if p.strip()]
+    unknown = [p for p in wanted if p not in _TRAFFIC_PATTERNS]
+    if unknown:
+        raise ValueError(f"Unknown traffic pattern(s) {unknown}; valid: {_TRAFFIC_PATTERNS}")
+    return [p for p in _TRAFFIC_PATTERNS if p in wanted]
+
+
+_ALL_SCENARIOS = ["logos-nosleep", "ollama", "logos-sleep"]
+
+
+def _resolve_scenarios(raw: Optional[str], only_ollama: bool) -> list[str]:
+    """Resolve the --scenarios selection for --run-all-scenarios.
+
+    only_ollama forces just ["ollama"]. Empty/None → all three. Unknown names
+    raise so a typo fails fast. Used to limit a quick debug run to e.g.
+    --scenarios logos-nosleep.
+    """
+    if only_ollama:
+        return ["ollama"]
+    if not raw or not str(raw).strip():
+        return list(_ALL_SCENARIOS)
+    wanted = [s.strip().lower() for s in str(raw).split(",") if s.strip()]
+    unknown = [s for s in wanted if s not in _ALL_SCENARIOS]
+    if unknown:
+        raise ValueError(f"Unknown scenario(s) {unknown}; valid: {_ALL_SCENARIOS}")
+    return [s for s in _ALL_SCENARIOS if s in wanted]
+
+
 async def _run_all_traffic_patterns(
     scenario: str,
     base_url: str,
@@ -3552,11 +3588,16 @@ async def _run_all_traffic_patterns(
     model_map: dict,
     args: argparse.Namespace,
 ) -> list:
-    """Run all four traffic patterns for a scenario; warmup is done only for the first."""
+    """Run the selected traffic patterns for a scenario; warmup is done only for the first.
+
+    The set of patterns is ``--patterns`` (comma-separated; default all four), so a
+    quick debug run can target a single pattern, e.g. ``--patterns mixed``.
+    """
+    selected = _resolve_patterns(getattr(args, "patterns", None))
     summaries = []
-    for i, pattern in enumerate(_TRAFFIC_PATTERNS):
+    for i, pattern in enumerate(selected):
         print(f"\n{'─' * 58}")
-        print(f"  Traffic pattern {i+1}/{len(_TRAFFIC_PATTERNS)}: {pattern.upper()}")
+        print(f"  Traffic pattern {i+1}/{len(selected)}: {pattern.upper()}")
         print(f"{'─' * 58}")
         summary = await _benchmark_scenario(
             scenario,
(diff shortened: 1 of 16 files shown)


Claudia-Anthropica (dismissed):
@wasnertobias The calibration change looks solid to me, and the focused worker calibration tests pass locally. The Docker vLLM/server build failure is "No space left on device" during dependency installation, so I am not holding that against this PR.

logos/benchmarks/benchmark_logos.py
```diff
@@ -3709,6 +3709,197 @@ async def _reset_and_calibrate_all_nodes(
...
+            host, ssh_user, ssh_key, profiles_path, unsupported_path, models, sudo, relay_host, relay_user
+        )
+        pending_by_host[host] = pending
+        line = f"  [calib] {host}: {len(done)}/{len(models)} calibrated"
+        if pending:
+            line += f"  | incomplete: {', '.join(pending)}"
+            pending_anywhere.update(pending)
+        print(line)
+
+    if not pending_anywhere:
```
Claudia-Anthropica, line 4082: @wasnertobias [medium] This can skip calibration for old calibrated profiles that do not have `kv_cache_to_max_model_len_pairs`. The worker explicitly treats those profiles as uncalibrated, but this status path goes through `_profile_is_calibrated()`, which only checks residency/sleep/envelope, so `not pending_anywhere` can return true and the benchmark runs without generating the new KV curve. Please make this status check match the worker and add a fixture for a calibrated profile with missing pairs.

In `logos/benchmarks/benchmark_logos.py`, `_profile_is_calibrated()` treats calibrated profiles with missing `kv_cache_to_max_model_len_pairs` as complete even though the worker recalibrates them. Fix this by returning `False` for `residency_source == "calibrated"` when `kv_cache_to_max_model_len_pairs` is missing or empty, and update the benchmark helper tests to cover that legacy profile shape.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The KV sweep optimization looks good overall and CI is green now. I found one gap in the new non-reset calibration path: it can skip legacy calibrated profiles that the worker still treats as uncalibrated, so please fix that before merging.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Looks good now, thanks.

Claudia-Anthropica (dismissed):
@wasnertobias All feedback addressed, nice work.

Claudia-Anthropica (changes requested):
@wasnertobias The latest calibration changes are close, and the focused benchmark/helper tests pass locally. The TP-rank KV subtraction comment from CodeRabbit looks valid: the calibrated base is measured across all TP GPUs, while the stored KV budget field is per rank, so TP>1 profiles still undercount available KV unless that budget is multiplied by TP. Please address that before merging.

Claudia-Anthropica (dismissed):
@wasnertobias All previously requested feedback is addressed, and the focused calibration/capacity tests pass locally. Approving this follow-up.

logos/benchmarks/benchmark_logos.py
```diff
@@ -4017,6 +4231,28 @@ def _cleanup(reason: str = "cleanup") -> None:
...
+                    logos_url,
+                    args.logos_key,
+                    getattr(args, "calibration_provider_ids", None) or [],
+                    args.logos_admin_port,
+                    use_sudo,
+                    relay_host,
+                    relay_user,
+                ):
+                    print(
+                        "  WARNING: calibration did not fully complete — continuing anyway.",
```
Claudia-Anthropica, line 4474: @wasnertobias [medium] This branch can return `False` because models are still pending and no provider IDs were provided, but the caller only prints this warning and then starts the benchmark anyway. That defeats the new “finish calibration before traffic” guarantee: the next steps can send requests to models the worker never calibrated. Treat this as fatal for the run, e.g. raise or `sys.exit(1)` when `_ensure_calibration_complete_all_nodes()` cannot complete calibration.

In `logos/benchmarks/benchmark_logos.py`, the non-reset calibration path logs a warning and continues after `_ensure_calibration_complete_all_nodes()` returns `False`. Change this branch to abort before the benchmark scenarios run, for example by raising `SystemExit(1)` after printing the error.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The calibration curve changes and focused tests look good, but the non-reset path still continues after it has proven calibration cannot be completed. Please make that case abort before the benchmark starts; otherwise the run can produce invalid failing results.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still seeing the same behavior here: when the non-reset ensure-calibration path returns `False`, this only prints a warning and then proceeds into Step 1. That needs to abort before any benchmark traffic starts, otherwise the failure mode from the previous comment is still possible.

Claudia-Anthropica (changes requested):
@wasnertobias The previous calibration-abort feedback is still open: the non-reset path still warns and continues into the benchmark after calibration cannot be completed. Please make that fatal before merging.

Claudia-Anthropica (changes requested):
@wasnertobias The non-reset calibration failure path is still unresolved: when `_ensure_calibration_complete_all_nodes()` returns `False`, the runner still only warns and proceeds into the benchmark. Please make that path fatal before merging; the existing open thread has the exact spot.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still unchanged on the latest commit: this path only logs the warning and then continues into Step 1. Please make the non-reset ensure-calibration failure fatal before benchmark traffic starts.

Claudia-Anthropica (changes requested):
@wasnertobias The previous non-reset calibration abort feedback is still open on the latest commit. Keeping this as changes requested until that path fails the run instead of continuing into benchmark traffic.

logos/benchmarks/benchmark_logos.py
Claudia-Anthropica: @wasnertobias Still open on current head: the non-reset branch still prints `WARNING: calibration did not fully complete` and then enters Step 1. Please make that path fatal before benchmark traffic starts.

Claudia-Anthropica (changes requested):
@wasnertobias The current head still leaves the non-reset calibration failure path non-fatal. Keeping changes requested until `_ensure_calibration_complete_all_nodes()` returning `False` aborts before benchmark traffic starts.

PR #617, opened 2026-06-20
`Logos`: Use sharded model checkpoint for TP>1 lanes

### LINKED ISSUE #615: `Logos`: Use sharded model checkpoint
> With tensor parallelism enabled, each process will read the whole model and split it into chunks, which makes the disk reading time even longer (proportional to the size of tensor parallelism).

Logos shall automatically use sharded model checkpoints (conversion needed before), when `tensor_parallel_size > 1` is used.

> You can convert the model checkpoint to a sharded checkpoint using [ examples/features/sharded_state/load_sharded_state_offline.py](https://github.com/vllm-project/vllm/blob/main/examples/features/sharded_state/load_sharded_state_offline.py). The conversion process might take some time, but later you can load the sharded checkpoint much faster. The model loading time should remain constant regardless of the size of tensor parallelism.

wasnertobias:
Closes #615.

## Problem

With tensor parallelism enabled, each vLLM rank reads the **whole** model checkpoint and slices out its own shard, so cold-start load time grows with the tensor-parallel size (proportional to model size × TP of disk reads).

## Change

Automatically use [vLLM sharded-state checkpoints](https://docs.vllm.ai/en/latest/configuration/conserving_memory/#tensor-parallelism-tp) for `tensor_parallel_size > 1` lanes. The model is pre-converted to a sharded checkpoint (each rank then reads only its own shard → load time roughly constant in TP) and served with `--load-format sharded_state`.

Conversion is triggered in two places, per the issue:
1. **After calibration**, when the calibrated `tensor_parallel_size` is `>1` — the GPU is already free at that point, so it's the cheapest time to convert.
2. **Lazily, right before a TP>1 lane is spawned**, if no converted checkpoint exists yet.

Both paths go through an idempotent `ensure_sharded_checkpoint()`:
- keyed by `(model, tp)` under `<cache_root>/.sharded_cache/<model>/tp<N>/`,
- guarded by a completion marker (written last) + an in-process lock so the two triggers never collide,
- supports cancellation (wired to `stop_calibration_session`) and a timeout,
- cleans up partial output on failure, and falls back to loading the full checkpoint if anything goes wrong.

The converter (`_sharded_convert.py`) mirrors vLLM's official `save_sharded_state_offline.py` for the pinned `vllm==v0.23.0`, and runs as a subprocess under the vLLM-equipped interpreter (which may be a different venv than the worker).

## Config

New `engines.vllm` options (all default to the feature being **on**):

| key | default | meaning |
|---|---|---|
| `sharded_checkpoint_enabled` | `true` | master switch for both triggers |
| `sharded_checkpoint_convert_on_spawn` | `true` | allow the lazy pre-spawn fallback conversion |
| `sharded_checkpoint_min_tensor_parallel_size` | `2` | min TP that triggers conversion |
| `sharded_checkpoint_max_file_size_bytes` | `5 GiB` | shard file size cap |

## Testing

- New `tests/test_sharded_checkpoint.py` (15 tests): path layout, readiness marker, cache-root/python resolution, conversion success/failure/cleanup, and `_build_cmd` / `_maybe_prepare` integration.
- Full workernode suite: **386 passed, 2 pre-existing skips**. `pre-commit` clean.

## Release Notes

* **New Features**
  * Added support for sharded checkpoints in tensor-parallel configurations to reduce cold-start time.
  * New configuration options to enable sharded checkpoint loading and control conversion timing.
  * Optional automatic checkpoint conversion during model calibration for tensor-parallel setups.

### DIFF
--- a/logos/logos-workernode/config.yml
+++ b/logos/logos-workernode/config.yml
@@ -51,6 +51,20 @@ engines:
     # lane and via sleep_mode_disabled in WorkerRuntimeStatus, so the
     # planner falls back to stop/start for VRAM reclamation here.
     disable_sleep_mode: false
+    # Sharded-checkpoint loading for tensor-parallel (TP>1) lanes. With TP>1
+    # every vLLM rank otherwise reads the whole checkpoint and slices out its
+    # shard, so cold-start load time grows with TP. When enabled, the worker
+    # pre-converts the model to a sharded_state checkpoint (each rank reads
+    # only its own shard → load time roughly constant in TP) and serves it
+    # with --load-format sharded_state. Conversion runs right after calibration
+    # when the calibrated TP is >1, or lazily before a TP>1 lane is spawned if
+    # not yet converted. Checkpoints live under
+    # <cache_root>/.sharded_cache/<model>/tp<N>/.
+    sharded_checkpoint_enabled: true
+    # Allow the lazy, spawn-time fallback conversion (set false to only use
+    # checkpoints converted by the post-calibration trigger, so a lane spawn
+    # never blocks on a conversion).
+    sharded_checkpoint_convert_on_spawn: true
     # Turing (SM 7.5) workaround for AWQ + TP=2 on Quadro RTX 5000:
     # force AWQ path, disable vLLM custom all-reduce kernels (PCIe, no NVLink),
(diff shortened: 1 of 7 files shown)


bassner:
image

PR #620, opened 2026-06-22
`Iris`: Add Course Memory — Weaviate collection, ingestion pipeline, hybrid retrieval, and agent tool

toukhi:
### ⚠️ This PR needs to be tested with (Artemis [PR#13002](https://github.com/ls1intum/Artemis/pull/13002) and [PR#13390](https://github.com/ls1intum/Artemis/pull/13390))
---

### Summary
  This PR introduces Course Memory, a persistent knowledge store that captures tutor-verified Q/A pairs from Artemis
  communication channels and makes them available to Iris during student interactions. When a relevant past answer is
  found, Iris reuses the verified response for consistency rather than generating a new one from scratch.

  ### Key Changes

  **New Features:**
  - `CourseMemory` Weaviate collection with HNSW cosine index; only the question field is embedded and BM25-indexed, all
  other fields are payload metadata
  - `CourseMemoryIngestionPipeline` — LLM-based Q/A extraction from resolved threads with deterministic UUID-keyed upsert
  (supports tutor corrections overwriting existing entries in place)
  - `POST /api/v1/webhooks/course-memory/ingest` webhook — single endpoint handling all source types (`IRIS_AUTO`,
  `TUTOR_WRITTEN`, `IRIS_CORRECTED`, `THREAD_RESOLVED`)
  - `CourseMemoryRetrieval` — hybrid search (BM25 + dense, configurable `alpha`, default 0.5) with optional query
  rewriting and score-threshold filtering
  - `course_memory_retrieval` agent tool wired into the Autonomous Tutor pipeline so Iris checks verified history before
  answering
  - `CourseMemoryIngestionStatus` callback for progress reporting back to Artemis

  **Domain DTOs:**
  - `CourseMemoryEntryDTO`, `CourseMemorySource` enum, `ThreadMessageDTO`
  - `CourseMemoryIngestionExecutionDTO` for the webhook payload

  **Configuration:**
  - New `course_memory` namespace in `application.local.yml`/`application.example.yml` with `similarity_threshold`,
  `result_limit`, `alpha`, and `query_rewrite_enabled`

  ### Fix: answer the newest message of a thread, not the thread root

  Found while testing Course Memory end to end. Asking a follow-up in a thread got the *opening* question answered a
  second time, word for word:

  > **Student:** What is a bridge pattern?
  > **Iris:** The Bridge Pattern is a design pattern used to separate an abstraction from its implementation…
  > **Instructor:** Then what is a strategy pattern
  > **Iris:** *The Bridge Pattern is a design pattern used to separate an abstraction from its implementation…*

  Artemis re-runs the pipeline on every new message and sends the whole thread, but the pipeline named `post.content` as
  "the student's post" and dumped every other message — including the one that triggered the run — into an unlabelled
  list of "existing replies". The agent answered the root post, exactly as instructed.

  Two independent causes, both fixed:

  - **The prompt pointed at the wrong message.** The thread is now turned into chat history — Iris's own replies as
  assistant turns, everyone else as user turns prefixed with their role (`[Student]`, `[Tutor]`, `[Instructor]`,
  `[Iris (you)]`) — so the newest message is the one the agent answers. This also fixes a latent gap: the DTO carries no
  `chat_history`, so `message_history` was empty and the retrieval query rewriter had no context to resolve a
  context-poor follow-up against.
  - **The retrieval query was the whole thread.** The opening question dominated the embedding, so course memory kept
  returning the *first* question's entry — and combined with "prefer reusing it for consistency" it was served back
  verbatim. The query is now scoped to the message being answered, and the prompt tells the agent to ignore a retrieved
  prior answer that matches an earlier message in the thread rather than the current one.

  Depends on Artemis [PR#13390](https://github.com/ls1intum/Artemis/pull/13390), which sends the thread ordered
  (oldest first) and annotates every message with an author role. Without it the replies arrive in arbitrary order and
  Iris cannot tell its own drafts from student messages.

  ### Testing

  **Course Memory**

  1. Start Iris locally with a running Weaviate instance
  2. Send a `POST /api/v1/webhooks/course-memory/ingest` request with a valid bearer token and a thread payload (`source:
  TUTOR_WRITTEN`)
  3. Confirm the entry appears in Weaviate (`CourseMemory` collection)
  4. Trigger the Autonomous Tutor pipeline with a matching student question
  5. Check logs — the `course_memory_retrieval` tool call should return the stored answer

  **Thread follow-ups** (needs Artemis [PR#13390](https://github.com/ls1intum/Artemis/pull/13390))

  6. Post a question in a course-wide channel and let Iris answer it
  7. Reply in the same thread with a *different* question
  8. Confirm Iris answers the follow-up instead of repeating its previous answer, and that it does not re-serve the
  stored course-memory entry for the opening question

## Summary by CodeRabbit

* **New Features**
  * Added Course Memory for storing and retrieving course-scoped, tutor-verified Q&A.
  * Added configurable retrieval, query rewriting, and context limits.
  * Added ingestion and deletion support for memory entries.
  * Tutor responses can reuse relevant verified answers with source links.
  * Responses now focus on the latest thread message while preserving context.
* **Bug Fixes**
  * Improved fallback handling for retrieval, embedding, rewriting, and extraction failures.
  * Safely skips private-channel or disabled-feature ingestion.
* **Documentation**
  * Added Course Memory integration and testing guides.


iris/src/iris/web/status/course_memory_ingestion_status_callback.py
```diff
@@ -0,0 +1,47 @@
+from typing import List
+
+from iris.common.logging_config import get_logger
+
+from ...domain.ingestion.ingestion_status_update_dto import (
+    IngestionStatusUpdateDTO,
+)
+from ...domain.status.stage_dto import StageDTO
```
Claudia-Anthropica: @toukhi [critical] This callback is built on the stage API that has already been removed from `main`: `stage_dto.py`/`stage_state_dto.py` no longer exist, `PipelineExecutionDTO` no longer exposes `initial_stages`, and `StatusCallback` now uses `update()`/`finish()`/`fail()`. The auto-merged tree therefore fails at import before Iris can start; migrate this callback and all course-memory call sites to the run-state API.

In `iris/src/iris/web/status/course_memory_ingestion_status_callback.py`, the callback imports deleted stage types and uses the status interface removed from `main`. Rebase onto `main`, implement `CourseMemoryIngestionStatus` using `IngestionStatusCallback` with the course-memory status URL, remove `initial_stages`, and update the course-memory pipeline and worker to call `update()`, `finish()`, and `fail()`. Add a regression test against the current run-state DTO contract.

</details>

iris/src/iris/pipeline/course_memory_ingestion_pipeline.py
```diff
@@ -0,0 +1,209 @@
...
+        return question, answer
+
+    def _format_thread(self) -> str:
+        """Render the thread as an ordered, role-tagged transcript."""
+        lines = []
+        for message in self.dto.thread:
+            role = message.author_role or "unknown"
+            if message.is_iris_draft:
+                role = f"{role} (iris draft)"
+            lines.append(f"[{role}]: {message.content}")
```
Claudia-Anthropica: @toukhi [high] `message_id` identifies the exact verified/resolving answer, but this transcript drops every message ID and sends only roles/content to the extractor. A thread with multiple tutor/Iris answers (or replies arriving before dashboard approval) gives the model no way to select the verified message, so it can persist the wrong answer under the target message's backlink. Mark the message whose `id == self.dto.message_id` as the verified answer and test a multi-answer thread.

In `iris/src/iris/pipeline/course_memory_ingestion_pipeline.py`, `_format_thread()` omits the message IDs needed to identify the verified answer referenced by `dto.message_id`. Tag the matching message explicitly as the verified/resolving answer in the extraction transcript, update the extraction prompt to require that message as the answer source, validate that it exists, and add a multi-answer regression test.

</details>

iris/src/iris/config.py
```diff
@@ -65,6 +65,28 @@ def validate_llm_configuration_when_enabled(self):
...
+    """Settings for the Course Memory feature.
+
+    The embedding/chat models are NOT configured here; they are resolved through
+    the variant roles of ``course_memory_ingestion_pipeline`` and
+    ``course_memory_retrieval_pipeline`` in ``llm_configuration``.
+    """
+
+    enabled: bool = Field(default=True)
+    alpha: float = Field(
+        default=0.5, description="Hybrid fusion weight (0=dense, 1=BM25)"
```
Claudia-Anthropica: @toukhi [medium] `alpha` is passed straight to Weaviate, where `0` is pure BM25 and `1` is pure vector search; this description advertises the exact opposite. Operators trying to tune toward dense retrieval would instead disable it. Align the public setting with Weaviate's convention (and the existing global-search docs), or invert the value before the query.

In `iris/src/iris/config.py` and `iris/application.example.yml`, the documented `course_memory.alpha` semantics are reversed relative to the value passed to Weaviate. Change the contract to `0 = BM25/keyword, 1 = dense/vector` everywhere, or transform the configured value before `CourseMemoryRetrieval` passes it to `query.hybrid`, and add a test that locks down the chosen semantics.

</details>

iris/src/iris/config.py
```diff
@@ -65,6 +65,28 @@ def validate_llm_configuration_when_enabled(self):
...
+    )
+    similarity_threshold: float = Field(
+        default=0.85, description="Minimum fused score for a retrieved entry"
+    )
+    result_limit: int = Field(default=5)
+    query_rewrite_enabled: bool = Field(default=True)
+    context_message_limit: int = Field(
```
Claudia-Anthropica, line 93: @toukhi [medium] `context_message_limit` is never read anywhere, while `_format_thread()` still sends the entire thread to the extraction LLM. The advertised limit therefore has no effect, so a long public thread can overflow the model context or make ingestion unnecessarily expensive. Apply it while preserving the root question and verified target message, or remove the setting.

In `iris/src/iris/config.py`, `course_memory.context_message_limit` is declared but unused by the ingestion pipeline. Apply the configured limit when building the extraction transcript while always retaining the root question and the message identified by `dto.message_id`, and add tests for long threads and boundary values.

</details>

Claudia-Anthropica (changes requested):
@toukhi The storage/retrieval structure is promising, but this branch's callback code no longer matches `main` and will break Iris after merge. The extractor also cannot identify the verified message, and two retrieval settings currently have misleading or ineffective behavior; see the inline comments.

iris/src/iris/config.py
Claudia-Anthropica: @toukhi This is still unresolved: context_message_limit remains unused and _format_thread still sends every thread message to the model.

iris/src/iris/web/status/course_memory_ingestion_status_callback.py
Claudia-Anthropica: @toukhi This is still unresolved after the merge: the callback still imports the deleted stage modules and calls the removed in_progress, done, and error API, so Iris cannot import this module on the current main tree.

iris/src/iris/pipeline/course_memory_ingestion_pipeline.py
Claudia-Anthropica: @toukhi This is still unresolved: _format_thread still omits message IDs and never marks the message matching dto.message_id, so the extractor still cannot reliably choose the verified answer.

iris/src/iris/config.py
Claudia-Anthropica: @toukhi This is still unresolved: the setting still documents zero as dense and one as BM25 while the value is passed unchanged to Weaviate, whose convention is the reverse.

iris/src/iris/pipeline/course_memory_ingestion_pipeline.py
```diff
@@ -0,0 +1,311 @@
...
+
+        For corrections (``IRIS_CORRECTED`` with an ``existing_answer`` provided),
+        the supplied tutor-edited answer is used directly and only the question is
+        derived from the thread.
+        """
+        thread_text = self._format_thread()
+        prompt = ChatPromptTemplate.from_messages(
+            [
+                ("system", course_memory_extraction_system_prompt),
+                ("user", thread_text),
```
Claudia-Anthropica: @toukhi [high] `thread_text` is inserted as a LangChain prompt template, so any discussion containing `{...}` (very common in code snippets) becomes an unresolved template variable and `.invoke({})` raises before extraction. Pass it as a `HumanMessage`, or escape braces before building the template, and cover a programming-thread example.

In `iris/src/iris/pipeline/course_memory_ingestion_pipeline.py`, the raw thread transcript is passed to `ChatPromptTemplate` as a template and braces in message content become missing variables. Pass the transcript as a non-template human message (or escape template braces) and add a regression test with a code snippet containing braces.

</details>

iris/src/iris/pipeline/course_memory_ingestion_pipeline.py
```diff
@@ -0,0 +1,311 @@
...
+        )
+        response = (prompt | self.pipeline).invoke({})
+        if self.llm.tokens is not None:
+            self._append_tokens(
+                self.llm.tokens, PipelineEnum.IRIS_COURSE_MEMORY_INGESTION
+            )
+
+        is_correction = (
+            self.dto.source == CourseMemorySource.IRIS_CORRECTED
+            and self.dto.existing_answer
```
Claudia-Anthropica, line 286: @toukhi [high] An `IRIS_CORRECTED` request with a missing or blank `existingAnswer` falls through to LLM extraction but is still persisted with source `IRIS_CORRECTED`; retrieval then labels model-generated text as tutor-verified. Reject such payloads so every corrected entry contains the tutor’s actual edit.

In `iris/src/iris/domain/ingestion/course_memory_ingestion_dto.py`, `IRIS_CORRECTED` accepts a missing or blank `existingAnswer`, causing LLM output to be stored as tutor-corrected. Add DTO validation requiring a non-blank `existingAnswer` for `IRIS_CORRECTED` and add acceptance and rejection tests.

</details>

iris/src/iris/pipeline/course_memory_ingestion_pipeline.py
```diff
@@ -0,0 +1,311 @@
...
+        if existing is None:
+            return False
+        existing_source = existing.properties.get(CourseMemorySchema.SOURCE.value)
+        return existing_source in TUTOR_VERIFIED_SOURCES
+
+    def delete_for_message(self, message_id: str, course_id: int) -> bool:
+        """Delete the entry for a given (course, answer-message) pair."""
+        try:
+            obj_uuid = self._deterministic_uuid(message_id, course_id)
+            self.collection.data.delete_by_id(obj_uuid)
```
Claudia-Anthropica: @toukhi [high] Ingestion and deletion run in independent background threads, so a delete arriving during extraction can finish first and the older ingestion worker then inserts the memory again. That leaves deleted or retracted content retrievable; coordinate these operations per `(course_id, message_id)` with cancellation/versioning or a tombstone checked atomically before the write.

In `iris/src/iris/pipeline/course_memory_ingestion_pipeline.py` and the course-memory webhook workers, deletion can be undone by an older in-flight ingestion for the same course/message key. Add per-key coordination or durable tombstone/version checks so an ingestion started before a delete cannot write afterward, and add a deterministic concurrency regression test.

</details>

iris/COURSE_MEMORY_ARTEMIS_INTEGRATION.md
```diff
@@ -0,0 +1,270 @@
...
+POST {artemisBaseUrl}/api/iris/internal/webhooks/ingestion/course-memory/runs/{runId}/status
+Authorization: Bearer {runId}          // runId == settings.authenticationToken
+Content-Type: application/json
+```
+
+Body is an ingestion status update (same shape as FAQ ingestion):
+
+```jsonc
+{
+  "stages": [
```
Claudia-Anthropica: @toukhi [medium] This “exact contract” still shows the removed stage payload, while `CourseMemoryIngestionStatus` now sends the run-state `IngestionStatusUpdateDTO`; an Artemis implementation following this will deserialize the wrong schema. Remove `initialStages`, document the current `runState`/`error`/`tokens` payload, and update the deletion and cosine-gating sections that are now stale too.

In `iris/COURSE_MEMORY_ARTEMIS_INTEGRATION.md` and `iris/COURSE_MEMORY_TESTING.md`, the documentation still describes the removed stage callback API and says current deletion/cosine-gating behavior does not exist. Synchronize both documents with the current run-state callback payload, deletion endpoint, and absolute cosine-certainty gate.

</details>

Claudia-Anthropica (changes requested):
@toukhi The run-state migration fixes the import blocker, but the ingestion path still has correctness and trust issues, and the hand-off contract no longer matches the code. Please address the four inline findings; I did not hold the unrelated root .dockerignore container failures against this PR.

iris/src/iris/config.py
Claudia-Anthropica: @toukhi Nice, the config and example documentation now match Weaviate’s alpha semantics.

iris/src/iris/config.py
Claudia-Anthropica: @toukhi Nice, truncation now honors the configured limit while retaining both the root and target messages.

iris/src/iris/pipeline/course_memory_ingestion_pipeline.py
Claudia-Anthropica: @toukhi Looks good—the transcript is now passed as a plain human message, and the regression test covers braces.

iris/src/iris/pipeline/course_memory_ingestion_pipeline.py
Claudia-Anthropica: @toukhi Nice, the DTO now rejects missing and blank corrected answers before they can be mislabeled as tutor-verified.

iris/src/iris/pipeline/course_memory_ingestion_pipeline.py
Claudia-Anthropica: @toukhi Nice fix—the generation check under the write/delete lock closes the background-thread resurrection race.

iris/COURSE_MEMORY_ARTEMIS_INTEGRATION.md
Claudia-Anthropica: @toukhi Looks good now; the hand-off docs describe the run-state payload, deletion endpoint, and cosine gate accurately.

iris/src/iris/pipeline/course_memory_ingestion_pipeline.py
Claudia-Anthropica: @toukhi The matching message is tagged now, but one part of this is still open: the DTO accepts an empty thread or a thread with no matching `messageId`, which produces zero `VERIFIED ANSWER` tags and leaves answer selection ambiguous again. Reject payloads unless exactly one thread message matches `messageId`, and cover the missing-target case.

Claudia-Anthropica (changes requested):
@toukhi Most of the prior feedback is addressed, and I resolved six threads. The ingestion DTO still needs to reject a thread that does not contain exactly one message matching `messageId`; otherwise the extractor again receives no unambiguous verified answer.

iris/src/iris/pipeline/course_memory_ingestion_pipeline.py
Claudia-Anthropica: @toukhi This is still open on the current head: the DTO validator only checks `existingAnswer` for corrections, so an empty thread or zero/multiple messages matching `messageId` still passes validation and leaves the extractor without exactly one verified target.

Claudia-Anthropica (changes requested):
@toukhi The previous course-memory feedback is addressed except for the ingestion target validation. The DTO still needs to require exactly one thread message whose ID matches messageId; requesting changes until that is fixed.

iris/COURSE_MEMORY_ARTEMIS_INTEGRATION.md
```diff
@@ -0,0 +1,277 @@
...
+  "settings": {
+    "authenticationToken": "<run id; echoed back as Bearer on the status callback>",
+    "artemisBaseUrl": "https://artemis.example.com",
+    "selection": "CLOUD_AI", // or "LOCAL_AI"
+    "variant": "default",
+  },
+
+  "courseId": 1, // int, REQUIRED — scopes storage + retrieval
+  "conversationId": "12345", // string, REQUIRED — originating thread id (backlink)
+  "messageId": "67890", // string, REQUIRED — answer message id; UPSERT KEY
```
Claudia-Anthropica: @toukhi [high] This advertised “exact contract” is incompatible with the current DTOs: ingestion now requires `postId` and at least one thread message flagged with `isVerifiedAnswer` or `resolvesPost`, while deletion requires `postId`; however, the examples, field table, trigger instructions, and checklist still omit those fields and describe `messageId` as the upsert/delete key. An Artemis client following this document will receive validation errors instead of storing or deleting course memory, so migrate this specification and the testing guide to the new wire contract.

In `iris/COURSE_MEMORY_ARTEMIS_INTEGRATION.md`, the ingestion and deletion examples omit required `postId`, omit the required answer-anchor flags, and still describe `messageId` as the deduplication/deletion key even though the current DTOs key entries on `postId`. Update every example, field table, trigger instruction, checklist item, and the corresponding testing guide to send `postId`, mark the appropriate thread messages with `isVerifiedAnswer` or `resolvesPost`, and use `postId` for upsert and deletion semantics.

</details>

Claudia-Anthropica (changes requested):
@toukhi The explicit answer flags and thread-keyed storage address the prior ambiguous `messageId` matching, and the Iris test check passed. However, the Artemis handoff contract was not migrated with the implementation, so clients following it will send invalid ingestion and deletion payloads; see the inline finding.

Claudia-Anthropica (changes requested):
@toukhi The new thread-targeting logic and its tests are internally consistent, but the existing high-severity Artemis contract thread remains unresolved: the documentation still omits required `postId` and answer-anchor fields and describes `messageId` as the storage/deletion key. Clients following that contract would fail DTO validation, so changes are still requested.

Claudia-Anthropica (changes requested):
@toukhi The latest commit only adds autonomous-tutor model-mode logging and does not address the outstanding Artemis handoff contract. The integration and testing examples still omit required `postId` and answer-anchor fields and incorrectly describe `messageId` as the upsert/deletion key, so the unresolved high-severity thread remains valid. The new logging hook itself conforms to the existing pre-agent hook contract.

Claudia-Anthropica (changes requested):
@toukhi The unresolved high-severity handoff-contract thread remains valid: the examples still omit required `postId` and answer-anchor flags and incorrectly describe `messageId` as the upsert/deletion key. Previously reported validation and ordering defects also remain, including coercible `isPublicChannel`, nullable settings that are dereferenced, and worker-start generation tracking that cannot prevent queued or cross-replica resurrection. The latest autonomous-tutor role-label changes and passing CI do not address these blockers.

iris/src/iris/domain/ingestion/course_memory_ingestion_dto.py
```diff
@@ -0,0 +1,94 @@
...
+        self,
+    ) -> "CourseMemoryIngestionExecutionDTO":
+        """A correction must carry the tutor's edited answer.
+
+        ``IRIS_CORRECTED`` marks an entry as tutor-verified. Without a non-blank
+        ``existingAnswer`` the pipeline would fall back to LLM extraction and
+        still store the result under the tutor-verified label, so reject the
+        payload instead of persisting model-generated text as tutor-corrected.
+        """
+        if self.source == CourseMemorySource.IRIS_CORRECTED and not (
```
Claudia-Anthropica: @toukhi [high] `IRIS_AUTO` represents a draft the tutor approved unchanged, but this validator permits it without `existingAnswer`; the documented Trigger A payload does exactly that. `extract_qa()` then stores the LLM's extracted paraphrase under the tutor-verified `IRIS_AUTO` source, so retrieval may present wording no tutor approved. Require a non-blank verbatim answer for `IRIS_AUTO` too, or deterministically copy the flagged approved draft, and align the contract and tests.

In `iris/src/iris/domain/ingestion/course_memory_ingestion_dto.py`, `IRIS_AUTO` payloads can omit the exact approved text and consequently store an LLM paraphrase as tutor-verified. Require a verbatim answer for `IRIS_AUTO` or derive it deterministically from the flagged approved draft, then update the wire contract and regression tests.

</details>

iris/src/iris/pipeline/shared/organizational_guard.py
```diff
@@ -0,0 +1,198 @@
...
+    """Whether a tool returned something that can ground an organizational fact.
+
+    Only two sources qualify. The course FAQ is what instructors maintain for exactly
+    these questions, and course memory holds answers a tutor verified. Lecture content
+    does not qualify: it describes what is taught, and "what the course teaches" is
+    precisely the evidence the model keeps mistaking for "what the exam covers". The
+    course-details tool does not qualify either — it is always available, so counting
+    it would mean the guard never fires.
+    """
+    return bool(faq_hits) or bool(memory_hits)
```
Claudia-Anthropica: @toukhi [high] This treats every course-memory hit as trusted organizational evidence, although `THREAD_RESOLVED` entries are explicitly community-resolved and not tutor-verified. A matching community claim about an exam or deadline therefore disables the confidence cap and can be auto-published as authoritative. Count only tutor-verified memory sources (`IRIS_AUTO`, `TUTOR_WRITTEN`, and `IRIS_CORRECTED`) and cover `THREAD_RESOLVED` with a regression test.

In `iris/src/iris/pipeline/shared/organizational_guard.py`, community-resolved `THREAD_RESOLVED` memories incorrectly bypass the organizational confidence cap. Filter memory evidence to tutor-verified sources only and add a test proving `THREAD_RESOLVED` hits do not remove the cap.

</details>

Claudia-Anthropica (changes requested):
@toukhi The current head addresses the old anchor and handoff-contract defects, and all captured checks pass. Two trust guarantees remain broken: `IRIS_AUTO` can store a model paraphrase as tutor-approved, and community-resolved memory can bypass organizational review; see the inline findings.

iris/src/iris/pipeline/course_memory_ingestion_pipeline.py
```diff
@@ -0,0 +1,765 @@
...
+    @staticmethod
+    def _is_verified(message) -> bool:
+        """Whether Artemis flagged this message as (part of) the verified answer.
+
+        Stated explicitly by the sender — never inferred from ``id``. Artemis
+        draws post and answer ids from two tables with independent sequences, so
+        a root post and one of its answers routinely share a number; matching on
+        it used to tag the student's question as the verified answer.
+        """
+        return message.is_verified_answer or message.resolves_post
```
Claudia-Anthropica, line 593: @toukhi [high] This tags every `resolvesPost` message as verified, although the contract derives the entry's single `TUTOR_WRITTEN` trust tier only from the anchoring answer's endorser. If a tutor endorses one answer while a student resolved another, both are merged and the combined output is stored as tutor-verified, allowing organizational claims from the community-resolved answer to bypass the confidence cap. For tutor-verified sources, require and extract only the single `isVerifiedAnswer` anchor, or carry per-message endorsement and downgrade mixed entries.

In `iris/src/iris/pipeline/course_memory_ingestion_pipeline.py`, every `resolvesPost` message is merged even when only one anchor earned the `TUTOR_WRITTEN` trust tier, so community content can be stored as tutor-verified. For tutor-verified sources, require and extract only the `isVerifiedAnswer` anchor and add a mixed-endorser regression test.

</details>

iris/src/iris/pipeline/shared/organizational_guard.py
```diff
@@ -0,0 +1,226 @@
...
+    tier of it counts. A ``THREAD_RESOLVED`` entry is offered to the agent as a hint
+    and labelled unverified; a hint cannot make an exam date authoritative, and
+    counting it would let one student's claim in a resolved thread lift the cap and
+    auto-publish that same claim to the whole course. Lecture content does not qualify:
+    it describes what is taught, and "what the course teaches" is precisely the
+    evidence the model keeps mistaking for "what the exam covers". The course-details
+    tool does not qualify either — it is always available, so counting it would mean
+    the guard never fires.
+    """
+    return bool(faq_hits) or bool(tutor_verified_memory_hits(memory_hits))
```
Claudia-Anthropica, line 226: @toukhi [high] `bool(faq_hits)` treats any returned FAQ as support for the organizational claim. `FaqRetrieval` returns the nearest results without an absolute relevance threshold, so an unrelated FAQ in the course can populate this list, disable the cap for an exam/deadline question, and permit an invented answer to auto-publish. Count only FAQ hits that passed an explicit relevance gate for the current question.

In `iris/src/iris/pipeline/shared/organizational_guard.py`, any non-empty FAQ retrieval disables the organizational confidence cap even though FAQ retrieval has no absolute relevance floor. Add an explicit FAQ relevance gate and count only qualifying hits as organizational evidence, with a regression test using an unrelated FAQ.

</details>

Claudia-Anthropica (changes requested):
@toukhi The previous `IRIS_AUTO` and `THREAD_RESOLVED` trust issues are fixed, and all captured checks pass. Two remaining paths can still let unsupported or community-authored organizational claims bypass tutor review; see the inline findings.

PR #622, opened 2026-06-22
`Iris`: Logprob uncertainty confidence scoring for the autonomous tutor

toukhi:
## Motivation

The autonomous tutor pipeline attaches a confidence score (0–1) to each generated response, which Artemis uses to route it (auto-post ≥ 0.85 / tutor-review 0.70–0.85 / discard < 0.70). Until now the pipeline only implemented the **verbalized** strategy: a confidence prompt is appended and the model self-reports a `Probability:` line.

This PR adds log-probability based confidence with **three strategies, selected by runtime capability**:

1. **Uncertainty scoring** (primary) — the method of *Logprobs Know Uncertainty* (Xu et al., FSE Companion '25): for each generated token, the top-k alternative candidates (`top_logprobs`) are classified against a curated lexicon into synonym / antonym / speculative pools; tokens whose prediction set contains contradictions or hedges accumulate into a response uncertainty `U = Σ β^(N−i)·σ*((p_a + γ·p_sp)/p_sy)` (γ=0.3, β=0.9), mapped to confidence via `exp(−k·U)`.
2. **Mean-logprob** (fallback) — `exp(mean(token logprobs))`, used when a backend honors `logprobs` but ignores `top_logprobs` (only discoverable after generation; without this tier such deployments would discard every response).
3. **Verbalized** (fallback) — unchanged, for models without logprob support (e.g. Ollama), selected pre-generation via the `supports_logprobs` capability.

Two documented adaptations of the paper (rationale in `uncertainty_scoring.py`): a rescaled sigmoid `σ*(x)=2·(σ(x)−0.5)` (the paper's plain σ ≥ 0.5 contradicts its own empirical distribution), and the `exp(−k·U)` confidence mapping calibrated to the routing tiers (0.85 ⇔ U≈0.163, 0.70 ⇔ U≈0.357).

## How it works

- Strategy tier 3 vs 1–2 is selected **before generation** (`supports_logprobs` capability, so the verbalized prompt can be omitted); tier 1 vs 2 is selected **after generation** based on whether top-k candidates came back.
- Rich per-token entries (token, logprob, top-k candidates) are plumbed through the LLM stack (`CompletionArguments.top_logprobs` → OpenAI client → `PyrisMessage.token_logprob_entries` → `IrisLangchainChatModel.last_token_logprob_entries`), with final-answer capture semantics (tool-call turns don't overwrite).
- The lexicon (`uncertainty_lexicon.py`) covers English **and German** general-discourse antonym classes plus domain classes for the Artemis catalog (`cs`, `ml`, `security`), a negation-prefix heuristic (un-/in-/im-/ir-/non-/dis-), and standalone negators.
- Routing thresholds stay in Artemis; Pyris keeps sending the same 0–1 float, so the callback contract is unchanged.

## Validation

A 10-prompt pilot (student queries across SE/algorithms/ML/security courses, gpt-4o-mini, both scores computed on identical generations — paired design) validated the method end-to-end: the uncertainty method separates reliable from unreliable answers (0.44–1.00, expected routing tier for **10/10** prompts), while mean-logprob compresses all responses into 0.68–0.84, never clears the auto-post bar, and ranks an unanswerable logistics question above a textbook-correct answer. The full experiment writeup is part of the thesis evaluation material.

## Testing

- 61 unit tests across `tests/test_uncertainty_scoring.py` (lexicon, classification, formula math incl. β ordering/γ scaling/fallback contract), `tests/test_logprob_confidence.py` (mean-logprob math, capability lookup, OpenAI `top_logprobs` request/clamping/extraction), and `tests/test_logprob_pipeline_wiring.py` (langchain capture, 3-tier strategy selection).
- Full suite: **329 passed** (after merging main). All commits pass the `pre-commit` gate (black, isort, flake8, bandit, detect-secrets, pylint).

## Summary by CodeRabbit

* **New Features**
  * Added optional token-level log-probability support for compatible chat models, including top alternative tokens.
  * Added model capability settings and configuration guidance for log-probability features.
  * Improved autonomous tutor confidence scoring using token probabilities and uncertainty detection, with fallback to existing verbalized confidence.
  * Added structured confidence data for downstream integrations.
* **Bug Fixes**
  * Preserved confidence data across tool-call responses and disabled unsupported or streaming log-probability requests.
* **Tests**
  * Added comprehensive coverage for log-probability handling, uncertainty scoring, model compatibility, and confidence estimation.


Claudia-Anthropica (dismissed):
@toukhi Looks good from my pass. The logprob strategy is wired through capability selection, prompt construction, OpenAI extraction, and post-agent confidence scoring, and CI is green.

iris/src/iris/llm/external/openai_chat.py
```diff
@@ -268,6 +311,18 @@ def chat(
...
 
+                    # Token-level log-probabilities are requested only when the
+                    # caller opts in and the model declares support. They are
+                    # used downstream to derive a confidence score.
+                    if arguments.logprobs and self.supports_logprobs:
+                        params["logprobs"] = True
+                        # Top-k alternatives per token feed the uncertainty
+                        # confidence method; the API caps top_logprobs at 20.
+                        if arguments.top_logprobs:
+                            params["top_logprobs"] = max(
```
Claudia-Anthropica: @toukhi [high] This assumes a backend that supports plain `logprobs` will silently ignore `top_logprobs`. Strict OpenAI-compatible servers instead return a 4xx for an unsupported parameter, and this client then retries the same invalid request five times, so the documented mean-logprob fallback is never reached. Gate `top_logprobs` behind its own capability (or retry once without it on the specific unsupported-parameter error) while keeping plain logprobs enabled.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `iris/src/iris/llm/external/openai_chat.py`, `top_logprobs` is sent for every model that supports plain logprobs, causing strict logprob-only backends to reject the request before mean-logprob fallback can run. Add a separate top-logprobs capability or a targeted retry without `top_logprobs`, wire it through model configuration, and test a backend that supports `logprobs` but not `top_logprobs`.\n\n</details>

Claudia-Anthropica (changes requested):
@toukhi The uncertainty scorer and its tests are thoughtfully put together, but the fallback path still breaks on strict logprob-only backends because `top_logprobs` is sent unconditionally. Please separate that capability so the mean-logprob tier is actually reachable.

iris/src/iris/llm/external/openai_chat.py
Claudia-Anthropica: @toukhi Nice, the separate capability gate makes the mean-logprob fallback reachable now. Thanks.

Claudia-Anthropica (dismissed):
@toukhi All feedback addressed, nice work.

.dockerignore
```diff
@@ -1,13 +1,20 @@
...
 !logos/db
 !logos/logos-orchestrator/pyproject.toml
 !logos/logos-orchestrator/src
 !logos/logos-orchestrator/tests
+!iris/pyproject.toml
+!iris/poetry.lock
+!iris/src/iris
+!iris/log_conf.yml
+!iris/models
+!memiris
```
Claudia-Anthropica: @toukhi [medium] `!memiris` re-includes every file under that directory after the root `*`, so gitignored files such as `memiris/.env`, `application.local.yml`, and `llm_config.local.yml` are uploaded and then copied into the image by `COPY memiris ../memiris`. Keep the dependency build working with a narrow allowlist for the Memiris packaging inputs instead of admitting the whole tree.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `.dockerignore`, the broad `!memiris` exception allows gitignored local secrets and configuration to enter the Iris Docker build context and image. Replace it with narrow exceptions for the Memiris packaging inputs required by `iris/Dockerfile` (including `pyproject.toml`, `README.MD`, `src`, and the currently packaged `tests`) while ensuring `.env` and local YAML configuration remain excluded.\n\n</details>

Claudia-Anthropica (changes requested):
@toukhi The logprob confidence implementation is solid, but the Docker-context fix can copy gitignored Memiris secrets into the Iris image. Please narrow that whitelist as noted inline.

.dockerignore
Claudia-Anthropica: @toukhi The merge from main leaves `!memiris` unchanged, so the original exposure remains. Keeping this open.

Claudia-Anthropica (changes requested):
@toukhi The merge from main does not change the open .dockerignore concern, so gitignored Memiris configuration can still enter the Iris Docker context. Please narrow that whitelist before this can be approved.

.dockerignore
Claudia-Anthropica: @toukhi Nice fix — the whitelist now includes only the Memiris packaging inputs and keeps local configuration out of the build context. Thanks.

Claudia-Anthropica (approved):
@toukhi All feedback addressed, nice work.

thynguyentumde (approved):
Approved.

bassner (approved):
Validated exact head 24ba3bc8c86015cf13be46574cc26126d4fd9835 in an isolated paired Artemis/Pyris stack. Confirmed logprobs=true and top_logprobs=10 are sent without the verbalized confidence suffix; the returned confidence (0.9713680090803797) produced an automatic Iris Bot reply in the live Artemis communication UI.

PR #624, opened 2026-06-22
`Logos`: Design refactor

alex7sz:
fixing some minors


logos/logos-ui/src/app/features/team-detail/team-detail.ts
```diff
@@ -0,0 +1,149 @@
...
+    this.activeTab.set(tab);
+  }
+
+  private loadAll(teamId: number): void {
+    this.loading.set(true);
+    this.loadError.set(false);
+
+    forkJoin({
+      membersRes: this.teamService.getTeamWithMembers(teamId),
+      apiKeys:    this.teamService.getTeamApiKeys(teamId),
```
Claudia-Anthropica: @alex7sz [medium] This makes team detail fail for app admins who are members but not owners. `/teams/{id}/members` allows that case, but `/admin/teams/{id}/api-keys` returns 403 unless the caller owns the team, and `forkJoin` turns that into a full page load error. Load the members/team first, derive `is_caller_owner`, then fetch keys/model permissions only for admins/owners, or handle the 403 without failing the page.

In `logos/logos-ui/src/app/features/team-detail/team-detail.ts`, `loadAll()` fetches owner-only API key data together with the basic team detail and fails the whole page for non-owner app admins. Fix this by loading `getTeamWithMembers()` first, deriving admin/owner access from the response, and only fetching API keys/model permissions when the caller can access them; non-owner members should still see the basic team detail.

</details>

logos/logos-ui/src/app/core/services/model-management.service.ts
```diff
@@ -0,0 +1,25 @@
...
+
+@Injectable({ providedIn: 'root' })
+export class ModelManagementService {
+  private http = inject(HttpClient);
+
+  getModels(): Observable<Model[]> {
+    return this.http.post<Model[]>('/api/logosdb/get_models', {});
+  }
+
+  addModel(payload: AddModelPayload): Observable<Model> {
```
Claudia-Anthropica: @alex7sz [medium] These endpoints are typed as returning a full `Model`, but the backend returns `{ result, model_id }` from `/add_model` and only `{ result }` from `/update_model_info`. The callers then read `newModel.id` / `updated.id`, so creating with weights sends `model_id: undefined` and edits cannot update the row. Map the create response's `model_id` explicitly and refresh the list after updates, or change the backend to return the full model.

In `logos/logos-ui/src/app/core/services/model-management.service.ts`, the add/update model methods are typed as returning `Model` even though the backend returns result maps. Fix the service and its callers so create uses the returned `model_id` and update refreshes or receives a real `Model` before updating UI state.

</details>

logos/logos-ui/src/app/features/my-keys/my-keys.ts
```diff
@@ -0,0 +1,198 @@
...
+
+  budgetPercent(key: MyKey): number {
+    if (!key.settings?.budget_limit_micro_cents) return 0;
+    return Math.min(100, (key.used_micro_cents / key.settings.budget_limit_micro_cents) * 100);
+  }
+
+  // ── Display helpers ───────────────────────────────────────────────────────
+  formatMicroCents(mc: number | null): string {
+    if (mc == null) return '∞';
+    return '$' + (mc / 1_000_000).toFixed(2);
```
Claudia-Anthropica: @alex7sz [medium] This is off by 100x: `micro_cents` are divided by `100_000_000` everywhere else in the UI, but this helper divides by `1_000_000`. A `$1.00` budget/usage value will render as `$100.00` on My Keys. Use the same conversion constant/helper as the team and billing pages.

In `logos/logos-ui/src/app/features/my-keys/my-keys.ts`, `formatMicroCents()` converts micro-cents to dollars with the wrong divisor. Fix it to divide by `100_000_000` or reuse the shared conversion used by the other Logos budget views.

</details>

logos/logos-ui/src/app/features/team-detail/tabs/app-keys/app-keys-tab.ts
```diff
@@ -0,0 +1,277 @@
...
+        local_tpm_limit:          this.parseLimit(this.cLocalTpm()),
+      },
+    };
+
+    this.svc.createApiKey(this.teamId, payload).subscribe({
+      next: (newKey) => {
+        this.createOpen.set(false);
+        this.resetCreate();
+        this.createLoading.set(false);
+        this.openModal(newKey);
```
Claudia-Anthropica: @alex7sz [medium] `createApiKey()` does not return a `TeamApiKey`; the backend returns `{ id, api_key }`. Opening the modal with that object means `copyKey()` reads `key.key_value` and no-ops, so the newly-created key cannot be copied from the modal. Map `api_key` to `key_value` and fill the fields the modal needs, or wait for the refreshed key list before opening it.

In `logos/logos-ui/src/app/features/team-detail/tabs/app-keys/app-keys-tab.ts`, the create-key response is passed to `openModal()` as if it were a full `TeamApiKey`. Fix this by mapping the backend's `api_key` response field to the modal's expected `key_value` shape, or by reopening the modal from the refreshed key list.

</details>

logos/logos-ui/src/app/features/providers/providers.ts
```diff
@@ -0,0 +1,391 @@
...
+  submitEdit(): void {
+    const target = this.editTarget();
+    if (!target || this.editLoading()) return;
+    this.editLoading.set(true);
+    this.editError.set('');
+    const payload: UpdateProviderPayload = {
+      provider_id: target.id,
+      name: this.editName().trim() || undefined,
+      base_url: this.editBaseUrl().trim() || undefined,
+      api_key: this.editApiKey().trim() || undefined,
```
Claudia-Anthropica: @alex7sz [medium] Clearing any of these fields in the edit form will not actually clear them in the backend. Blank input becomes `undefined`, the service sends `null`, and `ProviderService.updateProvider()` ignores null fields, so old API keys/auth headers remain even though the save appears successful. Preserve an explicit empty string or add a backend clear semantic for these fields.

In `logos/logos-ui/src/app/features/providers/providers.ts`, blank provider credential/auth fields are converted to `undefined` and then `null`, which the backend ignores. Fix provider editing so clearing `api_key`, `auth_name`, or `auth_format` actually clears the stored value, either by sending an explicit empty string or by adding backend support for explicit null clears.

</details>

Claudia-Anthropica (changes requested):
@alex7sz The Angular rewrite is headed in the right direction, but I found several response-shape and permission-flow bugs that will break normal UI paths. The main issues are team detail failing for non-owner app admins and model/app-key creation using backend responses as the wrong TypeScript shapes. See inline comments.

logos/logos-ui/src/app/features/team-detail/team-detail.ts
Claudia-Anthropica: @alex7sz Still seeing this in current `loadAll()`: it `Promise.all`s `getTeamApiKeys()` with `getTeamWithMembers()`, and the API-key endpoint still 403s for app-admin members who are not owners. Please load the basic team detail first and only fetch owner/admin-only data when allowed.

logos/logos-ui/src/app/core/services/model-management.service.ts
Claudia-Anthropica: @alex7sz Looks good now, thanks.

logos/logos-ui/src/app/features/my-keys/my-keys.ts
Claudia-Anthropica: @alex7sz Looks good now, thanks.

logos/logos-ui/src/app/features/team-detail/tabs/app-keys/app-keys-tab.ts
Claudia-Anthropica: @alex7sz This still uses `createApiKey()` as `Promise<TeamApiKey>`, but the backend still returns `{ id, api_key }`; `openModal(newKey)` still does not get a `key_value`. Please map the response before opening the modal, or refresh the list and open from the refreshed key.

logos/logos-ui/src/app/features/providers/providers.ts
Claudia-Anthropica: @alex7sz This is still present: `submitEdit()` turns blank credential fields into `undefined`, the service serializes them as `null`, and `ProviderService.updateProvider()` ignores nulls. Clearing `api_key`, `auth_name`, or `auth_format` still leaves the old values in the database.

logos/logos-ui/src/app/features/team-management/team-management.html
```diff
@@ -0,0 +1,180 @@
...
+
+  <!-- ── Toolbar ─────────────────────────────────────────────────────── -->
+  <div class="toolbar">
+    <app-search-input
+      [value]="search()"
+      placeholder="Search teams..."
+      (valueChange)="search.set($event)"
+    />
+
+    @if (isLogosAdmin()) {
```
Claudia-Anthropica: @alex7sz [medium] This hides team creation from `app_admin` users even though the route and `POST /teams` still allow them, and the old UI let app admins create a team with themselves as the default owner. With this guard, an app admin can manage existing teams but has no way to start a new one. Show this action for app admins too; the owner picker can stay logos-admin-only.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `logos/logos-ui/src/app/features/team-management/team-management.html`, the New Team button is gated by `isLogosAdmin()` so `app_admin` users cannot create teams. Fix this by showing the create-team action to app admins as well, while keeping any logos-admin-only owner selection behavior in the modal.\n\n</details>

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
```diff
@@ -30,6 +32,73 @@ boolean existsByTeamIdAndKeyTypeAndEnvironmentAndIsActive(
...
+          ON bu.api_key_id = ak.id
+          AND bu.month = DATE_TRUNC('month', CURRENT_DATE)::date
+        LEFT JOIN teams t ON t.id = ak.team_id
+        WHERE ak.user_id = :userId AND ak.is_active = true
+        ORDER BY ak.id
+        """, nativeQuery = true)
+    List<MyKeyProjection> findKeysForUser(@Param("userId") int userId);
+
+    @Query(value = """
+        SELECT m.name AS model_name, p.name AS provider_name, p.privacy_level::text AS provider_type
```
Claudia-Anthropica: @alex7sz [medium] This field is named `provider_type`, but both access queries return `privacy_level` values like `CLOUD_IN_EU_BY_US_PROVIDER`. `MyWorkspace.getModelGroups()` checks `provider_type === 'CLOUD'`, so cloud-backed models are classified as local and sorted/badged incorrectly. Return the actual provider type (`cloud`/`logosnode`) and align the UI check, or rename the DTO field and use privacy levels consistently.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java`, `findAccessibleModelsByTeam()` and `findAccessibleModelsByKey()` alias `p.privacy_level` as `provider_type`, while the Angular workspace uses that field to decide cloud vs local. Fix the contract by returning `p.provider_type::text` as `provider_type` from both queries and updating the workspace check to match the returned `cloud`/`logosnode` values, or rename the DTO field and update the UI to use privacy levels intentionally.\n\n</details>

Claudia-Anthropica (changes requested):
@alex7sz The Angular rewrite is much closer now, but I found two new permission/data-contract regressions in the current head. App admins lose the team-create action, and the My Workspace model metadata reports privacy levels as provider types.

logos/logos-ui/src/app/features/providers/providers.ts
Claudia-Anthropica: @alex7sz Looks good now, thanks.

logos/logos-ui/src/app/features/team-detail/tabs/app-keys/app-keys-tab.ts
Claudia-Anthropica: @alex7sz Looks good now, thanks.

logos/logos-ui/src/app/features/team-management/team-management.html
Claudia-Anthropica: @alex7sz Nice, that handles it.

logos/logos-ui/src/app/features/team-detail/team-detail.ts
Claudia-Anthropica: @alex7sz Looks good now, thanks.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
Claudia-Anthropica: @alex7sz This is still off in the current head: the queries now return `cloud`/`logosnode`, but `MyWorkspace.getModelGroups()` still checks for uppercase `CLOUD`, so cloud models are still grouped as local. Please update the workspace check to match the values the backend now returns.

Claudia-Anthropica (changes requested):
@alex7sz Most of the earlier feedback is addressed now, but one issue is still open. My Workspace still checks uppercase `CLOUD` while the backend now returns `cloud`/`logosnode`, so cloud models are still grouped as local.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
Claudia-Anthropica: @alex7sz Looks good now, thanks.

Claudia-Anthropica (dismissed):
@alex7sz All requested changes are addressed now; the code-side follow-up looks good. I only saw pending CI checks, not a failing result to hold this on.

wasnertobias:
@copilot resolve the merge conflicts in this pull request

wasnertobias:
Some minor fixes for a follow-up:
- Page title mentions "LogosUi": Could be changed to Logos only. Ideally, it also adapts to the navigated page.
- Text next to logo in page is "logos" => Uppercase L
- Statisics page: Shall show live information **only** in the Provider tab, currently it changes when selecting a different time range
- Statistics page: Remove refresh button
- Do not show "My Workspace" in navigation if "You are not a member of any team yet."
- OpenCode: Consolidate steps 4 and 5 into one. Do not override the current config, but adapt any existing one (if existing). Make it ephemeral.

PR #637, opened 2026-06-26
`Logos`: Fix Azure cloud routing (404) + auto-sync deployments + record failed-request status

wasnertobias:
Debugged live on the prod instance (`ssh logos`). Three related fixes for Azure cloud serving, in one PR.

## 1. Broken Azure cloud forward URL (the 404)
Someone could not call **gpt-4.1-mini** — `404 Resource not found`.

**Root cause:** for cloud providers, `ContextResolver._cloud_forward_url` rebuilt the upstream URL from `base_url` + the inbound request path, producing:
```
https://xxx.openai.azure.com/openai/deployments/v1/chat/completions
```
— missing the deployment name and `api-version`. Confirmed in the orchestrator logs:
```
executor  Sync request to https://xxx.openai.azure.com/openai/deployments/v1/chat/completions
executor  ... failed: status=404, body={'error': {'code': '404', 'message': 'Resource not found'}}
```
**Fix:** a fully-qualified per-model `endpoint` is now forwarded verbatim (Azure URLs can't be reconstructed from base_url + path). The like-for-like `request_path` rewrite still applies to OpenAI-shaped upstreams whose endpoint is relative/empty.

> A secondary 401 from an invalid per-model `api_key` override on the gpt-4.1 family was a data issue, cleared in prod separately.

## 2. Auto-sync deployed Azure models into the DB (startup + every 24h)
New `AzureDeploymentSyncService` lists the live deployments on each Azure provider (`GET /openai/deployments?api-version=2023-03-15-preview`) and upserts them into `models` + `model_provider`, so the catalogue mirrors what's actually deployed. Handles the Azure quirks found on prod:
- deployment id (URL path) can differ from the served `model` (e.g. id `gpt-4o` actually serves `gpt-5.1`) — models are named by the served model; the URL uses the deployment id;
- duplicate deployments for one model are deduped (prefer the id matching the model name);
- per-family operation paths/api-versions: chat/completions, `/responses` (gpt-5.x reasoning), embeddings, audio (whisper/tts), images;
- Responses-API models whose deployment id ≠ model name are skipped (Azure `/responses` resolves the deployment from the body's `model`, which Logos forwards unchanged) — verified live;
- stale `model_provider` links are pruned to mirror the resource.

Gated by `LOGOS_AZURE_SYNC_ENABLED` (default on); interval + per-operation api-versions are env-overridable. Failures never block startup; new models trigger a runtime refresh + classifier rebuild.

## 3. Failed cloud requests showed grey, not red
On the statistics page the failed call appeared grey rather than as an error. `log_entry.result_status` was only written via `record_completion`, which runs only when `scheduling_stats` is present (keyed off request_id). Cloud requests with no scheduling stats left `result_status` NULL → grey. Now the final status is persisted directly by `log_id`, independent of `scheduling_stats`.

## Tests
- `tests/unit/pipeline/test_cloud_forward_url.py` — Azure absolute endpoint used as-is; OpenAI-shaped upstreams still forward like-for-like.
- `tests/unit/sdi/test_azure_deployment_sync.py` — operation classification, endpoint building, dedup/naming, mismatch-skip.
- Full `tests/unit/pipeline`, `tests/unit/main/test_request_logging.py`, and the new suites pass (81 tests).

### DIFF
--- a/logos/logos-orchestrator/src/logos/sdi/azure_deployment_sync.py
+++ b/logos/logos-orchestrator/src/logos/sdi/azure_deployment_sync.py
@@ -49,11 +49,13 @@ class AzureOperation:
     """How to address a model family on Azure."""
 
     # Operation suffix appended after the deployment segment, e.g.
-    # "chat/completions". Empty for the Responses API, which is addressed at
-    # /openai/responses (no deployment in the path; model goes in the body).
+    # "chat/completions". For the Responses API this is "responses": Azure's
+    # real route is /openai/responses (no deployment in the path; the
+    # deployment is named by the request body's "model"), but we still store
+    # the deployment-scoped form so the id is recoverable — see
+    # build_azure_endpoint and ContextResolver._azure_responses_route.
     suffix: str
     api_version: str
-    uses_deployment_path: bool = True
 
 
 def classify_azure_operation(model_name: str) -> AzureOperation:
@@ -76,7 +78,7 @@ def classify_azure_operation(model_name: str) -> AzureOperation:
     # gpt-5.x reasoning models use the Responses API (matches existing DB
     # config for the gpt-5.4 family); gpt-5-chat is a normal chat model.
     if re.match(r"^gpt-5", m) and "chat" not in m:
-        return AzureOperation("", _RESPONSES_API_VERSION, uses_deployment_path=False)
+        return AzureOperation("responses", _RESPONSES_API_VERSION)
     return AzureOperation("chat/completions", _CHAT_API_VERSION)
 
 
@@ -92,10 +94,20 @@ def azure_host_from_base_url(base_url: str) -> str:
 
 
 def build_azure_endpoint(host: str, deployment_id: str, op: AzureOperation) -> str:
-    """Build the fully-qualified Azure endpoint URL for a deployment."""
+    """Build the per-model endpoint URL stored in the DB for a deployment.
+
+    Always deployment-scoped (``.../openai/deployments/<id>/<suffix>``) so the
+    deployment id is recoverable from the stored URL: the scheduler extracts it
+    for capacity tracking (``extract_azure_deployment_name``) and the forward
+    layer recovers it to address the deployment.
+
+    For the Responses API Azure's real route has no deployment segment — it
+    resolves the deployment from the request body's ``model`` field — so
+    ``ContextResolver._azure_responses_route`` collapses
+    ``.../openai/deployments/<id>/responses`` to ``.../openai/responses`` and
+    rewrites the body ``model`` to ``<id>`` at forward time.
+    """
     host = host.rstrip("/")
-    if not op.uses_deployment_path:
-        return f"{host}/openai/responses?api-version={op.api_version}"
     return f"{host}/openai/deployments/{deployment_id}/{op.suffix}?api-version={op.api_version}"
 
 
@@ -132,20 +144,6 @@ def plan_sync(host: str, deployments: List[Dict[str, Any]]) -> List[Dict[str, st
             deps_sorted[0],
         )
         op = classify_azure_operation(model_name)
-        # The Responses API addresses the deployment via the request body's
-        # "model" field, and Logos forwards the client's body verbatim for
-        # cloud providers (no model rewrite). So a Responses model is only
-        # routable when its deployment id matches the served model name —
-        # otherwise the body would carry a name Azure can't resolve. Skip
-        # those rather than create a broken entry.
-        if not op.uses_deployment_path and _norm(chosen["id"]) != _norm(model_name):
-            logger.warning(
-                "Azure deployment sync: skipping Responses-API model %r — served by deployment "
-                "%r whose name differs, so it can't be addressed via the request body",
-                model_name,
-                chosen["id"],
-            )
-            continue
         planned.append(
             {
                 "model_name": model_name,
@@ -170,7 +168,7 @@ class AzureDeploymentSyncService:
         self,
         interval_s: int = SYNC_INTERVAL_S,
         enabled: bool = SYNC_ENABLED,
-        on_models_changed: Optional[Callable[[bool], "asyncio.Future | Any"]] = None,
+        on_models_changed: Optional[Callable[..., "asyncio.Future | Any"]] = None,
     ):
         self._interval_s = interval_s
         self._enabled = enabled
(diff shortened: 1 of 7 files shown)


logos/logos-orchestrator/src/logos/sdi/azure_deployment_sync.py
```diff
@@ -0,0 +1,262 @@
...
+    if not parts.scheme or not parts.netloc:
+        raise ValueError(f"Cannot derive Azure host from base_url: {base_url!r}")
+    return f"{parts.scheme}://{parts.netloc}"
+
+
+def build_azure_endpoint(host: str, deployment_id: str, op: AzureOperation) -> str:
+    """Build the fully-qualified Azure endpoint URL for a deployment."""
+    host = host.rstrip("/")
+    if not op.uses_deployment_path:
+        return f"{host}/openai/responses?api-version={op.api_version}"
```
Claudia-Anthropica: @wasnertobias [high] The scheduler never gets a capacity entry for these `/openai/responses` models: `_register_models_with_facades()` only registers Azure entries when `extract_azure_deployment_name()` finds a `/deployments/...` segment, and `estimate_ettft_azure(None)` marks the candidate unavailable. A synced gpt-5 model can land in `model_provider` but proxy/resource mode will 503 before it ever reaches this URL. Register Responses models with a stable capacity key, or make Azure scheduling treat no-deployment Responses entries as available, and cover that path in a test.

In `logos/logos-orchestrator/src/logos/sdi/azure_deployment_sync.py`, Responses API endpoints are synced without a `/deployments/...` segment, but the Azure scheduler registration path drops such models and then treats them as unavailable. Fix this by updating the Azure registration/scheduling flow so Responses API models are registered and schedulable, and add a regression test that a synced gpt-5 Responses model is not rejected before execution.

</details>

logos/logos-orchestrator/src/logos/sdi/azure_deployment_sync.py
```diff
@@ -0,0 +1,262 @@
...
+            logger.debug("Azure deployment sync: no Azure providers configured")
+            return
+
+        any_new = False
+        async with httpx.AsyncClient() as client:
+            for provider in providers:
+                changed = await self._sync_provider(provider, client)
+                any_new = any_new or changed
+
+        if any_new and self._on_models_changed is not None:
```
Claudia-Anthropica: @wasnertobias [high] This only refreshes runtime state when a brand-new `models` row is inserted, but `sync_azure_deployments()` can add a `model_provider` link for an existing model, update a deployment endpoint, or prune a stale link with `newly == []`. In those cases the DB mirrors Azure while the scheduler registry/Azure facade keep the old registrations until restart, so newly synced existing models can still be unavailable. Return a separate DB-changed signal and refresh runtime state whenever links/endpoints change; reserve the classifier rebuild flag for actual new model rows.

In `logos/logos-orchestrator/src/logos/sdi/azure_deployment_sync.py`, runtime refresh is gated only on newly inserted model rows. Fix this by having the DB sync report whether any `model_provider` link was inserted, updated, or deleted, then call the runtime refresh whenever that changed while only rebuilding the classifier when model rows were newly inserted.

</details>

logos/logos-orchestrator/tests/unit/pipeline/test_cloud_forward_url.py
```diff
@@ -0,0 +1,43 @@
...
+    "https://ase-se01.openai.azure.com/openai/deployments/"
+    "gpt-41-mini/chat/completions?api-version=2025-01-01-preview"
+)
+
+
+def test_absolute_azure_endpoint_used_verbatim():
+    # Even with an inbound request path, an absolute per-model endpoint wins:
+    # the deployment name + api-version cannot be reconstructed from base_url.
+    assert (
+        ContextResolver._cloud_forward_url(AZURE_BASE, "/v1/chat/completions", AZURE_ENDPOINT)
```
Claudia-Anthropica: @wasnertobias [low] CI's Black hook reformats this assertion, so the `pre-commit` job is red. Run Black on the Logos orchestrator tests, or `pre-commit run --config logos/.pre-commit-config.yaml --all-files`, and commit the generated formatting.

In `logos/logos-orchestrator/tests/unit/pipeline/test_cloud_forward_url.py`, Black reformats the new assertion and CI fails. Run Black/pre-commit for Logos and commit the formatter output.

</details>

logos/logos-orchestrator/tests/unit/sdi/test_azure_deployment_sync.py
```diff
@@ -0,0 +1,85 @@
...
+    # gpt-5.1 served by deployment 'gpt-4o' routes to /responses, which keys off
+    # the body model name — unroutable, so it must be dropped.
+    planned = plan_sync(HOST, [{"id": "gpt-4o", "model": "gpt-5.1", "status": "succeeded"}])
+    assert planned == []
+
+
+def test_plan_keeps_responses_model_with_matching_deployment():
+    planned = plan_sync(HOST, [{"id": "gpt-5.4", "model": "gpt-5.4", "status": "succeeded"}])
+    assert len(planned) == 1
+    assert planned[0]["endpoint"].endswith("/openai/responses?api-version=" + classify_azure_operation("gpt-5.4").api_version)
```
Claudia-Anthropica: @wasnertobias [low] Same Black failure here: CI reformats this long assertion into the multiline layout. Please commit Black's output so the `pre-commit` check passes.

In `logos/logos-orchestrator/tests/unit/sdi/test_azure_deployment_sync.py`, Black reformats the long endpoint assertion and CI fails. Run Black/pre-commit for Logos and commit the formatter output.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The Azure URL fix itself makes sense, but the deployment sync has two scheduler integration gaps: Responses API models are synced into the DB but never registered with the Azure facade, and existing-model deployment changes do not refresh runtime state. CI is also red because Black reformats the new tests. See inline comments.

logos/logos-orchestrator/src/logos/sdi/azure_deployment_sync.py
wasnertobias: Fixed in 880990c5. Synced Responses endpoints are now stored deployment-scoped (`.../openai/deployments/<id>/responses?api-version=...`), so `extract_azure_deployment_name()` returns the id and `_register_models_with_facades()` registers them with the Azure facade rather than filtering them out. `get_capacity_info()` then reports `has_capacity=True` (no rate-limit headers yet) and `estimate_ettft_azure()` returns `WARM`, so they are scheduled as available instead of 503-ing. The deployment-scoped URL is collapsed to Azure's real `/openai/responses` route at forward time (`ContextResolver._azure_responses_route`). Regression test `test_synced_responses_model_is_schedulable` covers plan → extract → register → estimate.

logos/logos-orchestrator/src/logos/sdi/azure_deployment_sync.py
wasnertobias: Fixed in 880990c5. `sync_azure_deployments()` now returns `{"new_models": [...], "changed": bool}`, where `changed` is True whenever a `model_provider` link is inserted, an endpoint drifts, or a stale link is pruned. The sync service refreshes the runtime registry whenever `changed` is True and only sets `rebuild_classifier=True` when new model rows were inserted (`on_models_changed(rebuild_classifier=any_new_models)`), so the classifier rebuild stays reserved for actual new models.

logos/logos-orchestrator/tests/unit/pipeline/test_cloud_forward_url.py
wasnertobias: Done in 880990c5 — ran Black 25.1.0 + isort via the Logos pre-commit config on the touched tests; this file is now formatter-clean.

logos/logos-orchestrator/tests/unit/sdi/test_azure_deployment_sync.py
wasnertobias: Done in 880990c5 — Black + isort applied (isort also consolidated the new imports); formatter-clean.

logos/logos-orchestrator/src/logos/sdi/azure_deployment_sync.py
Claudia-Anthropica: @wasnertobias Looks good now, thanks.

logos/logos-orchestrator/src/logos/sdi/azure_deployment_sync.py
Claudia-Anthropica: @wasnertobias Nice, that handles it.

logos/logos-orchestrator/tests/unit/pipeline/test_cloud_forward_url.py
Claudia-Anthropica: @wasnertobias Looks good now, thanks.

logos/logos-orchestrator/tests/unit/sdi/test_azure_deployment_sync.py
Claudia-Anthropica: @wasnertobias Looks good now, thanks.

Claudia-Anthropica (approved):
@wasnertobias All previous feedback is addressed, nice work. The remaining red check is `Build Logos UI / finalize`, which failed while creating the Harbor manifest after the UI image build itself passed, so I’m treating that as CI/registry noise rather than code feedback.

PR #639, opened 2026-06-26
`Logos`: Drain in-flight requests before sleeping/evicting a lane

wasnertobias:
AdMITTED-during-swap requests were being dropped (client sees `RemoteProtocolError: peer closed connection without sending`, or a half-open hang), ~2/1000 under model-swap pressure. This deadlocked a long benchmark run.

## Root cause (TOCTOU race)
The capacity planner decides to sleep/evict a lane against a **heartbeat snapshot up to ~15s old**, drains against that *same stale snapshot* (`_drain_lane` reads the cached `latest_runtime`, never queries live), then issues the command. The worker then sleeps/unloads the lane with **no active-request guard of its own**, severing any in-flight generation. `LANE_MIN_TENURE` only shrinks the window probabilistically — which is why lowering it worsens the drops.

Confirmed locations: `capacity_planner.py` (`_find_eviction_set` ~1953, `_drain_lane` ~5756 reads `peek_runtime_snapshot`), `logosnode_registry.py` `peek_runtime_snapshot` ~925 (cached heartbeat), worker `lane_manager.sleep_lane` (no guard) + `logos_bridge._execute_stream_command` (stream breaks when vLLM is slept underneath it).

## Fix
The worker holds the only authoritative, non-stale per-lane active-request count, so gate teardown on it. New bounded, lock-free `_drain_lane_inflight()` (`LOGOS_LANE_SLEEP_DRAIN_TIMEOUT_S`, default 30s) called before teardown on all three paths:
- **sleep_lane** (`mode!=force`): drain + a final in-lock re-check that skips the sleep if a request slipped into the drain gap (lock is held through `handle.sleep()` and `increment_active_requests` needs the same lock → window fully closed); planner retries next cycle.
- **remove_lane**: bounded best-effort drain before removal.
- **apply_lanes**: pre-drain (lock-free) only the lanes the apply will REMOVE (covers logos-nosleep eviction swaps); no-op/add-only applies pay nothing.

## ⚠️ Not yet deployed
Needs review + a worker-image rebuild and testing on free nodes. A benchmark run is currently in progress and is protected meanwhile by the benchmark’s 30-min drain cap (separate change on `logos/benchmark-evaluation`). 

Follow-up (defense in depth, not required): make the planner’s `_drain_lane` query live worker status instead of the cached heartbeat.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lane lifecycle handling so active inference requests are less likely to be interrupted during lane sleep, stop, or removal.
  * Added a bounded “drain in-flight requests” step before taking lanes offline, reducing race conditions between dispatch and shutdown.
  * If new requests arrive during the sleep transition, the lane remains serving instead of shutting down mid-request.
  * Introduced a configurable drain timeout via `LOGOS_LANE_SLEEP_DRAIN_TIMEOUT_S` (with safe defaults).

### DIFF
--- a/logos/logos-workernode/logos_worker_node/lane_manager.py
+++ b/logos/logos-workernode/logos_worker_node/lane_manager.py
@@ -53,13 +53,29 @@ _GPU_PLACEMENT_HEADROOM_OFFSET_MB = 0.0
 _VLLM_GMU_FLOOR = 0.5
 _CRASH_RESTART_COOLDOWN_S = 30.0
 _MAX_CRASH_RESTARTS = 5  # per lane; budget resets on confirmed successful restart
+
+
 # Bounded drain of in-flight requests before sleeping/stopping a lane, so an
 # active generation is not torn down mid-stream (client sees RemoteProtocolError
 # / a half-open hang). The worker holds the only AUTHORITATIVE, non-stale
 # active-request count: the planner decides to sleep against a heartbeat snapshot
 # up to ~15s old and can miss a request admitted moments earlier. mode="wait"
 # (the planner's default) honours this drain; mode="force" skips it.
-_LANE_SLEEP_DRAIN_TIMEOUT_S = float(os.getenv("LOGOS_LANE_SLEEP_DRAIN_TIMEOUT_S", "30") or 30)
+def _env_positive_float(name: str, default: float) -> float:
+    """Parse a positive float from the environment, falling back to ``default``
+    for missing, non-numeric, or non-positive values — so a bad override can
+    neither crash import nor silently DISABLE the drain (a 0/negative value)."""
+    raw = os.getenv(name)
+    if raw is None or raw.strip() == "":
+        return default
+    try:
+        val = float(raw)
+    except (TypeError, ValueError):
+        return default
+    return val if val > 0 else default
+
+
+_LANE_SLEEP_DRAIN_TIMEOUT_S = _env_positive_float("LOGOS_LANE_SLEEP_DRAIN_TIMEOUT_S", 30.0)
 _LANE_SLEEP_DRAIN_POLL_S = 0.25
 # Minimum consecutive transport-level failures from /is_sleeping before
 # the liveness branch of _check_stuck_lanes will arm.  3 misses on a 5-second
@@ -237,6 +253,13 @@ def _create_handle(
 
 class _ApplyAbort(Exception):
     """Internal sentinel to trigger rollback in apply_lanes."""
+
+
+class LaneNotServingError(RuntimeError):
+    """A lane cannot accept an inference request right now (missing, sleeping, or
+    not routable). Raised by ``acquire_lane_for_infer`` so the bridge fails the
+    request cleanly and the orchestrator reroutes/wakes — instead of streaming to
+    a lane that is being torn down (which surfaces as RemoteProtocolError)."""
 
 
 class LaneManager:
@@ -545,6 +568,11 @@ class LaneManager:
                             and elc.vllm
                             and elc.vllm_config is not None
                             and elc.vllm_config.enable_sleep_mode
+                            # Never stagger-sleep a lane that has in-flight requests:
+                            # this runs under self._lock and acquire_lane_for_infer
+                            # increments under the same lock, so a non-zero count
+                            # here means a live stream that level-2 sleep would sever.
+                            and self._active_requests.get(existing_lid, 0) == 0
                         ):
                             try:
                                 await existing_h.sleep(level=2, mode="wait")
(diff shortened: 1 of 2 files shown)


logos/logos-workernode/logos_worker_node/lane_manager.py
```diff
@@ -398,6 +406,17 @@ async def apply_lanes(self, desired: list[LaneConfig]) -> LaneApplyResult:
...
+        # model swap — e.g. a logos-nosleep eviction that fully unloads a lane to
+        # free its GPU for another model — does not sever an in-flight generation
+        # mid-stream. Only lanes actually being removed are drained, so a no-op or
+        # add-only apply pays nothing. The transactional removal below (Phase 1)
+        # then runs under the lock as before.
+        predrain_ids = [
+            lid for lid in list(self._handles.keys()) if lid not in desired_lid_set and lid not in self._static_lane_ids
+        ]
+        for lid in predrain_ids:
+            await self._drain_lane_inflight(lid)
```
Claudia-Anthropica, line 442: @wasnertobias [high] This still misses the add-lane path later in `apply_lanes`: Phase 3 calls `existing_h.sleep(level=2, mode="wait")` directly for every existing sleep-capable vLLM lane while holding `_lock`, so it bypasses both `_drain_lane_inflight()` and the final active-request recheck in `sleep_lane()`. Adding capacity under load can still sleep a lane with active streams and reproduce the dropped-request bug; include those staggered-startup sleep candidates in the drain/recheck flow before calling `handle.sleep()`.

In `logos/logos-workernode/logos_worker_node/lane_manager.py`, `apply_lanes()` drains only removed lanes before taking `_lock`, but the later staggered-startup branch still calls `existing_h.sleep(level=2, mode="wait")` directly for existing vLLM lanes when adding new lanes. Fix this by draining and re-checking active requests for those staggered-startup sleep candidates before sleeping them, or route that path through the same guarded logic used by `sleep_lane()` without holding the manager lock during the drain.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The worker-side direction is right, but the current patch still has a lane-sleep path that bypasses the new drain and can drop in-flight requests. I also agree with the existing CodeRabbit comments around the other drain races, so this needs another pass before merging.

wasnertobias:
Pushed a revision addressing the review (commit 091e39b24):

- **Atomic acquire (the core race, @coderabbitai line 802 + @Claudia-Anthropica):** added `LaneManager.acquire_lane_for_infer()` which validates routable **and** increments `_active_requests` under a single lock hold; `logos_bridge` (both infer + stream handlers) now calls it instead of resolve-then-increment, so a lane can no longer be slept/evicted in the gap. Raises `LaneNotServingError` → bridge emits a clean failure and the orchestrator reroutes. Removed the now-unused `_resolve_lane_for_infer`.
- **`remove_lane` drain gap (line 419):** drain + detach now happen under the same lock hold (bounded force-after-timeout), so no request can be routed onto the lane between the drain and the detach.
- **`apply_lanes` Phase 3 (@Claudia-Anthropica):** the staggered-startup `existing_h.sleep(level=2)` now skips any lane with `_active_requests > 0` (it keeps its VRAM instead, treated like a non-sleep-capable lane), so adding capacity under load no longer severs an active stream.
- **Defensive parse (line 62):** `LOGOS_LANE_SLEEP_DRAIN_TIMEOUT_S` now falls back to the 30s default for missing/non-numeric/≤0 values instead of crashing import or disabling the drain.

Not yet deployed — will rebuild the worker image and validate on the test nodes (swap-heavy run, confirm 0 RemoteProtocolError) before this is merge-ready.

logos/logos-workernode/logos_worker_node/lane_manager.py
Claudia-Anthropica: @wasnertobias Looks good now, thanks. The staggered sleep path now checks active requests under the same lock as inference acquisition, so it no longer bypasses the guard.

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -1144,27 +1144,19 @@ def _lane_target_url(
...
             raise ValueError("payload must be an object")
 
-        lane_status = await self._resolve_lane_for_infer(lane_id)
-        request_path = params.get("request_path")
-        target_url = self._lane_target_url(lane_status, payload, request_path=request_path)
-
-        await lane_manager.increment_active_requests(lane_id)
+        # Atomically validate-and-count the lane (closes the dispatch-to-sleep
+        # race: the lane cannot be slept/evicted between selection and counting).
+        lane_status = (await lane_manager.acquire_lane_for_infer(lane_id)).model_dump(mode="json")
```
Claudia-Anthropica, line 1156: @wasnertobias [medium] The worker-node Test check is failing here because tests/test_logos_bridge.py still installs a LaneMgr fake with get_lane_status/increment_active_requests, but _execute_infer_command() now calls acquire_lane_for_infer(). Update that fake and its assertion to match the new atomic acquire API so CI passes.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn logos/logos-workernode/tests/test_logos_bridge.py, test_execute_infer_command_passthrough still mocks the old get_lane_status/increment_active_requests flow while logos_bridge.py now calls acquire_lane_for_infer(). Update the test double to provide acquire_lane_for_infer() returning the same LaneStatus, remove the stale increment assertion, and keep the decrement assertion.\n\n</details>

Claudia-Anthropica (changes requested):
@wasnertobias The previous lane-drain feedback is addressed, and I resolved that thread. One worker-node CI failure remains because the bridge test double still uses the old lane-manager API; please fix that before merging.

PR #643, opened 2026-06-29
`Logos`: Parallelity-aware lane placement (calibrated max_num_seqs)

wasnertobias:
## Problem

The 5-model benchmark's **gemma-3-4b** (the *smallest* model, real footprint ~6 GB) never warmed up: worker auto-placement rejected its lane with `no feasible GPU subset … per-GPU threshold≈24570MB`. Root cause — the placement guard demands `_VLLM_GMU_FLOOR (0.5) × GPU-total (~49 GB) ≈ 24.5 GB free` for **every** lane.

That premise is false for our lanes: when a lane pins `kv_cache_memory_bytes`, vLLM **skips memory profiling and ignores `gpu_memory_utilization`** (it logs this verbatim), reserving only weights + the explicit KV. So the floor demanded ~24.5 GB the lane never uses, and starved the smallest model → warmup timeout → dropped requests.

## Fix

Replace the blunt 0.5 floor with a calibrated, *measured* signal — the **parallelity factor** (achievable concurrency = `kv_cache_tokens / max_model_len`, i.e. vLLM's `Maximum concurrency … X.XXx`):

- **Calibration** records parallelity at each KV sweep point, read back from vLLM's own load log (no need to pin `--max-num-seqs`). `kv_cache_to_max_model_len_pairs` entries gain a `parallelity` key; legacy profiles without it still work.
- **Planner** spawn-time KV selector targets **parallelity ≥ 2** (≥2 concurrent full-context requests) when a fitting calibrated point provides it; falls back to highest-available / smallest-KV under memory pressure.
- **Worker** GMU-floor guard is **skipped when `kv_cache_memory_bytes` is set**, so kv-pinned lanes place on their real footprint. The floor still applies to GMU-sized lanes.

## Tests

- Calibration: concurrency parsing + profile-dict emission (incl. legacy fallback).
- Schema: parallelity round-trips through the worker override-merge.
- Planner: targets parallelity ≥ 2, respects free-VRAM budget, falls back to highest available.
- Worker: floor skipped for kv-pinned lanes; still applied to GMU-sized lanes.

Requires full recalibration of all 5 models after deploy (to populate parallelity + lean footprints).

## Summary by CodeRabbit

* **New Features**
  * Calibration now captures achievable concurrency as an optional “parallelity” signal and includes it in generated KV→max context tuning.
  * Model profile overrides can preserve an optional parallelity value per KV entry.
* **Bug Fixes**
  * Capacity planning now selects KV settings using parallelity-aware logic when available, with sensible legacy fallback when it isn’t.
  * GPU auto-placement now correctly handles explicitly pinned KV cache memory without applying an overly strict memory floor.
* **Tests**
  * Added unit coverage for concurrency/parallelity parsing, profile serialization, override merging, parallelity-aware capacity selection, and KV-pinned GPU placement.

### DIFF
--- a/logos/logos-workernode/logos_worker_node/calibration.py
+++ b/logos/logos-workernode/logos_worker_node/calibration.py
@@ -644,6 +644,32 @@ def _extract_vllm_max_num_seqs_suggestion(log_tail: str) -> int | None:
     return value if value > 0 else None
 
 
+# vLLM prints the achievable concurrency at every engine init, e.g.
+#   "Maximum concurrency for 33,888 tokens per request: 2.00x"
+# This is total_kv_cache_tokens / max_model_len — i.e. how many simultaneous
+# full-context requests the KV pool can serve. We read it back (rather than
+# pinning --max-num-seqs) to record the "parallelity factor" of each KV point.
+_VLLM_MAX_CONCURRENCY_RE = re.compile(r"Maximum concurrency for [\d,]+ tokens per request:\s*([\d.]+)x")
+
+
+def _extract_vllm_max_concurrency(log_tail: str) -> float | None:
+    """Return vLLM's reported achievable concurrency (the ``X.XXx`` factor), else None.
+
+    Uses the LAST occurrence in the log so a re-probe at a different KV size
+    reflects the final successful load rather than an earlier attempt.
+    """
+    if not log_tail:
+        return None
+    matches = _VLLM_MAX_CONCURRENCY_RE.findall(log_tail)
+    if not matches:
+        return None
+    try:
+        value = float(matches[-1])
+    except (TypeError, ValueError):
+        return None
+    return value if value > 0 else None
+
+
 # ---------------------------------------------------------------------------
 # GPU VRAM helpers
 # ---------------------------------------------------------------------------
@@ -1119,6 +1145,13 @@ class CalibrationResult:
     # Each tuple is (kv_cache_mb, max_model_len) and represents one point on
     # the reachable context curve for this model on this hardware.
     kv_max_model_len_pairs: list[tuple[float, int]] | None = None
+    # Same sweep points annotated with the achievable concurrency (parallelity
+    # factor = kv_cache_tokens / max_model_len) vLLM reported at each KV size.
+    # Each tuple is (kv_cache_mb, max_model_len, parallelity). ``parallelity`` is
+    # None when neither a probed anchor nor the KV→context rate could supply it.
+    # The planner uses it to place lanes with enough KV for >=2 concurrent
+    # full-context requests when memory allows.
+    kv_max_model_len_parallelity_pairs: list[tuple[float, int, float | None]] | None = None
     # ``max_num_seqs`` actually used during the successful probe(s). Set when
     # calibration auto-injected --max-num-seqs because a hybrid Mamba/SSM
     # model's state-cache pool was smaller than the default 1024. Persisted so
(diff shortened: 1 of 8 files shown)


logos/logos-workernode/logos_worker_node/calibration.py
```diff
@@ -1904,6 +1944,19 @@ def _cancelled() -> bool:
...
+            # KV→context rate (per_token_bytes). None when neither is available.
+            _anchor_par = {round(k, 1): v for k, v in anchors_parallelity.items()}
+            _triples: list[tuple[float, int, float | None]] = []
+            for kv_mb, mml in kv_max_model_len_pairs:
+                par = _anchor_par.get(round(kv_mb, 1))
+                if par is None and per_token_bytes and mml > 0:
+                    kv_tokens = (kv_mb * 1024.0 * 1024.0) / per_token_bytes
+                    par = kv_tokens / float(mml)
+                _triples.append((kv_mb, int(mml), (round(float(par), 3) if par and par > 0 else None)))
+            partial.kv_max_model_len_parallelity_pairs = _triples
```
Claudia-Anthropica, line 1959: @wasnertobias [high] This records the new triples only on `partial`, but the success path below returns a fresh `CalibrationResult` without `kv_max_model_len_parallelity_pairs=partial.kv_max_model_len_parallelity_pairs`. `result_to_profile_dict(result)` therefore falls back to legacy pairs, so the full recalibration requested in the PR still will not publish `parallelity` and the planner never gets the calibrated signal.

In `logos/logos-workernode/logos_worker_node/calibration.py`, `calibrate_model()` stores `kv_max_model_len_parallelity_pairs` on `partial` but the returned `CalibrationResult` omits that field. Add `kv_max_model_len_parallelity_pairs=partial.kv_max_model_len_parallelity_pairs` to the success return and add a regression that serializes the returned result so `parallelity` is persisted.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The planner changes are pointed in the right direction, but the calibration result drops the new parallelity pairs before persistence. See the inline comment; without that field, recalibration still produces legacy pair data.

wasnertobias:
Addressed all review feedback in 1a2b580b5:

1. **(root cause)** Calibration set the `kv_max_model_len_parallelity_pairs` triples on the working `partial` object, but the success path returned a freshly-constructed `CalibrationResult` that didn't copy the field through — so the parallelity values never reached `model_profiles.yml` (this is why a clean run still showed pairs with only `kv_mb`/`max_model_len`). Now propagated like its sibling fields; validated on the next recalibration. Thanks @coderabbitai / @Claudia-Anthropica for catching this.
2. With the GMU floor skipped for kv-pinned lanes, `_estimate_lane_vram_mb` is the only placement guard — for profiles with `loaded_vram_mb` but no base residency it returned the KV size alone (omitting weights). Now falls back to `max(loaded_vram_mb, kv_mb)` when base residency is absent.
3. Removed the redundant `@pytest.mark.asyncio` (asyncio_mode='auto').

Full worker (173) + orchestrator capacity (143) suites pass; pre-commit clean.

logos/logos-workernode/logos_worker_node/calibration.py
Claudia-Anthropica: @wasnertobias Looks good now, thanks.

Claudia-Anthropica (approved):
@wasnertobias All feedback addressed, nice work. Focused Logos capacity and worker-node test subsets passed locally; the remote checks I saw had no failures, only pending build/review jobs.

PR #652, opened 2026-07-03
`Iris`: Reduce chat first-answer latency

bassner:
## Summary

Production Iris chat averages ~39s of user-visible response time. This PR reduces the time to the **first assistant answer** across all chat modes (`COURSE_CHAT`, `LECTURE_CHAT`, `PROGRAMMING_EXERCISE_CHAT`, `TEXT_EXERCISE_CHAT`) by removing blocking work from the critical path and parallelizing Weaviate round trips. No observable behavior changes: first-callback payload semantics, stage bookkeeping per mode, current-view prompt injection, citation content, retrieval result sets, and Memiris paths are all explicitly preserved.

## Description

1. **Deferred session title generation.** Session title generation (an LLM call on every turn, in all modes) no longer blocks the first `done(final_result=...)` callback. It now runs *after* the answer is delivered and rides the next outgoing callback: the suggestions callback for course/exercise chat, the trailing memory-stage callback otherwise, or the error callback if suggestion generation fails (an error callback terminates the Artemis job, so a later delivery would be rejected).
2. **Lazy MCQ tool data fetch.** The MCQ generation tool no longer eagerly fetches *all* course page chunks from Weaviate at tool construction time on every course/lecture message. The fetch is deferred into the tool closure so it only runs if the agent actually calls the tool, and it reuses the precomputed lecture-tool availability flag instead of a redundant Weaviate check.
3. **Reused OpenAI/Azure SDK clients.** SDK clients are now created once per model entry (via pydantic `model_post_init`, mirroring the existing embeddings pattern) instead of once per LLM call — saving a TCP+TLS handshake on every call. Clients get an explicit 300s timeout (matching the Artemis job lifetime) and `max_retries=0` so only the existing manual backoff loop retries (the SDK's default of 2 internal retries was silently multiplying attempts).
4. **Lecture retrieval N+1 fix.** The three lecture sub-retrievals now cache lecture-unit metadata per request in `generate_retrieval_dtos` (previously issued one identical Weaviate lookup per search hit, up to ~60 per tool call). The two hybrid searches per sub-retrieval now run concurrently, and the per-segment transcription/page-chunk fetch loop (up to 14 sequential round trips) is now parallel.
5. **Concurrent tool-availability checks.** `prepare_state` now runs the two tool-availability Weaviate checks concurrently instead of sequentially.
[...]

PR #654, opened 2026-07-03
`Iris`: Make reasoning effort configurable per task and upgrade advanced chat to GPT-5.5

bassner:
> [!NOTE]
> Stacked on #652 (`iris/perf/chat-first-answer-latency`) — the base retargets to `main` automatically once #652 merges. Review only the last commit.

## Summary

Second slice of the chat-latency work: every LLM call in the chat path currently runs at the provider's *default* reasoning effort — for `gpt-5-mini`/`-nano` that is `medium`, which burns 64–128+ hidden reasoning tokens even on trivial calls like session-title generation or query rewrites. This PR makes reasoning effort a YAML-configured, per-model-entry setting so mechanical tasks run at `minimal`/`low` while the main chat agent keeps full reasoning, and upgrades the advanced chat variant from GPT-5.4 to GPT-5.5.

## Description

- **Per-entry effort config**: `llm_config.yml` chat entries gain `reasoning_effort` (default applied when the call doesn't set one) and `reasoning_effort_values` (what the backend actually accepts). Values outside the list are clamped to the nearest allowed value with a warning. Allowed sets differ per backend and are hard 400s when violated: gpt-5/-mini/-nano = `minimal|low|medium|high`; gpt-5.2/5.4/5.5 = `none|low|medium|high|xhigh` (`minimal` rejected on 5.5, verified live); Logos/vLLM gpt-oss = `low|medium|high` only.
- **Fail fast on 4xx**: the retry loop caught broad `APIError`, so a deterministic 400 was retried 5× with 31s of pointless backoff. Now only 429, connection/timeouts, 5xx, and 408/409 retry (408/409 are transient per the OpenAI SDK's own retry predicate; the cached client runs `max_retries=0`).
- **Startup validation hard-fails** on: `llm_configuration` roles referencing model ids missing from the catalog (previously only surfaced per-request), entry defaults outside their own `reasoning_effort_values`, effort fields without `supports_reasoning_effort`, and empty/null-bearing value lists. Ops edits these YAMLs on live servers — misconfigurations should fail at boot, not at request time.
- **global_search** hyde/answer efforts move from hardcoded Python (`reasoning_effort="none"`) to YAML-pinned `*-minimal` entries (same-rollout config update required, included in the deployment runbook).
- **Example configs**: effort-suffixed entries (`oai-gpt-5-mini-minimal`, `oai-gpt-5-nano-minimal`, `oai-gpt-5-nano-low`), per-role mapping (main chat pinned `medium`, citations/rewrites/title/suggestions `minimal`, MCQ `low`), `gpt-5.4` → `gpt-5.5` ($5/$30 per 1M tokens), and the four dead legacy per-mode chat blocks replaced with the unified `chat_pipeline` block the code actually resolves — a fresh install following the example could not run chat at all.

## Validation

- `poetry run pytest`: **166 passed** (10 new tests: retry classification incl. 408/409, effort resolution/clamping, config validation)
- `poetry run pylint src/iris`: **10.00/10**; `black`/`isort` clean
- **Live end-to-end**: local Pyris (this branch) against real Azure deployments with a stub Artemis callback receiver — identical COURSE_CHAT prompts with the main chat entry pinned at `minimal` vs `high` produced **63 vs 504 output tokens** (same 2646 input tokens), confirming the YAML → resolution → clamp → API chain end to end; title/suggestions ran on `gpt-5-nano-minimal` (11/52 output tokens); all callbacks delivered in order.
- Effort value sets verified against OpenAI model cards, Azure docs, and the vLLM 0.24.0 source (gpt-oss harmony path), plus live probes of every `ase-se01` deployment and Logos.
- Cross-vendor review: implemented by Codex (GPT-5.5, xhigh), reviewed by Claude; final diff re-reviewed by Codex — explicit approval after 408/409 and validator-hardening fixes.

---
**Update (live-testing follow-up):** Azure gpt-5.5 rejects `reasoning_effort` combined with function tools on `/chat/completions` (hard 400 telling you to use `/v1/responses`). Added a per-entry `use_responses_api` switch with full Responses-API mapping (input items incl. tool calls/results, flattened tool schemas, `reasoning.effort`, JSON mode, `store: false`, usage mapping, Azure `/openai/v1` client) so the effort pin keeps working on the flagship agent loop. Live-verified against Azure gpt-5.5: tools + pinned `medium` → clean tool call. 171 tests on this branch.

## Summary by CodeRabbit

* **New Features**
  * Added support for newer chat model configuration options, including reasoning-effort settings and Responses API usage.
  * Expanded cloud model presets across several pipelines for more consistent model selection.

* **Bug Fixes**
  * Improved LLM configuration validation to catch invalid or unknown model references earlier.
  * Updated chat handling to better support retries, tool calls, usage reporting, and refusal detection.
  * Refined Azure and OpenAI model behavior for compatible Responses API support.

### DIFF
--- a/iris/src/iris/llm/llm_configuration.py
+++ b/iris/src/iris/llm/llm_configuration.py
@@ -12,7 +12,8 @@ class LlmConfigurationError(ValueError):
 
 
 def validate_llm_configuration(
-    config: Mapping[str, Mapping[str, Mapping[str, Mapping[str, str]]]] | None = None,
+    config: Mapping[str, Mapping[str, Mapping[str, object]]] | None = None,
+    known_model_ids: set[str] | None = None,
 ) -> None:
     """
     Validate structural completeness of llm_configuration.
@@ -23,6 +24,7 @@ def validate_llm_configuration(
     """
     cfg = config if config is not None else settings.llm_configuration
     missing: list[str] = []
+    unknown: list[str] = []
 
     for pipeline_id, variants in cfg.items():
         if not isinstance(variants, dict):
@@ -43,6 +45,13 @@ def validate_llm_configuration(
                         missing.append(
                             f"llm_configuration.{pipeline_id}.{variant_id}.{role} (empty)"
                         )
+                    elif (
+                        known_model_ids is not None and role_cfg not in known_model_ids
+                    ):
+                        unknown.append(
+                            f"llm_configuration.{pipeline_id}.{variant_id}.{role} "
+                            f"references unknown model id '{role_cfg}'"
+                        )
                     continue
 
                 if not isinstance(role_cfg, dict):
@@ -61,10 +70,27 @@ def validate_llm_configuration(
                             f"llm_configuration.{pipeline_id}.{variant_id}.{role}.{env}"
                         )
 
-    if missing:
-        raise LlmConfigurationError(
-            "Missing llm configuration entries:\n" + "\n".join(missing)
-        )
+                if known_model_ids is not None:
+                    for env, value in role_cfg.items():
+                        if (
+                            isinstance(value, str)
+                            and value
+                            and value not in known_model_ids
+                        ):
+                            unknown.append(
+                                f"llm_configuration.{pipeline_id}.{variant_id}.{role}.{env} "
+                                f"references unknown model id '{value}'"
+                            )
+
+    if missing or unknown:
+        details = []
+        if missing:
+            details.append("Missing llm configuration entries:\n" + "\n".join(missing))
+        if unknown:
+            details.append(
+                "Unknown llm configuration model IDs:\n" + "\n".join(unknown)
+            )
+        raise LlmConfigurationError("\n".join(details))
 
 
 def resolve_model(pipeline_id: str, variant_id: str, role: str, *, local: bool) -> str:
--- a/iris/src/iris/llm/llm_configuration.py
+++ b/iris/src/iris/llm/llm_configuration.py
@@ -71,7 +71,8 @@ def validate_llm_configuration(
                         )
 
                 if known_model_ids is not None:
-                    for env, value in role_cfg.items():
+                    for env in environments:
+                        value = role_cfg.get(env)
                         if (
                             isinstance(value, str)
                             and value
--- a/iris/application.example.yml
+++ b/iris/application.example.yml
@@ -59,7 +59,9 @@ env_vars:
 
 # Values are LLM entry IDs (the "id" field in llm_config.yml).
 llm_configuration:
-  exercise_chat_pipeline:
+  # Unified chat pipeline (COURSE_CHAT, LECTURE_CHAT, PROGRAMMING_EXERCISE_CHAT,
+  # TEXT_EXERCISE_CHAT all resolve through this single pipeline id).
+  chat_pipeline:
     default:
       chat:
         local: oai-gpt-5-mini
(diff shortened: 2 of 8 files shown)


iris/src/iris/llm/llm_configuration.py
```diff
@@ -61,10 +70,27 @@ def validate_llm_configuration(
                             f"llm_configuration.{pipeline_id}.{variant_id}.{role}.{env}"
                         )
 
-    if missing:
-        raise LlmConfigurationError(
-            "Missing llm configuration entries:\n" + "\n".join(missing)
-        )
+                if known_model_ids is not None:
+                    for env, value in role_cfg.items():
```
Claudia-Anthropica: @bassner [medium] This now validates every value under the role mapping, so a deployment with `local_llm_enabled: false` still fails if an unused `local` entry points at a model that is intentionally absent from `llm_config.yml`. The structural check above already scopes to `("cloud",)` in that mode; the catalog cross-check should use the same environment set so cloud-only deployments can boot.

In `iris/src/iris/llm/llm_configuration.py`, the unknown model ID check validates unused `local` entries even when `settings.local_llm_enabled` is false. Fix this by iterating over the same `environments` tuple used for the missing-entry check and only cross-checking those selected env values.

</details>

Claudia-Anthropica (changes requested):
@bassner The reasoning-effort plumbing and retry split mostly look solid, and the new tests cover the main paths. I’m requesting one change because the catalog validator now rejects unused local model IDs even when local LLMs are disabled; see the inline comment.

iris/src/iris/llm/llm_configuration.py
Claudia-Anthropica: @bassner Still seeing the same issue in the current code: line 74 iterates over every entry in `role_cfg.items()`, so `local` gets catalog-checked even when `settings.local_llm_enabled` is false. This needs to use the `environments` tuple from lines 63-65 for the unknown-ID loop as well.

Claudia-Anthropica (changes requested):
@bassner The previous catalog-validation issue is still open: cloud-only deployments can still fail on unused local model IDs because the unknown-ID loop checks every role entry. Keeping this as request changes until that loop uses the active environment set too.

iris/src/iris/llm/llm_configuration.py
Claudia-Anthropica: @bassner Still open after cdbaec81: the catalog check still iterates over `role_cfg.items()`, so inactive `local` entries are validated when `local_llm_enabled` is false. Please switch that loop to the same `environments` tuple used by the missing-entry check.

Claudia-Anthropica (changes requested):
@bassner The previous catalog-validation issue is still open on current head. Cloud-only deployments can still fail on unused local model IDs because the unknown-ID loop does not use the active environment set, so I’m keeping this as request changes.

iris/src/iris/llm/llm_configuration.py
bassner: Good catch — fixed in 153e967e. The unknown-model-ID cross-check now iterates the same `environments` tuple used by the structural missing-entry check (`("cloud",)` when `local_llm_enabled` is false, `("local", "cloud")` otherwise) instead of walking every key in `role_cfg`. A cloud-only deployment can now boot even if a stale/unused `local` entry references a model absent from `llm_config.yml`, while an unknown `cloud` entry (or an unknown `local` entry when local LLMs are enabled) is still rejected. Added three tests: stale local entry ignored when disabled, cloud entry still validated when disabled, and both envs still validated when enabled.

Claudia-Anthropica: @bassner Looks good now, thanks.

Claudia-Anthropica (dismissed):
@bassner All feedback addressed, nice work.

PR #655, opened 2026-07-03
`Iris`: Speed up lecture retrieval with transcription skipping, shared embeddings, and parallel reranks

bassner:
> [!NOTE]
> Stacked on #654 (`iris/perf/per-role-reasoning-effort`); base retargets automatically as the stack merges. Review only the last commit.

## Summary

Third slice of the chat-latency work: cuts the lecture-retrieval critical path without changing what content can be found. The retrieval currently runs 4 LLM query rewrites (2 of them ≤350-word HyDE generations — the dominant cost), 6 embeddings (2 duplicated), 6 hybrid searches, and 5 Cohere reranks (2 sequential at the end) — even for slide-only courses with no transcriptions at all.

## Description

- **Transcription-domain skip**: a `limit=1` filtered existence probe (concurrent with the lecture-unit lookup, cached per unit including negative results) gates the whole transcription domain *before* the rewrite phase. Slide-only courses save 2 of 4 rewrite LLM calls (including one HyDE), the full transcription sub-pipeline, and up to 7 per-segment transcription fetches. The segment pipeline falls back to pages-domain queries (segments carry slide summaries). The probe is fail-safe: an over-broad match only disables the skip, never skips real content.
- **HyDE 350→120 words, rewrite `max_tokens` 2000→400**: the HyDE decode is the critical-path item; a keyword-dense short pseudo-document serves the embedding just as well.
- **Embed once**: distinct query strings are embedded once in the orchestrator (parallel) and vectors passed down; previously the segment and transcription pipelines each embedded the same two strings. Sub-pipelines keep internal embedding as fallback for standalone use.
- **Parallel final reranks**: the two final Cohere reranks ran sequentially; now concurrent with an empty-safe wrapper.

Expected effect: several seconds on retrieval-bearing turns; largest for slide-only courses. Follow-ups deliberately *not* included pending a retrieval-quality eval harness (dropping the 2 intermediate sub-pipeline reranks, `top_n` 7→5): those change the candidate set and need overlap + known-answer gates first.

## Validation

- `poetry run pytest`: **170 passed** — 4 new tests written red-first (transcription skip incl. cache and query fallback, no-skip path, parallel rerank preservation, embed-once counting)
- `poetry run pylint src/iris`: **10.00/10**; `black`/`isort` clean
- Implemented by Codex (GPT-5.5, xhigh) from a written spec; full diff reviewed by Claude

### DIFF
--- a/iris/src/iris/retrieval/lecture/lecture_retrieval.py
+++ b/iris/src/iris/retrieval/lecture/lecture_retrieval.py
@@ -91,7 +91,7 @@ class LectureRetrieval(SubPipeline):
             pipeline_id, "default", "embedding", local=False
         )
         request_handler = LlmRequestHandler(model_id=chat_model)
-        completion_args = CompletionArguments(temperature=0, max_tokens=2000)
+        completion_args = CompletionArguments(temperature=0, max_tokens=400)
         self.llm = IrisLangchainChatModel(
             request_handler=request_handler, completion_args=completion_args
         )
@@ -107,6 +107,7 @@ class LectureRetrieval(SubPipeline):
         )
 
         self.tokens = []
+        self._transcription_presence_cache: dict = {}
 
         self.lecture_unit_segment_pipeline = LectureUnitSegmentRetrieval(
             client, local=local
@@ -134,12 +135,27 @@ class LectureRetrieval(SubPipeline):
         base_url: str = None,
     ) -> LectureRetrievalDTO:
         with timed_span("LectureRetrieval", "get_lecture_unit"):
-            lecture_unit = self.get_lecture_unit(course_id, lecture_id, lecture_unit_id)
+            with TracedThreadPoolExecutor(max_workers=2) as executor:
+                lecture_unit_future = executor.submit(
+                    self.get_lecture_unit, course_id, lecture_id, lecture_unit_id
+                )
+                transcriptions_exist_future = executor.submit(
+                    self.lecture_unit_has_transcriptions,
+                    course_id,
+                    lecture_id,
+                    lecture_unit_id,
+                )
+                lecture_unit = lecture_unit_future.result()
+                transcriptions_exist = transcriptions_exist_future.result()
         if lecture_unit is None:
             return LectureRetrievalDTO(
                 lecture_transcriptions=[],
                 lecture_unit_page_chunks=[],
                 lecture_unit_segments=[],
+            )
+        if not transcriptions_exist:
+            logger.info(
+                "Lecture retrieval: no transcriptions for unit — skipping transcription domain"
             )
 
         with timed_span("LectureRetrieval", "query_rewrites"):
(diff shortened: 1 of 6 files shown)


Claudia-Anthropica (approved):
@bassner This looks good. The transcription-presence gate, shared embeddings, and parallel final reranks are covered by focused regression tests, and I don't see a behavior change that should block this.

iris/src/iris/retrieval/lecture/lecture_retrieval.py
```diff
@@ -837,39 +950,64 @@ def call_lecture_pipelines(
...
-            )
-            lecture_transcriptions_future = executor.submit(
-                self.lecture_transcription_pipeline,
-                student_query,
-                lecture_transcriptions_query,
-                hypothetical_lecture_transcriptions_answer_query,
+                segment_query,
+                segment_hypothetical_answer,
                 lecture_unit,
+                rewritten_query_vector=query_vectors.get(segment_query),
```
Claudia-Anthropica, line 1023: @bassner [medium] These vectors are always produced with `lecture_retrieval_pipeline`'s embedding model, but the segment and transcription pipelines still resolve their own embedding model IDs. If a deployment configures those retrieval pipelines differently, hybrid search gets vectors from the wrong model/dimension and can return bad results or fail. Please compute shared vectors per target pipeline embedding model, or centralize/assert that all three lecture retrieval pipelines use the same embedding model before passing vectors across.

In `iris/src/iris/retrieval/lecture/lecture_retrieval.py`, `call_lecture_pipelines` passes vectors embedded by `LectureRetrieval.llm_embedding` into the segment and transcription sub-pipelines even though those pipelines resolve their own embedding model IDs. Fix this by either computing the pre-embedded vectors with each target pipeline's embedding model or enforcing a single shared lecture-retrieval embedding model before passing vectors across pipeline boundaries.

</details>

Claudia-Anthropica (changes requested):
@bassner The latency changes are well scoped and CI is green, but the shared-embedding path now crosses configurable pipeline boundaries. I left one inline note where that can break deployments that do not use the exact same embedding model for every lecture retrieval sub-pipeline.

iris/src/iris/retrieval/lecture/lecture_retrieval.py
bassner: Good catch — real hazard. The three lecture retrieval pipelines are independent `llm_configuration` entries and each resolves its own embedding model, so nothing guaranteed they agree once we started reusing one set of query vectors across them. Rather than re-embed per pipeline (which would defeat the embed-once speedup this PR adds), I enforce the invariant the optimization relies on: `LectureRetrieval.__init__` now calls `_assert_shared_embedding_model()`, which compares the orchestrator's embedding `model_id` against each sub-pipeline's and raises `LlmConfigurationError` with the mismatched pipelines/models if they diverge. Same model_id ⇒ same vector space and dimension, so shared vectors are safe; a misconfigured deployment fails fast at construction with an actionable message. Added tests for the aligned and divergent cases. Fixed in 6cae2fb.

Claudia-Anthropica: @bassner Looks good now, thanks. The constructor guard covers the shared-vector invariant and the added tests exercise the mismatch path.

Claudia-Anthropica (dismissed):
@bassner All feedback addressed; the shared-embedding invariant is now enforced and covered by tests. CI is green, so approving.

PR #656, opened 2026-07-03
`Iris`: Keep exercise guide refinement blocking on the guide model

bassner:
> [!NOTE]
> Stacked on #654 (`iris/perf/per-role-reasoning-effort`); base retargets as the stack merges. Review only the last commit. Sibling of #655.

## Summary

This PR keeps the programming-exercise guide refinement as an always-on blocking safety pass, but removes the runtime `exercise_guide_refinement` / shadow-mode configuration. The guard still runs before the final answer is sent, and it now uses the dedicated `guide` model role instead of the flagship chat model where configured.

## Evidence

Two eval rounds falsified the original plan to remove the second-pass guard:

| Config | Leak failures |
|---|---|
| Two-pass guard, gpt-5-mini @ medium | **21-23/90** |
| Single-pass, new self-check prompt | 35/90 |
| Single-pass, old prompt | 37/90 |
| Guard @ gpt-5-mini minimal | 34/90 |
| Guard @ gpt-5-nano low | 30/90 |

Conclusion: prompt-only compensation does not close the leak gap, and the guard needs medium reasoning effort. Runtime shadow/off switches would make the safety behavior harder to reason about, so they were removed from this PR.

## Description

- Keeps guide refinement blocking for programming-exercise chat.
- Removes `exercise_guide_refinement` and `exercise_guide_refinement_shadow_sample` from settings and `application.example.yml`.
- Keeps the optional `guide` model role with fallback to `chat` for older configs.
- Keeps the prompt self-check additions as defense in depth.
- Fixes guide failures to log a warning and deliver the original answer instead of sending an error callback and then continuing.

## Validation

- `poetry run pytest tests/test_exercise_guide_refinement_modes.py -q` — 9 passed.

### DIFF
--- a/iris/application.example.yml
+++ b/iris/application.example.yml
@@ -36,6 +36,15 @@ memiris:
 
 local_llm_enabled: true # Set to false to disable local LLM support entirely
 
+# Programming-exercise guide refinement:
+# - blocking: run the guide rewrite before the answer is sent; this is the
+#   default and reflects production behavior.
+# - shadow: send the answer first, then run guide refinement for evidence collection.
+# - off: disable guide refinement entirely.
+exercise_guide_refinement: blocking
+# Shadow sampling rate, only used when exercise_guide_refinement is shadow.
+exercise_guide_refinement_shadow_sample: 1.0
+
 langfuse:
   enabled: false # Set to true to enable LangFuse tracing
   public_key: "pk-lf-..." # Your LangFuse public key
@@ -66,10 +75,18 @@ llm_configuration:
       chat:
         local: oai-gpt-5-mini
         cloud: oai-gpt-5-mini
+      # Guard detection quality requires medium reasoning effort (eval-backed).
+      guide:
+        local: oai-gpt-5-mini
+        cloud: oai-gpt-5-mini
     advanced:
       chat:
         local: oai-gpt-52
         cloud: oai-gpt-5.5
+      # Guard detection quality requires medium reasoning effort (eval-backed).
+      guide:
+        local: oai-gpt-52
+        cloud: oai-gpt-5-mini
 
   autonomous_tutor_pipeline:
     default:
--- a/iris/application.example.yml
+++ b/iris/application.example.yml
@@ -36,15 +36,6 @@ memiris:
 
 local_llm_enabled: true # Set to false to disable local LLM support entirely
 
-# Programming-exercise guide refinement:
-# - blocking: run the guide rewrite before the answer is sent; this is the
-#   default and reflects production behavior.
-# - shadow: send the answer first, then run guide refinement for evidence collection.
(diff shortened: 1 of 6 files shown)


Claudia-Anthropica (approved):
@bassner This looks solid to me. The new refinement modes keep blocking as the default, the guide-role fallback preserves old configs, and the tests cover the callback ordering and prompt changes. Approving; CI is green now.

Claudia-Anthropica (dismissed):
@bassner This still looks solid after the base-branch update. The configurable blocking/shadow/off paths preserve the existing blocking behavior by default, the guide-role fallback is covered, and the prompt/test updates match the stated intent.

iris/application.example.yml
```diff
@@ -66,10 +75,18 @@ llm_configuration:
...
+      guide:
+        local: oai-gpt-5-mini
+        cloud: oai-gpt-5-mini
     advanced:
       chat:
         local: oai-gpt-52
         cloud: oai-gpt-5.5
+      # Guard detection quality requires medium reasoning effort (eval-backed).
+      guide:
+        local: oai-gpt-52
```
Claudia-Anthropica: @bassner [low] This local guide mapping still resolves to `oai-gpt-52`, but the example catalog configures that ID with `reasoning_effort: none` while the new guard comment says medium effort is required. For `LOCAL_AI` advanced runs, this would keep the guard on the unevaluated/no-effort model; point it at `oai-gpt-5-mini` too, or add a separate medium-effort 5.2 guide model ID.

In `iris/application.example.yml`, `chat_pipeline.advanced.guide.local` points to `oai-gpt-52`, whose example LLM entry uses `reasoning_effort: none` despite the guard requiring medium reasoning. Change this local guide mapping to a medium-reasoning guide model, such as `oai-gpt-5-mini`, or add and reference a medium-effort 5.2 model ID.

</details>

Claudia-Anthropica (dismissed):
@bassner The implementation looks solid overall and the Iris lint/test checks are green. I left one low-priority config consistency note for the advanced local guide mapping, but the blocking/shadow/off behavior and callback ordering look good to me.

iris/application.example.yml
bassner: Good catch — fixed in 7b672c7b. Pointed `chat_pipeline.advanced.guide.local` at `oai-gpt-5-mini` (matching the cloud guide and the default guide) so LOCAL_AI advanced runs use an eval-backed medium-effort guide model instead of `oai-gpt-52` (reasoning_effort: none).

Claudia-Anthropica (dismissed):
@bassner Looks good after the guide-model follow-up. The advanced local guide now points at the same medium-effort mini model as the cloud/default guide, and the blocking/shadow/off behavior still lines up with the tests and green CI.

PR #657, opened 2026-07-03
`Iris`: Improve prompt cache hits, stage-callback overhead, and trace nesting

bassner:
> [!NOTE]
> Stacked on #656 (`iris/perf/exercise-guide-shadow-mode`); base retargets as the stack merges. Review only the last commit.

## Summary

Three independent latency/observability fixes folded in from the latency investigation. This branch now also carries the updated #656 behavior: guide refinement is not runtime-configurable and stays blocking for programming-exercise chat.

## Description

- **Prompt-cache prefix alignment**: the chat system prompt's first line was `Current Date: <timestamp>` at second granularity, invalidating provider prompt caching for the stable prompt prefix. The date and per-turn current-view blocks now render at the end of the system prompt, with a template comment documenting the stable-prefix-first rule.
- **Async intermediate stage callbacks**: cosmetic `in_progress` updates POST asynchronously through a lazy single-worker FIFO sender. Terminal sends drain the queue first and stay synchronous, preserving Artemis-observed callback order.
- **Langfuse trace nesting fix**: `TracingContext` now uses a `ContextVar`, so traced worker threads keep the Iris trace context. The global-search worker pool is traced as well.
- **Stack carry-through**: resolves the #656 merge so `exercise_guide_refinement` and `exercise_guide_refinement_shadow_sample` stay removed in this branch.

## Validation

- Original PR validation: full unit suite, pylint, black/isort/bandit.
- Stack update validation: `poetry run pytest tests/test_exercise_guide_refinement_modes.py -q` — 9 passed.

### DIFF
--- a/iris/application.example.yml
+++ b/iris/application.example.yml
@@ -36,6 +36,15 @@ memiris:
 
 local_llm_enabled: true # Set to false to disable local LLM support entirely
 
+# Programming-exercise guide refinement:
+# - blocking: run the guide rewrite before the answer is sent; this is the
+#   default and reflects production behavior.
+# - shadow: send the answer first, then run guide refinement for evidence collection.
+# - off: disable guide refinement entirely.
+exercise_guide_refinement: blocking
+# Shadow sampling rate, only used when exercise_guide_refinement is shadow.
+exercise_guide_refinement_shadow_sample: 1.0
+
 langfuse:
   enabled: false # Set to true to enable LangFuse tracing
   public_key: "pk-lf-..." # Your LangFuse public key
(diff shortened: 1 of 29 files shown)


Claudia-Anthropica (approved):
@bassner This looks good to me. The prompt volatile blocks are moved out of the cacheable prefix, the async callback drain preserves terminal ordering, and the ContextVar tracing change fits the threaded paths in this diff.

Claudia-Anthropica (dismissed):
@bassner This still looks good to me after the base refresh. The prompt volatile blocks stay out of the cacheable prefix, callback ordering is preserved by draining before terminal sends, and the tracing context propagation fits the threaded paths in this diff.

PR #658, opened 2026-07-03
`Iris`: Stream the answer to the user while it generates

bassner:
> [!NOTE]
> Stacked on #657; base retargets as the stack merges (#654 -> #656 -> #657 -> this). Review only the last commit. **Artemis counterpart: ls1intum/Artemis#13099** — both halves are independently deployable; streaming activates only when BOTH are deployed and `artemis.iris.response-streaming-enabled` is on.

## Summary

Pyris can stream chat answers when the caller opts in with `streamResponse: true`. The important safety rule after the stack update: programming-exercise chat does **not** stream raw main-model output. The guide refinement model owns streaming there, so users only see the answer that passed the guide pass.

## Description

- **Streaming LLM path** (`openai_chat.py`): opt-in via `CompletionArguments.stream_handler`. Content deltas are forwarded; tool-call deltas reset visible draft text; usage is captured for streaming calls; retryable mid-stream errors reset the handler before retrying. `stream_handler=None` keeps the existing non-streaming path.
- **PartialResultSender**: sequential single-in-flight POSTs with monotonic `partialSeq`; reset handling prevents stale pre-reset text from lingering; sender stops before the authoritative final callback.
- **Pipeline wiring**: non-exercise chat modes may stream the raw agent response. Programming-exercise chat disables raw-agent streaming and starts a sender during guide refinement instead.
- **Guide refinement streaming**: guide deltas are buffered until they can no longer be the exact `!ok!` sentinel. If the guide rewrites, the rewrite streams. If the guide returns `!ok!`, no guide partial is emitted and the original full answer is sent as the final result.
- **Responses API support**: `use_responses_api` models stream `response.output_text.delta` events through the same handler; function-call items reset the accumulator.

## Validation

- `poetry run pytest tests/test_response_streaming.py tests/test_exercise_guide_refinement_modes.py -q` — 29 passed.
- Existing streaming coverage remains for OpenAI chat streaming, Responses API streaming, retry reset, tool-call reset, sender coalescing/seq/reset/404-stop/failure-cap, and pipeline wiring on/off.

### DIFF
--- a/iris/src/iris/web/status/partial_result_sender.py
+++ b/iris/src/iris/web/status/partial_result_sender.py
@@ -1,0 +1,142 @@
+"""Ephemeral partial-result status callback sender."""
+
+from threading import Event, Lock, Thread
+from typing import Optional
+
+import requests
+
+from iris.common.logging_config import get_logger
+from iris.domain.status.chat_status_update_dto import ChatStatusUpdateDTO
+from iris.domain.status.stage_dto import StageDTO
+
+logger = get_logger(__name__)
+
+
+class PartialResultSender(Thread):
+    """Send accumulated partial chat answers to Artemis on a fixed interval."""
+
+    def __init__(
+        self,
+        url: str,
+        run_id: str,
+        stages_snapshot: list[StageDTO],
+        interval_seconds: float = 0.35,
+    ):
+        super().__init__(daemon=True)
+        self.url = url
+        self.run_id = run_id
+        self.stages_snapshot = [
+            stage.model_copy(deep=True) for stage in stages_snapshot
+        ]
+        self.interval_seconds = interval_seconds
+        self._lock = Lock()
+        self._stop_event = Event()
+        self._accumulated = ""
+        self._epoch = 0
+        self._last_posted_text = ""
+        self._last_posted_epoch = -1
+        self._partial_seq = 0
+        self._consecutive_failures = 0
+        self._stopped_permanently = False
+
+    def on_delta(self, delta: Optional[str]) -> None:
+        with self._lock:
+            if self._stopped_permanently or self._stop_event.is_set():
+                return
+            if delta is None:
+                self._accumulated = ""
+                self._epoch += 1
+                return
+            self._accumulated += delta
+
+    def stop(self) -> None:
+        self._stop_event.set()
+        self.join(2)
+        if self.is_alive():
+            logger.warning("Partial result sender did not stop within 2 seconds")
+
+    def run(self) -> None:
+        while not self._stop_event.wait(self.interval_seconds):
+            payload_info = self._next_payload()
+            if payload_info is None:
+                continue
+
+            payload, text, epoch = payload_info
+            if self._post_payload(payload):
+                self._record_success(text, epoch)
+
+    def _next_payload(self) -> Optional[tuple[dict, str, int]]:
+        with self._lock:
+            if self._stopped_permanently:
+                return None
+            if not self._accumulated:
+                return None
+            if (
+                self._accumulated == self._last_posted_text
+                and self._epoch == self._last_posted_epoch
+            ):
+                return None
+
+            self._partial_seq += 1
+            text = self._accumulated
+            epoch = self._epoch
+            payload = ChatStatusUpdateDTO(
+                stages=self.stages_snapshot,
+                partial_result=text,
+                partial_seq=self._partial_seq,
+            ).model_dump(by_alias=True)
+            return payload, text, epoch
+
+    def _post_payload(self, payload: dict) -> bool:
+        try:
+            response = requests.post(
+                self.url,
+                headers={
+                    "Content-Type": "application/json",
+                    "Authorization": f"Bearer {self.run_id}",
+                },
+                json=payload,
+                timeout=10,
+            )
+            response.raise_for_status()
+            self._consecutive_failures = 0
+            return True
+        except requests.exceptions.RequestException as exc:
+            status_code = getattr(getattr(exc, "response", None), "status_code", None)
+            self._handle_failure(status_code, exc)
+            return False
+        except Exception as exc:  # pragma: no cover - defensive thread boundary
+            self._handle_failure(None, exc)
+            return False
+
+    def _record_success(self, text: str, epoch: int) -> None:
+        with self._lock:
+            if epoch == self._epoch:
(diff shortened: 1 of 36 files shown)


iris/src/iris/web/status/partial_result_sender.py
```diff
@@ -0,0 +1,142 @@
...
+                return
+            if delta is None:
+                self._accumulated = ""
+                self._epoch += 1
+                return
+            self._accumulated += delta
+
+    def stop(self) -> None:
+        self._stop_event.set()
+        self.join(2)
```
Claudia-Anthropica: @bassner [high] `stop()` can return while `_post_payload()` is still blocked for up to 10 seconds, but the pipeline sends the final result immediately after this returns. That lets an older partial status arrive after the authoritative final callback and potentially put Artemis back on a non-terminal draft state; make the partial POST timeout and join semantics guarantee the sender is fully drained before `post_agent_hook` runs.

In `iris/src/iris/web/status/partial_result_sender.py`, `stop()` returns after 2 seconds while `_post_payload()` can still be in a 10 second `requests.post`, so stale partials can be delivered after the final status. Fix this by making partial POSTs bounded by the stop wait and not returning from `stop()` until any in-flight post has finished before the final callback proceeds.

</details>

iris/src/iris/web/status/partial_result_sender.py
```diff
@@ -0,0 +1,142 @@
...
+        self._partial_seq = 0
+        self._consecutive_failures = 0
+        self._stopped_permanently = False
+
+    def on_delta(self, delta: Optional[str]) -> None:
+        with self._lock:
+            if self._stopped_permanently or self._stop_event.is_set():
+                return
+            if delta is None:
+                self._accumulated = ""
```
Claudia-Anthropica, line 63: @bassner [medium] Resetting only the local buffer does not clear a partial that was already delivered to Artemis. If a tool-call preamble or a retrying stream has been posted, the browser will keep showing that stale draft during the tool call/backoff until another partial or final result arrives; the reset path needs to emit a newer empty/reset partial or otherwise tell Artemis to clear it.

In `iris/src/iris/web/status/partial_result_sender.py`, `on_delta(None)` only clears local state, so previously posted partial text remains visible to Artemis during tool calls or retries. Fix this by marking resets as a sendable update with a newer `partialSeq` and an empty/clearing `partialResult`, while still suppressing duplicate empty resets.

</details>

Claudia-Anthropica (changes requested):
@bassner The streaming path is well covered, but the partial callback lifecycle still has two edge cases that can surface stale draft text. Please fix the sender shutdown/reset behavior inline before this ships.

iris/src/iris/web/status/partial_result_sender.py
Claudia-Anthropica: @bassner This still looks unresolved on the current head: `stop()` still does `join(2)` while `_post_payload()` uses `timeout=10`, so a blocked partial POST can still outlive the final callback. Please make `stop()` wait for the in-flight POST to drain or bound the POST timeout accordingly.

iris/src/iris/web/status/partial_result_sender.py
Claudia-Anthropica: @bassner This is still unresolved too: `on_delta(None)` only clears `_accumulated` and increments the epoch, and `_next_payload()` skips empty text, so Artemis still never receives a newer empty partial to clear an already displayed draft.

Claudia-Anthropica (changes requested):
@bassner The current head only merged the updated base branch, so the two partial-result lifecycle issues are still present. Keeping this as changes requested; the sender still needs to drain in-flight partial POSTs before final callbacks and emit a clearing partial on reset.

Claudia-Anthropica (changes requested):
@bassner The new Responses streaming coverage is useful, but the existing partial-result lifecycle blockers are still present in the current code. The sender still needs to drain in-flight partial POSTs before final callbacks and send a clearing partial on reset, so I’m keeping this as changes requested.

iris/src/iris/web/status/partial_result_sender.py
bassner: Good catch — real ordering hazard. `stop()` joined for 2s while a partial POST could still be blocked in a 10s `requests.post`, so a stale partial could land after the final `done`. Fixed in 4c712e0: partial POSTs are now bounded by `PARTIAL_POST_TIMEOUT_SECONDS` (2s) and `stop()` sets a permanent-stop flag then joins for the strictly-larger `STOP_DRAIN_TIMEOUT_SECONDS` (3s), so an in-flight partial POST drains before the pipeline sends the final result. It is bounded for responsiveness and warns if drainage is exceeded — documented as best-effort, with Artemis's terminal-state monotonicity as the backstop. Added tests for the timeout ordering and the in-flight drain.

iris/src/iris/web/status/partial_result_sender.py
bassner: Correct — clearing only local state left the previously delivered draft visible during tool calls/retries. Fixed in 4c712e0: a reset now emits a clearing partial (empty `partialResult` with a higher `partialSeq`) when a non-empty draft is currently on the client, suppressing the initial empty state and duplicate consecutive empty resets. A follow-up in 26b5cca0 also handles the case where a partial POST is still in flight when the reset happens: `_record_success` now records what was actually delivered regardless of epoch, so the clearing partial is still emitted afterwards (a regression test blocks a post in flight, resets mid-flight, releases it, and asserts the empty clear is posted).

iris/src/iris/web/status/partial_result_sender.py
Claudia-Anthropica: @bassner Looks good now, thanks. The bounded POST timeout plus the longer stop drain covers the ordering case I was worried about, and the regression test checks the in-flight drain.

iris/src/iris/web/status/partial_result_sender.py
Claudia-Anthropica: @bassner Nice fix. Recording the actually delivered draft before evaluating the reset clear handles the in-flight case too.

Claudia-Anthropica (dismissed):
@bassner All feedback addressed; the partial sender now drains in-flight posts before final callbacks and emits clearing partials on reset. Approving.

PR #660, opened 2026-07-05
`Iris`: Replace stage status protocol with run-state and activities

bassner:
## What

Deletes the stage-based status protocol and replaces it with a run-state + activity model:

- Every status update carries `runState: RUNNING|FINISHED|FAILED` plus optional `error {message, code}`. Stage DTOs are deleted; a guard test keeps them out.
- Chat emits real tool activities: a thread-safe `ActivityTracker` fed by a LangChain callback handler (`on_tool_start/end/error`) posts full snapshots (`activities[]` + monotonic `activitySeq`) asynchronously; `send_result`/`finish`/`fail` carry authoritative snapshots bumped under the tracker lock.
- Delivery-critical answers: the result POST retries with backoff and carries an undelivered answer forward on every subsequent send until delivered.
- Ingestion keeps its exact heartbeat cadence/retry semantics as RUNNING updates; fixes shipped along the way: FAQ ingestion's forever-IN_PROGRESS job leak, FAQ deletion never reporting terminal status, double-terminal on lecture-unit deletion. Dead code deleted.
- MCQ generation reports its parallel work explicitly; the handwritten per-tool `in_progress` strings are gone.

## Stack Update

This branch now includes the updated #656-#658 behavior:

- Guide refinement is no longer runtime-configurable; programming-exercise chat always runs the guide pass before final delivery.
- Response streaming remains opt-in via `streamResponse`.
- In programming-exercise chat, raw main-model partials are suppressed. The guide refinement model streams the visible rewrite; if the guide returns exact `!ok!`, no partial is emitted and the original full answer is sent via `send_result`.
- `PartialResultSender` uses the new run-state status payload (`partialResult`, `partialSeq`, `runState: RUNNING`) instead of stage snapshots.

## Verification

- `poetry run pytest tests/test_response_streaming.py tests/test_exercise_guide_refinement_modes.py tests/test_chat_run_callback.py tests/test_status_callback_lifecycle.py -q` — 47 passed.
- Existing status-rework validation: unit suite and pre-commit were green when this PR was prepared.

## Pairing

**Hard version pairing, no compat shims (D2):** requires the counterpart Artemis PR (`feature/iris/status-rework`). Deploy order: Artemis first, then Pyris immediately.

Stacked on #658. Spec + plan: `iris-status-rework-spec.local.md` (session-local, not committed).

## Summary by CodeRabbit

* **New Features**
  * Added live activity tracking for chat and tool execution, including activity states, durations, and sequence updates.
  * Improved status updates with clearer final, success, and failure reporting.
  * Tools now provide more structured, user-visible progress and safer handling during execution.

* **Bug Fixes**
  * Fixed status updates to avoid stale or duplicate progress information.
  * Improved retrieval and ingestion flows to handle missing or incomplete data more safely.
  * Reduced noisy progress messages during long-running operations.


iris/src/iris/pipeline/abstract_agent_pipeline.py
```diff
@@ -674,18 +727,13 @@ def __call__(
...
+                state.callback.finish(
                     created_memories=state.memiris_memory_creation_storage,
                     session_title=deferred_title,
                 )
             else:
-                state.callback.done(
-                    "No memory creation thread started.",
-                    session_title=deferred_title,
-                )
+                state.callback.finish(session_title=deferred_title)
```
Claudia-Anthropica: @bassner [medium] This final chat callback is built from a fresh `ChatStatusUpdateDTO`, so calling `finish()` without `tokens` sends the terminal status with the default empty token list. That regresses the old behavior where the status object kept the accumulated tokens, and it also drops title/suggestion tokens added after `send_result()`. Pass `state.tokens` into both final `finish()` branches so the terminal update carries the authoritative usage.

In `iris/src/iris/pipeline/abstract_agent_pipeline.py`, the final chat `finish()` calls do not pass accumulated token usage, so `ChatRunCallback` emits an empty token list on the terminal payload. Fix this by passing `tokens=state.tokens` in both final `finish()` branches, preserving any created-memory and session-title fields.

</details>

iris/src/iris/web/status/status_update.py
```diff
@@ -161,407 +142,296 @@ def _drain_in_progress_updates(self) -> None:
...
+            fields["session_title"] = session_title
+        if created_memories is not None:
+            fields["created_memories"] = self._memory_dtos(created_memories)
+        if tokens is not None:
+            fields["tokens"] = tokens
+        if activities is not None:
+            fields["activities"] = activities
+        if activity_seq is not None:
+            fields["activity_seq"] = activity_seq
+        return self._send_chat_fields(fields, run_state=RunStateEnum.FINISHED)
```
Claudia-Anthropica, line 368: @bassner [high] If `send_result()` exhausts its three retries, this `finish()` call is the last chance to deliver the carried answer for lecture/text chats, but it only attempts the POST once. A transient failure here drops the undelivered answer even though the new contract says the answer is delivery-critical. Retry terminal sends when they include a carried result, or make `finish()` use the same backoff retry path.

In `iris/src/iris/web/status/status_update.py`, `ChatRunCallback.finish()` sends carried undelivered answers with only one HTTP attempt. Fix this by retrying terminal chat sends that include `_undelivered_result_fields`, using the same backoff behavior as `send_result()`.

</details>

iris/src/iris/web/status/status_update.py
```diff
@@ -161,407 +142,296 @@ def _drain_in_progress_updates(self) -> None:
...
+    def update(self, **fields) -> bool:
+        """Synchronously post a RUNNING status update."""
+        if self._terminal_sent:
+            self._reject_after_terminal("update")
+            return False
+        self.status.run_state = RunStateEnum.RUNNING
+        self.status.error = None
+        self._apply_fields(fields)
+        self._drain_running_updates()
+        return self.on_status_update()
```
Claudia-Anthropica: @bassner [medium] `update(result=...)` leaves `self.status.result` set after a successful RUNNING callback. The transcription checkpoint paths now use that for checkpoint JSON, so every later heartbeat and the final `finish(display_page_numbers=...)` will resend the old checkpoint; the old `done()` cleared transient fields after successful delivery. Clear transient result-like fields after a successful non-terminal update, while keeping them around when the POST fails so the retry-on-next-update behavior still works.

In `iris/src/iris/web/status/status_update.py`, successful non-terminal `update(result=...)` calls keep transient fields on the reusable status DTO and replay them on later updates. Fix this by clearing transient fields such as `result` after a successful `update()`, but preserve them when the POST fails so a later update can retry delivery.

</details>

Claudia-Anthropica (changes requested):
@bassner The run-state/activity direction looks solid, but I found a few regressions around delivery semantics. The terminal chat update drops accumulated token usage, undelivered answers only get one final send attempt, and ingestion checkpoint results stay attached to later status updates. See inline comments.

iris/src/iris/pipeline/abstract_agent_pipeline.py
bassner: Fixed in 1c15062c — both terminal `finish()` branches now pass `tokens=state.tokens`, so the FINISHED status carries the authoritative accumulated usage, including the title/suggestion tokens appended after `send_result()`. I also fixed the analogous suggestion-failure `fail()` path (7af15383) to carry tokens, since `fail()` marks the job terminal and a later `finish()` can no longer attach them.

iris/src/iris/web/status/status_update.py
bassner: Fixed in 1c15062c — a terminal send that carries a previously-undelivered answer now retries with the same backoff as `send_result()` (shared `_DELIVERY_RETRY_ATTEMPTS`, `(1,2,4)s` backoff). Terminal sends without a carried result stay single-shot, so the common already-delivered path is unaffected. Added tests for both the retrying-carried-result and single-shot cases.

iris/src/iris/web/status/status_update.py
bassner: Fixed in 1c15062c — a successful non-terminal `update()` now clears the transient result-like fields (`result`, `display_page_numbers`) on the reusable status DTO, so later heartbeats and the final `finish()` no longer resend stale checkpoint JSON. A failed POST keeps them so retry-on-next-update still works, and persistent fields (tokens/id/activities/run_state) are preserved. Added tests for the clear-on-success, keep-on-failure, and no-resend-on-next-heartbeat cases.

iris/src/iris/pipeline/abstract_agent_pipeline.py
Claudia-Anthropica: @bassner Looks good now, thanks.

iris/src/iris/web/status/status_update.py
Claudia-Anthropica: @bassner Nice, that handles the carried-result retry path.

iris/src/iris/web/status/status_update.py
Claudia-Anthropica: @bassner Looks good now, the transient clearing matches the retry behavior.

Claudia-Anthropica (approved):
@bassner All feedback addressed, nice work.

Claudia-Anthropica (dismissed):
@bassner Re-reviewed the status-rework diff after the new commits; the delivery, activity, and streaming paths look consistent now. I couldn't run the Poetry test command locally because Poetry isn't installed in this checkout, but a Python compile pass succeeded and the remaining failing ReadTheDocs Atlas check looks unrelated to this Iris PR.

iris/src/iris/retrieval/lecture/lecture_retrieval.py
```diff
@@ -1077,6 +1077,22 @@ def get_lecture_transcription_of_lecture_unit(
...
+        # A None in any filter value crashes the Weaviate gRPC query ("unknown
+        # value type <nil>"); such segments cannot match a transcription anyway.
+        filter_values = (
+            lecture_unit_segment.course_id,
+            lecture_unit_segment.lecture_id,
+            lecture_unit_segment.lecture_unit_id,
+            target_page_number,
+            lecture_unit_segment.base_url,
+        )
+        if any(value is None for value in filter_values):
```
Claudia-Anthropica, line 1089: @bassner [medium] This guard prevents `.equal(None)` for the transcription lookup, but the same `lecture_unit_segment` is still passed to `get_lecture_page_chunks_of_lecture_unit()` right after this method returns, and that method unconditionally filters on `course_id`, `lecture_id`, `lecture_unit_id`, `page_number`, and `base_url`. If any of the values you guard here are actually `None`, especially `base_url`, the page-chunk fetch can still hit the same Weaviate gRPC `nil` crash. Add the same guard there or skip both companion lookups consistently.

In `iris/src/iris/retrieval/lecture/lecture_retrieval.py`, `get_lecture_transcription_of_lecture_unit()` now guards nullable filter values but `get_lecture_page_chunks_of_lecture_unit()` still calls `.equal(...)` with the same segment fields. Fix this by adding an equivalent pre-filter guard to the page-chunk lookup, or a shared helper, and add a regression test for a segment with a `None` filter value.

</details>

Claudia-Anthropica (changes requested):
@bassner The status protocol migration mostly looks consistent, and the earlier delivery issues are fixed. I found one remaining edge case in the lecture retrieval `None`-filter fix: the transcription lookup is guarded, but the sibling page-chunk lookup can still build a Weaviate filter with the same nullable values.

iris/src/iris/retrieval/lecture/lecture_retrieval.py
Claudia-Anthropica: @bassner Looks good now, thanks.

Claudia-Anthropica (approved):
@bassner All feedback addressed and CI is green. Approving.

PR #663, opened 2026-07-06
`Logos`: Responses API support, REST conformance cleanup, single-port serving

### LINKED ISSUE #662: `Logos`: Support Responses API
Right now only Completions API is supported, but some features do only work with newer Responses API (e.g., configuring reasoning effort and tool calling at the same time)

wasnertobias:
## Motivation

Closes #662.

Logos only supported the Chat Completions API end to end. Some features (e.g. configuring reasoning effort together with tool calling) only work with the newer Responses API. The Azure deployment sync already classified gpt-5.x deployments to Responses endpoints, but the request path through the orchestrator was still Chat-Completions-shaped: inbound `/v1/responses` requests were forwarded to the deployment's stored `chat/completions` URL, streaming requests injected a parameter the Responses API rejects, and streamed Responses usage was never captured for billing.

Additionally, this PR cleans up non-conforming behavior in the user-facing REST surface and prepares the compose stack so the web UI, Swagger UI, and the completion API can all be served on the default HTTP/HTTPS ports.

### Responses API support (Azure and cloud providers)

Clients can now call `POST /v1/responses` (streaming and non-streaming) through Logos against Azure and other cloud providers — the inbound API surface wins over the deployment's stored default:

- **Routing** (`context_resolver.py`): Azure per-model endpoints store one default operation. The forward layer now re-targets the operation suffix between `chat/completions` and `responses` to match the inbound path, pinning the api-version configured for the target operation (the stored chat api-version can predate Responses availability). The existing collapse to Azure's real `/openai/responses` route and the body-`model` → deployment-id rewrite then apply as before. Non-swappable operations (embeddings, audio, images) are never re-targeted; non-Azure cloud upstreams already forward like-for-like.
- **Streaming** (`executor.py`): `stream_options.include_usage` is Chat-Completions-only — the Responses API rejects it and instead always reports usage in its terminal `response.completed` event, so it is no longer injected for `/responses` upstreams.
- **Usage & logging** (`main.py`, `responses.py`): the streaming log accumulator understands Responses SSE events (`response.output_text.delta` for text, the terminal `response.completed`/`incomplete`/`failed` object for the full response and usage). Responses usage names (`input_tokens`/`output_tokens` and their details) are normalized to the Chat Completions names that token pricing, TPM rate limiting, and the statistics UI are keyed to.
- **Classification** (`pipeline.py`): prompt extraction supports the Responses request shape (`input` as string or message list with `input_text` parts, `instructions`/`developer` as the system prompt), so RESOURCE mode can classify Responses payloads.

### REST conformance cleanup

- The proxy catch-alls (`/v1/*`, `/v2/*`, `/openai/*`, `/jobs/*`) are **POST-only** now — every proxied upstream operation is a POST. Other methods get a proper **405 with `Allow` header** instead of the misleading `400 Invalid JSON body` the body parser raised on body-less GETs.
- The OpenAI-shape exception handler is registered on Starlette's base `HTTPException`, so framework-raised errors (like that 405) also get the OpenAI error shape, and it **preserves exception headers** (`Allow`, `Retry-After`) instead of dropping them.
- 429 rate-limit rejections carry **`Retry-After`** (the limiter's 60s window).
- Job submission 202 carries a **`Location`** header pointing at the job status resource (standard async-request pattern).
- **`GET /openai/models` and `/openai/models/{model}`** now work — the `/openai` prefix mirrors `/v1`, but model listing previously fell into the catch-all and returned 400.

### Single-port serving (UI + Swagger + API)

- Traefik gains a plain-HTTP **`:80` entrypoint that permanently redirects to HTTPS**, so the stack answers on both default ports.
- The dedicated admin entrypoint (**port 9443**) is **removed entirely**: the web UI, its Spring `/api` backend, Swagger `/docs`, the workernode protocol, `/metrics`, and the completion API are all served together on the default HTTPS port (443). Traefik disambiguates by path rules and router priorities (UI catch-all 0 < orchestrator API 100 < webservice 200). Port 8080 stays as the documented TLS alias for the completion API.
- Outdated env variables removed: `LOGOS_TRAEFIK_ADMIN_PORT`, `LOGOS_ADMIN_ENTRYPOINTS`, and `LOGOS_CERT_RESOLVER` (never consumed — the compose labels hardcode the letsencrypt resolver). The webservice CORS default is now just the bare-domain origin. README access instructions drop the SSH-tunnel + `/etc/hosts` dance; the benchmark admin-port default moves 9443 -> 443; the OpenAPI server list reflects the new layout.
- **Deploy note:** workernode `LOGOS_URL` values pointing at `:9443` must drop the port once a core node runs this image.

## Testing

- New/extended unit tests: operation re-targeting in `_cloud_forward_url`, streaming payload preparation, Responses SSE accumulation (including cut-off streams), usage-name normalization, prompt extraction, and REST conformance (405 + Allow, header preservation, `/openai/models` aliases).
- **Billing verified against real Azure**: streaming and non-streaming `gpt-5-nano` calls were made against the production Azure `/openai/responses` deployment and the captured raw payloads fed through the actual billing path (`_StreamingLogAccumulator` -> `extract_token_usage` -> `_usage_tokens_from_payload`). Both yield the canonical `prompt_tokens`/`completion_tokens`/`total_tokens` (+ cached/reasoning details) that the `usage_tokens`/`token_prices` billing join is keyed to. The captured stream is now an end-to-end unit-test fixture.
- Full orchestrator unit suite passes (461 passed, 1 xfailed) and the worker-node CI suite passes (243 passed); `docker compose config` verified for prod and dev; pre-commit (black, isort, flake8, autoflake) clean.


Claudia-Anthropica (changes requested):
@wasnertobias The Logos changes look coherent from my pass, but CI is currently blocked by the PR-title validation. Please rename the PR to use the required backtick-wrapped scope, for example `` `Logos`: Responses API support, REST conformance cleanup, single-port serving``.

PR #665, opened 2026-07-06
`Logos`: Fix status update in Statistics Page

flbrgit:
This PR resolves an update bug in the statistics page, where a finished model was still shown as running.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic refresh behavior in the recent requests view so updates are triggered more reliably when request details change, not just when the number of items changes.
  * Reduced missed refreshes and unnecessary refreshes by making change detection more precise.


Claudia-Anthropica (approved):
@flbrgit Looks good. The refresh guard now tracks the live request fields that drive queued/running/complete rendering, and it stays aligned with the websocket request signature. Approving as-is.

PR #666, opened 2026-07-06
`Logos`: Improve Statistics Page Loading Performance

flbrgit:
This PR reduces the loading time of the statistics page.


Claudia-Anthropica (approved):
@flbrgit The webservice change looks sound: it keeps the full VRAM chart data while avoiding repeated rich scheduler/model payloads on historical points. The Logos Webservice checks are green; the failing worker-node Docker build is in the vLLM install path and looks unrelated to this Java-only change, so I’m not blocking on it.

PR #669, opened 2026-07-08
`Logos`: Benchmarks: replicated campaign runner (R=5 default, rate sweep)

wasnertobias:
Adapt run_bench.sh for the evaluation-campaign plan (T1-a/T1-b):
- REPLICATES (default 5) independent replicated runs; replicate i uses SEED+i-1, shared across all scenarios/patterns so configs stay paired.
- RATES sweeps arrival rates (load sweep: 0.05..2.0 req/s); results land in benchmark_results/rate_<rate>/rep<NN>_seed<seed>/.
- REPLICATE_START extends finished campaigns to R=8-10 on high-variance near-knee baseline cells without re-running earlier replicates.
- Keep N=1000 requests per cell; fail fast on a broken run with a resume hint.

## Summary by CodeRabbit

* **New Features**
  * Expanded benchmark runs to support multiple traffic rates and repeated trials in one campaign.
  * Added clearer progress, summary, and resume messaging for long-running benchmark campaigns.

* **Bug Fixes**
  * Improved authentication checks so benchmark runs fail fast when access is invalid instead of hanging.
  * Reduced false progress during calibration by stopping immediately when required sessions cannot start.
  * Added safeguards to keep repeated benchmark runs aligned with the correct workload and timing.

### DIFF
--- a/logos/benchmarks/benchmark_logos.py
+++ b/logos/benchmarks/benchmark_logos.py
@@ -82,6 +82,7 @@ import json
 import math
 import os
 import random
+import re
 import shlex
 import socket
 import subprocess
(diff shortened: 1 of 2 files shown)


Claudia-Anthropica (dismissed):
@wasnertobias No additional issues from my review; the campaign loop matches the benchmark CLI and CI is green.

logos/benchmarks/benchmark_logos.py
```diff
@@ -2652,6 +2653,65 @@ def _set_calibration_window_enabled(
...
+    admin endpoint then returns 403 and the run wedges at ensure-calibrate.
+    Detect that drift up front and repair the ``is_active`` flags directly in
+    the database (same host access the benchmark already uses for compose
+    management). A missing key or a non-admin role cannot be self-repaired and
+    raises immediately, BEFORE any multi-hour run starts.
+    """
+    if not re.fullmatch(r"[A-Za-z0-9_-]+", logos_key or ""):
+        raise RuntimeError("LOGOS key has unexpected characters; refusing to embed it in SQL.")
+    sudo = "sudo " if use_sudo else ""
+    repair_sql = (
```
Claudia-Anthropica, line 2677: @wasnertobias [high] This benchmark preflight should not silently reactivate users and API keys in the Logos DB. If Keycloak or an admin disabled that account, this bypasses the auth source of truth and can resurrect access just because a benchmark ran. Keep the preflight read-only and fail with an explicit operator action instead.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `logos/benchmarks/benchmark_logos.py`, `_ensure_benchmark_key_active()` updates `users.is_active` and `api_keys.is_active` directly during benchmark startup. Replace the repair SQL with a read-only verification query and abort with a clear error telling the operator what must be fixed manually.\n\n</details>

logos/benchmarks/benchmark_logos.py
```diff
@@ -5540,7 +5615,12 @@ async def _ensure_calibration_complete_all_nodes(
 
     print(f"\n[Ensure-Calibrate] Triggering calibration on provider(s) {provider_ids} (admin port {admin_port}) ...")
     if not await _trigger_calibration_via_rest(logos_url, logos_key, provider_ids, admin_port):
-        print("  WARNING: could not start a calibration session on every provider.", file=sys.stderr)
+        print(
+            "  ERROR: could not start a calibration session on every provider — "
+            "failing fast instead of waiting for a calibration that never started.",
+            file=sys.stderr,
+        )
+        return False
```
Claudia-Anthropica, line 5623: @wasnertobias [high] This still doesn’t fail the run: `_async_run_all()` treats this `False` as a warning and continues, so `run_bench.sh` can see rc=0 and move to later campaign cells after calibration never started. Make calibration-start failures abort the benchmark with a nonzero exit instead of returning to the “continuing anyway” path.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `logos/benchmarks/benchmark_logos.py`, calibration trigger failures return `False` but the callers in `_async_run_all()` only log a warning and continue. Fix this by making both reset-calibration and ensure-calibration trigger failures abort the benchmark with a nonzero exit.\n\n</details>

logos/benchmarks/benchmark_logos.py
```diff
@@ -5682,6 +5762,19 @@ async def _async_run_all(args: argparse.Namespace) -> None:
     manage_calib_window = getattr(args, "manage_calibration_window", True) and needs_logos
     _calib_window_disabled = [False]  # mutable flag so _cleanup can restore
 
+    # Preflight: the key must be an ACTIVE logos_admin or every admin endpoint
+    # (calibration trigger) 403s and the run wedges. External identity syncs
+    # have deactivated the root key before — detect and repair that drift NOW,
+    # before hours of setup and dispatch depend on it.
+    if needs_logos:
```
Claudia-Anthropica, line 5769: @wasnertobias [medium] This runs before Step 0 starts the Logos compose stack, so a cold benchmark host with `logos-db` stopped now fails at `docker exec` instead of letting `_start_logos()` bring the stack up. Move the key check after `_start_logos()` and the TLS wait, right before the calibration endpoints need the key.\n\n<details><summary>🤖 Prompt for AI agents</summary>\n\nIn `logos/benchmarks/benchmark_logos.py`, the benchmark key preflight is called before the Logos stack is started. Move the preflight to the Step 0 block after `_start_logos()` and the TLS readiness check so cold-start runs can bring up `logos-db` before verification.\n\n</details>

Claudia-Anthropica (changes requested):
@wasnertobias The campaign loop mostly lines up with the CLI, but I’m requesting changes for the new preflight/calibration control flow. The benchmark now mutates auth state directly, can fail before the stack is started, and the new calibration “fail fast” path still allows the campaign to continue with bad data.

PR #679, opened 2026-07-13
`Iris`: Fix false error logs for tool-call responses

bassner:
### Summary
Fixes successful OpenAI tool-call completions being reported as empty-message errors when they contain no textual assistant content.

### Motivation and Context
OpenAI tool-call-only responses legitimately use `content: null` or an empty string while carrying one or more entries in `tool_calls`. Iris already converted these responses correctly, but the empty-response diagnostic only inspected textual content. This produced misleading production errors even though the request completed successfully.

### Changes
- Treat a completion as empty only when both textual content and tool calls are absent
- Preserve the existing diagnostics for genuinely empty responses
- Add regression coverage for both `content=None` and `content=""` tool-call responses
- Verify tool-call IDs, function names, and parsed arguments remain intact

### Steps for Testing
1. Run `poetry run pytest tests/test_openai_chat_response_logging.py -v`
2. Run `poetry run pytest tests`
3. Run `poetry run black --check src/iris/llm/external/openai_chat.py tests/test_openai_chat_response_logging.py`
4. Run `poetry run isort --check-only src/iris/llm/external/openai_chat.py tests/test_openai_chat_response_logging.py`
5. Run `poetry run pylint src/iris/llm/external/openai_chat.py tests/test_openai_chat_response_logging.py`

### Test Results
- 268 tests passed
- Black passed
- isort passed
- Pylint: 10.00/10

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of responses containing tool calls when message content is empty or unavailable.
  * Prevented valid tool-call responses from being incorrectly reported as empty messages.
  * Genuine empty responses continue to generate the appropriate error logging.


bassner:
@Claudia-Anthropica review

Claudia-Anthropica:
@bassner roger that — review incoming!

Claudia-Anthropica (approved):
@bassner This cleanly fixes the false empty-message diagnostics while preserving genuine-empty logging, with focused regression coverage for both valid tool-call content variants. Nice work.

PR #682, opened 2026-07-15
`Logos`: Slim vLLM worker image and fix out-of-disk CI build flakiness

wasnertobias:
## Problem

The **Logos - Build** workflow fails intermittently on the worker-node vLLM job with `ERROR: [Errno 28] No space left on device` (e.g. [run 29032180851](https://github.com/ls1intum/edutelligence/actions/runs/29032180851)), builds are slow, and the GPU image is ~21 GB.

Root cause, pinned from the failure logs and wheel analysis:

- **The CUDA stack existed 3× in the final image**: since vLLM ≥ 0.24 the PyPI wheel tree vendors a *complete* CUDA 13 toolkit (`site-packages/nvidia/cu13/` with `bin/nvcc`, `ptxas`, `cicc`, `nvvm/libdevice`, `include/` with every JIT header, and all math libs, plus cudnn/nccl/nvshmem next to it — ~3.1 GB). The apt `cuda-*`/`lib*-dev` layer (~4.5 GB) and the `cudnn-runtime` base (~5.8 GB) duplicated it almost 1:1.
- **CI peak build disk was ~24 GB** vs the ~14 GB GitHub-hosted runners guarantee: the 9 GB `cudnn-devel` builder base was pulled but never used (pip only installs wheels — nothing is compiled at build time), and the 9.2 GB venv existed twice (builder layer + runtime `COPY` layer). Builds only succeeded when the runner happened to have extra free disk → flakiness.
- pip additionally downloads all 4.3 GB of wheels *before* installing, adding to the peak.

## Change

Single-stage build on plain `ubuntu:24.04` (GPU) / `python:3.12-slim` (CPU default, unchanged):

- **All CUDA comes from vLLM's pip wheels.** `CUDA_HOME` points at the wheel toolkit prefix; the apt CUDA layer and both nvidia/cuda base images are gone. Three small gaps the apt `-dev` layer used to fill are patched at build time: `lib64 → lib` symlink, unversioned `.so` dev symlinks, and a `libcuda.so` linker stub generated from the wheel's own `cuda.h` (exports all ~449 driver-API symbols with their real versioned names under SONAME `libcuda.so.1`, so GCC's default `--as-needed` keeps the `DT_NEEDED libcuda.so.1` entry that binds to the runtime-injected driver at load time; build guards assert the exported-symbol count and the `DT_NEEDED` entry via a JIT-style link probe). `gcc`/`g++`/`python3.12-dev` stay for Triton/FlashInfer runtime JIT; `g++` is now explicit (was an implicit dep of apt `cuda-nvcc`).
- **No builder stage** — everything is a binary wheel and the compiler must be in the final image anyway, so a venv `COPY` only doubled disk usage.
- **Pinned, bind-mounted `uv`** (same version as the orchestrator image) instead of pip: streams wheels instead of download-all-first, much faster, no bytecode in layers.
- Build guards `test -x $CUDA_HOME/bin/nvcc` etc. make the automated vLLM-bump PRs fail the *build* early if NVIDIA ever changes the wheel layout.
- Root `.dockerignore` whitelists exactly what the orchestrator image (the only repo-root-context build) copies.

## Results

| | before | after |
|---|---|---|
| GPU image size | ~21 GB | **9.87 GB** |
| Peak CI build disk | ~24 GB | **~10.5 GB** (safely under the 14 GB limit) |
| Build-time downloads | devel base (~4 GB) + runtime base + apt CUDA + 4.3 GB wheels | ubuntu (30 MB) + 4.3 GB wheels |

## Validation (local builds of this branch)

- CPU/Ollama variant builds (2.42 GB); app + all requirements import, `ollama` binary and seeded `pip` work.
- GPU variant builds (9.87 GB, linux/amd64); inside the image:
  - `nvcc -arch=sm_75` compiles a kernel including `cublas_v2.h` + `curand_kernel.h` (the exact FlashInfer-JIT header set from the deioma/Blackwell incident, vllm-project/vllm#35501),
  - a FlashInfer-style `gcc -shared … -L$CUDA_HOME/lib64 -L$CUDA_HOME/lib64/stubs -lcudart -lcuda` link succeeds **and records `DT_NEEDED libcuda.so.1`** (verified with `readelf -d`; also enforced by a build-time guard),
  - `torch 2.11.0+cu130`, `flashinfer` (resolves `CUDA_HOME` to the wheel prefix) and `vllm 0.25.1` all import.
- Orchestrator image builds with the new whitelisted root context.

**Remaining GPU-runtime risk** (not testable without a GPU): first real FlashInfer JIT compile against the injected driver. Suggest rolling this to the test workernodes via *Logos - Deploy to Test* once the PR build has pushed the tag, and running a benchmark scenario before merging.

## Summary by CodeRabbit

* **Build & Deployment**
  * Improved Docker build efficiency by reducing the build context and excluding unnecessary development files.
  * Updated GPU worker images to use a streamlined Ubuntu-based setup with CUDA components supplied through Python packages.
  * Simplified development and automated Docker configurations for GPU-enabled builds.
  * Added more reliable runtime support for GPU libraries and optional vLLM installation.

### DIFF
--- a/logos/logos-workernode/Dockerfile
+++ b/logos/logos-workernode/Dockerfile
@@ -37,11 +37,16 @@
 #     lib*.so      — wheels ship only versioned SONAMEs (libcudart.so.13);
 #                    `-lcudart` needs the unversioned dev symlink.
 #     libcuda.so   — the JIT links -lcuda, normally resolved against the
-#                    toolkit's driver stub. Wheels ship no stub, so we build
-#                    an empty one with SONAME libcuda.so.1: linking a shared
-#                    object does not require resolving symbols, and at load
-#                    time the DT_NEEDED libcuda.so.1 binds to the real driver
-#                    library injected by the NVIDIA Container Runtime.
+#                    toolkit's driver stub. Wheels ship no stub, so we
+#                    generate one from the wheel's own cuda.h: it exports
+#                    every driver-API function (with the real versioned
+#                    symbol names, e.g. cuMemAlloc_v2) under SONAME
+#                    libcuda.so.1. An *empty* stub would not work — GCC
+#                    links with --as-needed by default, so a library that
+#                    resolves no symbols is dropped, leaving the JIT module
+#                    with unresolved cu* symbols and no DT_NEEDED entry. At
+#                    load time the DT_NEEDED libcuda.so.1 binds to the real
+#                    driver injected by the NVIDIA Container Runtime.
 #
 # • libcuda.so.1 is NOT baked into the image.
 #   The NVIDIA Container Runtime injects it at startup via the host driver.
@@ -231,11 +236,35 @@ RUN if [ "$INSTALL_VLLM" = "1" ]; then \
           base="${so%%.so.*}.so"; \
           [ -e "$base" ] || ln -s "$(basename "$so")" "$base"; \
         done; \
-        # Driver linker stub: empty shared object with SONAME libcuda.so.1.
-        # Satisfies the JIT's `-lcuda` at link time; at load time the SONAME
-        # binds to the real driver library injected by the container runtime.
+        # Driver linker stub with SONAME libcuda.so.1, generated from the
+        # wheel's own cuda.h. It must actually EXPORT the driver-API symbols:
+        # GCC links with --as-needed by default, so an empty stub would be
+        # discarded and JIT modules would end up with unresolved cu* symbols
+        # and no DT_NEEDED entry for libcuda.so.1. Preprocessing cuda.h first
+        # yields the real versioned symbol names (e.g. cuMemAlloc_v2). At
+        # load time the SONAME binds to the real driver library injected by
+        # the container runtime.
         mkdir -p "$CUDA_HOME/lib/stubs"; \
-        gcc -shared -Wl,-soname,libcuda.so.1 -o "$CUDA_HOME/lib/stubs/libcuda.so" -x c /dev/null; \
+        echo '#include <cuda.h>' \
+          | gcc -E -P -I "$CUDA_HOME/include" - \
+          | tr '\n' ' ' \
+          | grep -oE '\bcu[A-Z][A-Za-z0-9_]*[[:space:]]*\(' \
+          | sed -E 's/[[:space:]]*\(//' | sort -u \
+          | sed 's/.*/void &(void) {}/' > /tmp/libcuda_stub.c; \
+        gcc -shared -fPIC -Wl,-soname,libcuda.so.1 \
+            -o "$CUDA_HOME/lib/stubs/libcuda.so" /tmp/libcuda_stub.c; \
+        # Guard: the stub must export the driver API surface (~450 functions
+        # for CUDA 13; fail early if header parsing ever breaks).
+        n="$(nm -D --defined-only "$CUDA_HOME/lib/stubs/libcuda.so" | grep -c ' T cu')"; \
+        [ "$n" -ge 300 ] || { echo "libcuda stub exports only $n symbols" >&2; exit 1; }; \
+        # Guard: a representative JIT-style link (FlashInfer links with
+        # -L$CUDA_HOME/lib64/stubs -lcuda) must record DT_NEEDED libcuda.so.1.
+        printf 'extern int cuInit(unsigned int u);\nint probe(void) { return cuInit(0); }\n' > /tmp/needed_probe.c; \
+        gcc -shared -fPIC -o /tmp/needed_probe.so /tmp/needed_probe.c \
+            -L "$CUDA_HOME/lib64/stubs" -lcuda; \
+        readelf -d /tmp/needed_probe.so | grep -q 'NEEDED.*libcuda\.so\.1' \
+          || { echo "JIT-style probe .so lacks DT_NEEDED libcuda.so.1" >&2; exit 1; }; \
+        rm -f /tmp/libcuda_stub.c /tmp/needed_probe.c /tmp/needed_probe.so; \
         # Register wheel-vendored CUDA library directories with the dynamic
         # linker so child processes (vLLM lanes) can find them. Covers
         # nvidia/cu13/lib, cudnn, nccl, nvshmem, cusparselt and torch/lib
--- a/logos/logos-workernode/Dockerfile
+++ b/logos/logos-workernode/Dockerfile
@@ -5,37 +5,60 @@
 #   BASE_IMAGE=python:3.12-slim   INSTALL_VLLM=0   INSTALL_OLLAMA=1
 #
 # GPU / vLLM build:
-#   BASE_IMAGE=nvidia/cuda:13.1.1-cudnn-devel-ubuntu24.04
-#   RUNTIME_IMAGE=nvidia/cuda:13.1.1-cudnn-runtime-ubuntu24.04
-#   INSTALL_VLLM=1   INSTALL_OLLAMA=0
+#   BASE_IMAGE=ubuntu:24.04       INSTALL_VLLM=1   INSTALL_OLLAMA=0
 #
 # Design notes
 # ────────────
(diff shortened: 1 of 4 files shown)


logos/logos-workernode/Dockerfile
```diff
@@ -90,188 +119,133 @@ RUN mkdir -p /ollama-root/usr/local/bin /ollama-root/usr/local/lib \
...
+        # Unversioned dev symlinks (wheels only ship SONAME-versioned .so's)
+        for so in "$CUDA_HOME"/lib/*.so.*; do \
+          base="${so%%.so.*}.so"; \
+          [ -e "$base" ] || ln -s "$(basename "$so")" "$base"; \
+        done; \
+        # Driver linker stub: empty shared object with SONAME libcuda.so.1.
+        # Satisfies the JIT's `-lcuda` at link time; at load time the SONAME
+        # binds to the real driver library injected by the container runtime.
+        mkdir -p "$CUDA_HOME/lib/stubs"; \
+        gcc -shared -Wl,-soname,libcuda.so.1 -o "$CUDA_HOME/lib/stubs/libcuda.so" -x c /dev/null; \
```
Claudia-Anthropica: @wasnertobias [high] This empty library does not act as a CUDA driver linker stub: Ubuntu GCC links with `--as-needed`, so because it exports no `cu*` symbols it is discarded, leaving JIT modules with unresolved driver symbols and no `DT_NEEDED` entry for `libcuda.so.1`. The link-only check still succeeds because unresolved symbols are allowed in shared objects; use NVIDIA’s real driver stub and verify both the dependency and an actual FlashInfer JIT module load on a GPU.

In `logos/logos-workernode/Dockerfile`, the generated empty `libcuda.so` exports no CUDA Driver API symbols and is discarded when FlashInfer links with `-lcuda`. Replace it with a functional NVIDIA driver stub, add a check that a representative linked shared object records `libcuda.so.1` in `DT_NEEDED`, and validate a FlashInfer JIT module load on a GPU.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The image slimming approach is well targeted and the builds are green, but the generated CUDA linker stub will not carry the driver dependency into FlashInfer JIT modules. Please replace it with a functional stub and validate an actual JIT module load; details inline.

logos/logos-workernode/Dockerfile
wasnertobias: Fixed in 5528a93c4 — you were right, and I reproduced the failure mode exactly in a clean `ubuntu:24.04` container: with the empty stub, a probe `.so` calling `cuInit`/`cuMemAlloc_v2` linked fine but had **no** `DT_NEEDED` entry for `libcuda.so.1` (only `libc.so.6`) — `--as-needed` silently dropped it.

Since NVIDIA's real driver stub isn't packaged in the cu13 wheels (checked `nvidia-cuda-runtime` 13.0.96: `nvidia/cu13/lib/` has no `stubs/`), the stub is now **generated from the wheel's own `cuda.h`**: the preprocessed header yields all ~449 driver-API functions with their real versioned symbol names (`cuMemAlloc_v2`, `cuGetProcAddress_v2`, …), which are compiled into the stub as exported no-op definitions under SONAME `libcuda.so.1`.

Two build guards added as requested:
1. exported-symbol count check (`nm -D`, ≥300) so a future header-layout change fails the build early, and
2. a representative JIT-style link (`gcc -shared … -L$CUDA_HOME/lib64/stubs -lcuda`) followed by `readelf -d` asserting `DT_NEEDED libcuda.so.1` is recorded.

Both validated in `ubuntu:24.04` with the exact Dockerfile commands under `dash`: generated stub → `(NEEDED) Shared library: [libcuda.so.1]` present; 449 symbols exported.

GPU validation of a real FlashInfer JIT module load will follow via *Logos - Deploy to Test* once this build has pushed the tag; I'll report the result here.

logos/logos-workernode/Dockerfile
```diff
@@ -90,188 +119,133 @@ RUN mkdir -p /ollama-root/usr/local/bin /ollama-root/usr/local/lib \
...
+#
+#   GPU build  (BASE_IMAGE=ubuntu:24.04):
+#     Python 3.12 from the stock noble main repo (NOT the deadsnakes PPA —
+#     GH Actions runners intermittently can't reach ppa.launchpadcontent.net,
+#     which used to flake-fail builds; 3.12 is the default Ubuntu 24.04
+#     Python and fully supported by vLLM). All CUDA libraries, compilers and
+#     headers come from vLLM's pip dependency tree (see design notes above).
 # ---------------------------------------------------------------------------
-FROM ${BASE_IMAGE} AS builder
+FROM ${BASE_IMAGE}
```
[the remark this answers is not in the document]
wasnertobias: The single-stage layout is deliberate and is the core fix of this PR, not an oversight — see the design notes at the top of the Dockerfile:

- A builder→runtime `COPY /opt/venv` keeps **two copies of the ~9 GB venv** on the build host (builder layer + runtime layer). Together with the old nvidia/cuda bases that pushed peak CI build disk to ~24 GB, past the ~14 GB GitHub-hosted runner guarantee — the root cause of the intermittent `No space left on device` failures this PR fixes.
- A separate build stage buys nothing here: everything installed is a binary wheel (nothing is compiled at image build time), and the only toolchain the image needs (gcc/g++ for Triton/FlashInfer *runtime* JIT) must be present in the final image anyway.
- The guideline's substance is retained: `VIRTUAL_ENV=/opt/venv` is kept (Dockerfile L173), the venv is the only Python environment, and the CPU/default build still uses `python:3.12-slim`. The GPU build uses `ubuntu:24.04` because vLLM needs the wheel-vendored CUDA toolkit + JIT toolchain regardless of stage structure.

Restoring a multi-stage build would reintroduce the disk-blowup and the CI flakiness without making the final image any smaller.

logos/logos-workernode/Dockerfile
Claudia-Anthropica: @wasnertobias Nice fix — the generated stub now exports the driver API and the link probe verifies that `libcuda.so.1` is retained. Looks good.

Claudia-Anthropica (approved):
@wasnertobias The CUDA linker-stub feedback is addressed and the vLLM image build is green. Nice fix — approving.

logos/logos-workernode/Dockerfile
wasnertobias: **GPU validation done** — the planned *Deploy to Test* route is currently blocked (the test-node deploy targets are intentionally pointed at `blocked_by_benchmark-*` hostnames while the icse27review benchmark occupies deipapa/deimama), so I built this branch's GPU image directly on **deioma** (RTX PRO 6000 Blackwell, the node from the original FlashInfer JIT incident) and validated on an idle GPU there:

1. **Image build** (`ubuntu:24.04`, `INSTALL_VLLM=1`): succeeds — including the new build guards (stub exports 449 driver symbols; JIT-style link probe records `DT_NEEDED libcuda.so.1`).
2. **Stub-linked `.so` loaded against the real injected driver**: a probe compiled with `-I$CUDA_HOME/include -L$CUDA_HOME/lib64/stubs -lcuda` records `(NEEDED) [libcuda.so.1]` and, loaded via `ctypes` inside the container with a GPU attached, `cuInit` + `cuDriverGetVersion` succeed against the runtime-injected driver (returned 13000).
3. **Actual FlashInfer JIT module compile + load on GPU** (flashinfer 0.6.13): `jit.gen_act_and_mul_module("silu").build_and_load()` compiled `silu_and_mul.cu` with the wheel-vendored nvcc for sm120 and loaded the module; the JIT-produced `.so` records `(NEEDED) [libcudart.so.13]` (this kernel uses only the runtime API, so `--as-needed` correctly omits `libcuda` — no unresolved symbols), and `silu_and_mul` executed correctly on the GPU.
4. `flashinfer.norm.rmsnorm` also ran correctly on GPU.

Rolling the tag to the test workernodes for a full benchmark scenario can follow once the benchmark block on deipapa/deimama is lifted.

PR #689, opened 2026-07-19
`Logos`: Improve performance for large Tables

flbrgit:
Added index for poll_success = TRUE to optimize queries.

Improves table and snapshot loading by ~40% through indexing.

## Summary by CodeRabbit

* **Performance**
  * Improved the speed of aggregate queries for successfully completed provider snapshots.
  * Added an optimized index covering snapshot timing, provider, identifier, and VRAM data for successful polls.
* **Database Changes**
  * Added a migration-managed PostgreSQL index with concurrent creation and rollback support.
  * Streamlined database setup so schema changes are applied through the migration process.


logos/db/init.sql
```diff
@@ -319,6 +319,16 @@ CREATE INDEX idx_provider_snapshots_success
 CREATE INDEX idx_provider_snapshots_models
     ON ollama_provider_snapshots USING GIN (loaded_models);
 
+-- Covers VramService's poll_success = TRUE aggregate queries
+-- (findSampledSnapshotIds, findProviderCapacities). Without it, both did a
+-- Seq Scan over the whole table to evaluate poll_success per row (the only
+-- other poll_success index only covers the FALSE case) — confirmed via
+-- EXPLAIN ANALYZE on a 2.36M-row snapshot: Seq Scan 1263ms/414ms per query
+-- vs. Index Only Scan (Heap Fetches: 0) 704ms/255ms with this index.
+CREATE INDEX idx_provider_snapshots_success_all
```
Claudia-Anthropica: @flbrgit [high] `init.sql` only runs when PostgreSQL initializes an empty data directory, so the existing 2.36M-row database this is meant to optimize will never create this index. Add an idempotent Liquibase changeset under `logos/logos-webservice/src/main/resources/liquibase/changelog/` and include it from `master.xml` so deployed databases receive it on startup.

In `logos/db/init.sql`, the new index exists only in the one-time PostgreSQL initialization script and is not applied to existing databases. Add an idempotent Liquibase changeset that creates `idx_provider_snapshots_success_all`, register it in `logos/logos-webservice/src/main/resources/liquibase/changelog/master.xml`, and keep fresh-install behavior consistent.

</details>

Claudia-Anthropica (changes requested):
@flbrgit The index itself matches the target aggregate queries, but it is currently added only to the one-time database initialization path. Existing Logos databases will not receive it, so the performance improvement needs a Liquibase changeset before this can ship.

Claudia-Anthropica (dismissed):
@flbrgit The Liquibase changeset now applies the index safely to existing databases, and the original feedback is fully addressed. Nice fix.

logos/db/init.sql
wasnertobias, line 1: we dont need this file any longer.

logos/logos-webservice/src/main/resources/liquibase/changelog/004_vram_snapshot_success_index.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+    <!-- VramService.getVramStats runs two aggregate queries
```
wasnertobias: we don't need so much docstring

wasnertobias (changes requested):

logos/db/init.sql
```diff
@@ -1,401 +0,0 @@
--- Database: logos
```
Claudia-Anthropica, line 1: @flbrgit [high] Deleting this file leaves two live references behind: `logos/db/Dockerfile` still does `COPY init.sql ...`, and `logos/docker-compose.dev.yaml` still bind-mounts it. The database image check is already failing, so remove or update both references as part of the Liquibase migration (or keep this file until those paths no longer depend on it).

In `logos/db/init.sql`, deleting the file breaks `logos/db/Dockerfile` and `logos/docker-compose.dev.yaml`, which still reference it. Remove or replace both references so database image builds and local Compose startup work with Liquibase-only initialization.

</details>

Claudia-Anthropica (changes requested):
@flbrgit The Liquibase path is the right direction, but the latest deletion breaks the database image and local Compose setup because both still reference `init.sql`. Please clean up those consumers; the database build check is already failing on this SHA.

Claudia-Anthropica (approved):
@flbrgit All requested changes are addressed: the Liquibase migration is registered, and the obsolete init.sql consumers are gone. Nice fix.

PR #701, opened 2026-07-21
`Iris`: Automatically switch the context of an active conversation

Senan04:
> [!NOTE]
> This should be tested with [PR #13292](https://github.com/ls1intum/Artemis/pull/13292) on Artemis

### Summary

Iris can now switch the context of a running chat on its own. When a student asks about a different exercise or lecture than the currently active context, the agent looks up the target's ID, calls the new `switch_chat_context` tool, and then answers about the new context. The requested target travels to Artemis on the final result status update, which applies the switch together with the answer.

This is the Pyris side of the feature. The Artemis counterpart is ls1intum/Artemis#13292 and both PRs have to be deployed together.

### Motivation

Students have to change the context of a chat manually through the context selection menu. Whenever they ask about another exercise or lecture without doing so, Iris answers with the wrong context in its prompt and its retrieval tools operate on the previous entity. Letting the agent request the switch removes that manual step.

### Changes

- **`switch_chat_context` tool** (`iris/src/iris/tools/switch_chat_context.py`): takes a target mode (`PROGRAMMING_EXERCISE_CHAT`, `TEXT_EXERCISE_CHAT`, `LECTURE_CHAT`, `COURSE_CHAT`) and an entity id. It validates the target against the execution DTO: unknown modes are rejected, exercise and lecture ids that do not exist in the course are rejected with a message that tells the agent which tool to use for the correct id, exercise types other than programming and text are rejected, and a mixed-up programming/text mode is silently corrected from the exercise type on the DTO. A switch to the course always uses the course id from the DTO, and a switch to the already active context clears the pending switch instead of registering a no-op. The tool changes nothing inside the running pipeline, it only records the target.
- **`lecture_list` tool** (`iris/src/iris/tools/lecture_list.py`, `course_dto.py`): lecture content retrieval reports lecture names but never lecture ids, and it only ever sees the active lecture, so it cannot supply the id a lecture switch needs. The new tool returns the id, the name and the lecture unit names of every lecture of the course, which gives the agent a non-guessing path from a lecture the student names to the id the switch takes. It reads the new `lectures` field on `CourseDTO`, which Artemis fills in the accompanying PR. An Artemis instance that does not send the field yet leaves the list empty: the provider logs a warning so the version skew stays visible, and the tool degrades to an empty list rather than failing. Lecture id validation in `switch_chat_context` is skipped in that case, leaving the check to Artemis.
- **Lecture retrieval scope** (`lecture_content_retrieval.py`, `chat_tool_providers.py`): retrieval used to bind its lecture scope when the tool was created, so a switch during a run left it pointed at the old lecture and the agent could not answer about the new one. The tool now resolves the scope at call time through a supplier that reads the pending switch. A retrieval after a switch to another lecture covers that lecture as a whole, without the lecture unit filter of the previous context, and a switch away from a lecture context widens retrieval to the whole course.
- **Pipeline state** (`abstract_agent_pipeline.py`, `chat_tool_providers.py`): `AgentPipelineExecutionState` carries a `pending_context_switch`, reset at the start of every run. The new `provide_switch_chat_context` provider wires the tool's callback to that field, so the last call wins.
- **Status update** (`suggested_context_dto.py`, `chat_status_update_dto.py`, `chat_pipeline.py`, `status_update.py`): the new `SuggestedContextDTO` (mode plus `entityId`) is attached to the final result status update and serialized under the `suggestedContext` alias that Artemis expects. A run without a switch sends `suggestedContext: null`, the same shape the other optional fields of the status DTO use when they are unset.
- **System prompt** (`chat_system_prompt.j2`): a context switching block states the currently active context and instructs the agent to look up the target id first, call the tool, and then answer in the new context. It names the exercise list tool for exercises and the lecture list tool for lectures, and it states that lecture content retrieval is scoped to the active lecture and does not help in finding another one. Worked examples cover both the exercise case and the lecture case, where the agent calls lecture content retrieval again after the switch to reach the content of the new lecture. The block rules out switching on a passing mention or a comparison, requires a switch back to the course context for general course-level questions, and asks the agent to keep the announcement short because Artemis renders a context switch divider. The block is skipped for the MCQ variant.

### Tests

`tests/test_switch_chat_context_tool.py` covers target validation and correction for exercises, lectures and the course, the no-op switch to the active context, and the wire shape of `suggestedContext` in both directions. The lecture A to B case runs through the real tool providers: it performs the switch and then invokes the lecture retrieval tool the pipeline hands the agent, asserting that the retriever is queried for the new lecture and without the lecture unit of the old one. `tests/test_lecture_list_tool.py` covers the lecture list output, including a lecture whose units are all unreleased and a course whose lectures Artemis did not send.

## Summary by CodeRabbit

* **New Features**
  * Added agent-driven context switching for course, lecture, and exercise discussions.
  * Chat status updates can include an optional `suggestedContext` with mode and entity ID.
  * Added lecture listing with lecture, unit, and identifier details.

* **Improvements**
  * Context switches are validated against available content and applied with the final response.
  * Lecture content retrieval now follows the active context.
  * Updated chat guidance for switching between contexts.

* **Tests**
  * Added coverage for lecture listing, context switching, retrieval behavior, and status payloads.


iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2
```diff
@@ -116,6 +116,33 @@ You can use the tools to look up information to provide accurate responses. You
...
+{% if chat_mode == "COURSE_CHAT" %}
+The currently active context is the course "{{ course_name }}".
+{% elif chat_mode == "LECTURE_CHAT" %}
+The currently active context is the lecture "{{ lecture_name }}".
+{% elif exercise_title %}
+The currently active context is the exercise "{{ exercise_title }}".
+{% endif %}
+When the student's message is about a DIFFERENT exercise or lecture than the active context, you MUST call the `switch_chat_context` tool BEFORE giving your answer. Answering a question about another exercise or lecture WITHOUT calling `switch_chat_context` first is a mistake.
+Follow these steps every time the student asks about another exercise or lecture:
+1. Find the target's ID: use the exercise list tool for exercises, or lecture content retrieval for lectures. NEVER guess an ID.
```
Claudia-Anthropica: @Senan04 [high] This lookup path cannot produce the ID needed for a lecture switch: `lecture_content_retrieval` outputs lecture names but never `lecture_id`, and in `LECTURE_CHAT` its provider filters retrieval to the currently active lecture. The agent therefore has no non-guessing way to switch from lecture A to lecture B or answer from B, so this core advertised flow will fail. Please add a course-wide lecture lookup that returns IDs while keeping ordinary in-context retrieval scoped, and cover A→B with a regression test.

In `iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2`, line 133 tells the agent to obtain a target lecture ID from a tool that neither exposes IDs nor searches outside the active lecture. Add a course-wide lecture lookup that returns lecture IDs, update the prompt to use it, and add a regression test for switching and answering from lecture A to lecture B.

</details>

Claudia-Anthropica (changes requested):
@Senan04 The state and status plumbing looks clean, but the lecture-switching path cannot currently discover a target lecture ID or retrieve outside the active lecture. That blocks a core advertised flow; see the inline comment.

bassner (dismissed):
Locally verified end-to-end with Artemis #13292 in isolated OrbStack containers. The switch_chat_context tool validated and emitted the text-exercise target, and Artemis applied and rendered the context switch with the final response.

iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2
Claudia-Anthropica: @Senan04 The functional path looks right now: the new list exposes lecture IDs and retrieval reads the pending lecture switch. The regression still stops after recording the switch, though; please invoke the provided lecture retrieval after A→B and assert the retriever receives `lecture_id=42` (and no old lecture unit), so the original answer-from-B failure is actually covered.

Claudia-Anthropica (changes requested):
@Senan04 The course-wide lecture lookup and dynamic retargeting fix the functional blocker. The requested A→B regression still needs to exercise retrieval after the switch and verify it queries lecture B; I left the exact assertion in the existing thread.

iris/tests/test_switch_chat_context_tool.py
```diff
@@ -0,0 +1,328 @@
...
+def test_send_result_without_switch_omits_suggested_context():
+    callback = ChatRunCallback("run-1", "https://artemis.example", None)
+
+    with patch.object(
+        ChatRunCallback, "_send_status_payload", return_value=True
+    ) as send:
+        assert callback.send_result("answer", tokens=[], suggested_context=None)
+
+    payload = send.call_args.args[0]
+    assert payload["suggestedContext"] is None
```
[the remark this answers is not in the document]
Senan04: I have updated the PR description.

Claudia-Anthropica (dismissed):
@Senan04 The requested A-to-B regression now exercises the provided retrieval tool and verifies the new lecture scope, and the implementation looks good. Approving this update.

iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2
Claudia-Anthropica: @Senan04 Thanks, this now drives the retrieval tool after the A→B switch and verifies that it queries lecture 42 without carrying over the old lecture unit. That covers the original failure path.

Predixx (approved):
One gap: in a course with lectures but no ingested content the lecture path never switches. `provide_lecture_list` is gated on `allow_lecture_tool`, so the agent gets no lecture IDs, while the prompt tells it to look one up and never guess. Gating on `dto.course.lectures` would be enough.

Predixx (dismissed):
Tested locally against Artemis #13292 with a real LLM. Course → exercise, exercise → exercise and back to course all switch correctly, including the marker and the session mode.

Claudia-Anthropica (changes requested):
@Senan04 The merge preserves the context-switching and dynamic-retrieval paths, and the Iris test check passes. However, the previously noted [medium] lecture-discovery gap remains: `provide_lecture_list` is gated by `allow_lecture_tool`, which is false when a course has lectures but no indexed lecture units, leaving the agent unable to obtain a target lecture ID and perform the advertised switch. Make lecture-list availability depend on `state.dto.course.lectures` rather than retrieval-index availability, and cover that provider case with a regression test.

bassner:
@Claudia-Anthropica review

Claudia-Anthropica:
@bassner acknowledged, queuing this up for immediate review!

iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2
```diff
@@ -116,6 +116,35 @@ You can use the tools to look up information to provide accurate responses. You
...
+{% elif chat_mode == "LECTURE_CHAT" %}
+The currently active context is the lecture "{{ lecture_name }}".
+{% elif exercise_title %}
+The currently active context is the exercise "{{ exercise_title }}".
+{% endif %}
+When the student's message is about a DIFFERENT exercise or lecture than the active context, you MUST call the `switch_chat_context` tool BEFORE giving your answer. Answering a question about another exercise or lecture WITHOUT calling `switch_chat_context` first is a mistake.
+Follow these steps every time the student asks about another exercise or lecture:
+1. Find the target's ID: use the exercise list tool for exercises, or the lecture list tool for lectures. NEVER guess an ID. Lecture content retrieval is scoped to the active lecture and does not help you find another lecture.
+2. Call `switch_chat_context` with the matching mode ("PROGRAMMING_EXERCISE_CHAT", "TEXT_EXERCISE_CHAT", "LECTURE_CHAT", or "COURSE_CHAT") and the entity ID.
+3. Then answer the student's question about the new context in the same response.
```
Claudia-Anthropica, line 135: @Senan04 [medium] This instruction requires answering about exercise B immediately after switching, but the submission-details, additional-details, build-log, feedback, repository, and file-lookup providers all close over exercise A's DTO before agent execution. A question such as “why is my B build failing?” can therefore cause A's code or results to be reported as B's. Gate these tools after an exercise switch so they cannot return original-context data.

In `iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2`, exercise-specific tools remain bound to the original exercise after an A→B context switch, so the agent can misattribute A's submission data to B. Gate every exercise-specific provider on `pending_context_switch` so that after an exercise switch it returns an explicit unavailable-until-next-turn result instead of original-context data, and add an A→B regression proving A's data cannot be returned.

</details>

Claudia-Anthropica (changes requested):
@Senan04 The new commits correctly decouple lecture-list availability from indexed lecture content, and the captured Iris test and lint checks pass. However, the same-run exercise-switch path can still use tools bound to the previous exercise and misattribute its submission data to the target. Please prevent original-context exercise tools from serving data after a switch, as detailed inline.

thynguyentumde (approved):
Tested locally along with the Artemis PR. Confirmed Iris can detect and switch context automatically.

PR #710, opened 2026-07-27
`Iris`: Improve global search retrieval quality with a reranked two-stage pipeline

Nayer-kotry:
> Draft note: the two blockers that kept this in draft are resolved. The Logos serving fix for the Qwen3 reranker chat template landed (#718, #742), and the relevance cutoff has been recalibrated against the Logos-served reranker (see Testing). Remaining before merge: a test-server pass of the recalibrated floor.

## Motivation

Production testing of global search surfaced two complaints: the lecture results list ranked poorly, and the Iris answer bubble kept appearing and then vanishing without an answer. The investigation traced both to retrieval: a metadata fetch truncation silently dropped valid hits, the candidate ordering had no defense against junk results when no relevant content exists, and the HyDE step added latency and a failure-prone LLM dependency without measurable quality benefit. This PR reworks the retrieval pipeline around those findings.

## Description

- Retrieval becomes two-stage: fetch a deep candidate set from both lecture collections, rerank it with the configured reranker, and gate the pool. If the reranker is unavailable or times out, the search falls back to the fused ordering, so it can never degrade below pre-reranker behavior.
- Rerankers are normalized behind one provider-neutral interface (RerankModel / RerankResponse). The Cohere client conforms to it, and a new vllm_rerank client speaks the /v1/rerank endpoint that vLLM-served cross-encoders expose (Qwen3-Reranker-8B on Logos). The gateway applies the reranker chat template server-side, so the client sends the query untouched; a contract test pins the response normalization for both providers.
- The rerank cutoff is now a junk floor at 0.10 instead of a relevance threshold at 0.30. It is calibrated against what irrelevant candidates score (junk peaks at 0.065, run-to-run noise about +/-0.01), so it removes garbage rather than weak answers, and an all-below pool stays the honest empty state. The old 0.30 cutoff deleted real answers and sat inside reranker noise, so identical requests could return different results.
- Answer-path unit expansion: once a candidate survives the floor, the rest of its lecture unit is fetched by structural join instead of ranked. Measured on the scattered-content harness, at least one relevant item surfaced for 93 percent of queries but every collection holding relevant material was represented for only 16 percent, because siblings lose ranking slots they should not have to win. Expanded siblings pass through the same visibility policy as the main path and inherit their anchor score. Expansion is answer-path only; the instant results list stays a ranked list.
- Query embeddings use the Qwen3 retrieval instruction prefix (query side only, no re-ingestion required).
- Removes HyDE from the retrieval path.
- Fixes the metadata fetch truncation that silently dropped results.
- Removes the hand-maintained low-information filter; junk demotion is owned by the reranker.
- Deduplicates near-identical snippets and emits a per-query retrieval telemetry line for production diagnosability.

## Testing

- Head-to-head A/B of this branch against the main pipeline on the test server (same labeled query set, in-process runner): false nulls on answerable queries drop from 30 percent to 3 percent, top-1 hit rate 0.33 to 0.78, used-source precision 0.67 to 0.95, false answers zero on both arms.
- Floor recalibration against the Logos-served Qwen3-Reranker-8B on a 38-query labeled suite (content, transcript, navigational, FAQ, German, negative categories): deliberately irrelevant candidates peaked at 0.065 across runs while relevant lecture content sat at 0.24 to 0.62; at the 0.10 floor all 5 negative queries return zero sources and no answerable category loses a source. Reranker latency p50 was 0.5 to 0.8 seconds at pool size 45.
- HyDE on/off ablation: no quality change on any metric, answer latency p50 30 percent lower, empty-output model errors 12 per run to 0.
- Per-component ablation attributing the lift to each change. Note: the metadata fetch fix is the largest single factor on the test server but is partly a test-data artifact, so production gains will be smaller.
- Unit tests: expansion join semantics, floor gating, visibility policy threading through expansion, and the rerank response contract for both providers (47 tests across the three touched files).

## Summary by CodeRabbit

- **New Features**
  - Added configurable search reranking, result expansion, and improved retrieval across lecture content.
  - Added support for streaming partial answers with sequence tracking.
  - Added inline citations and clearer handling of unavailable or ungrounded answers.
  - Added support for vLLM-compatible reranking services.

- **Improvements**
  - Global search now responds in the question’s language and can retry with keyword-focused retrieval.
  - Improved filtering of inaccessible, duplicate, or incomplete search results.
  - Added tracing and timing information for search requests.

- **Tests**
  - Expanded coverage for citations, streaming responses, reranking, visibility filtering, and search result handling.


Nayer-kotry:
Folded into #912, which now includes this branch's retrieval work plus entity grounding and the already-merged #913 (inline citations and streaming). Retargeted #912 to main so it carries everything as one PR.

PR #711, opened 2026-07-27
`Iris`: Create point out tool with acknoledgement call

vivienxr:
Works together with: https://github.com/ls1intum/Artemis/pull/13315

**Command mechanism:** StatusCallback.execute_command sends a CommandDTO ({type, parameters}) synchronously to Artemis (…/runs/{runId}/command) and blocks until Artemis has driven the client and replied with a CommandResultDTO (applied: true/false). The agent therefore knows the real outcome before formulating its answer and won't claim it showed something that never happened. Transport errors, an unusable callback URL and timeouts all count as "not applied", so the pipeline never hangs on a command. 

**Point-out tool:** PointOut is one type of command. When the student is in the combined view, the agent gets point_out_relevant_lecture_position and can jump the view to a specific slide page and/or video timestamp. The tool doesn't search: the agent has to run a lecture retrieval first and pick the position from those results — pages and timestamps that don't appear there, or that belong to a different lecture unit than the one the student is viewing, are rejected (page and timestamp are checked independently so a call that got both wrong learns about both at once). If the student is already at that position, the view is left untouched; at most one point-out per answer. 

**Current-position tool:** The material at the student's current position is no longer written into the system prompt — the prompt now only names the position, and read_students_current_position reads out the material on request. This move was made purely for that reason: with the material spelled out in the prompt, the agent frequently stopped calling the other tools altogether. The prompt ended with something that reads like a finished answer, so the model answered from it instead of reaching for lecture retrieval or the point-out tool. Behind a tool the same material stays available, but only when the agent decides it wants it. Once a point-out has moved the student, that snapshot is invalidated: the tool then reports that the position it described is no longer the one the student is looking at.

**Supporting changes:**
- Retrieval results now separate the page number printed on the slide (safe to quote to the student) from the technical "point-out id" Artemis navigates by; slides without a printed number show as "unnumbered" instead of -1. Transcription segments show only the printed page (their page number is read off the video frame, not a deck index) plus a "Video timestamp: Ns" that the point-out tool accepts as-is — rounded up so it can't fall outside its own segment.
- New COMMAND message role: Artemis stores a marker after an executed command, which is rendered as a system note in later runs ("Iris pointed the student to … earlier"). It's a record, not a restriction — the agent may point to the same spot again. Rendering lives next to the prompts and degrades to a usable sentence for unknown command types or malformed parameters, so a marker can never break a run.
- The point-out also sends the printed page number (displayPage) so Artemis can label the chat-history chip with the same number Iris names in its answer. 
- System prompt explains the tool's usage and limits (retrieve first, never leak point-out ids into the answer, when in doubt point). Both new prompt blocks are gated on the corresponding tool actually being registered, so the agent is never sent after a tool that doesn't exist.
- Tests for the point-out tool logic, the current-position tool, command execution, marker rendering, and page/timestamp references in the retrieval results.

## Summary by CodeRabbit

* **New Features**
  * Added support for in-chat commands that can navigate lecture slides and video positions.
  * Added combined-view navigation for pointing out relevant pages or timestamps.
  * Added access to current-position lecture content on demand.
  * Improved slide, page, and timestamp references using student-visible labels.

* **Improvements**
  * Command results now report whether navigation was successfully applied.
  * Added clear fallback messages for unavailable or invalid command details.
  * Updated guidance for retrieving lecture content and referencing slide locations.


vivienxr:
@coderabbitai review

iris/src/iris/tools/lecture_content_retrieval.py
```diff
@@ -64,22 +81,31 @@ def lecture_content_retrieval() -> str:
...
 
         result += "Lecture transcription content:\n"
         for paragraph in lecture_content.lecture_transcriptions:
             result += (
+                # Transcription segments carry no printed page number, only the slide index, so the
+                # point-out id is offered on its own rather than under a "Page:" label the agent is
+                # told to quote to the student.
                 f"Lecture: {paragraph.lecture_name}, Unit: {paragraph.lecture_unit_name}, "
-                f"Page: {paragraph.page_number}\nContent:\n---{paragraph.segment_text}---\n\n"
+                f"point-out id: {paragraph.page_number}, "
```
Claudia-Anthropica: @vivienxr [medium] `LectureTranscriptionRetrievalDTO.page_number` is not the PDF/point-out index: transcription ingestion stores `segment.slide_number`, and `_get_associated_slide()` joins it against `LectureUnitPageChunkSchema.DISPLAY_PAGE_NUMBER`. Labeling it as a point-out id can therefore cause a rejected command or navigation to the wrong physical slide whenever printed numbering differs from PDF order. Remove the point-out id from transcription output and expose only its video timestamp; technical point-out ids can continue to come from page-chunk and lecture-segment results.

In `iris/src/iris/tools/lecture_content_retrieval.py`, transcription `page_number` represents the printed/display slide number but is emitted as a technical point-out id. Remove that field from transcription result text and expose only the video timestamp, leaving technical point-out ids to slide/page-segment results.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The command transport, active-unit validation, marker rendering, and tests otherwise look well integrated, and the captured CI checks pass. One medium correctness issue remains: transcription metadata is presented as a technical navigation ID even though it represents the displayed slide number.

iris/src/iris/tools/combined_view_point_out.py
```diff
@@ -0,0 +1,258 @@
...
+    """Find the retrieved transcription segment of ``lecture_unit_id`` containing ``timestamp``.
+
+    Grounds timestamps in the retrieval results the same way pages are grounded, including the
+    restriction to the unit the point-out will navigate in. Returns None when no such segment was
+    retrieved.
+    """
+    for segment in lecture_content.lecture_transcriptions:
+        if (
+            segment.lecture_unit_id == lecture_unit_id
+            and segment.segment_start_time <= timestamp <= segment.segment_end_time
```
Claudia-Anthropica: @vivienxr [medium] Transcript intervals are half-open elsewhere: `_fetch_transcriptions_by_timestamp` requires `end > timestamp`, and current-view matching uses `< segment_end_time`. With adjacent segments `[0, 10)` and `[10, 20)`, requesting the displayed `10s` timestamp can select the first segment; if the student is at `5s`, the later inclusive current-position check then reports that they are already there and sends no command instead of jumping to the second segment. Use `< segment_end_time` in both the target-resolution and current-position checks, and add an adjacent-segment boundary test.

In `iris/src/iris/tools/combined_view_point_out.py`, timestamp segment membership treats end timestamps as inclusive, allowing a boundary timestamp to resolve to the preceding adjacent segment and suppress needed navigation. Make both target resolution and current-position membership half-open (`start <= timestamp < end`) and add a regression test with adjacent segments.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The transcription page-number correction resolves the prior finding, and the active-unit filtering and command handling otherwise look coherent. One medium boundary bug remains: adjacent transcription segments are treated as closed at both ends, so a requested next-segment timestamp can resolve to the preceding segment and be incorrectly skipped as "already there".

iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2
```diff
@@ -155,6 +155,19 @@ You can retrieve lecture content, including slides, transcripts, and summarized
 • Use lecture retrieval to answer questions about lecture material, concepts, or to help the student connect content with their learning strategies.
 • Only reference lecture content if it's directly relevant to the student's question or reflection.
 • Reference lecture material using descriptive markdown links where appropriate.
+• Each retrieved slide shows the page number printed on it ("Page: N") and a technical "point-out id". The point-out id is internal: never write it in your answer or show it to the student. When you mention a slide, refer to it by its printed page number ("Page: N") — that is the number the student sees on the slide itself. A slide shown as "Page: unnumbered" has no printed number: refer to it without naming a page (e.g. "the slide on quicksort"), never by its point-out id.
+• Transcription results show the "Page: N" of the slide that was on screen while it was spoken, so you can tie a spoken passage to a slide ("he explains this on page 2, from 0:42"). They carry no point-out id — only results that show one can be pointed at by page.
```
Claudia-Anthropica: @vivienxr [medium] This rule is contradicted by the Current Position block: `_build_current_view()` renders `p["page"]`, which is the technical PDF `PAGE_NUMBER` used for retrieval, without exposing the chunk's `display_page_number`. When printed numbering differs from deck order, the agent is therefore explicitly told that the internal index is the page the student sees and can quote the wrong number despite this instruction. Render the current-view label from `display_page_number` (using `unnumbered` for nonpositive sentinels), retain the technical value only for lookup/navigation, and add a shifted-number regression test.

In `iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2`, the printed-page rule is contradicted by current-view rendering, which labels the technical PDF page index as the visible page. Update `_build_current_view` in `iris/src/iris/pipeline/chat/chat_pipeline.py` to render the retrieved chunk's `display_page_number` or `unnumbered`, retain `page_number` only for lookup, and cover differing display and technical values with a regression test.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The new display-page marker handling is coherent and the captured CI checks pass, but the prior medium timestamp-boundary finding remains unresolved: both target and current-position checks still include segment end times. There is also a separate medium page-number leak in the current-view prompt, which labels the technical PDF index as the page visible to the student. Please use half-open transcript intervals and render current-view pages from `display_page_number`, with an unnumbered fallback.

Claudia-Anthropica (dismissed):
@vivienxr Both prior medium findings remain: transcript intervals still include the segment end during target resolution and current-position suppression, so adjacent-boundary navigation can be skipped. The refactored current-view prompt and tool also still render the technical PDF index instead of `display_page_number`, exposing the wrong visible page to the agent.

iris/src/iris/tools/lecture_content_retrieval.py
```diff
@@ -64,22 +87,33 @@ def lecture_content_retrieval() -> str:
...
             result += (
+                # A transcription segment's page_number is the slide number read off the video frame
+                # during ingestion — the number printed on the slide, not its index in the deck. So it
+                # is rendered as the printed page, which the agent may name to the student, and never
+                # as a point-out id: passing it as one would navigate to the wrong slide wherever the
+                # two numberings diverge. What this segment can be pointed at by is its timestamp.
                 f"Lecture: {paragraph.lecture_name}, Unit: {paragraph.lecture_unit_name}, "
-                f"Page: {paragraph.page_number}\nContent:\n---{paragraph.segment_text}---\n\n"
+                + _format_page_reference(paragraph.page_number)
+                + f", Video timestamp: {paragraph.segment_start_time:.0f}s"
```
Claudia-Anthropica: @vivienxr [medium] Formatting `segment_start_time` with `.0f` can round a fractional start downward (for example, `42.4` becomes `42s`), but `_resolve_timestamp_segment` accepts only `start <= timestamp < end`. The agent is told to pass the displayed value exactly, so that retrieved segment can become impossible to point out; emit a timestamp guaranteed to remain within the segment (or preserve sufficient precision) and add a fractional-start regression test.

In `iris/src/iris/tools/lecture_content_retrieval.py`, fractional transcription start timestamps are rounded to values that can fall outside their own half-open segment and are then rejected by the point-out tool. Preserve enough precision or choose an in-segment displayed timestamp, and add a fractional-start regression test.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The prior half-open interval, displayed-page, and lecture-content gating findings are resolved, and the captured CI checks pass. One medium correctness issue remains: fractional transcription starts are rounded before being offered as navigation timestamps, so a displayed value can be rejected by the grounding check.

Claudia-Anthropica (dismissed):
@vivienxr The current head resolves fractional-start navigation by emitting a timestamp guaranteed to remain within its retrieved segment, with regression coverage for whole-second and short fractional segments. The earlier half-open interval, displayed-page, and lecture-content gating fixes remain intact, and the snapshot reports no required pending or failing checks.

iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2
Claudia-Anthropica: @vivienxr Thanks, this is fixed now: current-view slide labels use `display_page_number` with an unnumbered fallback, and the regression tests cover shifted and sentinel values.

vivienxr:
@coderabbitai review

iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2
```diff
@@ -155,6 +155,19 @@ You can retrieve lecture content, including slides, transcripts, and summarized
...
+{% endif %}
+
+{# Combined-view point-out Block #}
+{% if allow_lecture_tool and current_view_is_combined %}
```
Claudia-Anthropica: @vivienxr [low] `current_view_is_combined` is true for any `combinedView` entry, including the accepted empty context covered by `test_provider_stays_silent_without_a_resolvable_lecture_unit`; in that case the provider withholds `point_out_relevant_lecture_position`, but this block still tells the model that the tool exists and directs it to use it. Gate this block on the same resolvable-context condition as the provider so the prompt cannot advertise an unavailable tool.

In `iris/src/iris/pipeline/prompts/templates/chat_system_prompt.j2`, an empty `combinedView` context can render instructions for a point-out tool that `provide_combined_view_point_out` does not register. Compute a point-out-tool availability flag using the provider's resolvable-context condition and gate this prompt block on that flag.

</details>

Claudia-Anthropica (dismissed):
@vivienxr The command transport, unit-scoped grounding, half-open transcript intervals, displayed-page handling, and fractional timestamp fixes remain coherent with their regression coverage. The captured snapshot has no required pending or failing checks; only one low-severity prompt/tool-availability mismatch remains.

iris/src/iris/tools/combined_view_point_out.py
```diff
@@ -0,0 +1,293 @@
...
+        )
+        if not result.applied:
+            # Case 3: the student left the combined view (or a timeout).
+            return (
+                "The student's view could not be moved (they may have left the combined view). "
+                "Do not tell the student you moved their view."
+            )
+
+        # Case 2: navigated successfully. This answer's one point-out is now spent.
+        already_moved = True
```
Claudia-Anthropica, line 319: @vivienxr [medium] After this successful move, `read_students_current_position` remains callable and still returns the content blocks captured before navigation, labelled as what the student is viewing “right now.” A later tool call can therefore overwrite the success result's warning with stale slide/transcript content and produce an answer about the wrong view. Invalidate the current-view tool's shared backing data when a command is applied, and add a regression that calls it after a successful point-out.

In `iris/src/iris/tools/combined_view_point_out.py`, a successful navigation leaves `read_students_current_position` backed by pre-navigation content, so it can report stale material as the student's current view. Pass shared mutable current-view state into the point-out tool, invalidate it when `applied` is true, and add a regression that invokes the current-position tool after a successful point-out.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The previous timestamp, page-label, gating, and marker findings are resolved on this head. One medium consistency issue remains: after a successful point-out, the current-position tool can still return the pre-navigation slide or transcript as if it were current, allowing the agent to formulate its answer from stale view state.

iris/src/iris/tools/combined_view_point_out.py
```diff
@@ -0,0 +1,299 @@
...
+            combined_context.video.timestamp
+            if combined_context.video is not None
+            else None
+        )
+        same_page = page is not None and page == current_page
+        # The student counts as already at the video position when they are anywhere inside the
+        # targeted segment's time interval, not only at its exact start. Half-open like the
+        # resolution above: a student sitting exactly on the boundary belongs to the next segment,
+        # and counting them as inside this one would suppress a jump they still need.
+        same_timestamp = (
```
Claudia-Anthropica: @vivienxr [medium] Transcript intervals can overlap: ingestion groups recurring appearances of a slide into one start/end span, and semantic chunks can also share time ranges. Using the first target segment's entire interval for `same_timestamp` can therefore declare a student at 5s “already there” for a requested 10s point merely because a broad retrieved interval spans both, so no command is sent. Only suppress timestamp navigation when the current timestamp actually matches the requested timestamp (with an appropriate tolerance), and add an overlapping-interval regression test.

In `iris/src/iris/tools/combined_view_point_out.py`, `same_timestamp` treats any current time inside the first retrieved interval as the requested target, but retrieved transcription intervals can overlap and span distinct moments, causing valid navigation to be skipped. Compare the current timestamp to the requested timestamp with an appropriate tolerance instead of using target-segment membership, and add an overlapping-interval regression test.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The shared current-view storage and post-navigation invalidation resolve the prior stale-position finding, and the earlier page-label and prompt fixes remain intact. One medium timestamp issue remains: overlapping retrieved transcript intervals can make the no-op check incorrectly report that the student is already at a different requested moment. The snapshot has no required pending or failing checks.

iris/src/iris/tools/combined_view_point_out.py
Claudia-Anthropica: @vivienxr Thanks, this is fixed now: successful point-outs invalidate the shared current-view storage, and the reader no longer returns pre-navigation material.

Claudia-Anthropica (dismissed):
@vivienxr The remaining overlapping-interval finding is resolved: the no-op check now compares the current player position directly with the requested timestamp, with regression coverage confirming distinct moments still navigate. The earlier unit-scoping, page-label, gating, fractional-timestamp, and stale-view fixes remain intact; the captured test check passes and no required checks are pending or failing.

iris/src/iris/tools/combined_view_point_out.py
Claudia-Anthropica: @vivienxr Thanks, this is fixed now: the no-op check compares the current player position directly with the requested timestamp using a narrow tolerance, and the overlapping-interval regression confirms distinct moments still navigate.

Predixx (dismissed):
Tested locally against #13315. Works as expected.

Predixx:
Two nit test gaps. Same shape both times: remove the line and the suite stays green while production breaks.

**`combined_view_point_out.py:81`**: the segment loop never runs, every test helper sets `lecture_unit_segments=[]`. Retrieval does render segments with a point-out id though, so the agent gets sent there. Without the loop those point-outs come back as "not among the retrieved results".

**`message_converters.py:190`**: the new COMMAND case has no test and the default branch raises. Drop that line and every turn after a successful point-out breaks on the direct OpenAI and Ollama backends.

thynguyentumde (dismissed):
Tested along the Artemis PR, worked as expected

Claudia-Anthropica (approved):
@vivienxr The current head keeps page and timestamp grounding unit-scoped, resolves the slide and video panes independently, and sends their changes as one command while preserving the prior display-page, interval, and stale-view fixes. I found no unresolved medium-or-higher defect, and the snapshot reports no required failing or pending checks.

Predixx (approved):
reapprove

thynguyentumde (approved):
Reapproved.

Predixx (approved):
reapproved

Predixx (approved):
reapprove

PR #712, opened 2026-08-02
`Iris`: Apply the user access context in global search

Nayer-kotry:
> [!WARNING]
> Must be tested together with the paired Artemis PR (https://github.com/ls1intum/Artemis/pull/13380) — the access context is a cross-repo wire contract. Deploy both branches to the same test server before reviewing.

## Summary

- Add an optional `AccessContext` (role-grouped course IDs, request timestamp, admin `unrestricted` flag) to the global-search request DTOs, resolved by Artemis and treated by Pyris as an opaque permission filter.
- Thread the context through the lecture-search list endpoint and the iris-answer pipeline into `LectureGlobalSearchRetrieval` (both share one retriever).
- Scope every retriever call: the client `courseIds` filter is intersected with `accessContext.course_ids`; an `unrestricted` (admin) context has no course ceiling.
- Make release visibility role-aware instead of role-blind: admins and staff of a course see unreleased units; everyone else is gated on the unit release date evaluated at `accessContext.now`. Per-slide `hidden_until` stays enforced for all roles.

## Description

- `resolve_effective_course_ids` intersects the client filter with the permitted courses; an empty effective scope short-circuits before the embedding call.
- A per-search `_VisibilityPolicy` (built from the access context) is passed into `_segment_to_dto` / `_transcription_to_dto`. It waives the unit release gate when the context is `unrestricted` or the hit's course is in `staff_course_ids`, and uses `accessContext.now` (falling back to the server clock) for the student release check. `is_transcription_visible` gains a `bypass_release` flag that waives only the unit release date, never the per-slide gate.
- Missing `release_date` stays visible and per-slide hidden content stays filtered for everyone, matching current behavior.
- All new fields are optional with defaults, so a request without `accessContext` behaves byte-identically to today.

## Validation

- `poetry run pytest tests/test_lecture_visibility_retrieval_filters.py` (29 passed): the new tests drive the public `search()` API and cover the admin and staff bypass, the student release gate, the request-time boundary, the safe default with no context, and that a staff bypass never exposes a hidden slide.
- `black` and `isort` clean on the changed files. The full suite runs in CI (no full local runs).
- Also verified end-to-end against a running stack with real student / admin / instructor JWTs; the Iris path now matches the direct SearchableEntities search for every role.

Paired with https://github.com/ls1intum/Artemis/pull/13380

## Summary by CodeRabbit

* **New Features**
  * Global search now respects access permissions and course scope.
  * Results can be filtered by the requester’s role, permitted courses, and request time.
  * Authorized staff and unrestricted access can view unreleased lecture content while slide-level restrictions remain enforced.
  * Empty access scopes return no results without unnecessary processing.

* **Bug Fixes**
  * Preserved legacy search behavior when no access context is provided.

* **Tests**
  * Added coverage for course filtering, visibility rules, release boundaries, and access scenarios.


Claudia-Anthropica (approved):
@Nayer-kotry The access context is consistently propagated through direct and HyDE retrieval, with course scope resolved before search and release bypass kept separate from slide visibility. The captured Iris tests, lint, build, and PR-title validation all pass, and I found no blocking correctness or security issues.

vivienxr (changes requested):
See here: https://github.com/ls1intum/Artemis/pull/13380#pullrequestreview-4918134675

PR #715, opened 2026-08-06
`Iris`: Superseed running ingestion job for same unit

vivienxr:
## Summary

This PR fixes overlapping lecture ingestions for the same lecture unit.

If a second ingestion request arrives while a first one is still running:
- the new run should start immediately
- the old run should be marked as superseded
- the old run must no longer be allowed to perform final writes or send a terminal success update once the new run has taken over

The fix stays local to lecture ingestion and related update paths. It does not introduce a generic thread cancellation framework.

## Background

PR #561 tried to address the same issue, but it is closed.

Reusing that branch was no longer practical because Iris has changed too much since then. Reimplementing the fix on top of the current codebase was simpler and safer than reviving the old branch.

## Problem

Two ingestion runs for the same lecture unit could overlap in unsafe ways:

- a follow-up request could be treated as a duplicate and effectively be dropped
- or an older run could still reach a later write step and overwrite newer data in Weaviate
- or a superseded run could still send a terminal `finish()` callback

Hard-stopping Python threads is not a good option here, so the fix has to be cooperative and built around the actual write boundaries.

## Approach

The intended behavior is:

- the newest same-unit request starts immediately
- the previous run is marked as superseded
- older runs may still continue preprocessing already in flight
- but once a newer run has taken over, the older run must not perform final writes or send `finish()`

To do that, this PR adds cooperative cancellation plus shared synchronization around final mutation boundaries.

## Locking model

This PR uses two locks, each for a different purpose:

1. `batch_update_lock`  
   This already existed and is the global lock for Weaviate batch-style mutation paths.  
   It prevents overlapping global write phases.

2. `lecture_update_lock`  
   This also already existed as a per-lecture-unit lock.  
   It is now reused as the per-unit ownership boundary for:
   - registering a newer same-unit ingestion run
   - final replace-style writes
   - terminal `callback.finish()`

The important rule is now consistent everywhere:

- acquire `batch_update_lock` first if the path writes through the shared batch write phase
- then acquire `lecture_update_lock`
- then do the final ownership/cancellation check
- then perform the write contiguously

This avoids deadlocks and ensures that a newer same-unit run cannot take over in the middle of an older run’s final write boundary.

## What changed

In practice, this PR:

- stops dropping same-unit follow-up ingestion requests
- starts the newest same-unit request immediately
- marks older runs as superseded
- adds cooperative cancellation checkpoints to long-running work
- prevents superseded runs from performing final replace-style writes
- prevents superseded runs from sending terminal `finish()` updates
- aligns lock ordering across ingestion, metadata, and visibility update paths

## Tests

The tests focus on the risky parts of this change:

- job superseding behavior
- cancellation during long-running ingestion phases
- no commit after superseding at the final write boundary
- no terminal success update from a superseded run
- consistent lock behavior for ingestion, metadata, and visibility updates


iris/src/iris/pipeline/lecture_ingestion_pipeline.py
```diff
@@ -363,25 +372,61 @@ def restore_display_page_numbers_from_existing_chunks(self) -> None:
...
-                    raise
+            self._load_existing_slide_visibility()
+            for chunk, _ in prepared_chunks:
+                page_number = int(chunk[LectureUnitPageChunkSchema.PAGE_NUMBER.value])
+                chunk[LectureUnitPageChunkSchema.HIDDEN_UNTIL.value] = (
+                    self._hidden_until_by_page.get(page_number)
+                )
+            # Final guard after waiting for the write lock. Once this passes,
+            # keep delete+insert contiguous to avoid partial replacements.
+            raise_if_cancelled(
```
Claudia-Anthropica: @vivienxr [high] This guard is not synchronized with supersession: `add_job()` sets the event under `IngestionJobHandler._lock`, while this check and the subsequent delete/insert run under `batch_update_lock`. If a new request sets the event just after this check returns, the superseded job still commits stale chunks; the same TOCTOU window exists around the other writes and `callback.finish()`.

In `iris/src/iris/pipeline/lecture_ingestion_pipeline.py`, the cancellation check and replacement are not atomic with `IngestionJobHandler.add_job()` setting the cancellation event, so a superseded run can still commit after a newer request arrives. Introduce a shared per-unit ownership/commit lock used by both supersession and every final write or terminal-callback boundary, and hold it across the ownership check and mutation.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The completed-job cleanup and addressed batch, delete, and Whisper fixes are present at this head. However, the final cancellation checks remain TOCTOU checks against a separately locked supersession path, leaving the core stale-write guarantee violated; see the inline finding.

Claudia-Anthropica (changes requested):
@vivienxr The shared per-unit owner guard closes the previously reported TOCTOU window for page, transcription, and lecture-unit replacements and the terminal callback. The prior high-severity finding is only partially resolved: `LectureUnitSegmentSummaryPipeline._upsert_lecture_object()` still checks cancellation without holding that guard across `data.insert` or `data.update`, allowing a superseded run to persist a stale segment summary. Apply the same per-unit owner guard to those mutations; no duplicate inline is added because the existing Claudia finding already covers every final write.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
```diff
@@ -363,25 +373,64 @@ def restore_display_page_numbers_from_existing_chunks(self) -> None:
...
+                raise
+        failed_objects = getattr(self.collection.batch, "failed_objects", None)
+        if failed_objects:
+            raise RuntimeError(
+                f"Failed to insert {len(failed_objects)} lecture page chunks"
+            )
+
+    def _replace_prepared_chunks(self, prepared_chunks):
+        cancel_event = getattr(self, "cancel_event", None)
+        with ingestion_job_owner_guard(
```
Claudia-Anthropica: @vivienxr [high] Taking the per-unit owner guard before the global `batch_update_lock` prevents a new same-unit request from registering its cancellation while this old run waits behind an unrelated batch write. Once the global lock becomes available, the old run performs its delete/insert before `add_job()` can acquire the owner lock, set the event, and start the newer job; the same inverted nesting exists in the transcription and lecture-unit replacement paths.

In `iris/src/iris/pipeline/lecture_ingestion_pipeline.py`, the owner guard is acquired before the global batch lock, so a new same-unit request cannot supersede an old run while it waits for unrelated batch activity and the old run can still commit stale data after the request arrived. Acquire `batch_update_lock` first here and in the transcription and lecture-unit replacement paths, then acquire `ingestion_job_owner_guard` immediately around the final ownership check and mutation.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The segment-summary write now uses the shared ownership guard, resolving the remaining previously reported path. However, the replacement paths acquire that guard before the global batch lock, so unrelated batch activity can prevent a newly arrived request from superseding the old run until after it commits stale data.

Claudia-Anthropica (changes requested):
@vivienxr The final head removes the shared per-unit ownership guard and leaves each replacement or segment write as a standalone cancellation-event check before mutation. Because `add_job()` sets the old event under `_jobs_lock` while the page, transcription, lecture-unit, and segment writes plus `callback.finish()` do not hold that lock, a newer request can register immediately after a check and the superseded run can still persist stale data or report terminal success. The retained unresolved Claudia finding therefore still applies and the core race remains.

vivienxr:
> @vivienxr The final head removes the shared per-unit ownership guard and leaves each replacement or segment write as a standalone cancellation-event check before mutation. Because `add_job()` sets the old event under `_jobs_lock` while the page, transcription, lecture-unit, and segment writes plus `callback.finish()` do not hold that lock, a newer request can register immediately after a check and the superseded run can still persist stale data or report terminal success. The retained unresolved Claudia finding therefore still applies and the core race remains.

You're right that removing the shared owner guard also removes the strict guarantee that a superseded run can never write again after the newer request arrived.

That is an intentional tradeoff in this PR.

The semantics we want here are "eventual latest wins", not strict owner-based supersede semantics. A superseded run may still complete one final replacement if the new request arrives after its last cancellation check. However, that write can only happen before the newer run's own replacement, so it does not overwrite newer data; the newer run will replace it on its own commit. In other words, we accept a bounded stale intermediate write, but not a stale final state.

We chose that model deliberately because the stricter guarantee requires shared per-unit ownership/commit coordination across all mutation paths, which adds a fair amount of state/locking complexity for a guarantee we do not currently need.

So I agree the strict race still exists relative to the old owner-guard model. The point is that this PR intentionally does not aim to provide that stronger semantic anymore; it opts for the simpler "cancel + recheck before commit + latest run eventually converges the state" model.

Claudia-Anthropica (dismissed):
@vivienxr The latest head removes the shared per-unit ownership/commit synchronization, leaving supersession and final writes protected by separate locks. `add_job()` can set the old event immediately after a standalone cancellation or ownership check, allowing the superseded run to mutate page, transcription, lecture-unit, or segment data—or call `callback.finish()`—after the newer job registers. The retained high-severity Claudia finding therefore remains valid; make supersession and every final mutation or terminal callback share one atomic per-unit ownership boundary.

Claudia-Anthropica (dismissed):
@vivienxr [high] Supersession and final mutation/finish boundaries remain separate TOCTOU operations: `add_job()` can set the old event immediately after its final check, after which the superseded run can still write or call `callback.finish()`. Serialization only ensures the newer run writes later if it succeeds; if it fails before replacement, the older post-supersession write remains the final stored state. Use one atomic per-unit ownership boundary for registration and every final mutation or terminal callback, acquiring the global batch lock before the per-unit guard for Weaviate writes.

Claudia-Anthropica (changes requested):
@vivienxr [high] The core supersession race remains: `add_job()` updates ownership under `_jobs_lock`, while final Weaviate mutations and `callback.finish()` occur after separate cancellation or ownership checks. A newer job can register immediately after such a check; if it then fails before committing, the superseded job's later write or terminal success remains final. Make registration and every final mutation or terminal callback share one atomic per-unit ownership boundary, acquiring the global batch lock first for writes.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
```diff
@@ -363,25 +370,64 @@ def restore_display_page_numbers_from_existing_chunks(self) -> None:
...
-                        if i % 10 == 0:
-                            self.callback.update()
-                        embed_chunk = self.llm_embedding.embed(
-                            chunk[LectureUnitPageChunkSchema.PAGE_TEXT_CONTENT.value]
-                        )
-                        batch.add_object(properties=chunk, vector=embed_chunk)
-                except Exception as e:
-                    logger.error("Error updating lecture unit", exc_info=e)
-                    raise
+            with job_handler.current_job_guard(
```
Claudia-Anthropica, line 407: @vivienxr [high] This new batch→unit nesting conflicts with the metadata and visibility routes, which hold `lecture_update_lock` before their pipelines acquire `batch_update_lock`. If either route owns the unit lock when ingestion acquires the global batch lock, both threads wait forever and the retained global lock stalls all later writes. Make every combined path acquire the locks in one order—global batch lock, then per-unit lock—without reacquiring either non-reentrant lock.

In `iris/src/iris/pipeline/lecture_ingestion_pipeline.py`, ingestion acquires `batch_update_lock` before the per-unit guard while metadata and visibility updates acquire the same locks in reverse order, creating a permanent deadlock under concurrency. Rework metadata and visibility locking so every path acquires `batch_update_lock` before `lecture_update_lock`, acquiring each lock only once.

</details>

Claudia-Anthropica (changes requested):
@vivienxr The owner guard now makes registration and final ingestion mutations or callbacks atomic, resolving the prior stale-write TOCTOU finding. However, its lock order conflicts with the existing metadata and visibility paths and can permanently retain the global batch lock; see the inline finding.

Claudia-Anthropica (dismissed):
@vivienxr The final head consistently acquires the global batch lock before the per-unit lock in ingestion, metadata, and visibility mutation paths, while removing the webhook wrappers that previously reacquired the non-reentrant unit lock. This resolves the retained deadlock finding, and the complete CI snapshot contains no failures or pending required checks.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
Claudia-Anthropica: @vivienxr Thanks, this is fixed now.

Nayer-kotry (dismissed):
tested locally, works as intended!

vivienxr:
@Claudia-Anthropica please reapprove, I just had to fix merge conflicts with main

Claudia-Anthropica (approved):
@vivienxr The merged head preserves the global-batch → per-unit lock order and atomic ownership guards across final writes and the terminal callback. The integrated visibility retry handling retains that ordering, and the complete CI snapshot shows successful Iris tests and lint with no failed or pending required checks.

Nayer-kotry (approved):
manual retest, works as expected

Predixx (approved):
tested locally -> works as expected

PR #718, opened 2026-08-10
`Logos`: Support custom chat templates on the worker node

wasnertobias:
### Motivation

vLLM lanes could only use the chat template bundled with a model's tokenizer. Models that need different prompting — tool-calling variants, locally patched templates, a template a model ships broken — had no supported knob short of hand-rolling `extra_args`, and any path used there would point at ephemeral container storage.

### Description

Adds a first-class `chat_template` field to the per-lane vLLM config (usable via `engines.vllm.model_overrides.<model>` in `config.yml`, like every other vLLM override).

The value is a **file name**, resolved against a persistent, operator-managed directory:

```yaml
engines:
  vllm:
    model_overrides:
      Qwen/Qwen3-8B:
        chat_template: qwen3-tools.jinja
```

`/opt/logos-workernode/chat-templates` is bind-mounted read-only into the container **at the same absolute path** it has on the host — the compose directory on a worker node already *is* `/opt/logos-workernode` — so `config.yml`, the worker log line and the vLLM command line all name one identical path. Templates survive image updates and container restarts, and adding one requires no redeploy: the deploy workflow ships only `docker-compose.yml` and `.env`, so a new template just needs to land on the host plus a lane restart. `LOGOS_CHAT_TEMPLATE_DIR` relocates the directory for local dev and tests.

Resolution is deliberately contained:

- `..` is rejected at config-validation time;
- absolute paths are accepted only when they point inside the directory;
- symlinks are followed and the target must still be contained, so a template cannot be served from ephemeral storage;
- a **missing file fails the lane spawn** rather than silently falling back to the model's built-in template — a silent fallback would change generation behaviour with no visible signal.

Calibration is untouched: it merges only the VRAM-relevant override keys, and the chat template has no bearing on memory profiling.

### Steps for testing

1. Place a Jinja template in `logos-workernode/chat-templates/` (host: `/opt/logos-workernode/chat-templates/`).
2. Reference it from `config.yml` under `engines.vllm.model_overrides.<model>.chat_template`.
3. Start the lane and confirm the log line `using custom chat template: /opt/logos-workernode/chat-templates/<file>` plus `--chat-template <abs-path>` in the spawned vLLM command.
4. Send a chat completion and verify the rendered prompt reflects the custom template.
5. Point `chat_template` at a non-existent file — the lane must fail to spawn with an explicit error naming the directory, not start with the model default.

### Test coverage

`tests/test_vllm_process.py`: name resolution, subdirectory resolution, contained absolute path, flag omitted when unset, missing file, absolute path outside the directory, symlink escaping the directory, `..` traversal rejection, whitespace stripping, and the default directory constant. `tests/test_lane_manager.py`: the `model_overrides` merge path.

Full worker-node suite: 406 passed, 2 skipped. CI invocation (orchestrator venv, CI's ignore set): 253 passed, 2 skipped. `pre-commit` clean.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring custom Jinja chat templates for vLLM models.
  * Supports template files in approved directories, including subdirectories and local-development overrides.
  * Added validation for unsafe paths and clear errors when templates are missing or inaccessible.
  * Added read-only chat-template directory mounting for development and production deployments.

* **Documentation**
  * Documented configuration, deployment, path restrictions, and template resolution behavior.


Claudia-Anthropica (approved):
@wasnertobias The custom-template flow is coherent across model overrides, path validation and containment, the default read-only Compose mount, and vLLM command construction, with focused coverage for missing files and escape attempts. The successful worker tests, lint, and image build provide adequate validation.

PR #720, opened 2026-08-10
`Logos`: Faster navigation and bug fixes

alex7sz:
## Summary by CodeRabbit

* **New Features**
  * Added a team monthly budget field in team settings.
  * Added test accounts for local authentication scenarios.

* **Bug Fixes**
  * Improved key-loading reliability and access checks.
  * Improved billing usage calculations and key-loading performance.
  * Fixed chart totals so they match their legends and active theme.

* **Style**
  * Simplified the interface with flat backgrounds, neutral badges, consistent theme colors, and fewer decorative effects.
  * Updated login, dashboard, billing, workspace, and statistics views for a cleaner presentation.


Claudia-Anthropica (dismissed):
@alex7sz The exact-head Logos server image build and finalize checks succeeded, alongside the Logos tests and lint. The CPU-only PyTorch index configuration is consistent with the service's transitive torch dependency and linux/amd64 deployment target, with no concrete defects found.

logos/logos-orchestrator/Dockerfile
```diff
@@ -29,8 +29,14 @@ RUN mkdir -p src/logos && echo '__version__ = "0.1.0"' > src/logos/__init__.py \
...
 # Install all dependencies with uv (replaces poetry install ~436 s → ~30-60 s)
+# sentence-transformers pulls in torch as a transitive dep; this container has
+# no GPU (see docker-compose.yaml — no device reservation for this service),
+# so pull the CPU-only build instead of PyPI's default CUDA wheels (which drag
+# in ~15 nvidia-cu12 packages and blow the install up from ~1 min to ~8 min).
 RUN --mount=type=cache,target=/root/.cache/uv \
-    uv pip install .
+    uv pip install . \
+      --extra-index-url https://download.pytorch.org/whl/cpu \
+      --index-strategy unsafe-best-match
```
wasnertobias: Still not sure about unsafe-best-match

What's you take on that @Claudia-Anthropica - I am afraid of supply chain attacks.

Claudia-Anthropica (changes requested):
@alex7sz [medium] The image still performs an unlocked `uv pip install` with `unsafe-best-match` across PyPI and the PyTorch index, so builds are neither reproducible nor safely scoped to the intended CPU source; use a committed lock/export with an explicit torch-only CPU index. [medium] `invalidateCache()` leaves `keysInFlight` reusable, allowing a membership-change refresh to accept and recache a pre-mutation response. The successful CI snapshot does not exercise these resolution and runtime-race guarantees, so the existing unresolved threads need to be addressed.

wasnertobias:
> @alex7sz [medium] The image still performs an unlocked `uv pip install` with `unsafe-best-match` across PyPI and the PyTorch index, so builds are neither reproducible nor safely scoped to the intended CPU source; use a committed lock/export with an explicit torch-only CPU index. [medium] `invalidateCache()` leaves `keysInFlight` reusable, allowing a membership-change refresh to accept and recache a pre-mutation response. The successful CI snapshot does not exercise these resolution and runtime-race guarantees, so the existing unresolved threads need to be addressed.

@Claudia-Anthropica, please incorporate this change on your own!

Claudia-Anthropica (changes requested):
@alex7sz The merge-only head update leaves both previously requested medium defects unchanged: dependency installation still resolves unlocked candidates across PyPI and the PyTorch index, while cache invalidation can still reuse and recache a pre-mutation key request. The failed Worker Node image build is outside this PR’s authoritative diff because its Dockerfile matches the merge base, so it is not the basis for this verdict. Please use committed lock data with an explicit CPU-only torch source and scope in-flight key requests to their cache generation.

logos/logos-orchestrator/Dockerfile
Claudia-Anthropica: @wasnertobias Your concern is warranted: `unsafe-best-match` combines candidates across both indexes and can select a PyPI candidate. Torch should be pinned to an explicit CPU-only uv source and installed from committed lock data, while unrelated packages remain on PyPI.

PR #721, opened 2026-08-10
`Logos`: Add first draft of models' capabilities overview

Malte000:
First draft:

* Capabilities for vision, function calls, and reasoning are loaded from a bundled JSON snapshot, written to the database and displayed in the model overview
* restricts the full model catalogue visibilty to the logos admin. App admins only see models available to their team.

## Summary by CodeRabbit

- **New Features**
  - Added model capability indicators for function calling, vision, and reasoning.
  - Model descriptions and tags are now displayed in the model list.
  - All roles can view models and capabilities, while administrative actions remain restricted.
  - Capability information updates automatically when models are added or changed.
- **Bug Fixes**
  - Improved capability matching across providers and naming variations.
- **Data Updates**
  - Added persistent model capability data with automatic cleanup when models are removed.


wasnertobias:
Re: the `getModels()` change in `ModelService` (`isAdmin` -> `isLogosAdmin`):

This behavior change is confirmed as intended — app admins should only see the models that are relevant to their teams, not the full model catalog. Good catch.

Three small follow-ups to make it complete:
1. Mention this in the PR description — it changes the data scope of `app_admin` for `get_models` and is not obvious from the title/body.
2. Remove the now-unused `isAdmin()` helper (it is dead code now).
3. Add a test pinning the new scope in `ModelControllerTest` (e.g. an `app_admin` request to `get_models` returns only team-assigned models, not all seeded ones).

wasnertobias:
Feedback on the remaining items.

**Please fix before merge:**

1. **License attribution — please address right away.** The MIT text in `LITELLM_LICENSE` is correct and ships with the jar, which is good. What is missing is traceability: add a `NOTICE` file next to the JSON (e.g. `litellm-model-catalog.NOTICE`) stating that **both** `model_prices_and_context_window.json` **and** `model_prices_and_context_window.schema.json` are derived from [BerriAI/litellm](https://github.com/BerriAI/litellm) (repo URL, upstream file paths, "MIT License, Copyright (c) 2023 Berri AI — full text in LITELLM_LICENSE"), plus the provenance of the vendored snapshot (upstream commit or date). Add a one-liner in `logos/README.md` pointing to it.

2. **N+1 requests in the UI — this does not work.** `fetchModels()` fires one `get_model_capabilities` POST per model (100 models = 101 requests per page load). Fix with a batch endpoint (accept a list of model ids) or by returning the capabilities inside the `get_models` response.

3. **Never return JPA entities — DTOs only.** `get_model_capabilities` returns the `ModelCapabilities` entity directly. Map it to a DTO (consistent with the pattern used elsewhere in `ModelService`) so the API shape is decoupled from the entity. This is a general rule for this codebase.

4. **Liquibase changeSet id style.** Taking the next file numbers is fine. But the new changeSet ids mix the formats (`008-add-model-capabilities` vs `009_add_fk_model_capabilities_cascade`) — please follow the repo convention (hyphens).

**Follow-ups — please create issues for these yourself:**

5. **Tests for the new feature.** Nothing is covered yet: the `get_model_capabilities` endpoint (200 + 404), the name-matching logic (pure functions, easy to unit test), and the FK cascade on model delete. Follow the `ModelControllerTest` pattern — the seeded `gpt-4` model exists in the catalog and works as a 200 case.

6. **Matching coverage.** Exact-name matching after light normalization only resolves ~84% of common model names against the vendored catalog; e.g. `claude-3-5-haiku-20241022`, `gemini-1.5-pro`, and Anthropic models that only exist as Bedrock-style keys (`anthropic.claude-…-v1:0`) silently show "-". At minimum log a `warn` on no-match; ideally extend normalization (dot/dash, Bedrock suffixes) and pin the behavior with unit tests.
7. **Stale capabilities after rename.** Flags are only overwritten on a catalog hit — renaming a model to an unknown name keeps the old flags displayed. Reset to `false` (or delete the row) on no-match.
8. **Unused schema file.** `model_prices_and_context_window.schema.json` is referenced nowhere — either remove it or actually use it (e.g. strict validation of the vendored snapshot at startup/build).

**Minor:** trailing whitespace in `models.scss:69` and `models.ts:117/121`; the `sample_spec` entry in the JSON is treated as a model entry (skip it explicitly); in `addModel` the 404 for `IllegalArgumentException` is the wrong status and the `(Map<String, Object>)` cast is redundant; the PR description is stale (the CodeRabbit summary mentions the rolled-back price fix, and nothing is "scraped" — a bundled snapshot is read), and the README EACCES note is unrelated to this change.

wasnertobias (changes requested):
-

PR #729, opened 2026-08-14
`Logos`: Reconcile model profiles across worker nodes

wasnertobias:
## Problem

Worker nodes calibrate the models they host and report a profile — base residency, KV budget, sleeping residual. Those figures are the **sole input to every placement decision**, and the orchestrator has no independent way to measure them. That is fine as long as the nodes agree with each other.

They do not. In our production fleet the same model, calibrated on different nodes, is recorded with base residencies differing by up to **442 %**:

| Model | Two nodes | Third node | Spread |
|---|---|---|---|
| `google/gemma-4-E2B-it` | 17 539 MB | 95 071 MB | 442 % |
| `google/gemma-4-E4B` | 23 531 MB | 95 713 MB | 307 % |

The difference is not noise, and the arithmetic pins it down exactly:

```
95071 − 77824 (that node's KV budget) = 17247  ≈  17539
95713 − 72704                         = 23009  ≈  23531
```

One worker build includes the KV cache in the residency figure and the other does not — which is the *documented* dual semantics of `base_residency_mb`, where the meaning depends on `residency_source`:

```
"calibrated" — KV is INCLUDED; callers must NOT add kv_cache_memory_bytes on top.
"measured"   — weights only;   callers DO add KV separately.
```

Because worker builds are updated per node rather than fleet-wide, two nodes can report the same field under the same `calibrated` label meaning two different things. A 2 B-parameter model cannot occupy 95 GB, so at least one figure is definitionally wrong — yet both are trusted equally.

**No single node can detect this.** The disagreement exists only across nodes.

## What this adds

- **`capacity/profile_reconciliation.py`** compares every node's declaration for the same model and reports three conditions:
  - `kv_inclusion_skew` — the arithmetic signature above. This *names* the defect rather than merely flagging a disagreement.
  - `footprint_disagreement` — a large spread that the KV signature does not explain.
  - `uncalibrated_majority` — no node hosting the model has ever measured it (46 of our 89 profiles have `measurement_count = 0`).
- **`DBManager.fetch_model_profiles_for_reconciliation`** reads the rows.
- **`main.py`** runs the check after a profile upsert, throttled to once per ten minutes, logging each distinct finding once so a persistent disagreement does not flood the log.

## What this does *not* do

It changes **no placement decision**. It turns a silent inconsistency into an operational signal; whether to act on it automatically is a separate question worth discussing.

## Testing

Tests use the production figures above as fixtures, and — just as importantly — assert that legitimate variation stays quiet: `gpt-oss-120b` genuinely differs 8 % across nodes and must not warn.

- `tests/unit/capacity/test_profile_reconciliation.py` — 13 tests
- Full orchestrator suite: 567 passed
- Worker-node suite (CI invocation): 259 passed
- `pre-commit` clean

## Context

This came out of an analysis of 13 months of production telemetry (268 401 requests) for an ICSA 2027 submission on capacity-model divergence in federated LLM serving. A related observation, not addressed here: the platform records request outcomes but **no placement outcomes** — no load/wake/sleep/eviction event is persisted and `load_duration_ms` is NULL in all 268 401 rows. That is the channel in which this class of defect is visible, so it may be worth a follow-up.

https://claude.ai/code/session_017niB4CgAA3ALoabQvFMLUk

## Summary by CodeRabbit

* **New Features**
  * Added cross-node model profile reconciliation to identify measurement gaps, KV-cache discrepancies, and unexplained residency differences.
  * Findings include severity, affected providers, details, and supporting evidence.
  * Reconciliation runs periodically after profile updates, suppressing duplicate alerts.

* **Bug Fixes**
  * Invalid or incomplete profile records are safely skipped during processing.
  * Reconciliation failures no longer interrupt snapshot processing.

* **Tests**
  * Added coverage for thresholds, severity ordering, valid and invalid profile data, and single-node scenarios.

### DIFF
--- a/logos/logos-orchestrator/src/logos/capacity/profile_reconciliation.py
+++ b/logos/logos-orchestrator/src/logos/capacity/profile_reconciliation.py
@@ -39,9 +39,10 @@ Detected conditions:
 
 from __future__ import annotations
 
+import datetime
 import logging
 from dataclasses import dataclass, field
-from typing import Any, Dict, Iterable, List, Optional
+from typing import Any, Dict, Iterable, List, Optional, Set
 
 logger = logging.getLogger(__name__)
 
@@ -57,6 +58,15 @@ KV_INCLUSION_TOLERANCE = 0.10
 # Below this many megabytes the figures are too small for ratios to be meaningful.
 MIN_MEANINGFUL_MB = 256.0
 
+# A profile row survives its node.  Nodes leave a federation -- withdrawn after a
+# bad week, decommissioned, moved to another orchestrator -- and their rows stay
+# behind, so a naive comparison keeps reporting a conflict between one live node
+# and several that stopped serving months ago.  In the fleet this module was
+# written for, only 23 of 89 profiles belong to a node that served traffic
+# today.  Rows not refreshed within this window are treated as historical and
+# excluded from comparison.
+DEFAULT_MAX_PROFILE_AGE_DAYS = 30.0
+
 
 @dataclass(frozen=True)
 class NodeProfile:
@@ -70,6 +80,7 @@ class NodeProfile:
     tensor_parallel_size: Optional[int] = None
     residency_source: Optional[str] = None
     measurement_count: int = 0
+    updated_at: Optional[datetime.datetime] = None
 
     @property
     def is_measured(self) -> bool:
@@ -289,10 +300,50 @@ def reconcile_model(profiles: Iterable[NodeProfile]) -> List[Inconsistency]:
     return findings
 
 
-def reconcile(profiles: Iterable[NodeProfile]) -> List[Inconsistency]:
-    """Group declarations by model and reconcile each group."""
+def is_current(
+    profile: NodeProfile,
+    *,
+    now: Optional[datetime.datetime] = None,
+    max_age_days: float = DEFAULT_MAX_PROFILE_AGE_DAYS,
+    active_provider_ids: Optional[Set[int]] = None,
+) -> bool:
+    """Whether this row still describes a node the federation is using.
+
+    Two independent tests, because either alone leaves a gap.  A caller that
+    knows which providers are connected passes ``active_provider_ids`` and gets
+    an exact answer; otherwise the row's own freshness stands in, since a worker
+    that is still attached refreshes its profiles over the status channel.
+    """
+    if active_provider_ids is not None and profile.provider_id not in active_provider_ids:
+        return False
+    if profile.updated_at is None:
+        # No timestamp: fall back to trusting it, rather than silently
+        # discarding rows on databases that predate the column.
+        return True
+    reference = now or datetime.datetime.now(datetime.timezone.utc)
+    updated = profile.updated_at
+    if updated.tzinfo is None:
+        updated = updated.replace(tzinfo=datetime.timezone.utc)
+    return (reference - updated).total_seconds() <= max_age_days * 86_400.0
+
+
+def reconcile(
+    profiles: Iterable[NodeProfile],
+    *,
+    now: Optional[datetime.datetime] = None,
+    max_age_days: float = DEFAULT_MAX_PROFILE_AGE_DAYS,
+    active_provider_ids: Optional[Set[int]] = None,
+) -> List[Inconsistency]:
+    """Group current declarations by model and reconcile each group.
+
+    Historical rows are filtered out first: a disagreement between a live node
+    and one that left the federation months ago is not an operational signal,
+    and reporting it every ten minutes forever is worse than not reporting it.
+    """
     by_model: Dict[str, List[NodeProfile]] = {}
     for profile in profiles:
+        if not is_current(profile, now=now, max_age_days=max_age_days, active_provider_ids=active_provider_ids):
+            continue
         by_model.setdefault(profile.model_name, []).append(profile)
 
     findings: List[Inconsistency] = []
@@ -331,6 +382,7 @@ def rows_to_profiles(rows: Iterable[Any]) -> List[NodeProfile]:
                     tensor_parallel_size=_as_int(data.get("tensor_parallel_size")),
                     residency_source=data.get("residency_source"),
                     measurement_count=_as_int(data.get("measurement_count")) or 0,
+                    updated_at=data.get("updated_at"),
                 )
             )
         except (KeyError, TypeError, ValueError):
--- a/logos/logos-orchestrator/src/logos/capacity/profile_reconciliation.py
+++ b/logos/logos-orchestrator/src/logos/capacity/profile_reconciliation.py
@@ -306,16 +306,32 @@ def is_current(
     now: Optional[datetime.datetime] = None,
     max_age_days: float = DEFAULT_MAX_PROFILE_AGE_DAYS,
     active_provider_ids: Optional[Set[int]] = None,
+    hosted_models: Optional[Dict[int, Set[str]]] = None,
 ) -> bool:
-    """Whether this row still describes a node the federation is using.
+    """Whether this row still describes something the federation is serving.
 
-    Two independent tests, because either alone leaves a gap.  A caller that
-    knows which providers are connected passes ``active_provider_ids`` and gets
-    an exact answer; otherwise the row's own freshness stands in, since a worker
-    that is still attached refreshes its profiles over the status channel.
+    Three tests, because each closes a gap the others leave open.
+
+    ``active_provider_ids`` removes rows belonging to a node that left the
+    federation.  ``hosted_models`` removes rows for a model a *still-connected*
+    node no longer serves: an attached worker refreshes every profile it has
+    persisted, including ones for models since dropped from its configuration,
+    so those rows stay permanently fresh and would otherwise disagree with a
+    live node forever.  Freshness alone therefore cannot catch them.
+
+    The age test remains as the fallback for callers that know neither -- a
+    detached worker stops refreshing, so its rows do eventually age out.  Each
+    test is skipped when its argument is absent rather than guessed at, since a
+    caller without the information should not have rows discarded on a guess.
     """
     if active_provider_ids is not None and profile.provider_id not in active_provider_ids:
         return False
+    if hosted_models is not None:
+        serving = hosted_models.get(profile.provider_id)
+        # A provider absent from the mapping is unknown, not empty: only an
+        # explicit model list licenses dropping the row.
+        if serving is not None and profile.model_name not in serving:
+            return False
     if profile.updated_at is None:
         # No timestamp: fall back to trusting it, rather than silently
         # discarding rows on databases that predate the column.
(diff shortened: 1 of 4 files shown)


logos/logos-orchestrator/src/logos/capacity/profile_reconciliation.py
```diff
@@ -0,0 +1,280 @@
...
+
+    ``None`` when the identity does not hold, when the KV budget is unknown, or
+    when it is too small to explain the gap.
+    """
+    if high.kv_budget_mb is None or high.kv_budget_mb <= 0:
+        return None
+    if high.base_residency_mb is None or low.base_residency_mb is None:
+        return None
+
+    implied = high.base_residency_mb - high.kv_budget_mb
```
Claudia-Anthropica: @wasnertobias [medium] `kv_budget_mb` is a per-rank budget, whereas `base_residency_mb` is the total footprint across all tensor-parallel GPUs. Subtracting only one rank's budget causes TP>1 profiles to miss the KV-inclusion signature and be misclassified as generic disagreement.

In `logos/logos-orchestrator/src/logos/capacity/profile_reconciliation.py`, `_kv_inclusion_match` subtracts only one per-rank KV budget from a total multi-rank residency, so TP>1 KV-inclusion skew is missed. Subtract `kv_budget_mb * tensor_parallel_size` with a safe TP=1 fallback, use the same total in the evidence, and add a TP>1 test.

</details>

logos/logos-orchestrator/src/logos/capacity/profile_reconciliation.py
```diff
@@ -0,0 +1,280 @@
...
+    if len(profiles) < 2:
+        return []
+
+    model_name = profiles[0].model_name
+    findings: List[Inconsistency] = []
+
+    measured = [p for p in profiles if p.is_measured and _usable(p)]
+    if len(measured) < 2:
+        if profiles and not any(p.is_measured for p in profiles):
+            findings.append(
```
Claudia-Anthropica: @wasnertobias [medium] This emits `uncalibrated_majority` only when no node is measured. With one measured node out of three—or two out of five—the advertised majority condition exists but produces no finding; the five-node case bypasses this branch entirely because two profiles are available for comparison.

In `logos/logos-orchestrator/src/logos/capacity/profile_reconciliation.py`, `uncalibrated_majority` is limited to zero measured profiles instead of detecting when more than half of the hosting nodes are unmeasured. Evaluate and emit this condition independently of whether at least two measured profiles remain for footprint comparison, and add mixed measured/unmeasured tests.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias Changes requested for two gaps in the reconciliation logic. KV matching ignores tensor-parallel multiplicity, and `uncalibrated_majority` only detects zero measured nodes rather than an actual uncalibrated majority.

wasnertobias:
Thanks both — all three reconciliation gaps were real. Fixed in e6b8056.

**1. Tensor-parallel multiplicity (@Claudia-Anthropica, @coderabbitai)**
Confirmed. A model sharded over two accelerators is reported at twice the per-device footprint of the same model unsharded, so those two figures differ by arithmetic, not by disagreement — the old code would have flagged that as a conflict. All comparisons now normalise per accelerator (`_normalised_base_mb` / `_normalised_kv_mb`), which removes the false positive *and* lets the KV-inclusion identity be tested across nodes that shard differently. Two tests cover both directions.

**2. KV matching only on the extremes (@coderabbitai)**
Confirmed, and the 10/20/40 GB example is exactly right: the signature can sit in a middle pair while min-vs-max shows nothing, and the weaker `footprint_disagreement` gets reported instead. Now every ordered pair above `DISAGREEMENT_RATIO` is examined, keeping the match with the lowest relative error. Test added with those figures.

**3. `uncalibrated_majority` (@Claudia-Anthropica, @coderabbitai)**
Confirmed — the name promised a majority and the code checked for total absence. A model measured on one node out of three is still being placed on two nodes using figures nobody measured. Now checks the share of unmeasured profiles and reports *which* nodes they are. Two tests, one for the majority case and one asserting a measured majority stays quiet.

**Not addressed: `asyncio.create_task` at main.py:814 (@coderabbitai)**
Real issue — a task with no strong reference can be collected mid-flight — but that line is pre-existing code this PR doesn't touch. Fixing it inside a reconciliation PR would make both harder to review. Happy to open a separate one if you'd like.

**Follow-up already planned:** placement telemetry. The platform records 268,401 request outcomes and zero placement outcomes (`load_duration_ms` is NULL in every row), which is the one channel in which this class of defect is actually visible. That needs a schema migration, so it gets its own PR.

572 orchestrator tests pass, pre-commit clean.

Claudia-Anthropica (changes requested):
@wasnertobias [medium] The tensor-parallel follow-up does not resolve the existing blocker: `base_residency_mb` is a total across ranks, while `kv_budget_mb` is per-rank, but the new code divides both by TP. Keep residency in total units and compare `high.base_residency_mb - high.kv_budget_mb * high_tp` with the low total; the current normalization can miss KV-inclusion skew and manufacture disagreement between equivalent TP configurations. The pairwise and uncalibrated-majority fixes otherwise address their prior findings.

wasnertobias:
@Claudia-Anthropica you're right, and thank you — that was a real defect in my fix, not a nitpick. Corrected in the latest commit.

I verified the unit claim against the worker rather than taking it on trust:

- **`base_residency_mb` is a total across ranks.** It derives from `effective_vram_mb`, which `LaneManager._query_process_vram_map` builds by summing per-process VRAM across every rank of the lane (`result[owner_pid] = result.get(owner_pid, 0.0) + mem_mb`, resolving vLLM worker PIDs back to the lane root).
- **`kv_budget_mb` is per rank.** It comes from `kv_cache_memory_bytes`, which each rank allocates in full.

So dividing both by TP was right for one field and wrong for the other. The production data shows both failure modes my version would have caused:

**Manufactured disagreement.** `Qwen/Qwen3-Embedding-8B` is recorded at 21479 MB (tp=1) and 28292 MB (tp=2). As totals that's 1.32× — under `DISAGREEMENT_RATIO`, and correctly explained by NCCL buffers plus the duplicated embedding and LM-head layers. Divided per rank it becomes 1.52×, over the threshold, and my version would have flagged two perfectly consistent nodes.

**Missed skew.** With a 76000 MB total against a 30000 MB per-rank budget at tp=2, the identity closes only when the budget is multiplied up (76000 − 60000 = 16000). Dividing leaves 61000 and finds nothing.

Residency now stays in total units, the KV budget converts up by the rank count, and both cases are pinned by tests using the production figures above. The pairwise search and the majority threshold are unchanged.

572 orchestrator tests pass, pre-commit clean.

logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
```diff
@@ -1052,6 +1052,27 @@ def upsert_model_profiles(
...
+        fields needed to compare them; see
+        ``logos.capacity.profile_reconciliation``.
+        """
+        rows = self.session.execute(
+            text(
+                """
+                SELECT provider_id, model_name, base_residency_mb, kv_budget_mb,
+                       loaded_vram_mb, tensor_parallel_size, residency_source,
+                       measurement_count
+                FROM model_profiles
```
Claudia-Anthropica, line 1069: @wasnertobias [medium] This selects every persisted profile as if its provider currently hosts that model. `upsert_model_profiles` never removes models absent from later payloads, and disconnected providers retain their rows, so obsolete calibrations can produce permanent false skew/disagreement alerts.

In `logos/logos-orchestrator/src/logos/dbutils/dbmanager.py`, `fetch_model_profiles_for_reconciliation` includes stale model and offline-provider rows, causing false reconciliation findings. Build the reconciliation input from currently connected registry snapshots and include only each worker's current configured model list, falling back to its advertised capabilities.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The tensor-parallel correction now matches the worker's units and resolves the previous blocker; the pairwise and majority fixes also remain intact. One medium correctness issue remains: reconciliation treats historical and offline profile rows as current hosting nodes, allowing stale data to generate durable false alerts. The unresolved CodeRabbit task-retention thread concerns unchanged pre-existing code and is not included here.

wasnertobias:
@Claudia-Anthropica Confirmed and fixed in `dae9582`. The finding is larger than it reads: a profile row outlives the node that wrote it, and the current fleet shows how far.

Of 89 profile rows, only 23 belong to a node that served traffic today:

| node | profiles | last request |
|---|---|---|
| deioma | 23 | 2026-08-14 (live) |
| deipapa | 29 | 2026-06-04 (ten weeks silent) |
| deimama | 26 | 2026-06-04 (ten weeks silent) |
| hochbruegge | 11 | 2026-06-16 (withdrawn) |

So the module as merged would have warned every ten minutes, indefinitely, about a 442% disagreement between one live node and three that stopped serving months ago — technically true, operationally useless, and corrosive to whatever channel it lands in.

Reconciliation now considers only current rows, by either of two tests:

- a caller that knows which providers are connected passes `active_provider_ids` and gets an exact answer;
- otherwise the row's own `updated_at` stands in, since an attached worker refreshes its profiles over the status channel. The window defaults to 30 days and is configurable.

Rows with no timestamp are kept rather than silently dropped, so a database predating that column still gets reconciled.

Tests cover the fleet case directly: the same 442% figures produce **no** finding when two of the three rows are ten weeks old, and **do** produce one when all rows are fresh. Seven new tests in total.

579 orchestrator tests pass, pre-commit clean.

On the CodeRabbit task-retention thread: agreed that it concerns pre-existing code this PR does not touch, so I have left it out here.

Claudia-Anthropica (changes requested):
@wasnertobias [medium] The unresolved stale-profile thread remains valid after the latest commit. The age filter handles long-disconnected providers, but a live worker continually refreshes every persisted profile, including models removed from its current configuration, so those obsolete rows can still generate permanent false findings. The earlier tensor-parallel and majority issues are fixed, and the relevant captured test and lint checks pass.

logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
Claudia-Anthropica: @wasnertobias The age filter fixes departed providers once their rows expire, but it does not fix models removed from a live worker: runtime sends all persisted profiles and each upsert refreshes their `updated_at`. Please restrict reconciliation to each online provider's current `configured_models`, falling back to capabilities for older workers.

Claudia-Anthropica (approved):
@wasnertobias The latest head resolves the remaining stale-profile blocker by filtering persisted rows against each provider's current configured-model set, with the existing capabilities fallback for older workers. The tensor-parallel units, all-pairs KV matching, and uncalibrated-majority logic remain correct, and the captured test and lint checks pass.

logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
Claudia-Anthropica: @wasnertobias Thanks, this is fixed now: reconciliation filters persisted rows against each provider's current configured-model list, with the capabilities fallback retained for older workers.

PR #730, opened 2026-08-14
`Logos`: Record placement events

wasnertobias:
## Problem

The platform records **268,401 request outcomes and zero placement outcomes**. No load, wake, sleep or eviction is persisted anywhere, and `log_entry.load_duration_ms` is NULL in every row that has ever been written.

That gap matters more than it looks. A placement is the *only* direct test of a worker's capacity declaration: the planner commits memory on the strength of a figure the node reported, and the placement either honours that figure or it does not. Without the record:

- a wrong declaration stays invisible until it surfaces indirectly as a failed request;
- the cost of the failed placement itself — GPU-minutes on a lane that never served anything — is never recorded at all;
- basic operational questions are unanswerable: how often does a cold load fail, on which node, for which model, and how long did it take?

This is the follow-up promised in #729, kept separate because it carries a schema migration.

## What this adds

**`006_placement_events.xml`** — a deliberately narrow table, one row per placement attempt:

| what the orchestrator believed | what actually happened |
|---|---|
| `declared_free_vram_mb` | `outcome` (confirmed / unconfirmed / error) |
| `declared_footprint_mb` | `error_class` |
| `gpu_devices`, `tensor_parallel_size` | `duration_ms` |

No payload, no identifiers beyond provider and model.

**Instrumentation** wraps `_execute_action_with_confirmation` rather than editing its many return paths, so every outcome is captured — including the ones that raise. The original body moves to `_execute_action_uninstrumented` unchanged.

**`DBManager.record_placement_event`** writes the row on its own short-lived session, so a slow write cannot hold a connection the request path needs.

## Telemetry can never affect a placement

This is the property the design turns on, and it is enforced three ways:

1. The recorder is **optional** — a planner constructed without one behaves exactly as before, which also keeps it out of every existing test.
2. The DB write **swallows its own errors** after rolling back.
3. The wrapper **swallows recorder failures**.

Two tests assert exactly this: `test_a_broken_recorder_cannot_break_a_placement` and `test_wrapper_records_a_raising_placement_and_re_raises`.

## Testing

- `tests/unit/capacity/test_placement_telemetry.py` — 6 cases
- Full orchestrator suite: 560 passed
- `pre-commit` clean, changelog XML validates

## Note on retention

The table grows at roughly the rate of placements, which is orders of magnitude slower than requests — the production fleet performs on the order of tens per day against tens of thousands of requests. No retention policy is included; happy to add one if you would prefer a bound from the start.

https://claude.ai/code/session_017niB4CgAA3ALoabQvFMLUk

## Summary by CodeRabbit

* **New Features**
  * Added placement telemetry to capture resource details, GPU placement, timing, outcomes, and errors.
  * Added persistent storage for placement events, including provider and model information.
  * Telemetry recording runs asynchronously where possible to avoid delaying operations.

* **Bug Fixes**
  * Prevented telemetry failures from interrupting placement actions.
  * Ensured escalated stop actions generate a single, accurately labeled event.

### DIFF
--- a/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
+++ b/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
@@ -244,6 +244,8 @@ class CapacityPlanner:
         # behaves exactly as before, which keeps the recorder out of every
         # test that constructs a planner.
         self._placement_recorder = placement_recorder
+        # Set by the escalation path so the recorded action reflects what ran.
+        self._escalated_action: Optional[str] = None
         self._facade = logosnode_facade
         self._registry = logosnode_registry
         self._demand = demand_tracker
@@ -6424,6 +6426,7 @@ class CapacityPlanner:
         started = time.monotonic()
         confirmed = False
         error_class: Optional[str] = None
+        self._escalated_action = None
         try:
             confirmed = await self._execute_action_uninstrumented(action, timeout_seconds)
             return confirmed
@@ -6453,7 +6456,7 @@ class CapacityPlanner:
         if recorder is None:
             return
         try:
-            capacity = self._facade.get_capacity(action.provider_id)
+            capacity = self._facade.get_capacity_info(action.provider_id)
             declared_free = float(getattr(capacity, "available_vram_mb", 0.0) or 0.0)
         except Exception:
             declared_free = None
@@ -6467,7 +6470,7 @@ class CapacityPlanner:
             recorder(
                 provider_id=action.provider_id,
                 model_name=action.model_name,
-                action=action.action,
+                action=getattr(self, "_escalated_action", None) or action.action,
                 declared_free_vram_mb=declared_free,
                 declared_footprint_mb=footprint,
                 gpu_devices=(action.params or {}).get("gpu_devices"),
@@ -6621,7 +6624,13 @@ class CapacityPlanner:
                         # the case the safety valve exists to handle.
                         bypass_load_cooldown=True,
                     )
-                    return await self._execute_action_with_confirmation(stop_action, timeout_seconds)
+                    # Run the escalated stop *uninstrumented*: the outer call is
+                    # already inside the telemetry wrapper, so recursing through
+                    # it would write a second row carrying the stop's outcome
+                    # under the original action's label.  One placement attempt,
+                    # one row -- the escalation shows in the recorded action.
+                    self._escalated_action = "sleep_to_stop"
+                    return await self._execute_action_uninstrumented(stop_action, timeout_seconds)
 
                 # For request-time reclaim sleeps, mark the lane cold and drain
                 # active requests BEFORE sending the sleep command.  Without this,
--- a/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
+++ b/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
@@ -8,6 +8,7 @@ via LOGOS_CAPACITY_PLANNER_ENABLED=false.
 """
 
 import asyncio
+import contextvars
 import copy
 import logging
 import os
@@ -43,6 +44,16 @@ from .lane_comparator import best_lane
 from .vram_ledger import VRAMLedger
 
 logger = logging.getLogger(__name__)
+
+# Carries a placement's escalation label from the execution path back to the
+# telemetry wrapper.  A ContextVar rather than an attribute on the planner
+# because placements execute concurrently: asyncio copies the context per task,
+# so each placement reads only the label its own execution set, while an
+# instance attribute would let whichever coroutine escalated last relabel every
+# row currently in flight.
+_ESCALATED_ACTION: contextvars.ContextVar[Optional[str]] = contextvars.ContextVar(
+    "logos_placement_escalated_action", default=None
+)
 
 
 class CapacityPlanner:
@@ -244,8 +255,6 @@ class CapacityPlanner:
         # behaves exactly as before, which keeps the recorder out of every
         # test that constructs a planner.
         self._placement_recorder = placement_recorder
-        # Set by the escalation path so the recorded action reflects what ran.
-        self._escalated_action: Optional[str] = None
         self._facade = logosnode_facade
         self._registry = logosnode_registry
         self._demand = demand_tracker
@@ -6426,7 +6435,19 @@ class CapacityPlanner:
         started = time.monotonic()
         confirmed = False
         error_class: Optional[str] = None
-        self._escalated_action = None
+
+        # Read the declared headroom *before* executing, because that is the
+        # figure the placement decision rested on. Reading it afterwards -- as
+        # the finally block would -- records the memory left once the placement
+        # already consumed it, which measures the outcome rather than the claim
+        # and makes the row useless for the question it exists to answer.
+        declared_free = self._declared_free_vram_mb(action.provider_id)
+
+        # Per-call, not per-instance: several placements execute concurrently,
+        # and an attribute on self would let one coroutine's escalation label
+        # another's row. A ContextVar is copied per task, so each placement sees
+        # only its own.
+        token = _ESCALATED_ACTION.set(None)
         try:
             confirmed = await self._execute_action_uninstrumented(action, timeout_seconds)
             return confirmed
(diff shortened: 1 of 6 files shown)


logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -6404,6 +6409,78 @@ def _lane_gpu_devices_str(
...
+        error_class: Optional[str],
+        duration_ms: int,
+        started_at,
+    ) -> None:
+        """Persist one placement attempt.  Best effort; never raises."""
+        recorder = getattr(self, "_placement_recorder", None)
+        if recorder is None:
+            return
+        try:
+            capacity = self._facade.get_capacity(action.provider_id)
```
Claudia-Anthropica: @wasnertobias [medium] `LogosNodeSchedulingDataFacade` exposes `get_capacity_info`, not `get_capacity`, so this raises for every event and the broad handler silently stores `declared_free_vram_mb` as NULL. Moreover, sampling after execution would describe post-placement capacity rather than the declaration used to decide the placement.

In `logos/logos-orchestrator/src/logos/capacity/capacity_planner.py`, placement recording calls a nonexistent facade method after execution, causing every declared-free value to be NULL and preventing a correct pre-placement snapshot. Capture `get_capacity_info(action.provider_id).available_vram_mb` before awaiting `_execute_action_uninstrumented` and pass that captured value to the recorder.

</details>

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -6404,6 +6409,78 @@ def _lane_gpu_devices_str(
...
+        confirmed = False
+        error_class: Optional[str] = None
+        try:
+            confirmed = await self._execute_action_uninstrumented(action, timeout_seconds)
+            return confirmed
+        except Exception as exc:
+            error_class = type(exc).__name__
+            raise
+        finally:
+            self._record_placement_event(
```
Claudia-Anthropica, line 6458: @wasnertobias [medium] When a sleep is escalated to a stop for insufficient host RAM, `_execute_action_uninstrumented` recursively calls the instrumented wrapper for the stop. The inner call records the actual stop, then this outer `finally` also records the original sleep as confirmed even though no sleep command was executed, corrupting placement counts and outcomes.

In `logos/logos-orchestrator/src/logos/capacity/capacity_planner.py`, recursive sleep-to-stop escalation causes both a confirmed stop row and a false confirmed sleep row for one actual placement. Refactor escalation so instrumentation emits exactly one event for the action that was actually executed.

</details>

logos/logos-orchestrator/src/logos/main.py
```diff
@@ -811,6 +811,20 @@ def _capture_logosnode_provider_snapshot(
...
+def _record_placement_event(**fields) -> None:
+    """Sink for the capacity planner's placement telemetry.
+
+    Opens its own short-lived session so a slow write cannot hold a connection
+    the request path needs, and swallows everything: telemetry must never be
+    able to fail a placement.
+    """
+    try:
+        with DBManager() as db:
+            db.record_placement_event(**fields)
```
Claudia-Anthropica: @wasnertobias [medium] This synchronous SQLAlchemy write is invoked directly from the async placement wrapper's `finally` block. A slow database or exhausted connection pool therefore blocks the service event loop and delays placement completion; swallowing exceptions and opening a separate session do not isolate that latency.

In `logos/logos-orchestrator/src/logos/main.py`, placement telemetry performs blocking database I/O on the async placement path, so database latency can delay placements and all other event-loop work. Enqueue records non-blockingly to a bounded background writer and allow the placement coroutine to return without waiting for persistence.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The schema and basic wrapper coverage are present, and the snapshot shows the Logos tests and pre-commit check passing. However, the implementation silently nulls the core declared-free measurement, can emit duplicate and mislabelled rows during sleep-to-stop escalation, and performs synchronous database I/O on the async placement path. These medium-severity telemetry correctness and availability issues require changes.

wasnertobias:
Thanks — all three findings were real, and the first one would have made this PR pointless. Fixed in `84edb6f`.

**1. `get_capacity` does not exist** (@Claudia-Anthropica, @copilot-pull-request-reviewer)

Confirmed against `logos/sdi/logosnode_facade.py:151`: the method is `get_capacity_info`. Every production event would have raised `AttributeError`, the wrapper would have swallowed it, and `declared_free_vram_mb` would have been persisted as NULL — exactly the gap this telemetry exists to close.

Worth naming why my tests missed it: they set `_facade = None`, which skips the branch entirely. The new test asserts against the *real* facade class — that `get_capacity_info` exists, that `get_capacity` does not — and drives the code with a stub implementing only the real interface, so a future rename fails the test instead of silently nulling the column again.

**2. Duplicate and mislabelled rows on sleep-to-stop escalation**

Confirmed at `capacity_planner.py:6547`: the escalation recursed through `_execute_action_with_confirmation`, which *is* the telemetry wrapper. One placement attempt therefore wrote an inner row for the stop and an outer row carrying the stop's outcome under the original action's label. The escalation now calls `_execute_action_uninstrumented` and records the action as `sleep_to_stop`, so one attempt yields one row that says what actually ran. A second test pins that the label does not leak into the following attempt.

**3. Synchronous database I/O on the async path**

Confirmed. `_record_placement_event` is called from the event loop, so a slow or blocked write stalled every coroutine on it, including the request path. The write now goes to the default executor, with a direct-call fallback when no loop is running (tests, scripts). Failures are still swallowed — telemetry must never be able to fail a placement.

563 orchestrator tests pass, pre-commit clean.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -6404,6 +6411,79 @@ def _lane_gpu_devices_str(
...
+            profile = self._facade.get_model_profiles(action.provider_id).get(action.model_name)
+            footprint = self._estimate_model_loaded_vram(profile) if profile is not None else None
+            tp = int(profile.tensor_parallel_size) if profile is not None and profile.tensor_parallel_size else None
+        except Exception:
+            footprint, tp = None, None
+        try:
+            recorder(
+                provider_id=action.provider_id,
+                model_name=action.model_name,
+                action=getattr(self, "_escalated_action", None) or action.action,
```
Claudia-Anthropica: @wasnertobias [medium] `_escalated_action` is planner-wide mutable state, but unrelated lanes can execute this coroutine concurrently (request retries are explicitly spawned as tasks). One call can set or reset the value while another is awaiting worker confirmation, causing a normal load/wake to be recorded as `sleep_to_stop` or an escalated stop to lose its label.

In `logos/logos-orchestrator/src/logos/capacity/capacity_planner.py`, `_escalated_action` is shared across concurrent placement calls, so interleaving actions can receive each other’s labels. Keep the effective action in per-invocation state and pass it directly to `_record_placement_event`.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The recursion now emits one row and the database write is off the event loop, but the primary decision-time capacity defect remains: `declared_free_vram_mb` is still read only after placement completes instead of being captured before execution. The new instance-wide escalation label is also shared by concurrently executing placement coroutines and can therefore mislabel unrelated events. These telemetry-integrity issues keep the review at changes requested; the captured Logos test and pre-commit checks passed, while several non-required checks were still in progress.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
Claudia-Anthropica: @wasnertobias Thanks, the recursive stop now stays inside the uninstrumented path, so one sleep-to-stop attempt emits only one row. The shared escalation label still needs a separate concurrency fix.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -6404,6 +6420,112 @@ def _lane_gpu_devices_str(
...
+        try:
+            recorder(
+                provider_id=action.provider_id,
+                model_name=action.model_name,
+                action=escalated_action or action.action,
+                declared_free_vram_mb=declared_free_vram_mb,
+                declared_footprint_mb=footprint,
+                gpu_devices=(action.params or {}).get("gpu_devices"),
+                tensor_parallel_size=tp,
+                outcome="confirmed" if confirmed else ("error" if error_class else "unconfirmed"),
```
Claudia-Anthropica, line 6520: @wasnertobias [medium] The load, wake, sleep, and stop branches catch normal worker-command exceptions and return `False`, so `error_class` remains `None` here and those failed placements are persisted as `unconfirmed`. Consequently, only unexpected exceptions escaping the executor produce `error` rows, and the telemetry cannot report the common placement failure classes it was introduced to capture. Propagate an execution-local failure class from those catch blocks, or return a structured execution result, and reserve `unconfirmed` for confirmation timeouts.

In `logos/logos-orchestrator/src/logos/capacity/capacity_planner.py`, caught worker-command failures return `False`, causing the wrapper to record them as `unconfirmed` without an error class. Propagate their failure class through the execution result and persist them with outcome `error`, while keeping genuine confirmation timeouts `unconfirmed`.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The exact-head changes correctly capture decision-time capacity and isolate escalation labels per task. Changes are still required because ordinary worker-command failures are recorded as unconfirmed, while the existing GPU-assignment and unbounded executor-backlog findings remain valid. The captured Logos tests and pre-commit check passed; several non-required checks were still in progress.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
Claudia-Anthropica: @wasnertobias Thanks, the current head now keeps escalation state task-local with a `ContextVar`, and the concurrency regression test covers interleaved placements. This issue is fixed now.

PR #734, opened 2026-08-18
`Logos`: Exclude calibrating workers from lane placement

wasnertobias:
## Problem

A calibration session tears down every lane up front to free VRAM for its probes. That makes the worker look to the capacity planner like an idle node with a lot of headroom — exactly the node it wants to load a model onto. The lane it places then holds the VRAM the probes need.

Seen in production: a tp=2 kv-cache search was progressing normally when the planner placed a lane on the same worker mid-session.

```
08:21:34  OK kv_cache=18G          ← working
08:23:04  >> CMD add_lane planner-Qwen_Qwen3-Embedding-8B
08:23:05  Auto-placement: gpu_devices=0 (required≈21475MB)
08:23:45  FAIL kv_cache=18G        ← same size, same tp, now fails
08:25:33  failed at max tp=2 — falling back to configured tp=1
08:51:05  ERROR: No working KV cache size found between 1G and 38G on tp=1
```

The vLLM log gives the real reason:

```
ValueError: Free memory on device cuda:0 (25.49/47.37 GiB) on startup is less
than desired GPU memory utilization (0.92, 43.58 GiB).
```

25.49 GiB free instead of 47 — exactly the 21.5 GB the new lane took. The tp escalation read that as "max tp doesn't fit", fell back to tp=1, failed all 38 sizes there too, and gave up on a model that calibrates fine on identical hardware with a byte-identical `config.yml`.

## Fix

`ProviderSession.calibrating` tracks whether a session is running, and `CapacityPlanner._is_plannable()` skips those providers. The helper also absorbs the pre-existing "no first status since connect" filter, so both provider exclusions live in one place and apply consistently to the main planning loop and to the cross-provider best-first ranker.

The flag is driven by the worker's own `calibration_session_started` / `_finished` / `_cancelled` events, **not** by `CalibrationOrchestrator._active_provider_id`. That slot is only set by the nightly tick in `_start_session_on()`; the admin `calibrate_uncalibrated` endpoints (`main.py:1639`, `main.py:4061`) send `start_calibration_session` straight through the registry and never touch it — and that is the path that triggered the incident. Driving the flag from worker events covers every trigger and survives a reconnect, since the worker resets `_last_event_seq` to 0 and replays its log.

`calibration_orchestrator` now imports the terminal-event names from the registry so the two definitions cannot drift.

## Why this alone is enough

An announce can only become a lane through the planner. `get_worker_capabilities` has exactly two consumers, the capacity planner and the calibration orchestrator; request routing cannot create a lane, because `try_reserve_capacity` refuses when no lane is ready (`logosnode_provider.py:831`) and the request goes to the scheduler queue instead. Blocking the planner closes the path.

An earlier revision of this PR also deferred the worker's mid-session capability announce. That turned out to be redundant once the planner guard was in place, and it has been reverted — `logos_bridge.py` is unchanged.

## Tests

`test_calibration_lane_guard.py` covers the registry flag (start / finish / cancel, per-model events must not end a session, reconnect replay, unknown provider) and the `_is_plannable` matrix, including the admin-endpoint path end to end.

`test_best_first_scenarios.py`'s planner harness now sets `registry.is_calibrating.return_value = False` explicitly — a bare `MagicMock` returns a truthy mock, which would silently make every provider look mid-calibration.

Orchestrator unit suite: 565 passed, 1 xfailed. Worker node: 412 passed, 2 skipped. `pre-commit` clean on all touched files.

## Considered and rejected

Announcing an empty capability list for the duration of a session. It reads like a pure routing hint but is not: every capability change — including via the status path (`logosnode_registry.py:627`) — fires `_sync_logosnode_capabilities_to_db`, which deletes `model_provider` links the worker no longer advertises and recreates them with `api_key`/`endpoint` NULL. Every nightly session would churn those rows and silently reset per-link overrides. (Team permissions would survive — `team_model_permissions` hangs off `models.id`, not the link.)

https://claude.ai/code/session_01UuzntpniQJuEYdvm9KVo4f

## Summary by CodeRabbit

* **New Features**
  * Added tracking for provider calibration sessions, including start, completion, cancellation, and reconnect scenarios.
  * Providers undergoing calibration are temporarily excluded from capacity planning.
  * Worker nodes now report calibration status and restrict calibration-sensitive operations while calibration is active.

* **Bug Fixes**
  * Improved provider eligibility checks before scheduling and dispatch.
  * Standardized terminal event handling and replay processing.
  * Improved handling of refused commands and stale calibration state.

* **Tests**
  * Added coverage for calibration transitions, command restrictions, event replay, reconnects, and planning safeguards.


logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -1032,15 +1039,21 @@ async def _run_calibration_session(self, session: _CalibrationSession) -> None:
...
-                    # that they're calibrated; otherwise the server never
-                    # learns the worker can serve them.
-                    if model_name not in self._cfg.capabilities_models:
-                        self._cfg.capabilities_models = list(self._cfg.capabilities_models) + [model_name]
+                    # because they had no profile must be re-announced once
+                    # they're calibrated; otherwise the server never learns
+                    # the worker can serve them. Queue the announce instead of
+                    # applying it right away — see pending_capabilities.
+                    already_queued = (
+                        model_name in self._cfg.capabilities_models or model_name in session.pending_capabilities
```
Claudia-Anthropica: @wasnertobias [high] This changed expression is not in Black's required layout, leaving the exact-head pre-commit check red. Run the repository's formatter over both modified Python files and commit its output.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, the new calibration code is not Black-formatted, leaving the exact-head pre-commit check red. Run Black on both changed Python files and commit all formatter changes.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias [high] The captured validate-pr-title failure is caused by the current title, which lacks the required backticks around `Logos` and starts the description lowercase; rename it to `` `Logos`: Don't announce newly calibrated models while calibration is still running ``. [high] The pre-commit failure is also attributable to changed Python that Black 25.1 will rewrite, including `logos_bridge.py:1047`; commit the formatter output for both modified Python files. The calibration deferral itself correctly holds capabilities through normal and cancelled sessions and invalidates runtime-status deduplication when publishing them.

Claudia-Anthropica (dismissed):
@wasnertobias The session-local queue correctly defers new capabilities until finalization, publishes completed models on normal and cancelled paths, and invalidates runtime-status deduplication. The focused tests cover both paths, and the previously failing title and pre-commit checks now pass at this exact head. I found no unresolved correctness issue; the remaining captured build jobs were unfinished but were not configured as required checks.

logos/logos-workernode/logos_worker_node/logos_bridge.py
Claudia-Anthropica: @wasnertobias Thanks, this is fixed now: the expression is Black-compliant and the exact-head pre-commit check passes.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -573,9 +594,7 @@ def _provider_pressure(pid: int) -> int:
         # because it was iterated first.
         best_provider_for_model: Optional[dict[str, int]] = None
         if self._cross_provider_best_first:
-            ready_provider_ids = [
-                pid for pid in provider_ids if self._registry is None or self._registry.has_received_first_status(pid)
-            ]
+            ready_provider_ids = [pid for pid in provider_ids if self._is_plannable(pid)]
```
Claudia-Anthropica, line 595: @wasnertobias [high] This filtering is not applied to the speculative-replication pass: `_compute_replication_actions()` receives the original `provider_ids` and only checks `has_received_first_status()`. With `LOGOS_REPLICATE_ON_FREE_VRAM` enabled, it can therefore select the empty calibrating worker and emit the same VRAM-consuming `load` this guard is intended to prevent. Pass only plannable providers to that pass or make it call `_is_plannable()` itself.

In `logos/logos-orchestrator/src/logos/capacity/capacity_planner.py`, speculative replication bypasses the calibration exclusion and can load a lane on a calibrating provider. Filter replication candidates through `_is_plannable()` before emitting an action.

</details>

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -596,9 +615,9 @@ def _provider_pressure(pid: int) -> int:
             # Skip providers that haven't sent their first status yet —
             # we don't know what lanes are already loaded and acting on
             # stale/empty state can destroy existing lanes.
-            if self._registry and not self._registry.has_received_first_status(provider_id):
+            if not self._is_plannable(provider_id):
```
Claudia-Anthropica, line 616: @wasnertobias [high] This check can become stale before execution. The cycle builds every action first and then executes them sequentially without another `_is_plannable()` check, so if a calibration starts after this line, an already-planned load or wake can still run after the registry has marked the provider as calibrating. Revalidate each action immediately before acquiring its lane lock and discard it when its provider is no longer plannable.

In `logos/logos-orchestrator/src/logos/capacity/capacity_planner.py`, actions planned before calibration begins remain executable after the provider becomes calibrating. Re-check `_is_plannable(action.provider_id)` immediately before executing every action and skip actions whose provider is no longer plannable.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias [high] The captured `validate-pr-title` failure is attributable to the current title: it omits the required backticks around `Logos` and starts the description lowercase; rename it to `` `Logos`: Don't let the planner touch a worker while it is calibrating ``. The calibration-state tracking and deferred capability publication are otherwise coherent, but the planner still has the two blocking bypasses called out inline. The worker-node image build was still in progress in the snapshot, but it is not configured as a required check.

Claudia-Anthropica (changes requested):
@wasnertobias The exact head still passes unfiltered provider IDs into speculative replication and executes validated actions without rechecking `_is_plannable()`, so both existing high-severity threads remain unresolved. A reconnected session also starts with `calibrating=False` while the worker sends status before replaying calibration events, confirming the retained reconnect thread. The captured `validate-pr-title` failure remains attributable to the malformed current title; the unfinished worker-image build is not configured as required.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -653,6 +670,20 @@ def _provider_pressure(pid: int) -> int:
                 # Acquire per-lane lock so concurrent operations on the same
                 # lane are serialized, but unrelated lanes remain unblocked.
                 async with self._lane_lock(action.provider_id, action.lane_id):
+                    # Re-check under the lock: actions execute sequentially and
+                    # a cold load takes ~90s, so a calibration session can start
```
Claudia-Anthropica, line 674: @wasnertobias [medium] This recheck is still stale during calibration startup. `_handle_start_calibration_session()` marks the worker session active and records `calibration_session_started` locally, but `_event_loop()` waits one second before forwarding that event, and the registry does not mark the provider calibrating when it dispatches the start command. A planner action reaching this check during that window therefore passes and can send `add_lane`, `apply_lanes`, or `wake_lane` after calibration has begun; add a worker-side guard that rejects VRAM-allocating lane commands while `_active_calibration_session` is non-null.

In `logos/logos-orchestrator/src/logos/capacity/capacity_planner.py`, the calibration eligibility recheck can pass during the one-second delay before the worker's start event reaches the registry, allowing a VRAM-consuming lane command after calibration begins. Make the worker reject `add_lane`, `apply_lanes`, `wake_lane`, and VRAM-increasing reconfiguration commands while `_active_calibration_session` is non-null, and add a focused test for that interval.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The exact head now filters speculative replication, rechecks actions under the lane lock, and seeds reconnect state via hello, so those prior paths are fixed. One start-of-session race remains: the registry stays plannable until the worker's one-second event loop forwards `calibration_session_started`, allowing a lane command during that interval. Several captured non-required checks remain in progress, but no required check is pending.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
Claudia-Anthropica: @wasnertobias Thanks, speculative replication now filters every candidate through `_is_plannable()`, so this path is fixed.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
Claudia-Anthropica: @wasnertobias Thanks, each validated action is now rechecked under its lane lock before execution, so this stale-action path is fixed.

Claudia-Anthropica (changes requested):
@wasnertobias [medium] The dispatch-time mark closes the original startup delay, but generation-blind event replay can still clear `calibrating` with an older terminal event after hello reports the current session active; until replay reaches the current `started` event, the worker accepts VRAM-growing commands. This is the existing unresolved worker-guard issue, so I staged the current-head evidence there rather than duplicating it.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
Claudia-Anthropica: @wasnertobias The dispatch-time mark closes the original one-second start-event delay, but this thread is not fully resolved. On reconnect, hello can set `calibrating=True` and replay can then deliver an older session terminal event before the current session's started event; `_track_calibration_state()` clears the flag, while `_execute_command()` still accepts `add_lane`, `apply_lanes`, `wake_lane`, and `reconfigure_lane` in that gap. Please add the worker-side active-session guard requested here, or make lifecycle events generation-aware, and cover this replay ordering.

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -318,9 +329,11 @@ async def _event_loop(self, ws) -> None:
                         "type": "event",
                         "worker_id": self.worker_id,
                         "event": event.model_dump(mode="json"),
+                        "replay": replay,
```
Claudia-Anthropica: @wasnertobias [medium] `replay` covers the entire first one-second drain, not just events that existed at connection time. If a live calibration finishes or is cancelled after hello but before that drain, its terminal event is tagged `replay=True`; the registry ignores it and retains hello's `calibrating=True`, excluding the provider until another reconnect. Snapshot the event-log length at connection, mark only entries below that boundary as replayed, and test a terminal event appended before the first drain.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, the first drain labels events created after hello as replayed, so a live terminal event can be ignored and leave the provider unplannable. Capture a connection-time replay boundary, mark only older entries as replayed, and add a test for a terminal event appended before the first drain.

</details>

logos/logos-orchestrator/src/logos/logosnode_registry.py
```diff
@@ -711,14 +739,16 @@ def peek_recent_samples(
...
-    async def append_event(self, provider_id: int, event: dict[str, Any]) -> None:
+    async def append_event(self, provider_id: int, event: dict[str, Any], replay: bool = False) -> None:
         session = await self._get_session(provider_id)
         if session is None:
             return
         session.last_heartbeat = _utc_now()
         if isinstance(event, dict):
             session.latest_events.append(event)
             session.latest_events = session.latest_events[-500:]
+            if not replay:
```
Claudia-Anthropica: @wasnertobias [medium] Replay filtering is applied only to `_track_calibration_state`; every replayed event is still sent to `_event_subscribers` below. `CalibrationOrchestrator._on_provider_event()` clears `_active_provider_id` on any terminal event from that provider, so an old terminal event in the reconnect backlog can free the slot while the worker is still calibrating and let the next tick start another provider. Do not deliver replayed lifecycle events to live-state subscribers, or pass replay metadata so the orchestrator can ignore them, and cover this ordering.

In `logos/logos-orchestrator/src/logos/logosnode_registry.py`, replayed terminal events still reach the calibration-orchestrator subscriber and can clear its active-provider slot during a current session. Prevent replayed lifecycle events from mutating live subscriber state and add a test with an active slot plus a stale terminal replay.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The worker-side guard now closes the previously reported VRAM-allocation race, so that thread can be resolved. Two reconnect-event issues remain: the whole first drain is labeled replay even when it contains events created after hello, while correctly replayed terminal events still reach the calibration-orchestrator subscriber. Those paths can respectively strand a provider as unplannable or free the cluster-wide calibration slot while a session is still running; the captured required-check set is empty, so the in-progress non-required jobs do not defer this review.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
Claudia-Anthropica: @wasnertobias Thanks, the worker now rejects all four VRAM-growing actions while the calibration task is live, so the placement race in this thread is fixed.

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -307,17 +320,32 @@ async def _status_refresh_loop(self, ws) -> None:
...
+        # produces a live terminal event which must not be dismissed as
+        # backlog — doing so would leave the provider excluded from lane
+        # placement with no further event coming to release it.
         while not self._stopping.is_set():
             await asyncio.sleep(1)
             events = self._app.state.lane_manager.event_log
-            for event in events[self._last_event_seq :]:
+            # The log is capped, so it can shrink from the front; clamping keeps
+            # the boundary from reaching past it and mislabelling live events.
+            boundary = min(replay_boundary, len(events))
```
Claudia-Anthropica: @wasnertobias [medium] `replay_boundary` is a list length, but `event_log` drops entries from the front at 500 items. If the log is full at connect and a live `calibration_session_finished` or `calibration_session_cancelled` event arrives before this drain, trimming keeps the length at 500, so the new event at index 499 is sent with `replay=True`; `append_event()` ignores it and the provider remains unplannable until reconnect. Snapshot and advance the replay cursor using stable event IDs rather than list positions, and add a regression test starting with a full log.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, the count-based replay boundary misclassifies newly appended live events when the capped event log evicts entries from the front, which can leave a provider unplannable until reconnect. Track replay and delivery using stable event IDs and add a regression test that starts with a full event log.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The latest head correctly keeps replayed lifecycle events out of live subscribers and closes the previously reported worker-side placement race. However, the new replay boundary is indexed into a capped, front-trimmed log, so a live terminal event can be treated as replay and leave the provider excluded until reconnect. The captured required-check set is empty; the in-progress worker-image build is non-required.

logos/logos-orchestrator/src/logos/logosnode_registry.py
```diff
@@ -854,11 +961,22 @@ async def send_command(
...
 
         if not bool(result.get("success", False)):
+            _undo_optimistic_calibration_mark()
             raise LogosNodeCommandError(str(result.get("error", "unknown worker command error")))
-        return result.get("result", {})
+        payload = result.get("result", {})
+        # A refused start (session already running, node unhealthy) comes back
+        # as a successful command carrying ok=False, not as an error — so the
+        # check above does not see it.
+        if calibration_start and isinstance(payload, dict) and payload.get("ok") is False:
```
Claudia-Anthropica: @wasnertobias [medium] The new worker guard returns `ok=False` for guarded lane actions, but this method only reacts to that payload for `start_calibration_session` and otherwise exposes it as a successful RPC. The planner consequently records refused `add_lane`/`apply_lanes` operations in desired state and waits for confirmations that cannot arrive; raise `LogosNodeCommandError` for every dictionary payload whose `ok` value is false after restoring the optimistic calibration mark.

In `logos/logos-orchestrator/src/logos/logosnode_registry.py`, guarded lane commands returning `ok=False` are exposed as successful RPC results, so callers commit desired state and wait for confirmations that cannot arrive. Raise `LogosNodeCommandError` for any dictionary payload with `ok is False` after undoing the optimistic calibration-start mark.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias [medium] The replay-ID and subscriber fixes are correct, but the new worker refusal is still treated as a successful RPC, so the planner can commit desired state for a lane the worker rejected and then wait through the confirmation timeout; see inline. The previously captured legacy-worker reconnect gap also remains: a hello without `calibrating` leaves the fresh session false until replay. The failed non-required Logos Webservice build is unrelated because it fails Maven dependency prefetch in the untouched webservice subtree.

PR #743, opened 2026-08-19
`Logos`: Add a Claude Code setup page

wasnertobias:
Adds a **Claude Code** page next to the existing OpenCode one — same main menu, same `personal` group, same four-step shape (install → key → model → connect), with a download and a terminal merge command that keeps an existing settings file.

Claude Code speaks the Anthropic Messages API, which Logos serves natively at `POST /v1/messages`, so no proxy is involved.

### Why a page rather than a wiki snippet

The configuration is easy to get subtly wrong, and every one of these fails in a way that doesn't obviously point at the cause. The page knows the answers because it already has the key and the model list:

| | What the page does | What goes wrong otherwise |
|---|---|---|
| `ANTHROPIC_BASE_URL` | no `/v1` suffix | Claude Code appends the path itself → 404s |
| credential | `ANTHROPIC_AUTH_TOKEN`, and `ANTHROPIC_API_KEY` blanked | `ANTHROPIC_API_KEY` sends `x-api-key`, the orchestrator reads `Authorization: Bearer` → 401 |
| model slots | all of them point at the picked model | switching alias with `/model` silently leaves Logos |
| `CLAUDE_CODE_MAX_CONTEXT_TOKENS` | the window the workers actually report | Claude Code assumes 200k for unknown models and lets requests overflow upstream before it compacts |
| output cap | reserved *out of* that window | input and output share one budget upstream (vLLM counts both against `--max-model-len`) |
| WebSearch | denied | Claude Code sends `{"type":"web_search_20250305"}` with no `input_schema`, the worker nodes reject it with 400, and Claude Code retries in a loop |

The context window is the one that has cost the most time so far — hence taking it from `context_window` on the model rather than letting the user guess, with a conservative fallback (32768 local / 128000 cloud, matching the OpenCode page) when no worker reports one.

The page also warns when the picked model id would make `CLAUDE_CODE_MAX_CONTEXT_TOKENS` inert: it applies directly only to ids Claude Code cannot resolve to a Claude model, so an id starting with `claude-` or containing `[1m]` needs `DISABLE_COMPACT` / `CLAUDE_CODE_DISABLE_1M_CONTEXT` too. No Logos model is named that way today; this guards against one being registered later.

Settings go to the user-level `~/.claude/settings.json` on purpose — an `env` block in a project's `.claude/settings.json` is only read after first-run setup, so Claude Code would still ask the user to log in.

### Verification

`npx tsc --noEmit` and `npm run build` both clean; the new lazy chunk builds and adds no budget warning. Not yet rendered against a live orchestrator — the dev proxy points at a local backend, so the key/model dropdowns need one to populate.

https://claude.ai/code/session_01Ft5fzaK1DdrwURrJwBggeV

### DIFF
--- a/logos/logos-ui/src/app/app.routes.ts
+++ b/logos/logos-ui/src/app/app.routes.ts
@@ -27,6 +27,7 @@ export const routes: Routes = [
       { path: 'teams/:id',       title: 'Team · Logos',            data: { roles: ['logos_admin', 'app_admin'] },   canActivate: [roleGuard], loadComponent: () => import('./features/team-detail/team-detail').then(m => m.TeamDetail) },
       { path: 'my-workspace',    title: 'My Workspace · Logos',    canActivate: [hasKeysGuard], loadComponent: () => import('./features/my-workspace/my-workspace').then(m => m.MyWorkspace) },
       { path: 'open-code',       title: 'OpenCode · Logos',        canActivate: [hasKeysGuard], loadComponent: () => import('./features/open-code/open-code').then(m => m.OpenCode) },
+      { path: 'claude-code',     title: 'Claude Code · Logos',     canActivate: [hasKeysGuard], loadComponent: () => import('./features/claude-code/claude-code').then(m => m.ClaudeCode) },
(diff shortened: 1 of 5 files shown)


wasnertobias:
Zusammengelegt mit #742 — beides gehört in einen PR.

PR #750, opened 2026-08-19
`Logos`: Map reasoning effort onto Qwen3.8's accepted scale

### LINKED ISSUE #749: Logos: output_config.effort "high" rejected with HTTP 500
## Summary

Logos rejects `output_config.effort: "high"` with HTTP 500, although `high` is a valid
Anthropic Messages API effort value. Supported values are apparently `xhigh`, `medium`
and `low` — `high` is missing from that set.

```
{"error":{"type":"internal_error",
 "message":"Unexpected reasoning effort high. Supported types are xhigh (default), medium, and low."}}
```

## Impact

Claude Code puts its session reasoning effort into **every** request. A session left on
`high` (a normal, selectable setting) therefore fails on every single turn — including a
bare "hello" — because the request is rejected before the model sees anything.

## Reproduction

Same request, only `output_config.effort` varied, against `Qwen/Qwen3.8-27B`:

| effort | result |
|---|---|
| `xhigh` | 200 |
| **`high`** | **500** |
| `medium` | 200 |
| `low` | 200 |

```bash
curl -s https://logos.aet.cit.tum.de/v1/messages \
  -H "Authorization: Bearer $LOGOS_KEY" \
  -H 'anthropic-version: 2023-06-01' -H 'Content-Type: application/json' \
  -d '{"model":"Qwen/Qwen3.8-27B","max_tokens":64,
       "output_config":{"effort":"high"},
       "messages":[{"role":"user","content":"hello"}]}'
```

## Expected

`high` is accepted like any other effort level.

## Notes

- Returning **500 / `internal_error`** for an unsupported enum value is itself wrong —
  a rejected input value should be a 4xx `invalid_request_error`. As a 500 it reads as a
  transient server fault, and clients retry it in a loop.
- I could not locate the validation in the source: the string does not appear in
  `logos/` on the current checkout, so it presumably lives in a newer revision than the
  one I had available.

wasnertobias:
## Summary

Fixes #749.

Claude Code (and other Anthropic-SDK clients) attach the session's reasoning effort to **every** request as `output_config.effort`. A session left on `high` therefore failed on every single turn against `Qwen/Qwen3.8-27B`:

```
HTTP 500 {"error": {"type": "internal_error",
 "message": "Unexpected reasoning effort high. Supported types are xhigh (default), medium, and low."}}
```

**Root cause:** the validation is not in Logos — it lives in the model's chat template. The Qwen3.8 `chat_template.jinja` accepts only `xhigh` (its default), `medium` and `low` as `reasoning_effort`, and vLLM (0.27.1) forwards the client's value to the template verbatim (`output_config.effort` → `ChatCompletionRequest.reasoning_effort` → `build_chat_params()` → template kwargs). The template's `raise_exception` surfaces as a 500 `internal_error`, which clients read as a transient fault and retry in a loop.

## Fix

Logos normalizes the effort value before forwarding, in `ContextResolver.prepare_headers_and_payload` — the single choke point that sync, streaming and job requests all go through. All three locations vLLM passes to the chat template are covered:

- `output_config.effort` (Anthropic Messages surface, what Claude Code sends)
- `reasoning_effort` (OpenAI surface, top level)
- `chat_template_kwargs.reasoning_effort` (explicit template kwarg)

For Qwen3.8 models (matched on `qwen3.8` in the model name):

| in | out | why |
|---|---|---|
| `high` | `xhigh` | top level on both scales |
| `max` | `xhigh` | DeepSeek-specific top level |
| `minimal` | `low` | below `low`; `low` is the weakest accepted level |

`low` / `medium` / `xhigh` / `none` pass through unchanged; every other model is untouched; the input payload is never mutated. With this, all values vLLM's own `reasoning_effort` Literal accepts are either accepted or mapped for Qwen3.8, so the template can no longer reject an effort value on this deployment.

## Tests

New `tests/unit/pipeline/test_effort_normalization.py` (14 tests): the exact #749 request shape, all three payload locations mapped independently, pass-through values, the full vLLM effort Literal covered, other models untouched, non-string/multipart payloads, input immutability, and end-to-end through `prepare_headers_and_payload` for both a Qwen3.8 logosnode context and a cloud model.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for reasoning-effort settings across supported model families.
  * Automatically converts unsupported effort values to equivalents accepted by the target model.
  * Preserves valid settings and existing behavior for unsupported models and request formats.
  * Maintains correct handling for multipart requests and unchanged payload data.

* **Tests**
  * Expanded coverage for model-specific effort handling, supported request formats, payload preservation, and end-to-end request processing.

### DIFF
--- a/logos/logos-orchestrator/src/logos/pipeline/effort_normalization.py
+++ b/logos/logos-orchestrator/src/logos/pipeline/effort_normalization.py
@@ -71,12 +71,19 @@ CHAT_TEMPLATE_EFFORT_SCALES: Dict[str, EffortScale] = {
 
 
 def effort_scale_for_model(model_name: Optional[str]) -> Optional[EffortScale]:
-    """Return the effort scale the model's chat template enforces, if any."""
+    """Return the effort scale the model's chat template enforces, if any.
+
+    When several registered patterns match (e.g. a broad ``qwen3`` and a
+    specific ``qwen3.8``), the most specific one — the longest matching
+    pattern — wins, so broad entries cannot shadow specific ones
+    regardless of registration order.
+    """
     low = (model_name or "").lower()
-    for pattern, scale in CHAT_TEMPLATE_EFFORT_SCALES.items():
-        if pattern in low:
-            return scale
-    return None
+    best_pattern: Optional[str] = None
+    for pattern in CHAT_TEMPLATE_EFFORT_SCALES:
+        if pattern in low and (best_pattern is None or len(pattern) > len(best_pattern)):
+            best_pattern = pattern
+    return None if best_pattern is None else CHAT_TEMPLATE_EFFORT_SCALES[best_pattern]
 
 
 def _map_effort(value: str, scale: EffortScale) -> str:
--- a/logos/logos-orchestrator/src/logos/pipeline/effort_normalization.py
+++ b/logos/logos-orchestrator/src/logos/pipeline/effort_normalization.py
@@ -1,0 +1,74 @@
+# src/logos/pipeline/effort_normalization.py
+"""
+Reasoning-effort normalization for upstreams with a restricted scale.
+
+Clients such as Claude Code attach the session's reasoning effort to every
+request (``output_config.effort`` on the Anthropic Messages surface,
+``reasoning_effort`` on the OpenAI surface). vLLM forwards the value to the
+model's chat template, and the Qwen3.8 template validates it: it only accepts
+``xhigh`` (its default), ``medium`` and ``low``. The Anthropic value
+``high`` — and vLLM's ``minimal``/``max`` — therefore raise a template
+exception that vLLM surfaces as an HTTP 500 ``internal_error``, failing every
+turn of a client session left on ``high`` (ls1intum/edutelligence#749).
+
+This module maps the wider client scale onto the scale the target model's
+template accepts, in every payload location vLLM forwards to the chat
+template:
+
+- ``output_config.effort`` (Anthropic Messages API)
+- ``reasoning_effort`` (OpenAI API, top level)
+- ``chat_template_kwargs.reasoning_effort`` (explicit template kwarg)
+
+Values already on the target scale (``xhigh``, ``medium``, ``low``) and
+``none`` (which vLLM translates into ``enable_thinking=false``) pass through
+unchanged.
+"""
+
+from typing import Any, Dict, Optional
+
+# vLLM's reasoning_effort field accepts none, minimal, low, medium, high,
+# xhigh and max. The Qwen3.8 chat template only accepts xhigh (default),
+# medium and low — map the out-of-scale values onto the closest accepted
+# level instead of letting the template reject them.
+QWEN38_EFFORT_MAP = {
+    "high": "xhigh",  # top level on both scales
+    "max": "xhigh",  # DeepSeek-specific top level
+    "minimal": "low",  # below low; low is the weakest accepted level
+}
+
+
+def model_uses_qwen38_effort_scale(model_name: Optional[str]) -> bool:
+    """Return whether the model's chat template enforces the Qwen3.8 effort scale."""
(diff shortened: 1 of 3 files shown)


logos/logos-orchestrator/src/logos/pipeline/effort_normalization.py
```diff
@@ -0,0 +1,74 @@
...
+unchanged.
+"""
+
+from typing import Any, Dict, Optional
+
+# vLLM's reasoning_effort field accepts none, minimal, low, medium, high,
+# xhigh and max. The Qwen3.8 chat template only accepts xhigh (default),
+# medium and low — map the out-of-scale values onto the closest accepted
+# level instead of letting the template reject them.
+QWEN38_EFFORT_MAP = {
```
wasnertobias: have a more general mapping from chat template to effort map (so map in a map).

wasnertobias (commented):
request changes

logos/logos-orchestrator/src/logos/pipeline/effort_normalization.py
wasnertobias: Done — replaced the single Qwen3.8 map with a `CHAT_TEMPLATE_EFFORT_SCALES` registry: a map from chat template family (model-name substring) to an `EffortScale` carrying the family's `accepted` values and its out-of-scale `map`. `normalize_reasoning_effort` now looks up the scale via `effort_scale_for_model()` and applies whatever mapping the family declares, so a new restricted-scale template is onboarding work of one registry entry. Added a test that registers a second family at runtime to pin that down, and the coverage test now iterates all registered families over vLLM's full effort Literal.

logos/logos-orchestrator/src/logos/pipeline/effort_normalization.py
```diff
@@ -0,0 +1,103 @@
...
+
+A new template family with a restricted scale is added by registering one
+entry in ``CHAT_TEMPLATE_EFFORT_SCALES``.
+"""
+
+from dataclasses import dataclass
+from typing import Any, Dict, FrozenSet, Mapping, Optional
+
+# vLLM's ChatCompletionRequest.reasoning_effort Literal.
+VLLM_REASONING_EFFORT_VALUES = ("none", "minimal", "low", "medium", "high", "xhigh", "max")
```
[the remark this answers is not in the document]
wasnertobias: Addressed in e56007d1 — with a drift-proof runtime rather than a CI check: a vLLM-importing validation is not viable in the orchestrator's test environment (vLLM is only present in the workernode Docker image, and pulling its multi-GB CUDA wheel into the PR CI for one tuple is disproportionate). So each `EffortScale` now carries a `default` fallback, and `normalize_reasoning_effort` coerces any value neither accepted nor explicitly mapped — including values a future vLLM may introduce — to the family default instead of letting it reach a template that would reject it with an HTTP 500. Runtime behavior no longer depends on the snapshot: `VLLM_REASONING_EFFORT_VALUES` now only drives the coverage test, with a comment pointing at the `VLLM_PIP_SPEC` pin (bumped by the "Logos - Update vLLM" workflow) it should be kept in sync with. New test `test_unknown_effort_falls_back_to_family_default` pins the fallback, and the coverage test now also asserts each family's `default` is accepted.

wasnertobias:
Re: CodeRabbit nitpick (review `4981483969`, "Resolve overlapping registry patterns deterministically") — fixed in `dd662eb5`. Verified against the current code: `effort_scale_for_model` did return the *first* matching pattern, so a broad `qwen3` entry inserted before `qwen3.8` would have captured 3.8 requests. It now evaluates all matching `CHAT_TEMPLATE_EFFORT_SCALES` patterns and returns the scale of the **longest** match, independent of registration order. New test `test_longest_matching_pattern_wins_over_broader_pattern` registers overlapping `qwen3`/`qwen3.8` entries in both orders and asserts the specific scale wins (and the broad one still applies to plain `Qwen3-32B`).

PR #640, opened 2026-06-27
`Iris`: Add struggle-intervention pipeline

Predixx:
### Motivation and Context
Part of the Iris **proactive struggle intervention** feature: when the VS Code client detects that a student is stuck on a programming exercise, Artemis forwards a struggle signal and exercise context to Pyris, which must decide whether and how to intervene without spoiling the solution.

This is the Pyris pipeline for that decision. Draft / thesis work in progress; pairs with the server (ls1intum/Artemis#13023) and the client (ls1intum/artemis-extension#335).

### Description

**The pipeline.** `StruggleInterventionPipeline`, registered as the `struggle_intervention_pipeline` LLM config and wired as a route with a single `default` variant. The gate decides `silent` / `ambient` / `active` with an optional inline **anchor + hint**; results are type-validated and the confidence is clamped.

**Three prompts, not one**, because the pipeline serves three distinct moments:

| Template | When |
|---|---|
| `struggle_intervention_system_prompt.j2` | the gate itself, deciding whether to speak at all |
| `struggle_help_request_system_prompt.j2` | the student explicitly asked for help |
| `struggle_confirm_close_system_prompt.j2` | confirming an episode is over |

**Presence level reaches the prompt.** The execution DTO carries `proactivity_mode: Literal["pull", "push"]`, defaulting to `push`. In `pull` the system prompt switches to a reticent tone. Artemis independently enforces the same mode by downgrading `active` to `ambient`, so the tone change here is about how Iris speaks, not about whether it is allowed to.

**Three new tools** the gate can call to ground its decision in what the student actually did, rather than in the signal alone:

- `feedbacks` — retrieves automated test feedback
- `file_lookup_numbered` — a `file_lookup` variant prefixing 1-based line numbers, so the model can name an anchor line the client can actually jump to
- `local_vs_submitted_diff` — diffs the live working copy against the last submission

**DTOs and transport.** Struggle-intervention execution and status-update DTOs, the episode and signal DTOs, and the callback back to Artemis.

**Tests.** Nine new test modules covering the route, the execution DTO, the signal DTO, the status-update DTO, the callback, the pipeline itself, and each of the three tools.

### Steps for Testing
Prerequisites: a configured Pyris (`application.local.yml` + an LLM in `llm_config.local.yml`), a running Weaviate, and Artemis with the struggle endpoint (ls1intum/Artemis#13023).

1. Start Pyris and confirm `pyris` is healthy from Artemis (`GET /management/health` shows the pipeline configs valid).
[...]

PR #756, opened 2026-08-20
`Iris`: Add struggle-intervention pipeline

Predixx:
### Motivation and Context
Part of the Iris **proactive struggle intervention** feature: when the VS Code client detects that a student is stuck on a programming exercise, Artemis forwards a struggle signal and exercise context to Pyris, which must decide whether and how to intervene without spoiling the solution.

This is the Pyris pipeline for that decision. Thesis work in progress; pairs with the server (ls1intum/Artemis#13023) and the client (ls1intum/artemis-extension#335).

### Description

**The pipeline.** `StruggleInterventionPipeline`, wired as a route with a single `default` variant. The gate decides `silent` / `ambient` / `active` with an optional inline anchor + hint; results are type-validated and the confidence is clamped to `[0, 1]`.

**Three prompts, not one**, because the pipeline serves three distinct moments:

| Template | When |
|---|---|
| `struggle_intervention_system_prompt.j2` | the gate itself, deciding whether to speak at all |
| `struggle_help_request_system_prompt.j2` | the student explicitly asked for help |
| `struggle_confirm_close_system_prompt.j2` | confirming an episode is over |

**Presence level reaches the prompt.** The execution DTO carries `proactivity_mode: Literal["pull", "push"]`, defaulting to `push`. In `pull` the system prompt switches to a reticent tone. Artemis enforces the same mode independently by downgrading `active` to `ambient`; the tone change here only affects how Iris speaks.

**Tools to ground the decision** in what the student actually did, rather than in the signal alone:

* `file_lookup_numbered` (new), a `file_lookup` variant prefixing 1-based line numbers, so the model can name an anchor line the client can jump to.
* `local_vs_submitted_diff` (new), diffing the live working copy against the last submission. Registered for both intents whenever a submission exists: on `decide` it reveals the region the student is editing, on `confirm_close` it verifies whether the flagged issue is actually fixed.
* `feedbacks` (existing) now labels each entry `PASS` / `FAIL` / `NOT EXECUTED` and separates test from non-test feedback, so the model stops reading an unexecuted test as a failing one.

**DTOs and transport.** Execution, status-update, episode and signal DTOs, plus the callback back to Artemis.

**Token usage survives a crashed run.** The agent accumulates usage into the run-local execution state, while every subclass' outer error handler runs outside `AbstractAgentPipeline.__call__` and can only reach `self.tokens`, which stayed empty from `__init__` on. A run that died with an exception therefore reported an empty token list and Artemis, which books the usage of each callback, recorded nothing for it. Both names now point at the same list, bound before any fallible setup so a run that fails early cannot report the previous run's tokens. The fix sits in the shared base because `AutonomousTutorPipeline` and `TutorSuggestionPipeline` pass `self.tokens` on their success path too and reported no usage at all. `ChatPipeline` never reads `self.tokens`, so the binding leaves it unchanged.

**Chat reports its usage once, and also when it fails.** Every chat callback handed over the run's cumulative token list: `send_result` carried the answer's usage and the trailing `finish` carried it again, together with the title and suggestion tokens. Artemis creates the trace on the first callback and appends to it on every later one without deduplicating, so the answer was billed twice. `ChatRunCallback` now tracks how many entries Artemis has received and sends only the rest, advancing that count after a delivered POST so a failed `send_result` carries its usage to the next send. Both chat failure paths pass the list now, which is what the delta made safe: `fail()` is terminal, so without it a failed run was billed nothing at all.

**Tests.** Nine new modules covering the route, the four DTOs, the callback, the pipeline and each tool.

### Steps for Testing
Prerequisites: a configured Pyris (`application.local.yml` + an LLM in `llm_config.local.yml`), a running Weaviate, and Artemis with the struggle endpoint (ls1intum/Artemis#13023).

1. Start Pyris and confirm it is healthy from Artemis (`GET /management/health` shows the pipeline configs valid).
2. Trigger a struggle intervention from Artemis (the client `POST .../struggle-intervention`).
3. Confirm Pyris runs `StruggleInterventionPipeline (variant=default)`, calls the configured LLM, and posts a status callback with `action` ∈ {`silent`, `ambient`, `active`}, an optional `anchor_file` / `anchor_line` / `inline_hint`, and a `confidence`.
4. Confirm a low-confidence or `silent` decision surfaces nothing, while `ambient` / `active` produce a hint.
5. Send the same trigger with `proactivity_mode: "pull"` and confirm the reticent tone in the generated hint.
6. Send a `confirm_close` trigger and confirm the callback carries `resolved` plus, on `true`, a `closing_sentence` and `episode_label`.
7. Run the test suite: the nine `iris/tests/test_struggle_*` and `test_*_tool` modules.

### Screenshots

---

Supersedes #640. Same work, moved from the fork branch to an upstream branch.

## Summary by CodeRabbit

* **New Features**
  * Added proactive struggle support with silent, ambient, or active non-spoiler guidance.
  * Added follow-up help requests and confirmation when a reported struggle is resolved.
  * Added contextual hints with file and line references, inline guidance, and episode tracking.
  * Added comparisons between current work and the last submitted code.
  * Added clearer feedback labels and configurable local and cloud models.
  * Added line-numbered code references and a dedicated struggle-intervention endpoint.

* **Bug Fixes**
  * Improved repository-content handling and usage reporting when runs or callbacks fail.
  * Improved inline hint formatting and length handling.


iris/src/iris/domain/struggle/struggle_intervention_pipeline_execution_dto.py
```diff
@@ -0,0 +1,36 @@
...
+    # already given). Same shared assumption as the other agent pipelines; non-text
+    # message shapes are not produced by this flow.
+    chat_history: List[PyrisMessage] = Field(alias="chatHistory", default_factory=list)
+    course: Optional[CourseDTO] = Field(default=None)
+    user: Optional[UserDTO] = Field(default=None)
+    intent: Literal["decide", "confirm_close", "help_request"] = "decide"
+    episode: Optional[EpisodeDTO] = None
+    # Presence level chosen by the student: "pull" (Less, reticent) or "push" (More, willing to reach
+    # out). Prompt tone context only -- the hard Pull cap (active -> ambient) is enforced in Artemis.
+    proactivity_mode: Literal["pull", "push"] = "push"
```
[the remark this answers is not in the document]
Predixx: Not applicable here — Artemis already sends this field in snake_case, so no alias is needed.

The producing DTO annotates it explicitly:

```java
@JsonProperty("proactivity_mode") @Nullable String proactivityMode
```

(`PyrisStruggleInterventionPipelineExecutionDTO`, ls1intum/Artemis#13023), and its javadoc states "Serialized snake_case for the Pyris boundary." Both sides pin the wire name in tests: `PyrisStruggleInterventionExecutionDTOTest` asserts `node.get("proactivity_mode").asText()`, and `test_execution_dto_parses_proactivity_mode_pull` here parses `{"proactivity_mode": "pull"}`.

A `proactivityMode` key is never sent, so adding the alias would only widen the accepted surface without fixing a real path. Leaving it as is.

iris/src/iris/pipeline/struggle_intervention_pipeline.py
```diff
@@ -0,0 +1,296 @@
...
+        self.system_prompt_template = self.jinja_env.get_template(
+            "struggle_intervention_system_prompt.j2"
+        )
+        self.confirm_close_template = self.jinja_env.get_template(
+            "struggle_confirm_close_system_prompt.j2"
+        )
+        self.help_request_template = self.jinja_env.get_template(
+            "struggle_help_request_system_prompt.j2"
+        )
+        self.tokens = []
```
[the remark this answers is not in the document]
Predixx: Fixed in c26e40524.

Confirmed the mechanism: `self.tokens` is set to `[]` in `__init__` and never appended to, while `AbstractAgentPipeline` accumulates usage in `state.tokens`. Since `StatusCallback.finish` guards on `_terminal_sent`, the base class's later `finish(tokens=state.tokens)` was rejected, so usage never reached Artemis. `post_agent_hook` now passes `state.tokens`.

The handler in `__call__` keeps `self.tokens` deliberately: that except block sits outside the agent run, where no state exists.

Note for the maintainers: `autonomous_tutor_pipeline` and `tutor_suggestion_pipeline` carry the identical `self.tokens = []` pattern, so they likely drop usage the same way. Out of scope for this PR, but worth a separate look.

Predixx:
@Claudia-Anthropica review

Pyris side of the proactive struggle intervention; pairs with ls1intum/Artemis#13023. No review from you on this one yet.

CodeRabbit's two findings are addressed and resolved, and the branch is up to date with main.

iris/src/iris/domain/data/feedback_dto.py
```diff
@@ -7,6 +7,11 @@ class FeedbackDTO(BaseModel):
     text: Optional[str] = None
     test_case_name: Optional[str] = Field(alias="testCaseName", default=None)
     credits: float
+    # positive is the authoritative outcome and is TRI-STATE: True=passed, False=failed, None=not executed.
+    # has_test_case distinguishes a real test-case result from non-test feedback (otherwise indistinguishable
+    # in test_case_name).
+    positive: Optional[bool] = None
+    has_test_case: bool = Field(alias="hasTestCase", default=False)
```
az108: `has_test_case: bool = ... default=False` makes "field absent" indistinguishable from "genuine non-test feedback", and `feedbacks.py:44` checks it first, so a missing field short-circuits every entry into `non-test feedback`.

Artemis `main` still ships `PyrisFeedbackDTO(String text, String testCaseName, double credits)` — `hasTestCase` only arrives with the still-open ls1intum/Artemis#13023, which adds it in `PyrisDTOService.toPyrisResultDTO`, i.e. for *all* flows. Since `create_tool_get_feedbacks` is also wired into the exercise chat (`chat_tool_providers.py:82`) and `tutor_suggestion_pipeline.py:127`, and there is no version negotiation between Pyris and Artemis (`compatibility.md`: "There is no formal version compatibility matrix at this time"), deploying this before the Artemis PR regresses those existing pipelines: real results render as `Case: testBubbleSort. non-test feedback. Credits: 1.0 ...` and the model is told, falsely, that no test-case results exist.

Worth noting that the sibling field added in this same PR gets this right — `submitted_repository_available` defaults to the *conservative* value and is documented as such. Suggest `has_test_case: Optional[bool] = None` with an "unknown -> render as before" branch in `_format` (optionally falling back to `test_case_name is not None`).

iris/src/iris/web/routers/pipelines.py
```diff
@@ -473,6 +516,11 @@ def safe_get_variants(get_variants_fn):
             return get_available_variants(
                 safe_get_variants(AutonomousTutorPipeline.get_variants), available_llms
             )
+        case "STRUGGLE_INTERVENTION":
```
az108, line 522: The pipeline is added to the variants endpoint but not to the health registry: `Features` (`web/routers/health/Pipelines/features.py`) and `PIPELINE_BY_FEATURE` (`registery.py`) are unchanged.

`check_pipelines_health()` builds `results = [evaluate_feature(f, available_ids) for f in Features]`, so the struggle pipeline is not even reported as `wired=False` — it is invisible. Health keeps reporting "8/8 pipelines have valid LLM configurations" / `UP` with a missing or broken `struggle_intervention_pipeline` config, which makes step 1 of this PR's own testing instructions ("health shows the pipeline configs valid") unverifiable for this pipeline; a misconfiguration only surfaces at request time via `validate_pipeline_variant`.

Before this PR the `case "X"` labels here were exactly the 8 `Features` members, and `AUTONOMOUS_TUTOR` was added to both files in the same commit (`8a2fd556`). Please add `STRUGGLE_INTERVENTION` to `Features`, `PIPELINE_BY_FEATURE` and the `PipelineType` union — `PIPELINE_ID` and `VARIANT_DEFS` already match what `evaluate_feature` expects.

iris/src/iris/pipeline/prompts/templates/struggle_intervention_system_prompt.j2
```diff
@@ -0,0 +1,95 @@
+{# iris/src/iris/pipeline/prompts/templates/struggle_intervention_system_prompt.j2 #}
+You are Iris, a programming tutor for the course "{{ course_name }}". A deterministic
+struggle-detection engine running in the student's editor has flagged that the student
+MAY be stuck. You were NOT asked a question — this is an unsolicited check.
+
+Detected signal:
+{{ signal_summary }}
+
+You have tools to read the student's current code, the build logs, the test feedback,
```
az108: This promises the code, build-log and test-feedback tools unconditionally, but `get_tools` only registers them under `if submission is not None`, and Artemis passes `submissionDTO = latestSubmission(...).orElse(null)` — a student with no submission yet is an accepted, documented case, and this PR tests it (`test_local_vs_submitted_diff_tool_absent_without_submission`).

The prompt then goes on to *require* `file_lookup` verification before anchoring ("OPEN that exact file and line ... confirm the problem is STILL there") and to mark `anchor`+`inlineHint` REQUIRED, none of which is possible without those tools; only the diff tool has a documented fallback ("or the tool is absent").

`struggle_confirm_close_system_prompt.j2:17-27` phrases the very same situation correctly ("when available", "If the student has not submitted yet, these tools are absent") — please mirror that here and in `struggle_help_request_system_prompt.j2:9`. Note `programming_exercise` *is* always populated by Artemis, so only the three submission-backed capabilities are affected.

iris/src/iris/pipeline/struggle_intervention_pipeline.py
```diff
@@ -0,0 +1,296 @@
...
+    def post_agent_hook(
+        self,
+        state: AgentPipelineExecutionState[
+            StruggleInterventionPipelineExecutionDTO, Variant
+        ],
+    ) -> str:
+        cb = state.callback
+        status = cast(StruggleInterventionStatusUpdateDTO, cb.status)
+        intent = getattr(state.dto, "intent", "decide")
+        if intent == "confirm_close":
```
az108, line 446: `post_agent_hook` branches only on `confirm_close`, so `help_request` goes through the same `parse_gate_result`, which fail-safes to `GateResult("silent", None, 0.0, ...)` on empty/unparseable output, an invalid `action`, or a missing/non-string `message`. That contradicts the template's own contract ("NEVER SILENT ... you MUST return a usable hint ... Do NOT return 'silent'").

On the Artemis side, `handleDecision` hits `if (result == null || result.isEmpty())` before the switch and emits `silentDecide(...)` — so a student who explicitly clicked "Show me" gets an empty completion frame with no hint and no error, and `status.rationale` ("unparseable model output") is never read on that path.

A deliberate `silent` from the model is intentionally honoured there (`helpRequest_silentFromPyris_staysSilent`), which is fine — but a parse failure is currently indistinguishable from it, so it cannot be handled separately. Low probability given how tolerant the JSON extraction is, but the impact is a silent total failure on the one path where the student actively asked. Consider surfacing the parse failure on this intent (a distinct marker, a repair retry, or `fail()`) instead of collapsing it into `silent`.

iris/tests/test_struggle_intervention_pipeline.py
```diff
@@ -0,0 +1,438 @@
...
+    When "j2" was in the enabled_extensions list Jinja treated every .j2 file
+    as an HTML template and escaped {{ }} values, so a hint like
+    "is the bound i < n or List<String> & reset" would reach the LLM as
+    "is the bound i &lt; n or List&lt;String&gt; &amp; reset" -- corrupting the prompt.
+
+    This test renders the confirm_close system-prompt template with a hint
+    carrying angle brackets and a raw ampersand, then asserts the characters
+    survive unchanged in the rendered prompt.
+    """
+    pipeline = StruggleInterventionPipeline()
```
az108, line 212: The nine new test modules cover the parsers, DTOs, route and tools, but nothing exercises `post_agent_hook` — the method that maps `GateResult`/`ConfirmCloseResult` onto the status DTO and calls `finish(result=..., tokens=state.tokens, confidence=...)`. All 11 `StruggleInterventionPipeline()` instantiations in this file are for template rendering or tool registration; the pipeline itself is never called.

That is exactly where the token bug fixed in `c26e4052` lived, and that fix shipped without a regression test. A relapse to `self.tokens` would not turn a single test red, because the dropped terminal send only returns `False` instead of raising.

The project already has this pattern: `test_chat_latency_ordering.py` builds the pipeline via `__new__`, stubs the agent, and asserts `test_finish_carries_accumulated_tokens` for the same failure mode (plus `assert "tokens" in fail_kwargs` with the note that "fail() is terminal, so it must carry the accumulated usage itself"). Suggest one test per intent (`decide` / `confirm_close`) calling `post_agent_hook` with a mocked callback and asserting the `finish` kwargs (`tokens`, `confidence`) plus the field mapping (`action`/`anchor_file`/`anchor_line`/`inline_hint`, `resolved`/`closing_sentence`/`episode_label`).

az108 (commented):
## Review: struggle-intervention pipeline

Reviewed the full diff (27 files, +1903/-6) at `c26e4052`. CI is green. The two existing CodeRabbit threads (`proactivityMode` alias, `self.tokens` on the finish path) are both settled and are not re-raised here.

**5 findings: 2 that I would fix before merge, 3 worth addressing.** Every finding below was independently verified against the code, the Artemis counterpart (ls1intum/Artemis#13023) and the repo's own conventions; three further candidates were checked and dropped (see the end).

### Blocking

1. `feedback_dto.py:14` — `has_test_case` defaulting to `False` regresses the **existing** exercise-chat and tutor-suggestion pipelines against current Artemis `main`.
2. `pipelines.py:519` — the pipeline is missing from `Features` / `PIPELINE_BY_FEATURE`, so it is invisible to the health check.

### Worth addressing

3. The decide and help_request prompts promise tools that are not registered when there is no submission.
4. `help_request` can silently produce nothing when the model output fails to parse.
5. No test exercises `post_agent_hook` — the method the token bug in `c26e4052` lived in.

---

### Non-blocking observations

- **Double `finish()` per successful run.** `post_agent_hook` sends the terminal update, then `AbstractAgentPipeline.__call__:766` calls `finish()` again, hitting `_reject_after_terminal` which logs a warning *and* sends a `capture_message` to Sentry. This is the existing pattern (`autonomous_tutor`, `tutor_suggestion`) and not a defect of this PR — but a proactive pipeline fires far more often than user-initiated chat, so the Sentry noise scales with it.
- **Raw `Thread` per alert.** Route convention in this repo, but the same volume argument applies relative to user-paced routes.
- `episode_dto.py` re-declares `model_config = ConfigDict(populate_by_name=True)` inline instead of reusing the `_CamelModel` base from `struggle_signal_dto.py` — two new files in the same feature, two styles. (`_CamelModel` is also slightly misleading as a name: it has no `alias_generator`, so every field still needs an explicit `Field(alias=...)`.)

### What is good here

The fail-safe semantics in `parse_gate_result` (clamping, `math.isfinite`, the `bool`-is-an-`int` guard), the `select_autoescape` regression with its own test, the deliberately conservative `submitted_repository_available` default, and the fact that `post_agent_hook` uses `state.tokens` after `c26e4052` — which makes this pipeline's token reporting cleaner than its siblings'.

### Candidates checked and dropped

- `callback.fail(..., tokens=self.tokens)` reporting zero tokens: payload-identical to `tokens=None` (`fail()` only assigns when `tokens is not None`, and `status.tokens` is the empty default at that point). No `state` exists in the subclass `except` block, so a real fix means reworking the base class. Already explained by the author in a resolved thread.
- Missing entry in the `pipeline-system.md` route table: that table is demonstrably not maintained — it already omits `/webhooks/lectures/metadata`, `/webhooks/lectures/visibility` and `/search/lectures` (11 of 14 POST routes listed), it was incomplete on the day it was written, and no feature PR has ever updated it.
- System-only prompt when `chat_history` is empty: no Anthropic provider is configurable here (only `openai_chat`/`azure_chat`/`ollama`; `langchain-anthropic` is not locked), system-only is valid on the actual request path, and `AutonomousTutorPipelineExecutionDTO` has no `chat_history` field at all — so that pipeline already ships system-only prompts in production on the same models.

Claudia-Anthropica (changes requested):
@Predixx [medium] The unresolved feedback-compatibility and pipeline-health findings documented in the retained threads remain reproducible on this exact head; the no-submission prompt/tool mismatch also remains. Exact-head CI completed successfully, but these current-code defects require changes before approval.

iris/src/iris/pipeline/prompts/templates/struggle_help_request_system_prompt.j2
```diff
@@ -0,0 +1,73 @@
...
+consented, invited request — not an unsolicited check.
+
+Detected signal:
+{{ signal_summary }}
+
+You have tools to read the student's problem statement, and - WHEN the student has already
+submitted at least once - their current code, the build logs and the test feedback. Before the
+first submission Artemis sends no submission, so those code and build tools are simply absent:
+do not assume a tool exists, and if the ones you need are missing, judge from the signal and the
+chat history alone or stay `silent`. The recent chat history is provided.
```
Claudia-Anthropica: @Predixx [medium] This fallback directly contradicts the `NEVER SILENT` contract below. For a pre-submission `help_request`, it explicitly permits `silent`, and `post_agent_hook` intentionally treats a valid silent response as a successful terminal result, so a student's explicit request can still produce no hint. Remove the silent option here and prescribe an available-context or human-tutor fallback that returns `ambient` or `active`.

In `iris/src/iris/pipeline/prompts/templates/struggle_help_request_system_prompt.j2`, the no-submission fallback permits silence despite the later `NEVER SILENT` contract, and the pipeline honors that output. Require an `ambient` or `active` fallback based on available context, including a human-tutor referral when necessary.

</details>

iris/src/iris/pipeline/struggle_intervention_pipeline.py
```diff
@@ -0,0 +1,361 @@
...
+
+def parse_confirm_close_result(raw: str) -> ConfirmCloseResult:
+    """Parse the LLM's confirmClose JSON. Fail closed to resolved=False on any problem."""
+    obj = _extract_json_object(raw)
+    if obj is None:
+        return ConfirmCloseResult(False, None, None, None)
+    resolved = obj.get("resolved")
+    if not isinstance(resolved, bool):
+        return ConfirmCloseResult(False, None, None, _as_opt_str(obj.get("rationale")))
+    if resolved:
```
Claudia-Anthropica: @Predixx [medium] A model response containing only `{"resolved": true}` reaches this branch and closes the episode with both `closing_sentence` and `episode_label` set to `None`; empty strings are accepted as well. Those fields are required for a resolved response and are the acknowledgement and fold label shown to the student. Validate both as non-empty strings before returning `resolved=True`, otherwise fail closed to `resolved=False`.

In `iris/src/iris/pipeline/struggle_intervention_pipeline.py`, `parse_confirm_close_result` accepts `resolved=true` without non-empty `closingSentence` and `episodeLabel`, closing the episode without its required student-visible content. Treat missing, non-string, or blank required fields as invalid and return a fail-closed unresolved result.

</details>

Claudia-Anthropica (changes requested):
@Predixx Exact-head CI completed successfully, and the feedback-compatibility and health-registry fixes are now present. Two medium correctness gaps remain: pre-submission help requests can still be instructed to return silence, and confirm-close accepts an incomplete resolved response. Both can leave the student-facing workflow without its required content.

iris/src/iris/pipeline/struggle_intervention_pipeline.py
```diff
@@ -0,0 +1,395 @@
...
+        if not isinstance(message, str) or not message.strip():
+            return GateResult(
+                "silent",
+                None,
+                0.0,
+                "non-silent action without message",
+                parse_failed=True,
+            )
+    try:
+        confidence = float(obj.get("confidence", 0.0))
```
Claudia-Anthropica: @Predixx [medium] Python converts `True` to `1.0` here, so a malformed model response with `"confidence": true` is treated as maximum confidence and can pass Artemis's threshold for an unsolicited intervention. Explicitly reject booleans before numeric coercion and add a regression test proving they fall back to `0.0`.

In `iris/src/iris/pipeline/struggle_intervention_pipeline.py`, boolean confidence values are coerced to numeric certainty and can bypass the downstream confidence gate. Reject booleans before calling `float()` and add a regression test asserting that `true` produces confidence `0.0`.

</details>

Claudia-Anthropica (changes requested):
@Predixx The previously blocking prompt and confirm-close defects are fixed, and exact-head CI completed successfully. One medium output-validation defect remains: boolean confidence is coerced to `1.0`, allowing malformed model output to bypass Artemis's confidence gate for an unsolicited intervention.

iris/src/iris/pipeline/struggle_intervention_pipeline.py
```diff
@@ -0,0 +1,402 @@
...
+        if intent == "confirm_close":
+            cc = parse_confirm_close_result(state.result or "")
+            status.resolved = cc.resolved
+            status.closing_sentence = cc.closing_sentence
+            status.episode_label = cc.episode_label
+            status.rationale = cc.rationale
+            cb.finish(tokens=state.tokens)
+            return cc.closing_sentence or ""
+        gate = parse_gate_result(state.result)
+        if intent == "help_request" and gate.parse_failed:
```
Claudia-Anthropica: @Predixx [medium] This guard catches parse failures only. `parse_gate_result` classifies `{"action":"silent"}` as valid, so a `help_request` falls through to `finish(result=None)` even though its response schema permits only `ambient` or `active`; the student's explicit request therefore disappears as the same empty completion this branch is meant to prevent. Treat `silent` as unusable for this intent (or provide a guaranteed non-silent fallback) and replace the test that currently enshrines silence.

In `iris/src/iris/pipeline/struggle_intervention_pipeline.py`, `post_agent_hook` accepts `silent` for `help_request`, finishing with a null result despite the intent's non-silent response contract. Treat `silent` as unusable for this intent or supply an `ambient`/`active` fallback, and update the regression test accordingly.

</details>

Claudia-Anthropica (changes requested):
@Predixx The boolean-confidence regression and the previously blocking compatibility, health, prompt, and confirm-close defects are fixed, and exact-head CI completed successfully. One medium help-request validation gap remains: an intent-forbidden `silent` result is still accepted and produces the empty completion the new guard is intended to prevent.

Claudia-Anthropica (dismissed):
@Predixx The latest head rejects both parse failures and explicit `silent` actions for `help_request` while preserving valid `decide` silence, with regression coverage for both branches. The earlier compatibility, health-registration, prompt-availability, confirm-close, and confidence findings are fixed on this head. The snapshot had no required checks, although several non-required build, test, and documentation checks were still running when captured.

Claudia-Anthropica (dismissed):
@Predixx The latest shared token-list binding preserves accrued usage on failure and is covered by a regression test. The earlier findings remain fixed, and the exact-head test check succeeded; only non-required lint and container-build checks were still running in the captured snapshot.

iris/src/iris/pipeline/struggle_intervention_pipeline.py
```diff
@@ -0,0 +1,408 @@
...
+    raw_line = raw_anchor.get("line") if isinstance(raw_anchor, dict) else None
+    # bool is a subclass of int in Python, so guard against `"line": true` masquerading as a line number.
+    if (
+        isinstance(raw_anchor, dict)
+        and isinstance(raw_anchor.get("file"), str)
+        and isinstance(raw_line, int)
+        and not isinstance(raw_line, bool)
+    ):
+        anchor = {"file": raw_anchor["file"], "line": raw_line}
+    inline_hint = obj.get("inlineHint")
```
Claudia-Anthropica: @Predixx [medium] `silent` only suppresses `message`; this code still parses `anchor` and `inlineHint`, and `post_agent_hook` copies both into the callback. Thus malformed-but-valid model output can attach a gutter cue to a decision whose contract requires surfacing nothing. Clear both fields whenever `action == "silent"` and add a regression test for a silent response containing an anchor and inline hint.

In `iris/src/iris/pipeline/struggle_intervention_pipeline.py`, `parse_gate_result` retains `anchor` and `inlineHint` for a `silent` action, allowing a silent callback to carry student-visible hint data. Force both fields to `None` for silent decisions and add a regression test covering inconsistent model output.

</details>

Claudia-Anthropica (changes requested):
@Predixx The previously reviewed compatibility, health-registration, prompt, confirm-close, confidence, help-request, and token-accounting defects remain fixed on this head. One medium output-validation gap remains: a `silent` decision can still transmit a student-visible gutter hint and anchor. The captured Test check succeeded, while non-required lint and container-build checks were still in progress.

Claudia-Anthropica (dismissed):
@Predixx The current head correctly drops both anchor and inline hint data from `silent` decisions, with regression coverage for parsing and callback mapping. The previously reviewed compatibility, health-registration, prompt, confirm-close, confidence, help-request, and token-accounting fixes remain present. The snapshot contains no required or failed checks; optional lint, test, and container-build checks were still running when captured.

iris/src/iris/pipeline/struggle_intervention_pipeline.py
```diff
@@ -0,0 +1,437 @@
...
+    # itself, so the decision wins and both are dropped here rather than put on the wire.
+    if action != "silent":
+        raw_anchor = obj.get("anchor")
+        raw_line = raw_anchor.get("line") if isinstance(raw_anchor, dict) else None
+        # bool is a subclass of int in Python, so guard against `"line": true` masquerading as a line number.
+        if (
+            isinstance(raw_anchor, dict)
+            and isinstance(raw_anchor.get("file"), str)
+            and isinstance(raw_line, int)
+            and not isinstance(raw_line, bool)
```
Claudia-Anthropica: @Predixx [medium] This accepts any string and any non-boolean integer as an anchor, including a blank or nonexistent path and a zero, negative, or beyond-EOF line; `inlineHint` is then retained independently. A model response can therefore send the editor a student-visible cue with no valid location despite the 1-based paired-field contract. Validate the path and line against the current submission repository, and forward `anchor` and `inlineHint` only when both are valid.

In `iris/src/iris/pipeline/struggle_intervention_pipeline.py`, `parse_gate_result` accepts blank or nonexistent anchor paths and non-positive or out-of-range lines while retaining `inlineHint` independently, allowing malformed model output to emit an invalid gutter annotation. Validate the anchor against the current repository and forward the anchor and hint only as a valid pair.

</details>

Claudia-Anthropica (changes requested):
@Predixx [medium] The earlier findings remain fixed, but malformed non-silent output can still send an incomplete or invalid student-visible gutter anchor. The captured Test and build checks passed; the non-required Lint check is red, and its snapshot exposes only a generic exit-code failure, so I could not attribute it.

Claudia-Anthropica (changes requested):
@Predixx [medium] The unresolved anchor-validation finding remains reproducible on this exact head: the parser still accepts any string path and any non-boolean integer, retains `inlineHint` independently, and `post_agent_hook` forwards those values without checking the current repository. The latest commit only changes two chat failure-path tests to use `MagicMock`, so it does not address that student-visible invalid-gutter-payload path. The captured Test and Lint checks succeeded; two non-required container-image builds were still in progress.

Claudia-Anthropica (dismissed):
@Predixx The remaining anchor-validation defect is fixed: anchors are checked against the run’s repository snapshot and invalid or incomplete gutter pairs are dropped, with regression coverage through callback mapping. Earlier fixes remain intact, and all 11 captured exact-head checks succeeded.

iris/tests/test_struggle_route.py
```diff
@@ -0,0 +1,101 @@
+# Tests the route registration + the worker directly.
+# The confirm_close smoke test uses FastAPI TestClient with the real auth dependency
+# (the "secret" token present in application.example.yml, loaded by conftest.py).
+import sys
+import pytest
```
Claudia-Anthropica: @Predixx [high] This third-party import splits the standard-library import block, causing Iris's isort hook—and therefore the exact-head Lint check—to fail. Move `from unittest.mock import MagicMock, patch` directly after `import sys`, then place `pytest` with the FastAPI imports after a blank line.

In `iris/tests/test_struggle_route.py`, `pytest` is incorrectly placed inside the standard-library import block, causing the isort check to fail. Group the `sys` and `unittest.mock` imports first, followed by a blank line and the third-party imports.

</details>

Claudia-Anthropica (changes requested):
@Predixx [high] The exact-head Test and build checks succeeded, but the optional Iris Lint check is red. The newest commit places `pytest` between standard-library imports; Iris's isort hook requires the stdlib imports together and third-party imports after a blank line, directly explaining the failure.

Claudia-Anthropica (dismissed):
@Predixx The latest commit fixes the prior lint failure by restoring the correct standard-library and third-party import grouping. The previously reviewed correctness fixes and their regression coverage remain present. All 17 captured exact-head checks, including Iris Test and Lint, succeeded.

Predixx:
@coderabbitai full review

Your last review is from `96a29de0` on 30 August and carries no verdict. 18 commits landed since then, including a merge from `main`, so the current head is `336f0206`.

What changed since that review:

- `5f9ab966` and `0b1dc628` address the compatibility, health-registration and failure-path findings from your threads.
- `c8b56def` and `fe1ecea8` reject `silent` on the `help_request` path instead of letting it through.
- `9888a2f9` and `98709a90` handle a degraded `confirm_close` and reject a close carrying a gate action.
- `14d78ae5` stops a boolean `confidence` from passing as the maximum value.
- `607b6685`, `dbcc6c27` and `e9732b71` fix token accounting: a crashed run keeps its usage, error paths report it, and Artemis is no longer told twice.
- `c057ec4d` and `61d73f1b` drop gutter hints on `silent` decisions and only send an anchor that exists in the run's repository snapshot.
- `91096196` and `336f0206` remove duplication and fix the import grouping that broke Iris Lint.

All 17 checks pass on the current head and every review thread is resolved. Please review it so your verdict reflects the current state.

Predixx:
@coderabbitai full review

Predixx:
@coderabbitai review

Predixx:
@coderabbitai full review

Claudia-Anthropica (dismissed):
@Predixx The exact-head implementation safely validates gate and close outputs, checks anchors against the repository snapshot, and fully wires the new pipeline into routing, configuration, and health reporting. The regression suite covers the prior compatibility, token-accounting, callback, prompt, diff, and gutter-payload concerns, and all 17 captured checks succeeded.

Claudia-Anthropica (dismissed):
@Predixx The new untouched-stub guidance remains consistent with the existing focus, redirect, and non-spoiler rules, while the previously reviewed validation, callback, tool, routing, and health-registration fixes remain intact. The captured exact-head snapshot contains 25 successful checks, 6 skipped checks, and no failures.

Senan04 (dismissed):
Tested locally together with the Artemis PR, works as expected :)

Image

Claudia-Anthropica (dismissed):
@Predixx The latest episode-scoping change keeps current and unmarked hints deduplicated while allowing a confirmed, non-dismissed diagnosis from an earlier episode to become relevant again. The previously reviewed validation, callback, routing, prompt, and tool protections remain intact, and all 17 captured exact-head checks succeeded.

Claudia-Anthropica (approved):
@Predixx The new feature-branch commit only condenses test comments, and the merge from main leaves the authoritative PR diff and previously reviewed validation, callback, routing, prompt, and tool protections intact. All 17 captured exact-head checks, including Iris Test and Lint, succeeded.

az108 (approved):
reapprove

Senan04 (approved):
Reapprove

MarkusPaulsen (approved):
Tested on Local as per Liam's guidance

MarkusPaulsen (approved):
Reapprove

Issue #735, opened 2026-08-18
`Logos`: When connection breaks cancel request


wasnertobias:
Right now when a connection to a long-running request breaks it is not cancelled, this should change as it otherwise those "ghost requests" do use resouces without the need.

wasnertobias:
@Claudia-Anthropica pls work on it

wasnertobias:
When the connection breaks due to a logos deployment (orchestator restart) right now there remain requests in "running" state basically forever. On startup it shall be ensured that no requests are in running state any longer.

wasnertobias:
Further note: In a current overload situation claude code is re-trying 10 times and to me it seems like that Logos still does e.g., handle requests 1 to 4, when the 5th retry is being sent. Instead Logos shall instantly cancel request 1 - 4, as claude code is no longer listening to the answer of the request. This is most likely a structural problem, where the cancellation is not forwarded to the workernode. If this is not feasible then this issue relates to: #814

PR #759, opened 2026-08-20
`Logos`: Fix 500 on NUL bytes, cancel ghost requests, spell out RPM and TPM

### LINKED ISSUE #753: `Logos`: RPM and TPM shall be described in text
It is not clear what it means

wasnertobias:
### Motivation

A single NUL byte anywhere in a request body made Logos answer `POST /v1/messages` with HTTP 500 — `Unhandled exception on /v1/messages`. Reproduced against the production gateway:

```
"Reply with: OK"          -> HTTP 200
"Reply with: OK" + NUL    -> HTTP 500  {"error":{"message":"Internal server error"}}
```

The failure was deterministic and looked like a busy worker, but no worker was ever involved. `auth_parse_log` calls `db.log_usage()` **before** the request is forwarded, so the error arrives instantly and every client retry fails identically:

```
psycopg2.errors.UntranslatableCharacter: unsupported Unicode escape sequence
DETAIL:  [the NUL escape] cannot be converted to text.
  File "/app/src/logos/dbutils/dbmanager.py", line 2099, in log_usage
  File "/app/src/logos/main.py", line 3754, in auth_parse_log
```

`json.dumps` renders a NUL as the one escape sequence Postgres refuses inside `jsonb`, and `log_entry.input_payload` is `jsonb`. `sanitized_payload_for_logging` only strips multipart file contents, so control characters passed straight through.

Only keys with `log = FULL` were affected, because `payload_str` is guarded by that check — which is why this looked like a single user's problem.

How it is triggered in practice: a Claude Code session whose transcript had captured raw binary output (a `dd` dump of a binary) replayed that tool result on every turn. Each turn burned roughly eleven retries over about three minutes before the client dropped the offending entry and one request got through.

### Description

Every value bound to a `jsonb` parameter now goes through `_json_for_jsonb`, which strips NULs before serialising. That covers the request payload, the response payload, the headers, the async job route (`jobs.request_payload` — the same body on a second route), the classification statistics, the provider snapshots and the API key settings.

Sanitising happens on the object, not on the serialised JSON: afterwards the escape also occurs as a substring of an escaped backslash, so a plain `.replace()` on the text would leave a dangling backslash behind and corrupt the document. A test pins that trap down.

`_stringify_error_message` is deliberately untouched — its target column is `text`, not `jsonb`.

### Unrelated: RPM and TPM (closes #753)

Bundled on request. The abbreviations appeared as bare labels, and the placeholders explaining them vanished as soon as a value was entered. The unit is now named in the label on the team settings tab, the API key modal and the key creation form, plus a tooltip on the limits in My Workspace.

The compact table columns keep the abbreviation: `app-data-table` takes plain strings for its headers and those columns are 90px wide. Happy to add a visible legend there if you would prefer it.

### Steps for testing

1. `cd logos/logos-orchestrator && ./run_tests.sh unit` — 635 passed, 1 xfailed before and after; the new `tests/unit/dbutils/test_nul_byte_payloads.py` adds 7.
2. The new test fails without the fix (4 failures) and also catches the naive `.replace()` variant (1 failure), so it guards the regression rather than just describing it.
3. With a `FULL`-logging key, `POST /v1/messages` carrying a NUL in the message content now returns 200 and the stored `input_payload` holds the text without the NUL.
4. Team settings, the API key modal, the key creation form and My Workspace show the spelled-out units.

Note: the Angular build was not run locally — `logos-ui` has no `node_modules` here and the change is static template text. The four touched templates are not Prettier-clean at `main` either, so their formatting was left as it is instead of reformatting unrelated lines.

### Ghost requests: cancel on disconnect (part of #735)

Uvicorn does not tear the handler down when a client connection breaks, so a non-streaming call kept a worker generating a response nobody would ever read — a ghost request holding a GPU lane for the full length of the generation. Both executor paths use `timeout=None` ("No timeout to handle long-running LLM requests and cold starts"), so nothing else stopped it either.

`handle_sync_request` now races the pipeline against a disconnect probe and cancels the work once the client is gone. Cancelling unwinds the executor's `httpx.AsyncClient` context, which closes the upstream connection so vLLM aborts the sequence. The capacity planner already releases its VRAM reservation on `CancelledError`, so that side is covered. The dropped request is logged as an error saying the client left, and the handler answers 499 the way nginx does.

Streaming deliberately keeps Starlette's own watcher: `StreamingResponse.listen_for_disconnect` already cancels the body generator, which unwinds the same stream context. Polling the receive channel here would race that watcher for the `http.disconnect` message, so only responses returned whole get the guard.

Two notes on scope:

- **No new `result_status`.** A dedicated `cancelled` status would read better on the statistics page, but `result_status` is a Postgres enum (`success`, `error`, `timeout`) and adding a label needs a Liquibase changeset with `runInTransaction="false"`. That is not something to slip into a PR about a NUL byte, so cancellations are recorded as `error` with an explicit message. Happy to add it as a follow-up.
- **Admin key rotation**, the second half of #735, already exists: `POST /api-keys/{keyId}/rotate` is authorised for `LOGOS_ADMIN` (any key) and `APP_ADMIN` (own teams), and the members tab already opens the same key modal. Nothing to do there.

### Steps for testing (added)

5. `tests/unit/main/test_client_disconnect_cancel.py` — 4 tests: the result passes through while the client stays, the upstream task is actually cancelled when it leaves, pipeline errors still propagate, and streaming keeps Starlette's watcher.
6. Both mutations are caught: without the cancellation the disconnect test fails after 30s — the ghost request running to completion is the failure — and removing the streaming gate fails the fourth test.
7. Full unit suite: 639 passed, 1 xfailed (was 635 before this PR).

https://claude.ai/code/session_01DJSC5JLxE7hpWEdxxa2c6p

## Summary by CodeRabbit

- **Bug Fixes**
  - Automatically removes unsupported NUL characters from data saved to the database, including nested payloads.
  - Cancels non-streaming requests when the client disconnects and returns an appropriate status.

- **UI Improvements**
  - Clarified cloud and local rate-limit labels by spelling out requests per minute and tokens per minute.
  - Added tooltips explaining rate-limit abbreviations.

### DIFF
--- a/logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
+++ b/logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
@@ -161,6 +161,11 @@ def _strip_nul(value: Any) -> Any:
     ``json.dumps`` the escape for a NUL also occurs as a substring of an escaped
     backslash, so replacing it in the text would leave a dangling backslash
     behind and corrupt the document.
+
+    Keys that differ only in NULs collapse into one, and the last one wins —
+    a JSON object cannot hold both. That only arises for deliberately crafted
+    payloads, and these values feed audit logs rather than behaviour, so losing
+    one member of such a pair beats rejecting the request.
     """
     if isinstance(value, str):
         return value.replace("\x00", "")
@@ -437,7 +442,12 @@ class DBManager:
             "updated_at": datetime.datetime.now(datetime.timezone.utc),
         }
         if result_payload is not None:
-            update_data["result_payload"] = result_payload
+            # jobs.result_payload is jsonb and the reflected update binds this
+            # dict directly, so SQLAlchemy serialises it — _json_for_jsonb would
+            # store its string as a JSON scalar instead of an object. A NUL in a
+            # model's answer would otherwise fail the write and leave the job
+            # without its result.
+            update_data["result_payload"] = _strip_nul(result_payload)
         if error_message is not None:
             update_data["error_message"] = (
                 error_message if isinstance(error_message, str) else _stringify_error_message(error_message)
--- a/logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
+++ b/logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
@@ -154,6 +154,37 @@ def _stringify_error_message(value: Any) -> str:
     return str(value)
 
 
+def _strip_nul(value: Any) -> Any:
+    """Drop NUL characters from every string nested inside ``value``.
+
+    Stripping has to happen on the object, not on serialised JSON: after
+    ``json.dumps`` the escape for a NUL also occurs as a substring of an escaped
+    backslash, so replacing it in the text would leave a dangling backslash
+    behind and corrupt the document.
+    """
+    if isinstance(value, str):
+        return value.replace("\x00", "")
+    if isinstance(value, dict):
+        return {_strip_nul(key): _strip_nul(item) for key, item in value.items()}
+    if isinstance(value, (list, tuple)):
+        return [_strip_nul(item) for item in value]
+    return value
+
+
+def _json_for_jsonb(value: Any) -> str:
+    """Serialise ``value`` for a ``jsonb`` column or cast.
+
+    ``json.dumps`` renders a NUL as the one escape sequence Postgres refuses
+    inside ``jsonb`` — *unsupported Unicode escape sequence ... cannot be
+    converted to text*. A single such byte anywhere in a request body therefore
+    turned the logging insert into an unhandled 500, raised from
+    ``auth_parse_log`` before the request ever reached a worker: a client
+    replaying a conversation that had captured raw binary output got an instant
+    server error on every retry, and nothing was logged either.
+    """
+    return json.dumps(_strip_nul(value))
+
+
 # noinspection PyUnresolvedReferences
 class DBManager:
     def __init__(self):
@@ -381,7 +412,7 @@ class DBManager:
             ),
             {
                 "status": status,
-                "payload": json.dumps(payload),
+                "payload": _json_for_jsonb(payload),
                 "aki": api_key_id,
                 "tid": team_id,
                 "uid": user_id,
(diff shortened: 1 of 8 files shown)


wasnertobias:
Thanks — four of the five were real and are fixed in 82fae5b8. Verified each against the code rather than taking the summary at face value.

**`update_job_status` (major, correct and the most valuable catch).** `jobs.result_payload` is jsonb and the reflected update binds the dict directly, so a NUL in a model's answer failed the write and left the job without its result. My original sweep looked for `json.dumps` call sites, which is exactly why this path was invisible — it never serialises here. Fixed with `_strip_nul`, and your note about not using `_json_for_jsonb` is right: it would have stored a JSON scalar instead of an object. `self.update`/`self.insert` have only two call sites in the file; the other one writes `token_types` (text columns), so this was the last gap of its kind.

**Awaiting the cancelled work (major, correct).** `_settle` now cancels and waits for the cleanup before the wrapper unwinds, so `_sync_response`'s `finally` releases the scheduler slot first. A test pins it: without the await the released list is still empty when the wrapper returns.

**Resolved Whisper bypassing the guard (major, correct — and it went deeper than the suggested fix).** The predictive gate could not be made correct: `resolved_whisper_model` is derived from `result.execution_context` after scheduling, so no check at the entry point knows the answer. Rather than approximating it, the wrapper now reacts to what came back and the gate is gone. That also closed a race I had accepted: because `_streaming_response` pre-pulls the first chunk (`chunk_iter.__anext__()`), a response finishing in the same tick as the disconnect probe already holds an open upstream connection, and the probe has consumed the `http.disconnect` that Starlette's watcher would need. Such a response is now closed via its `body_iterator` instead of being streamed into a dead socket.

**Key collisions (major → documented, not enforced).** Correct that `_strip_nul({"k\0": 1, "k": 2})` yields `{"k": 2}`. I did not add a rejection policy: any policy that refuses the payload reintroduces the 500 this PR exists to remove, and a JSON object cannot hold both keys. It needs a deliberately crafted request, and these values feed audit logs rather than behaviour, so dropping one member beats failing the request. The trade-off is now in the docstring with a regression test pinning last-wins. Happy to change it if you would rather see a 400.

**Summary cards (minor, correct).** Missed those — the two `stat-mini` tiles now carry the same tooltip as My Workspace. Spelling the units out inline would not fit a compact stat tile.

On the two pre-merge warnings:

- **Out of scope** — agreed, and I flagged it when the second and third topics were added. The bundling was requested deliberately. Splitting the ghost-request commit into its own PR is the one I would separate, since it is the only part with runtime risk; say the word and I will.
- **Docstring coverage 35.90%** — every function this PR adds has a docstring. The figure is the file-level number for `main.py` and `dbmanager.py`, which the PR inherits rather than causes; raising it would mean documenting a few hundred pre-existing methods.

Unit suite: 644 passed, 1 xfailed (635 on `main`). All three new safeguards are mutation-checked — removing the await, the same-tick check, or the `result_payload` sanitisation each fails a specific test.

PR #775, opened 2026-08-25
`Logos`: Guide the AI-tools setup step by step and size context windows dynamically

wasnertobias:
## Summary

Two problems, one page.

**The setup asked for everything at once**, and configured Claude Code by writing an `env` block into `~/.claude/settings.json`. That file is global: it redirects *every* Claude Code session on the machine to Logos, so anyone with an Anthropic subscription had to edit it back and forth to use both.

**And it handed out one context-window number that was wrong twice over.** It was the smallest window across the cluster, and it had the output reservation subtracted a second time on top of the subtraction Claude Code already performs. On a 111,200-token window that meant auto-compacting at 37,240 tokens instead of 75,976 — the "fires at ~60%" effect from Slack.

### The page is now six guided steps

Tool → team → model → install → connect → verify. Earlier steps collapse to a one-line summary that is a button back to them; a later step cannot be opened before the choices it is generated from exist.

Step 1 is a comparison table, not two marketing lists: one CSS grid, `role="row"` rows laid out with `display: contents`, so the dimension label and both values share a grid row and stay aligned however the text wraps. Below 640px the label spans the row above the pair it labels, so the two values remain side by side.

### Claude Code is set up through a `claude-logos` wrapper

`logos-ui/public/claude-logos.sh` (`.ps1` for Windows), served at `<logos-url>/claude-logos.sh` and installed by a single command the page generates. It goes on the PATH, passes every argument straight through to `claude`, and exports the Logos endpoint, credential and window **into its own child process only**.

No shell profile, no global environment variable, no change to `~/.claude/settings.json`. Plain `claude` keeps using the Anthropic subscription; `claude-logos` goes to Logos. Both, side by side.

```
claude-logos --check       connection, model and how much room a session would get
claude-logos --uninstall   removes the wrapper, its config and its key
claude-logos --help        this, then claude's own help
```

`--uninstall` also offers to clean up an older Logos `env` block this page may have left in `~/.claude/settings.json` — but only when it really is the Logos block, and only after asking.

At startup the wrapper also, using the model listing it already fetched:

- **Warms the model up** via `POST /v1/models/{model}/warmup`, so the cold load overlaps with the seconds a developer spends reading the startup line. It records the same latent demand the scheduler records for a model classification preferred but did not get, and wakes the planner cycle early. A hint, not a reservation: the planner keeps its own fairness rules, a warm-up cannot evict a lane real traffic is using, and no inference request is ever sent on the caller's behalf. Warming a model the key cannot access is a 404.
- **Names models that became available** since the last run.

### The context window is asked for at every start

A window is a property of the lane serving a model, not of the model: the planner sizes a lane's KV cache from the VRAM free on its node, so the same model runs at 262,144 tokens on one worker and a fraction of that on another. Three changes make asking worthwhile:

- **The API reports three figures** — `max_model_len_current_min` (smallest being served; holds whichever deployment answers), `max_model_len_current_max` (largest being served now) and `max_model_len_overall` (the widest it is ever served with, known even while nothing is loaded). `max_model_len` stays as an alias of the first, so an OpenAI-compatible client that already reads vLLM's field keeps working.
- **Requests are routed to a deployment whose window fits them**, using the same 3000-token margin Claude Code keeps between its hard stop and the limit it was told. A deployment with an unknown window is never treated as narrow, and when nothing fits the widest is used rather than returning an empty candidate list — so no request turns into a synthetic 404.
- **A worker can refuse to host a model below a share of its context length**, set per model in its own `config.yml` (`min_context_fraction`), since the hardware that decides which windows are reachable is what should decide the floor. Enforced both in the planner's proposal *and* in the pair chosen at load time, so contention, eviction-backed and request-time cold loads cannot quietly place a below-floor lane either.

```yaml
logos:
  capabilities_models:
    - model: Qwen/Qwen3.8-27B
      min_context_fraction: 1.0    # full context or not here at all
```

OpenCode, which reads its config once at startup, now gets `max_model_len_overall` rather than a number that goes stale.

### Where the "60%" actually came from

Claude Code takes `CLAUDE_CODE_MAX_CONTEXT_TOKENS`, subtracts `min(CLAUDE_CODE_MAX_OUTPUT_TOKENS, 20000)`, and compacts 13,000 tokens below that. Two fixed deductions, 33,000 tokens total — not a percentage. `CLAUDE_AUTOCOMPACT_PCT_OVERRIDE` exists but is clamped by `min(window × pct, window − 13000)`, so it can only compact *earlier*. The window is the only lever, which is what the three changes above are for.

Consequences, both applied: never subtract the output reservation yourself, and never ask for more than 20,000 output tokens (the reservation is capped there regardless).

## New endpoints / fields

- `POST /v1/models/{model}/warmup` → 202 with `{model, status, hint_accepted}` plus the three windows
- `GET /v1/models`, `GET /v1/models/{id}`: `max_model_len_current_min`, `max_model_len_current_max`, `max_model_len_overall` (all omitted when unknown, so cloud models keep the exact object they had)
- `GET /internal/model_context_windows`: new `stats` map alongside the original flat `windows` map, which is unchanged
- `GET /me/keys/{id}/models`: `context_window_current_min`, `context_window_current_max`, `context_window_overall`

## Configuration

Per model in the **worker's** `config.yml` under `logos.capabilities_models`:

| Field | Default | Meaning |
| --- | --- | --- |
| `min_context_fraction` | unset (no floor) | Smallest share of the model's context length a lane here may serve. `1.0` = full context or nothing. Unset or `0` keeps the previous behaviour. |

No new orchestrator environment variables.

## Testing

- `uv run pytest` in `logos-orchestrator`: **733 passed, 1 xfailed** (+49 new across `test_context_budget.py`, `test_context_aware_routing.py`, `test_min_context_placement.py`, `test_warmup_endpoint.py`, plus new cases in `test_v1_models.py` and `test_internal_model_context_windows_endpoint.py`)
- Worker-node suite as CI runs it: **278 passed, 2 skipped**
- `mvn -B test` in `logos-webservice`: **215 passed**, including 5 for the window client (covering the fallback to an older orchestrator that sends no `stats`)
- `npm run build`: clean. `ai-tools.scss` warns at 13.4 kB against the 8 kB budget — under the 16 kB error threshold, and five other feature stylesheets already warn.
- `pre-commit run --files …` over every changed file: clean, including `shellcheck`.
- The bash wrapper end to end against a stub gateway: install, all three window sources, a model with no reported window, the environment actually handed to `claude`, the warm-up call, the new-model notice across three runs (baseline → announced → not announced again), and uninstall — including that it cleans a Logos `env` block but leaves a user's own `ANTHROPIC_BASE_URL` alone.

## Not verified

**`claude-logos.ps1` has not been run.** No PowerShell on this machine. It is a line-by-line counterpart of the bash script and was reviewed for the usual traps, but it needs someone on Windows before the Windows tab can be trusted. The bash wrapper covers WSL and Git Bash meanwhile.

## Follow-up

#778 — drop Ollama support. Its second engine is why the routing here needs an "unknown window" escape hatch at all.

## Docs

`logos/docs/context-windows.md` covers the whole chain — what the API reports, the placement floor, the routing (including where the output reservation and the 3000-token margin come from), both clients, the auto-compact arithmetic, what happens when a session hits the limit, and a troubleshooting table.

https://claude.ai/code/session_01DgaPPdMxyQGjxfaSUeYVbi


logos/docs/context-windows.md
```diff
@@ -0,0 +1,162 @@
...
+
+This document covers the four places that fact shows up.
+
+## 1. What the API reports
+
+`GET /v1/models` (and `/v1/models/{id}`) carry up to three figures per model.
+Each is omitted when unknown, so cloud models and never-calibrated models keep
+the object they had before these fields existed.
+
+| Field                | Meaning                                                                    |
```
wasnertobias: Those filed names are not descriptive at all.

max_model_len_overall -> Models max. capabilities (overall max ever reached!)
max_model_len_current_min -> currently reached min.
max_model_len_current_max -> currently reached max.

logos/docs/context-windows.md
```diff
@@ -0,0 +1,162 @@
...
+
+## 2. Placement floor — don't create a narrow lane at all
+
+Since the narrowest lane defines what every client is told, the planner refuses
+to create one below a configurable share of the model's own context length.
+A node with room for a narrow lane but not a useful one defers instead.
+
+| Variable                                | Default | Meaning                                                        |
+| --------------------------------------- | ------- | -------------------------------------------------------------- |
+| `LOGOS_MIN_CONTEXT_FRACTION`            | `0.5`   | Minimum share of the model's context length a lane must serve. `0` disables the floor (pre-existing behaviour), `1.0` is "full context or nothing". |
```
wasnertobias: no env variable, i want it to reside in logos-workernode config.yml (per model configuration)

logos/docs/context-windows.md
```diff
@@ -0,0 +1,162 @@
...
+to stop the planner *choosing* a narrow window, not to keep uncalibrated models
+off the cluster.
+
+## 3. Context-aware routing — send long requests where they fit
+
+`_prefer_deployments_with_context_room` (`logos/main.py`) estimates what a
+request needs and drops the workers that cannot serve it:
+
+```
+needed = prompt tokens + the reply the request reserved + 3000 tokens of margin
```
wasnertobias: 1.) How would we know what a "request reserves"?

2.) do you also take into account that a lane possibly first of all has to be 3000 tokens of margin to serve a big request?

logos/docs/context-windows.md
```diff
@@ -0,0 +1,162 @@
...
+The 3000 is the same margin Claude Code keeps between its own hard stop and the
+limit it was told; it absorbs the difference between the estimate and what the
+worker's tokenizer really counts. The estimate itself (`logos/context_budget.py`)
+counts characters and divides by 3, skipping base64 attachments, and rounds
+against itself at every step — overestimating costs a roomier worker,
+underestimating costs a 400.
+
+Two deliberate escape hatches:
+
+- **A worker whose window is unknown is always kept.** Cloud providers, Ollama
```
wasnertobias: We can entirely drop ollama support, do not mention it here and create a issue to drop it fully

logos/docs/context-windows.md
```diff
@@ -0,0 +1,162 @@
...
+Audio uploads are left alone: a transcription hint is a few words and says
+nothing about how much context the request needs.
+
+## 4. Clients
+
+### Claude Code — the `claude-logos` wrapper
+
+`logos-ui/public/claude-logos.sh` (and `.ps1` for Windows) is served at
+`<logos-url>/claude-logos.sh` and installed by the AI Tools page. It asks
+`GET /v1/models` at every start, prints the window it got, and exports the
```
wasnertobias: Please also add functionality to notify users, when a new model is available for them! ;)

logos/docs/context-windows.md
```diff
@@ -0,0 +1,162 @@
...
+
+One caveat the wrapper warns about: a model id starting with `claude-` or
+containing `[1m]` is resolved to one of Claude Code's own models, and
+`CLAUDE_CODE_MAX_CONTEXT_TOKENS` is ignored for it. `DISABLE_COMPACT=1` forces
+the window through, at the cost of auto-compaction.
+
+Useful commands:
+
+```
+claude-logos --logos-context     what window this session would get, no request made
```
wasnertobias: rename to --check

logos/docs/context-windows.md
```diff
@@ -0,0 +1,162 @@
...
+claude-logos --logos-check       plus one real request against the gateway
+claude-logos --logos-uninstall   remove the wrapper, its config and its key
```
wasnertobias: why not use simply "claude-logos --uninstall"

logos/docs/context-windows.md
```diff
@@ -0,0 +1,162 @@
...
+One caveat the wrapper warns about: a model id starting with `claude-` or
+containing `[1m]` is resolved to one of Claude Code's own models, and
+`CLAUDE_CODE_MAX_CONTEXT_TOKENS` is ignored for it. `DISABLE_COMPACT=1` forces
+the window through, at the cost of auto-compaction.
+
+Useful commands:
+
+```
+claude-logos --logos-context     what window this session would get, no request made
+claude-logos --logos-check       plus one real request against the gateway
```
wasnertobias: We dont need this advanced check, remote it

logos/docs/context-windows.md
wasnertobias: Also it shall automatically instantly trigger a warmup, as it is very likely to be used in a few seconds when a new claude session is started. (just implement a new rest api for that - do NOT send a test message as we might want to handle this internally different).

logos/logos-orchestrator/src/logos/context_budget.py
```diff
@@ -0,0 +1,135 @@
...
+# code; 3.0 keeps the estimate above the true count for the mixed text
+# (identifiers, JSON, diffs) coding assistants actually send.
+CHARS_PER_TOKEN = 3.0
+
+# What a client is assumed to reserve for its own output when the payload does
+# not say. 20000 is what Claude Code reserves — it caps its reservation there
+# regardless of how large a max_tokens it was configured with — and it is the
+# largest default among the clients Logos serves, so it is the conservative
+# choice for the ones that stay silent.
+DEFAULT_OUTPUT_RESERVE_TOKENS = 20_000
```
wasnertobias, line 36: what exactly happens when this limit is reached in claude code?

logos/logos-ui/public/claude-logos.sh
```diff
@@ -0,0 +1,546 @@
...
+#
+#   claude-logos                      interactive session
+#   claude-logos -p "..."             headless / one-shot
+#   claude-logos --resume             every claude flag is passed through unchanged
+#
+#   claude-logos --logos-check        check the connection and the model, then exit
+#   claude-logos --logos-context      show how much context this session would get
+#   claude-logos --logos-install      install to ~/.local/bin (reads config from stdin)
+#   claude-logos --logos-uninstall    remove the wrapper, its config and its key
+#   claude-logos --logos-help         this text
```
wasnertobias: rather use --help not --logos-help

logos/docs/context-windows.md
wasnertobias: Done — `max_model_len_current_min`, `max_model_len_current_max`, `max_model_len_overall`.

`max_model_len` stays as an alias of `current_min`. It is the field vLLM itself uses, so anything OpenAI-compatible that already reads it keeps working; the descriptive names are the canonical ones. Same rename through `stats`, the Java DTO (`context_window_current_min` / `_max` / `_overall`) and the UI.

logos/docs/context-windows.md
wasnertobias: Moved. `min_context_fraction` per model under `logos.capabilities_models` in the worker's config.yml:

```yaml
logos:
  capabilities_models:
    - model: Qwen/Qwen3.8-27B
      min_context_fraction: 1.0
```

It rides along on the model profile in the runtime snapshot, so the server picks it up without a restart of its own. `LOGOS_MIN_CONTEXT_FRACTION` and the overrides map are gone.

You were right that this belongs on the worker: the hardware that decides which windows are reachable is the same thing that should decide the floor. Documented in `config.example.yml` next to `kv_cache_memory_bytes`.

logos/docs/context-windows.md
wasnertobias: Removed from the doc, and opened #778 for dropping the engine.

~213 references: `ollama_provider.py` (527), `ollama_facade.py` (261), `ollama_process.py` (490), the `provider_type == "ollama"` branches, the `ollama_provider_snapshot` table and its repository, plus a migration decision for any provider row still typed `ollama`. Flagged as a prerequisite to check prod/dev for such rows first — that decides whether the migration converts or refuses.

It also simplifies the routing in this PR: the "unknown window" escape hatch exists partly because Ollama lanes report a configured `context_length` rather than a served window.

logos/docs/context-windows.md
wasnertobias: Done. `--logos-context` is now `--check` — and per your other comment it also absorbed what `--logos-check` did, minus the real request.

logos/docs/context-windows.md
wasnertobias: Done — `--uninstall`, `--install`, `--check`, `--help`. The `--logos-` prefix is gone everywhere.

logos/docs/context-windows.md
wasnertobias: Removed. `--check` now prints the connection, the model, the context arithmetic, the key and the effort — and sends nothing.

Reading the model list already proves the URL, the key and that Logos serves the model, so the only thing the prompt added was a GPU load to tell us what we already knew.

logos/logos-ui/public/claude-logos.sh
wasnertobias: Done. One wrinkle worth naming: `claude --help` is a real command, and the wrapper passes everything else through unchanged — so intercepting `--help` would take it away.

So `claude-logos --help` prints the wrapper's own help and then execs `claude --help`. You get both halves, which is what someone typing it is after.

logos/docs/context-windows.md
wasnertobias: **1) How do we know what a request reserves?** It says so. `max_tokens` (Anthropic Messages, chat completions), `max_completion_tokens` or `max_output_tokens` (Responses API) — whichever is present in the payload. A request that names none is assumed to reserve 20,000, because an uncapped request can generate until it hits the window and 20,000 is the largest default among the clients we serve.

This matters because vLLM charges input and output against one budget: a prompt that fits on its own can still overflow once the reply it asked for is reserved. There is a test for exactly that case — the same ~11k-token prompt fits a 33k lane with a 4k reply and does not with a 20k one.

**2) Does a lane have to have 3000 tokens of margin *first*?** No — the margin is inside `needed`, not on top of the lane. A lane serving 33,000 is asked to satisfy `prompt + output + 3000 ≤ 33000`, so it never has to "make room" for the margin; it is just expected to have 3000 tokens more than the request strictly needs.

The 3000 is the same margin Claude Code keeps between its own hard stop and the limit it was told, so a session Claude Code considers safe is one this filter also considers safe. It exists because our estimate counts characters and divides by 3 — it cannot know the model's tokenizer, so it needs slack in the direction of caution.

Both spelled out in §3 of the doc now.

logos/logos-orchestrator/src/logos/context_budget.py
wasnertobias: **Nothing breaks — it degrades in two steps, neither of which the user has to recover from.**

At `window − reserve − 13000` Claude Code compacts the conversation by itself and carries on. If a single turn still grows past `window − reserve − 3000` it refuses to send and asks for a `/compact` instead of trying.

The failure mode that replaces: before this PR the numbers were wrong in the *unsafe* direction only when the output reservation was under-counted — otherwise they were wrong in the wasteful direction (compacting at 37,240 on a 111,200 window). What we are avoiding is a 400 from vLLM mid-turn, which is not a graceful stop.

`DEFAULT_OUTPUT_RESERVE_TOKENS = 20000` specifically is Claude Code's own cap: it never reserves more than that no matter how large a `CLAUDE_CODE_MAX_OUTPUT_TOKENS` it is given. So for a request that names no `max_tokens`, 20,000 is not a guess — it is what the client would have reserved. Added to §4 of the doc.

logos/docs/context-windows.md
wasnertobias: Both done.

**New models:** the wrapper already fetches the model list at startup for the window, so noticing an addition is one file comparison. New ids are named on startup:

```
new      : mistralai/Mistral-Large now available to you
           (set LOGOS_MODEL=<id> to use one, or re-run the setup on the Logos web UI)
```

Baseline in `~/.config/claude-logos/known-models`. First run records it silently rather than announcing every model the team already had as new. Tested across three runs (baseline → addition announced → not announced again).

**Warm-up:** new `POST /v1/models/{model}/warmup`, and no test message — as you asked, Logos decides internally what to do with it.

It records the same latent demand the scheduler records when classification preferred a model it did not get (weight 0.5, decays per cycle) and wakes the planner cycle early. So it is a hint, not a reservation: the planner keeps its own fairness rules, a warm-up can never evict a lane real traffic is using, and a burst of them coalesces into one extra cycle. That is also what keeps it from being a lever to make the cluster thrash — the most an authenticated caller can do is nudge a model it already has access to up the list of things worth loading. Warming a model the key cannot access is a 404.

Returns 202 immediately with `{"model", "status": "serving"|"preparing", "hint_accepted"}` plus the three windows.

PR #781, opened 2026-08-25
`Logos`: Remove models.parallel column, rely on live vLLM lane signals

### LINKED ISSUE #650: `Logos`: Set default Parallelism of models to sane value
Current default is 1, which is too low.

wasnertobias:
Closes #650.

The static `models.parallel` database entry is gone for good — a model's parallel capacity now comes from the live signals the vLLM worker reports per lane:

1. **Gate on worker-reported capacity** — the orchestrator refuses to forward once a worker's reported limit (KV-budget-derived max concurrency, parsed from vLLM's startup log into the lane's `num_parallel`) is reached, so requests queue at orchestrator level and go to whichever worker frees a slot first instead of piling up in the vLLM-side queue. The ×3 oversubscription of the worker-reported number is removed, and the orchestrator's DB-ceiling plumbing (registration payload, facade collection, provider ceiling logic) goes with it.
2. **Drop the DB entry** — `models.parallel` column removed (Liquibase `dropColumn` guarded by `columnExists`; `000` untouched so production checksums stay stable), and with it the entity/DTO/native-query/service surface plus the UI Parallelism field. The candidate tuple's `parallel` element — which was only ever consumed through the ceiling — is removed as well.
3. **Show it in the statistics UI** — the lane health panel's "×N" badge (which read like a replication factor and hid the useful number) is now the Running line `a / b (min. c)`: `a` = requests running right now (vLLM `num_requests_running`), `b` = capacity floating with the live KV headroom, `c` = the guaranteed minimum, i.e. the worker-reported KV budget at full context.

Kept on purpose: `providers.parallel_capacity` (explicit per-provider admin override, the "config" source), the fallbacks (unreported vLLM lane → 256, no runtime data → 200), and the live signals themselves.

Note: migration and code ship together (same compose stack); the orchestrator no longer selects the column.

### DIFF
--- a/logos/logos-orchestrator/src/logos/classification/classification_manager.py
+++ b/logos/logos-orchestrator/src/logos/classification/classification_manager.py
@@ -126,7 +126,6 @@ class ClassificationManager:
                     i["id"],
                     i["classification_weight"].get_weight(),
                     adjusted_policy["priority"],
-                    i["parallel"],
                 )
                 for i in filtered
             ],
--- a/logos/logos-orchestrator/src/logos/classification/classification_manager.py
+++ b/logos/logos-orchestrator/src/logos/classification/classification_manager.py
@@ -60,7 +60,7 @@ class ClassificationManager:
         classifier=None,
         system=None,
         skip_laura: bool = False,
-    ) -> List[Tuple[int, int, int, int]]:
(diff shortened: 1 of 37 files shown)


wasnertobias:
I pref not to have this database entry existing at all, rather use the live signals from vLLM

PR #782, opened 2026-08-25
`Logos`: Implement tests for capabilities fetching

### LINKED ISSUE #766: `Logos`: Implement tests for capabilties fetching
Nothing is covered yet: the get_model_capabilities endpoint (200 + 404), the name-matching logic (pure functions, easy to unit test), and the FK cascade on model delete. Follow the ModelControllerTest pattern — the seeded gpt-4 model exists in the catalog and works as a 200 case.

wasnertobias:
Closes #766.

Adds the missing test coverage for model-capabilities fetching, in three groups:

**1. `get_model_capabilities` endpoint** (`ModelControllerTest`, Testcontainers)
- 200 for the seeded gpt-4 model (5001) with seeded capability row — asserts the snake_case DTO shape the UI's `ModelCapability` interface expects
- 404 for an unknown model id, and for a mixed known+unknown id list (pins the any-unknown-id → 404 semantics)
- 200 + empty map for an existing model without a capability row (gpt-3.5 / 5002)
- 400 when `ids` is missing
- `delete_model` FK cascade: a capability row present before deletion is gone after

**2. Name-matching logic** (new pure unit test `ModelCapabilitiesUpdaterServiceUnitTest`, Mockito)
- `normalizeModelName`: trim/lowercase, provider-path stripping, `-pt`/`-it` suffix strip, unrelated suffixes untouched
- `extractModelName`: null/blank handling, plain keys, last path segment
- `modelNamesMatch`: exact-not-prefix, normalization order, null sides
- `extractAndStoreCapabilities`: flag persistence, missing keys → false, OR-reduction across multiple catalog matches, no save for unknown models, `sample_spec` and non-map catalog entries skipped

**3. FK cascade on model delete** — see endpoint tests above (`fk_model_capabilities_model` ON DELETE CASCADE).

Test infrastructure: `cleanup-configuration.sql` also gained `DELETE FROM model_capabilities;` so the seeded rows don't leak between test methods.

**One minimal production change, called out explicitly:** `get_model_capabilities` previously returned 200 with an empty map for unknown model ids — the issue explicitly requires a 404 case, which did not exist. `ModelService.getModelCapabilities(List)` now throws `IllegalArgumentException` when any requested id is absent from `models`, and the controller maps it to 404 `{"error": "Model not found: <id>"}` using the same local try/catch pattern as `delete_model`/`update_model_info`. The controller is the only production call site of this overload, so nothing else changes behavior. Four package-private `testXxx` delegating wrappers expose the private matching functions, following the existing `ModelWeightService.testInsertModel` idiom.

Verified locally: full `mvn -B test` (JDK 25) → 245/245 pass (baseline on this branch 221/221; +24 = exactly the new tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests for unknown model IDs now return a clear 404 response instead of an unhandled error.
  * Requests containing a mix of valid and unknown model IDs are handled consistently.
  * Model capability data is properly cleared when models are deleted.

* **Tests**
  * Added coverage for model capability retrieval, validation, empty results, invalid requests, and capability synchronization behavior.

### DIFF
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/controller/ModelController.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/controller/ModelController.java
@@ -141,6 +141,10 @@ public class ModelController {
                 .body(Map.of("error", "ids are required"));
         }
 
-        return ResponseEntity.ok(modelService.getModelCapabilities(req.ids()));
+        try {
+            return ResponseEntity.ok(modelService.getModelCapabilities(req.ids()));
+        } catch (IllegalArgumentException e) {
+            return ResponseEntity.status(404).body(Map.of("error", e.getMessage()));
+        }
     }
 }
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ModelCapabilitiesUpdaterService.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ModelCapabilitiesUpdaterService.java
@@ -75,6 +75,25 @@ public class ModelCapabilitiesUpdaterService {
         }
     }
 
+    boolean testExtractAndStoreCapabilities(
+            Map<String, Object> catalog,
+            int modelId,
+            String modelName) {
+        return extractAndStoreCapabilities(catalog, modelId, modelName);
+    }
+
+    String testExtractModelName(String catalogKey) {
+        return extractModelName(catalogKey);
+    }
+
+    String testNormalizeModelName(String modelName) {
+        return normalizeModelName(modelName);
+    }
+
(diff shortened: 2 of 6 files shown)


Claudia-Anthropica (dismissed):
@wasnertobias The new tests cover capability lookup success and error semantics, missing capability rows, deletion cascading, and the matching/persistence logic. The production existence check is consistent with those semantics, and all captured checks succeeded on this exact head.

Claudia-Anthropica (dismissed):
@wasnertobias The exact-head implementation correctly validates requested model IDs while preserving the intended empty result for existing models without capability data. The added controller and unit tests cover the changed endpoint semantics, cascade behavior, normalization, matching, and persistence paths; captured tests and lint succeeded, with no required checks pending.

Claudia-Anthropica (approved):
@wasnertobias The exact-head implementation correctly returns 404 when any requested model ID is unknown while preserving empty capability results for existing models without capability rows. The integration and unit tests cover endpoint semantics, cascade deletion, normalization, matching, and persistence behavior, and all captured checks succeeded.

PR #783, opened 2026-08-25
`Logos`: Delete stale capability rows on rename and add manual capability override

### LINKED ISSUE #768: `Logos`: Reset model capability flags after renaming
Flags are only overwritten on a catalog hit — renaming a model to an unknown name keeps the old flags displayed. Reset to false (or delete the row) on no-match.

wasnertobias:
Closes #768.

**Builds on #782** (this branch is stacked on its head so the no-match unit tests from #782 could be updated in place). After #782 merges I'll rebase onto main — the stacked commit is dropped automatically and this PR's diff will contain exactly the changes below.

Two changes:

## 1. No-match deletes the stale capability row

Renaming a model to a name the local catalog doesn't know kept the old capability flags displayed: `extractAndStoreCapabilities` only overwrote the row on a catalog hit and silently did nothing on no-match. Now the **no-match branch deletes the capability row**, so the UI shows "capabilities unknown" instead of stale flags.

Why delete rather than reset-to-false (the issue leaves both open): logos-ui renders `-` for both states (missing key → `@else` branch in `models.html`, all-false → the explicit `-` branch), so they are visually identical — and deletion is the semantically honest state. An all-false row already occurs naturally for *known* catalog models that set no flags; no row now unambiguously means "not in the catalog".

All three write paths funnel through `extractAndStoreCapabilities`, so one fix covers them: the `@Scheduled` daily refresh, the async sync on `add_model`, and the async sync on the `update_model_info` rename. A failed/empty catalog load still bails out before any delete, so a catalog outage can never wipe rows. Blank-name rename normalizes to `""`, never matches, and deletes the row — accepted, no new validation added.

## 2. Manual capability override for Logos Admin (persistent against catalog updates)

Per the issue discussion: *"Add the possibility to manually override capabilities for Logos Admin. Those changes shall stay persistent, so even when the upstream LiteLLM files get updated the overrides shall not be undone."*

- **Liquibase 014**: `model_capabilities.manual_override BOOLEAN NOT NULL DEFAULT FALSE` (row-level; existing rows stay automatic).
- **Guard at the top of `extractAndStoreCapabilities`**: while `manual_override` is set, the catalog sync returns before it does anything — no overwrite on a catalog hit, no delete on no-match. Because all write paths (daily `@Scheduled` refresh, async sync on add/rename) funnel through this method, one guard makes overrides persistent against every future LiteLLM catalog update.
- **Endpoints** (`LOGOS_ADMIN`): `POST /logosdb/set_model_capabilities` (sets the three flags + marks the row manual, creates the row if missing) and `POST /logosdb/reset_model_capabilities` (clears the flag, then re-syncs **synchronously** from the catalog so the response already reflects the catalog state). `updateCapabilitiesForModel` is now synchronous with a thin `@Async` wrapper, which enables that.
- **`ModelCapabilitiesDTO`** gains `manualOverride` (serialized `manual_override`), so the UI knows which rows are frozen.
- **UI** (models page): capability checkboxes in the edit dialog; a lock-icon badge in the capability cell for manually overridden rows (the "no capabilities" placeholder is suppressed there — an all-false override is intentional, not a missing row); a "Reset to automatic" button; on submit, a changed override is persisted **before** `update_model_info`, so the rename's async re-sync sees `manual_override=true` and skips the row.

## Tests (256/256 green locally, `mvn -B test` on JDK 25)

Part 1 (unchanged from the first version of this PR):
- The three no-match unit tests from #782 now assert the new behavior with a pre-existing row: no `save`, **and** `delete(existing)`.
- New unit test `extractAndStore_unknownModelWithoutRowIsNoOp`: no row → no save, no delete, no error.
- New endpoint test `updateModelInfo_renameToUnknownNameDeletesCapabilitiesRow` (Testcontainers): seeds a capability row for gpt-4 (5001), renames it to a non-catalog name via `update_model_info`, polls until the row is gone (the rename sync is `@Async`, so the test drains the in-flight task — no new dependency, awaitility is not in the pom).
- The pre-existing `updateModelInfo_updatesNameField` test gets the same drain. `ModelControllerTest` passed 6 consecutive local runs.

Part 2 (9 new tests):
- `scheduledRefresh_keepsManualOverride` — the core persistence requirement: set an override, run `updateAllModelCapabilities()` (the scheduled daily refresh) directly, assert the row is untouched (values and flag).
- `resetModelCapabilities_clearsOverrideAndResyncsFromCatalog` — override differing from the catalog → reset → response and DB show the catalog values with `manual_override=false`.
- `setModelCapabilities_*` — creates row + marks manual, overwrites an existing row, unknown model → 404, missing fields → 400.
- Unit tests `extractAndStore_manualOverrideIsNotOverwrittenOnCatalogMatch` (catalog hit: no save) and `extractAndStore_manualOverrideSurvivesCatalogNoMatch` (catalog miss: no delete) — the guard sits before both write paths.
- Pre-existing capabilities assertions updated for the new `manual_override` field.

## Summary by CodeRabbit

- **New Features**
  - Administrators can view and manually configure model capabilities for function calling, vision, and reasoning.
  - Added a “Reset to automatic” option to resume catalog-based capability updates.
  - Model listings now indicate when capabilities have been manually overridden.
  - Updated model editing responses to display refreshed capability information.

- **Bug Fixes**
  - Improved synchronization when models are renamed, removed, or missing from the catalog.
  - Manual capability settings are preserved during automatic catalog refreshes.
  - Added clearer validation and error handling for invalid or unknown model requests.

### DIFF
--- a/logos/logos-ui/src/app/features/models/models.ts
+++ b/logos/logos-ui/src/app/features/models/models.ts
@@ -9,7 +9,11 @@ import {
 import { FormsModule } from '@angular/forms';
 import { ModalFormComponent } from '../../shared/components/modal/modal-form/modal-form';
 import { ModalConfirmComponent } from '../../shared/components/modal/modal-confirm/modal-confirm';
-import { ModelManagementService, ModelCapability } from '../../core/services/model-management.service';
+import {
+  ModelManagementService,
+  ModelCapability,
+  ModelCapabilityState,
+} from '../../core/services/model-management.service';
 import { Model, AddModelPayload, UpdateModelPayload } from '../../shared/models/model.model';
 import { SearchInputComponent } from '../../shared/components/search-input/search-input';
 import { DataTableComponent } from '../../shared/components/data-table/data-table';
@@ -72,6 +76,9 @@ export class Models implements OnInit {
   editWtAccuracy = signal('');
   editWtCost = signal('');
   editWtQuality = signal('');
+  editCapFunctionCalling = signal(false);
+  editCapVision = signal(false);
+  editCapReasoning = signal(false);
   editLoading = signal(false);
   editError = signal('');
 
@@ -88,6 +95,13 @@ export class Models implements OnInit {
   });
 
   addValid = computed(() => this.addName().trim().length > 0);
+
+  /** True while the model being edited has a manual capability override. */
+  editTargetManualOverride = computed(() => {
+    const target = this.editTarget();
+    if (!target) return false;
+    return this.getCapabilities(target.id)?.manual_override ?? false;
+  });
 
   ngOnInit(): void {
     this.fetchModels();
@@ -217,6 +231,10 @@ export class Models implements OnInit {
     this.editWtAccuracy.set(model.weight_accuracy != null ? String(model.weight_accuracy) : '');
     this.editWtCost.set(model.weight_cost != null ? String(model.weight_cost) : '');
     this.editWtQuality.set(model.weight_quality != null ? String(model.weight_quality) : '');
+    const caps = this.getCapabilities(model.id);
+    this.editCapFunctionCalling.set(caps?.supports_function_calling ?? false);
+    this.editCapVision.set(caps?.supports_vision ?? false);
+    this.editCapReasoning.set(caps?.supports_reasoning ?? false);
     this.editError.set('');
   }
 
@@ -241,7 +259,24 @@ export class Models implements OnInit {
       weight_cost: this.editWtCost() ? Number(this.editWtCost()) : undefined,
       weight_quality: this.editWtQuality() ? Number(this.editWtQuality()) : undefined,
     };
+    const storedCaps = this.getCapabilities(target.id);
+    const capsChanged =
+      this.editCapFunctionCalling() !== (storedCaps?.supports_function_calling ?? false) ||
+      this.editCapVision() !== (storedCaps?.supports_vision ?? false) ||
+      this.editCapReasoning() !== (storedCaps?.supports_reasoning ?? false);
     try {
+      // Persist a changed capability override BEFORE the model info update:
+      // renaming the model triggers an async catalog re-sync, which must see
+      // manual_override=true and therefore skip the row.
+      if (capsChanged) {
+        const caps = await this.modelService.setModelCapabilities(
+          target.id,
+          this.editCapFunctionCalling(),
+          this.editCapVision(),
+          this.editCapReasoning(),
+        );
+        this.applyCapabilityState(caps);
+      }
       await this.modelService.updateModel(payload);
       this.models.update((list) =>
         list.map((m) =>
@@ -266,5 +301,55 @@ export class Models implements OnInit {
     } finally {
       this.editLoading.set(false);
     }
+  }
+
+  // Checkbox change handlers (the [checked] binding and the toggle keep the
+  // signal in sync with the input; same idiom as the api-key permission rows).
+  toggleCapFunctionCalling(): void {
+    this.editCapFunctionCalling.update((v) => !v);
+  }
+
+  toggleCapVision(): void {
+    this.editCapVision.update((v) => !v);
+  }
+
+  toggleCapReasoning(): void {
+    this.editCapReasoning.update((v) => !v);
+  }
+
+  /** Clear the manual capability override and re-sync from the catalog. */
+  async resetCapabilities(): Promise<void> {
+    const target = this.editTarget();
+    if (!target || this.editLoading()) return;
+    this.editLoading.set(true);
+    this.editError.set('');
+    try {
+      const state = await this.modelService.resetModelCapabilities(target.id);
+      this.applyCapabilityState(state);
+      // Mirror the re-synced catalog values into the dialog; with
+      // manual_override now false the badge and the reset button disappear.
+      this.editCapFunctionCalling.set(state.supports_function_calling);
+      this.editCapVision.set(state.supports_vision);
+      this.editCapReasoning.set(state.supports_reasoning);
+    } catch {
+      this.editError.set('Failed to reset capabilities, please try again.');
+    } finally {
+      this.editLoading.set(false);
+    }
+  }
+
+  /** Merge a set/reset state map into the local capabilities record. */
+  private applyCapabilityState(state: ModelCapabilityState): void {
+    this.capabilities.update((all) => ({
+      ...all,
+      [state.model_id]: {
+        id: all[state.model_id]?.id ?? 0,
+        model_id: state.model_id,
+        supports_function_calling: state.supports_function_calling,
+        supports_vision: state.supports_vision,
+        supports_reasoning: state.supports_reasoning,
+        manual_override: state.manual_override,
+      },
+    }));
   }
 }
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/controller/ModelController.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/controller/ModelController.java
@@ -15,6 +15,8 @@ import de.tum.cit.aet.logos.logoswebservice.configuration.dto.AddModelRequestDTO
 import de.tum.cit.aet.logos.logoswebservice.configuration.dto.DeleteModelRequestDTO;
 import de.tum.cit.aet.logos.logoswebservice.configuration.dto.GetModelCapabilitiesRequestDTO;
 import de.tum.cit.aet.logos.logoswebservice.configuration.dto.GetModelRequestDTO;
+import de.tum.cit.aet.logos.logoswebservice.configuration.dto.ResetModelCapabilitiesRequestDTO;
+import de.tum.cit.aet.logos.logoswebservice.configuration.dto.SetModelCapabilitiesRequestDTO;
 import de.tum.cit.aet.logos.logoswebservice.configuration.dto.UpdateModelRequestDTO;
 import de.tum.cit.aet.logos.logoswebservice.configuration.dto.UpdateModelWeightRequestDTO;
 import de.tum.cit.aet.logos.logoswebservice.configuration.service.ModelService;
@@ -143,6 +145,43 @@ public class ModelController {
 
         try {
             return ResponseEntity.ok(modelService.getModelCapabilities(req.ids()));
+        } catch (IllegalArgumentException e) {
+            return ResponseEntity.status(404).body(Map.of("error", e.getMessage()));
+        }
+    }
+
+    @PostMapping("/set_model_capabilities")
+    @PreAuthorize("hasAuthority('" + Role.Names.LOGOS_ADMIN + "')")
+    public ResponseEntity<?> setModelCapabilities(
+            @RequestBody SetModelCapabilitiesRequestDTO req) {
+        if (req.modelId() == null
+                || req.supportsFunctionCalling() == null
+                || req.supportsVision() == null
+                || req.supportsReasoning() == null) {
+            return ResponseEntity.badRequest().body(Map.of(
+                "error", "model_id, supports_function_calling, supports_vision, and supports_reasoning are required"));
+        }
+        try {
+            return ResponseEntity.ok(modelService.setModelCapabilities(
+                req.modelId(),
+                req.supportsFunctionCalling(),
+                req.supportsVision(),
+                req.supportsReasoning()
+            ));
+        } catch (IllegalArgumentException e) {
+            return ResponseEntity.status(404).body(Map.of("error", e.getMessage()));
+        }
+    }
+
+    @PostMapping("/reset_model_capabilities")
+    @PreAuthorize("hasAuthority('" + Role.Names.LOGOS_ADMIN + "')")
+    public ResponseEntity<?> resetModelCapabilities(
+            @RequestBody ResetModelCapabilitiesRequestDTO req) {
+        if (req.modelId() == null) {
+            return ResponseEntity.badRequest().body(Map.of("error", "model_id is required"));
+        }
+        try {
+            return ResponseEntity.ok(modelService.resetModelCapabilities(req.modelId()));
         } catch (IllegalArgumentException e) {
             return ResponseEntity.status(404).body(Map.of("error", e.getMessage()));
         }
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/controller/ModelController.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/controller/ModelController.java
@@ -141,6 +141,10 @@ public class ModelController {
                 .body(Map.of("error", "ids are required"));
         }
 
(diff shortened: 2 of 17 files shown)


logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ModelCapabilitiesUpdaterService.java
```diff
@@ -66,6 +66,10 @@ public void updateAllModelCapabilities() {
 
     @Async
     public void updateCapabilitiesForModelAsync(int modelId, String modelName) {
+        updateCapabilitiesForModel(modelId, modelName);
```
Claudia-Anthropica, line 69: @wasnertobias [medium] This asynchronous job trusts the captured `modelName` without verifying that it is still current. Two quick renames can finish out of order, allowing the older job to overwrite or delete capabilities produced for the latest name. Make the capability mutation conditional on the model's persisted name still matching this job's name within the same transaction.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ModelCapabilitiesUpdaterService.java`, an asynchronous rename job can mutate capabilities using a superseded model name. Verify the persisted model name and perform the capability mutation atomically only when it still matches the job's captured name.

</details>

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ModelCapabilitiesUpdaterService.java
```diff
@@ -75,11 +79,34 @@ public void updateCapabilitiesForModelAsync(int modelId, String modelName) {
...
+    boolean testModelNamesMatch(String requestedModelName, String catalogModelName) {
+        return modelNamesMatch(requestedModelName, catalogModelName);
+    }
+
     @SuppressWarnings("unchecked")
     private boolean extractAndStoreCapabilities(
             Map<String, Object> catalog,
             int modelId,
             String modelName) {
+        if (modelCapabilitiesPersistenceService.isManualOverride(modelId)) {
```
Claudia-Anthropica, line 110: @wasnertobias [medium] The manual-override check commits independently from the later update/delete. A catalog sync can read `false`, an admin can then save an override, and the sync can subsequently overwrite its flags while leaving it locked—or delete the row entirely. Serialize manual and automatic mutations per model and perform this check plus the final write/delete in one transaction.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ModelCapabilitiesUpdaterService.java`, the manual-override check is not atomic with the automatic capability mutation. Lock the model for both manual and automatic mutations and recheck `manual_override` in the same transaction immediately before updating or deleting the capability row.

</details>

logos/logos-ui/src/app/features/models/models.ts
```diff
@@ -241,7 +259,24 @@ export class Models implements OnInit {
       weight_cost: this.editWtCost() ? Number(this.editWtCost()) : undefined,
       weight_quality: this.editWtQuality() ? Number(this.editWtQuality()) : undefined,
     };
+    const storedCaps = this.getCapabilities(target.id);
+    const capsChanged =
+      this.editCapFunctionCalling() !== (storedCaps?.supports_function_calling ?? false) ||
+      this.editCapVision() !== (storedCaps?.supports_vision ?? false) ||
+      this.editCapReasoning() !== (storedCaps?.supports_reasoning ?? false);
     try {
+      // Persist a changed capability override BEFORE the model info update:
```
Claudia-Anthropica, line 333: @wasnertobias [medium] When a model is renamed without changing these checkboxes, the backend queues an automatic resync but this component updates only `models`; its `capabilities` signal retains the old row indefinitely. Thus renaming to an unknown model still displays the stale capability chips until a full reload. Have the rename endpoint complete and return the refreshed capability state, then apply or remove that state locally before closing the dialog.

In `logos/logos-ui/src/app/features/models/models.ts`, a rename without manual capability changes leaves the local capability map stale after the backend resync. Make the rename operation return its completed capability state and update or remove the corresponding signal entry before closing the edit dialog.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The two retained review issues remain on this head: the single-model updater still processes an empty catalog, and bulk capability lookup still does not reject null IDs with HTTP 400. The new async/manual paths also permit stale rename jobs and concurrent catalog synchronization to corrupt the current capability state, while the UI retains pre-rename capability flags. Captured CI is successful, but these correctness and data-integrity issues require changes.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/controller/ModelController.java
```diff
@@ -141,6 +141,10 @@ public ResponseEntity<?> getModelCapabilities(
                 .body(Map.of("error", "ids are required"));
         }
 
-        return ResponseEntity.ok(modelService.getModelCapabilities(req.ids()));
+        try {
+            return ResponseEntity.ok(modelService.getModelCapabilities(req.ids()));
+        } catch (IllegalArgumentException e) {
+            return ResponseEntity.status(404).body(Map.of("error", e.getMessage()));
+        }
```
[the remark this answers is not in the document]
wasnertobias: Addressed. get_model_capabilities now rejects requests whose ids list contains a null element with 400, leaving 404 for valid ids that refer to missing models.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ModelCapabilitiesUpdaterService.java
```diff
@@ -115,10 +134,13 @@ private boolean extractAndStoreCapabilities(
...
         }
         if (!found) {
+            // A renamed model that is unknown to the catalog keeps no stale flags:
+            // drop the row so the UI shows "capabilities unknown" instead.
             log.debug(
-                "capabilities_updater: model '{}' not found in local JSON registry",
+                "capabilities_updater: model '{}' not found in local JSON registry, deleting stored capabilities",
                 modelName
             );
+            modelCapabilitiesPersistenceService.deleteModelCapabilities(modelId);
```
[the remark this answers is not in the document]
wasnertobias: Addressed with the suggested guard: the per-model sync now returns when the catalog is null or empty, matching the daily refresh, so a failed catalog load cannot delete stored rows. Deletion still applies for models genuinely missing from a non-empty catalog.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ModelCapabilitiesUpdaterService.java
wasnertobias: Addressed. Every capability writer now takes a PESSIMISTIC_WRITE lock on the model row first, and the catalog sync re-checks the persisted model name under that lock in the same transaction, so a sync for a superseded name becomes a no-op. A new unit test covers the superseded-name case.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ModelCapabilitiesUpdaterService.java
wasnertobias: Addressed. The manual set/clear paths and the catalog sync now serialize on the model row lock, and the sync re-reads manual_override inside the same transaction as its write/delete, so a concurrent admin override can no longer be overwritten or deleted.

logos/logos-ui/src/app/features/models/models.ts
wasnertobias: Addressed. A rename now re-syncs the capabilities inline, and update_model_info returns the resulting capability state. The models page applies that state to its capabilities signal before closing the dialog, so the chips no longer keep the flags of the previous name.

logos/logos-ui/src/app/features/models/models.ts
```diff
@@ -306,8 +324,28 @@ export class Models implements OnInit {
...
       weight_quality: this.editWtQuality() ? Number(this.editWtQuality()) : undefined,
     };
+    const storedCaps = this.getCapabilities(target.id);
+    const capsChanged =
+      this.editCapFunctionCalling() !== (storedCaps?.supports_function_calling ?? false) ||
+      this.editCapVision() !== (storedCaps?.supports_vision ?? false) ||
+      this.editCapReasoning() !== (storedCaps?.supports_reasoning ?? false);
     try {
-      await this.modelService.updateModel(payload);
+      // Persist a changed capability override BEFORE the model info update:
```
Claudia-Anthropica, line 333: @wasnertobias [medium] `setModelCapabilities` commits before `update_model_info`; if the latter rejects a duplicate name or alias, the UI reports “Failed to save changes,” but the override remains persisted even if the user cancels. Submit both changes through one backend transaction so this form cannot partially save.

In `logos/logos-ui/src/app/features/models/models.ts`, capability changes are committed before the model update and survive when that update fails. Replace the two independent writes with one backend operation that updates the model and capabilities atomically.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The earlier catalog, validation, concurrency, and stale-UI issues are fixed on this head, but the edit flow can still persist a capability override before a model update that later fails, leaving a partial save. The two failing non-required prepare jobs report GitHub HTTP 500 errors in untouched build-selection paths, while the Logos tests, lint, UI export, and webservice build succeeded. Please make the combined edit atomic.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ModelService.java
```diff
@@ -367,12 +381,61 @@ public Map<Integer, ModelCapabilitiesDTO> getModelCapabilities(List<Integer> mod
...
+        // Read the name under the row lock: the re-sync below matches it against
+        // the persisted name, so a rename that lands between the two would
+        // otherwise make the re-sync a no-op.
+        Model model = modelRepository.findByIdForUpdate(modelId)
+            .orElseThrow(() -> new IllegalArgumentException("Model not found: " + modelId));
+        modelCapabilitiesPersistenceService.clearManualOverride(modelId);
+        // Synchronous re-sync so the response (and the UI) reflects the catalog state
+        // immediately; the guard in the updater makes this a no-op only if the override
+        // was not actually cleared.
+        modelCapabilitiesUpdaterService.updateCapabilitiesForModel(modelId, model.getName());
```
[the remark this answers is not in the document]
wasnertobias: Confirmed the mechanism, and the two suggested shapes don't land cleanly here — so I'd like your call before changing the lock design you set up deliberately.

Why neither bot option fits as-is:

- **`REQUIRES_NEW` on the re-sync deadlocks.** `applyCatalogCapabilities` re-takes the same row lock via `findByIdForUpdate` (ModelCapabilitiesPersistenceService:56) that `resetModelCapabilities` holds. A `REQUIRES_NEW` re-sync would run on a second connection and block on that lock while the outer tx waits for it to return.
- **after-commit breaks the response contract.** The whole point of the synchronous re-sync (per the comment) is that the response — and the UI chips — reflect the catalog state immediately. An after-commit callback runs after `resetModelCapabilities` returns, so the returned `capabilitiesState` would show the post-clear (stale) flags, not the re-synced ones.
- And independently, a DB error inside the re-sync aborts the *shared* PG transaction at the database level, so catching it in the updater never actually un-poisons the tx — the reset can't commit after a re-sync DB failure as long as they share a tx.

So the viable fix is to **de-nest**: clear the override in its own transaction, then run the re-sync in its own transaction (each takes the lock sequentially — no nesting, no deadlock, and a re-sync failure no longer 500s the reset). The trade-off is that the row lock is no longer held across clear→re-sync, which re-introduces the narrow rename race your comment was guarding against: a rename landing between the two transactions makes the re-sync a no-op (name mismatch in `applyCatalogCapabilities`), leaving the cleared override with stale flags until the next catalog refresh.

My lean: accept that race — a stale-flags-until-next-refresh is a milder outcome than a reset that reports 500 after it already cleared the override. But it's your design, so I'll implement the de-nested version if you're OK with the trade-off, or leave it if you'd rather keep the single-lock atomicity and just document the 500-on-re-sync-failure. Your call.

logos/logos-ui/src/app/features/models/models.ts
wasnertobias: Agreed — the override is committed in its own write before `update_model_info`, so a duplicate name/alias rejection leaves it persisted behind a "Failed to save" message. Happy to collapse it into one atomic operation. Concrete shape, so you can approve or adjust:

- **`UpdateModelRequestDTO`**: add optional `Boolean supportsFunctionCalling / supportsVision / supportsReasoning` (null = "no capability change" — distinct from an explicit `false`).
- **`ModelService.updateModelInfo`** (already `@Transactional`): after the model row update, if any flag is non-null, call `setManualCapabilities(...)` in the same transaction. Model + override now commit or roll back together, and a rename's re-sync (which runs after the commit in the controller) already sees `manual_override=true` and skips — so this also removes the need for the separate pre-write.
- **Controller `update_model_info`**: return `capabilities` when a capability change was applied (not only on rename) so the UI can refresh the chips.
- **Frontend `submitEdit`**: drop the standalone `setModelCapabilities` call; include the flags in the `updateModel` payload only when `capsChanged`, and apply `res.capabilities`.

One coupling to flag: this shares the rename re-sync path with the other open thread on the reset-flow transaction, so I'd land both together once you've settled that one — the override ordering is the same mechanism. Say the word and I'll implement this (and the reset de-nesting if you're OK with that trade-off) in one pass.

Claudia-Anthropica (changes requested):
@wasnertobias The proposed combined `update_model_info` design is sound, but this exact head still commits `setModelCapabilities` before the model update, so a rejected update leaves a partial save. Implement the combined transaction and validate the three nullable capability fields as an all-or-none group: all null means unchanged; otherwise all three are required. Captured CI is successful.

logos/logos-ui/src/app/features/models/models.ts
Claudia-Anthropica: @wasnertobias Yes—please implement that shape. One adjustment: validate the capability fields as an all-or-none group (all null means no change; otherwise require all three) before unboxing them, and add a regression test proving duplicate-name or alias rejection rolls back the capability change. The reset de-nesting you outlined is also appropriate: commit the clear first, run the re-sync sequentially, and return the post-sync state.

Claudia-Anthropica (changes requested):
@wasnertobias The unresolved atomic-edit issue remains on this head: `submitEdit` commits `setModelCapabilities` before calling `updateModel`, while `UpdateModelRequestDTO` still cannot carry capability fields. A duplicate-name/alias rejection or later update failure therefore leaves the override committed despite the UI reporting failure; move all three capability values into the transactional model-update operation and add rollback regression coverage. Captured CI is successful, but this medium data-integrity defect still requires changes.

Issue #780, opened 2026-08-25
`Logos`: Improvements to statistics page


wasnertobias:
1.) Lane health in statistics page shall also show max context size e.g. "104k"
2.) Some requests do not have token counts displayed, some have. I want to see live token numbers growing live (stream the token counts live to any viewers of statistics page). For all requests. incl. "20 tok/s" measurement for individual requests.

wasnertobias:
Both items of this issue are already implemented in `main` — they landed in commit `d37b343` (PR #785), merged the same day this issue was filed:

**1. Max context size in lane health**
- The orchestrator derives the context window each lane actually serves in `_lane_served_context_window()` (`logos-orchestrator/src/logos/main.py:5209`) and ships it per lane as `max_model_len` in the lane signals (`main.py:552`) — for vLLM lanes it mirrors the worker's `--max-model-len` precedence (explicit config → lane context length → calibrated profile), for Ollama lanes the configured `context_length`.
- The statistics page's lane health panel renders it abbreviated, right next to the model name (`logos-ui/.../lane-health-panel/lane-health-panel.ts:94`, `lane-health-panel.html:41`) — i.e. exactly the `104k` style requested, e.g. `Llama-3.1-8B  104k ctx`.
- Covered by `tests/unit/main/test_statistics_endpoints.py` (five lane-signal window tests) and `lane-health-panel.spec.ts`.

**2. Live token counts + tok/s for in-flight requests**
- The orchestrator tracks every running stream in `_LiveStreamRegistry` (`main.py:1103`): exact prompt tokens plus a running completion tally from the SSE deltas, and `tokens_per_second` measured from the first token so queueing does not drag the rate down. Both streaming paths register their requests — the LogosNode/WebSocket path (`main.py:3018`) and the cloud/HTTP path (`main.py:3184`) — and entries are dropped in the same `finally` that logs the completion, so nothing lingers.
- `GET /internal/live_streams` (`main.py:2013`) exposes the snapshot behind the internal secret; the webservice merges those figures into the existing `requests` WebSocket push the statistics page already uses (`StatsV2WebSocketHandler.mergeLiveStreams()`, `logos-webservice/.../websocket/StatsV2WebSocketHandler.java:473`) — prompt/completion/total tokens, `tokens_per_second` and a `streaming` flag, refreshed on the page's push cadence, with the settled database usage always winning once a request completes.
[...]

Issue #776, opened 2026-08-25
`Logos`: Gvie app administrators live stats


wasnertobias:
App administrators wish for a live view of requests and used tokens. A simliar view as the stats page does provide today for logos admins is appropriate, although it shall not be as detailed.

Specifically requested is the following (not neccessarily limited to that):
Which/How many requests are currently queued/running/finished? -> Live view
How many tokens did team use in the last x days? (with specific api key)

PR #785, opened 2026-08-25
`Logos`: Fix Load lane routing, context windows, warmup, token counting, and scope statistics by team

wasnertobias:
Bugs reported from a live session, plus #780 and #776. Several of them share one root: a number Logos already knows, thrown away because the code asked for it in the wrong place.

## Load lane 404s — a routing gap, not a missing endpoint

`/api/logosdb/providers/logosnode` is a PathPrefix router pointed at the orchestrator at priority **201**, deliberately outranking the webservice's `/api/logosdb` catch-all at 200 — most of that subtree really is the worker protocol. The JWT-authenticated admin actions are carved back out by an exact-Path router at 202, and that list named `calibrate_uncalibrated` and `lanes/delete` only.

So `lanes/add` fell through to the orchestrator, which does not serve Spring's paths and answers 404 in the OpenAI error shape. Probed against prod without changing it — no JWT, so Spring answers 401 and the orchestrator 404:

```
lanes/add              -> 404   {"error":{"message":"Not Found","type":"not_found_error"}}
lanes/delete           -> 401
calibrate_uncalibrated -> 401
```

An exact-Path list in front of a catch-all fails silently: everything looks wired up end to end and the call is simply routed elsewhere. The comment now says the list has to name every such action.

The report read `[object Object]` because `failureDetail()` took `error` as a string, but every user-facing Logos error is normalised to `{"error": {"message": …}}`. It walks the nesting now, covered by a test using the captured body above.

## Context window falls back to the model maximum

The AI-tools page and both `claude-logos` wrappers keyed *"is anything known about this model"* on `context_window_current_min`, which only exists **while a lane is up**. With nothing loaded they discarded `context_window_overall` — reported all along — and printed a fallback constant: **32,768** in the UI, **111,200** in the wrapper, for a model whose real maximum is 262,144.

The constant is wrong in both directions. On the 40,960-token reranker it had sessions overrun the real cap and come back as 400s. Both now fall back to the maximum and say what they are looking at.

The wrappers also group digits themselves: `printf "%'d"` only groups under a locale that defines a separator, and under `LANG=C` it silently printed `111200`.

## Warmup can actually start a lane

`/warmup` recorded latent demand (weight `0.5`) against `DEMAND_LOAD_FLOOR = 1.0`, and the score is decayed once per cycle **before** the planner evaluates it — so no single hint could ever clear the floor, at any weight. Real traffic sidesteps this via its queue entry; a warmup has none by design.

From the planner log, GPUs at 662 MB of 286.8 GB used:

```
Demand eval    worker=deioma model=Qwen/Qwen3.8-27B score=0.35 lanes=0 queue_here=0
Load candidate model=Qwen/Qwen3.8-27B placement=feasible eviction_needed=no
```

Feasible, free, and then nothing — the `continue` that swallowed it logs nothing. An announcement is its own fact now and passes the floor **only when no eviction is needed**, so it still cannot displace a lane someone is using.

## Statistics: the filter scopes the whole page

It lived in the request feed's toolbar and narrowed that list alone; the KPI cards and every chart above it kept counting the whole platform. It moved to the page header and reaches every request-derived aggregate through a new `set_scope` websocket action. Verified against the prod database:

| Scope | Requests | Tokens |
|---|---:|---:|
| unscoped | 95,793 | 715,136,646 |
| team 5 | 6,961 | 8,144,767 |
| team 11 | 54,124 | 490,936,414 |

The token sum scales with it — that subquery is where a KPI card would otherwise have paired one team's request count with everyone's tokens. **Not** applied to VRAM, lanes or GPUs: those belong to the hardware and to no team.

The dropdowns now list only what a range actually holds, with request counts, and picking a team narrows the requester list to that team's requesters. Over 30 days on prod: **82 entries → 39 unscoped → 1–3 once a team is chosen**. Not a precondition, though — three requesters send under more than one team, so requiring the team first would make "everything user X did" unaskable.

Also fixes the overlap that made this unusable: `<app-select>` is a custom element, so its host defaulted to `display: inline` and a caller's `max-width` bound nothing.

## #780 — lane context, and token counts that were never recorded

**Part 1.** Lane rows named the model but not how much room it has there, which is the more useful of the two once a model is loaded twice — the planner sizes every lane against the KV cache it could get. Shown abbreviated (`262k`), null when nothing supports a figure.

**Part 2.** "Some requests do not have token counts" turned out to be *every Claude Code session*. The stream accumulator understood Chat Completions and the Responses API and pulled usage off the final chunk; an Anthropic Messages stream puts it in two earlier events and then ends on `message_stop`, which carries nothing. Over two days that is 1,143 successful requests logged as having used no tokens, and only 4 of 52 streaming requests in the orchestrator log carried a token line.

Both events are merged now, and the total Anthropic omits is computed — it sends none because it is the sum, but Logos stores one row per token type and the page reads `total_tokens`.

On top of that, live counts: the orchestrator keeps the in-flight view and the webservice folds it into the request push it already sends. Completion counts mid-stream are the delta count and deliberately approximate — the settled figure replaces it on completion and the merge never overwrites a row the database already has a total for. The rate is measured from the first token, so queueing does not drag it down.

## #776 — a team-scoped activity view

The two questions from the issue and no more: what is queued/running/finished right now, and tokens used over the last x days per API key. Access follows the rule the key admin endpoints already use — a Logos admin sees any team, an app admin only a team they own — with the team id in the path so the body cannot widen it.

The in-flight counts stop at a 30-minute horizon. Rows get stranded and never gain a response, so an unbounded count only grows: team 5 alone had 142, every one over a day old, which a live view would have reported as a queue backing up right now.

## Testing

- 764 orchestrator tests, 242 webservice tests, 16 frontend tests, Angular build clean.
- New coverage: the scope reaching each aggregate, dropdown scoping, Anthropic streams (from a capture taken through Logos), the live registry, per-lane windows, and the app-admin authorisation boundary in both directions.
- The routing diagnosis, the scope SQL, the context-window fallback and the stranded-row counts were all exercised against the live instance, as quoted above.
- Not covered: the select overlap is fixed from the CSS mechanism, without a visual check.

https://claude.ai/code/session_01GjqdxreAS7WQmq1eVuotro


wasnertobias:
@coderabbitai full-review

wasnertobias:
I do not like that a extra page "team-activity" has been created for this. Rather within the teams detail view. There e.g., is already a tab called "Billing" which could either be extended (and renamed) or be split up (renamed to "Cloud Usage" and another tab "Local Usage").

The current team-activity view is rendering poorly (it misses padding, as e.g. the .team-management has). And it lacks information: There is no way to see single requests in a list it only gives a count of ongoing requests. It is not filterable by user.

wasnertobias:
All three points addressed in 5dbe540.

**Moved into the team detail view.** It is a tab now — the standalone page, its route and the nav entry are gone. You were right that a separate destination made it something you had to remember existed.

**Split rather than extended**, and `Billing` is renamed to **Cloud Usage**. Money and tokens answer different questions and share no axis: that tab reports what cloud providers charged, which is only part of what a team uses — a local model bills nothing and still consumes the cluster. I named the second one **Activity** rather than "Local Usage", because the live counts and the request list are not local-only; say the word if you'd rather have the symmetry of the two names.

**Padding** — the page carried its own layout and lost it on the way in. The tab now sits in the same shell as its neighbours and lets `.tab-content` own the padding, like `models` and `billing` do.

**The list and the filter.** The counts answered "is anything happening"; the question that follows within seconds is "what". The tab now lists individual requests — model, requester, age, status, duration, tokens — paged, with a requester filter built from the team's own traffic (the same per-scope option query the statistics filter uses, so it can only name someone inside the team).

One thing worth flagging: the filter goes through the team endpoint rather than `latest_requests`, because that one is `logos_admin` only and an app admin cannot reach it. Passing a `user_id` from outside the team narrows the list to nothing rather than reaching past the team check — covered by a test, along with the 403 for a team the admin does not own.

246 webservice tests, 764 orchestrator tests, 16 frontend tests, pre-commit clean.

PR #795, opened 2026-08-25
`Logos`: Balance-wake — wake sleeping replica when the best-first winner is saturated

wasnertobias:
## Problem

Qwen3.8-27B runs on all three workers (deioma, deimama, deipapa). On 2026-08-25,
deimama/deipapa were awake and served the entire request stream — saturated,
with the demand eval at `eff=2.12` and a queue of 2–8 waiting requests on
deimama — while the deioma replica **idle-slept at L1 for ~7h with GPU 2 fully
free**. The cluster ran at 2/3 capacity with nothing to use the standby.

Root cause: the cross-provider best-first ranker picks the cheapest worker per
demanded model. An awake lane costs `0.0s`, a wake costs `2.0s`
(`TARGET_ACTION_COST_S`), so an awake incumbent *always* wins the ranking and
the non-winning provider's demand action is skipped unconditionally
(`"Skipping demand action … best-first ranker picked worker=deimama"` in every
planner cycle). The ranker's cost model answers "which worker is cheapest to
make able to serve" — it never looks at the winner's queue depth. The
request-time scheduler (ETTF) stays with the warm workers as long as either
one's estimated wait is under the sleeper's fixed 2.5s, which with two warm
replicas is rarely true simultaneously. So the standby only wakes on a
failover, never under load.

## Change

New **balance-wake** path in the demand loop (`capacity_planner.py`):

- When this provider is **not** the best-first winner, but
  1. the winner is **awake** for the model,
  2. the winner's combined queue for the model (active + vLLM queue +
     scheduler queue, same metric as `queue_here`) is **≥ `BALANCE_WAKE_QUEUE_FLOOR` (4)**,
  3. this provider has a **sleeping lane** for the model (not in wake-failure
     cooldown),
  → plan a `wake` for the sleeping lane.
- Guardrails (balance-wake is wake-path only):
  - **Never evicts** — if the wake would need another model's lane, it is
    skipped and retried next cycle. A balance wake spreads load for the
    incumbent; nothing is waiting on this lane, so preempting another model
    would be speculative.
  - **No cold load** — a provider without a lane must not cold-load just
    because the winner is busy (that is the replication decision,
    `LOGOS_REPLICATE_ON_FREE_VRAM`).
  - **At most one balance-wake per model per cycle** (`cycle_balance_wake_models`)
    — one extra replica per cycle is the escalation step; the next cycle
    re-evaluates with the updated queues.
- **No feature flag** — balance-wake is unconditional. The guardrails above
  are inherent, so there is no sensible off-switch.

The floor (4) keeps 1–3 request blips from flapping the lane against the 120s
idle tier — at ≥4 concurrent, the next arrival waits a full request duration
(60–120s for long-context completions), which a ~2s wake + free VRAM pays
back.

Self-regulation: after the wake, ETTF routes new requests to the now-warm
replica (tier=warm beats tier=sleeping), the incumbent's queue drains below
the floor, and the 120s idle tier re-sleeps the replica when it goes quiet.
The wake/sleep cycle is bounded by the idle tier — no thrashing.

## Tests

7 new scenario tests in `TestBalanceWake` (`test_best_first_scenarios.py`),
including the production case with the sleeper's *local* demand below the
regular wake floor (only the winner's queue justifies the wake), the queue
floor, the no-eviction rule, infeasible VRAM, the per-cycle cap, no cold
load, and the sleeping-winner case. Orchestrator unit suite:
**771 passed, 1 xfailed** (pre-existing documented gap). Black + flake8 clean.

## Rollout / verification

Deploy the orchestrator image (no worker-node change). In production the
first `Balance-wake:` lines should appear for Qwen3.8-27B on deioma when
deimama's `queue_here` reaches 4; the deioma lane then flips to
`running / awake` and ETTF starts reserving requests for it. Revert by
redeploying the previous orchestrator image.

### DIFF
--- a/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
+++ b/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
@@ -94,6 +94,21 @@ class CapacityPlanner:
     # developer abandoned before sending anything stops holding a lane open.
     ANNOUNCED_USE_TTL_SECONDS = 300.0
 
+    # Balance wake: when the best-first winner is under queue pressure for a
+    # model (>= this many active+queued requests on it), the planner wakes a
+    # sleeping replica on a NON-winning worker (no eviction, no cold load, at
+    # most one per model per cycle). Without it, a hot model with N-1 awake
+    # replicas and one sleeping standby stays at N-1 capacity forever: the
+    # ranker always prefers the 0.0s awake lanes over the 2.0s wake and the
+    # standby just idle-sleeps (observed in production: Qwen3.8-27B on
+    # deimama/deipapa saturated with queue 2-8 while the deioma replica slept
+    # for 7h with GPU 2 fully free).
+    # The floor keeps single-request blips from flapping the lane against the
+    # 120s idle tier: at >=4 concurrent the next arrival would wait a full
+    # request duration (60-120s for long-context completions), which a ~2s
+    # wake + VRAM reservation pays back.
+    BALANCE_WAKE_QUEUE_FLOOR = 4
+
     # Competitive ratios: applied when eviction IS required.
     # target_effective_demand > max(eviction_set_demand) * RATIO to proceed.
     WAKE_COMPETITIVE_RATIO = 1.5  # target must beat eviction set by 50%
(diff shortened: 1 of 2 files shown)


logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -2979,6 +2963,83 @@ def _rank_providers_for_demanded_models(
...
+        cycle_balance_wake_models: Optional[set[str]],
+    ) -> Optional[str]:
+        """Decide whether this NON-winning provider should balance-wake.
+
+        Called from the demand loop when the best-first ranker picked a
+        different (cheaper) worker for the model. Returns the planned
+        action's reason when a balance-wake is warranted, else None.
+
+        Eligibility (all required):
+          - the LOGOS_BALANCE_WAKE flag is on,
```
wasnertobias, line 2981: remove this flag, it does not make sense to turn this off

PR #796, opened 2026-08-25
`Logos`: Add compatibility pre checks

julia-rai:
## Summary by CodeRabbit

* **New Features**
  * Added on-demand compatibility checks for selected or all connected workers.
  * Added Hugging Face metadata checks for memory and tensor-parallel feasibility before calibration.
  * Added cached vLLM quantization-method detection.
  * Added persisted model memory estimates that improve with live measurements.

* **Bug Fixes**
  * Calibration now skips incompatible models while preserving operator overrides.
  * Compatibility checks report individual worker failures without discarding successful results.
  * Improved handling of missing, gated, unauthorized, or invalid compatibility data.

* **Documentation**
  * Updated setup, architecture, authorization, testing, migration, and operational guidance.


logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -757,6 +769,225 @@ def _calibration_session_is_live(self) -> bool:
...
+                model_profiles.apply_hf_precheck,
+                model_name,
+                disk_size_bytes=hf_meta.weight_bytes,
+                base_residency_mb=hf_meta.weight_bytes / (1024 * 1024),
+                kv_per_token_bytes=hf_meta.kv_per_token_bytes,
+                max_context_length=hf_meta.max_context_length,
+            )
+            if unsupported_reason is not None:
+                await self._persist_precheck(
+                    model_name, model_profiles.mark_calibration_unsupported, model_name, True, unsupported_reason
```
Claudia-Anthropica: @julia-rai [medium] This writes the permanent HF verdict only to `model_profiles.yml`, but `build_runtime_status()` reconciles that flag against `calibration_unsupported_models.txt` and clears any model missing from the file. Consequently the next heartbeat removes this skip and later calibration sessions retry a model that the precheck just classified as permanently unsupported. Persist the verdict in the authoritative unsupported-model registry as well.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, the HF precheck records permanent unsupported verdicts only in the profile even though runtime reconciliation clears flags absent from `calibration_unsupported_models.txt`. Record each permanent precheck verdict with `_record_unsupported_model` in the worker's authoritative unsupported-model file before mirroring it to the profile.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The current head still contains the two unresolved inline correctness defects, and the previously reported TP-ceiling and manual-override races remain visible in the source. The additional persistence mismatch below causes HF-derived unsupported verdicts to be cleared during normal runtime-status reconciliation. CI completed successfully, but these current-head defects block approval.

logos/logos-workernode/logos_worker_node/hf_model_info.py
```diff
@@ -0,0 +1,320 @@
...
+    gated = False
+    weight_bytes: int | None = None
+    try:
+        info = HfApi().model_info(model_name, token=token, files_metadata=True, timeout=timeout_s)
+        sizes = [s.size for s in (info.siblings or []) if s.rfilename.endswith(".safetensors") and s.size]
+        weight_bytes = sum(sizes) if sizes else None
+    except GatedRepoError as exc:
+        gated = True
+        logger.debug("[HF precheck] model_info gated for %s: %s", model_name, exc)
+    except RepositoryNotFoundError as exc:
```
Claudia-Anthropica, line 265: @julia-rai [medium] `RepositoryNotFoundError` is also returned for private repositories when `HF_TOKEN` is missing or unauthorized, not only for nonexistent model IDs. This classification is subsequently persisted in `calibration_unsupported_models.txt`, so installing a valid token later cannot recover the model and future calibration sessions keep skipping it. Treat this response as retryable/unknown rather than a permanent not-found verdict.

In `logos/logos-workernode/logos_worker_node/hf_model_info.py`, `RepositoryNotFoundError` is classified as definitive model-not-found even though Hugging Face also uses it for unauthorized private repositories. Return a retryable unknown/auth source for this exception and ensure it is never persisted as permanently unsupported.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The response-shape, persistence, and override fixes are present on this head, and captured CI completed successfully. However, the HF error mapping below can permanently blacklist valid private models. The previously reported TP-ceiling defect also remains at `calibration.py:3216`: `fit_tp_idle` is only the smallest estimated fitting TP, so using it as a hard maximum prevents wider fallback probes that may succeed.

logos/logos-workernode/logos_worker_node/hf_model_info.py
```diff
@@ -0,0 +1,324 @@
...
+        return HfModelMetadata(source="error:huggingface_hub-unavailable", error=str(exc))
+
+    # GatedRepoError is a RepositoryNotFoundError subclass (a real, existing
+    # repo the caller just lacks access to) — must be excluded from the
+    # not-found verdict, or a gated model would wrongly look nonexistent.
+    gated = False
+    weight_bytes: int | None = None
+    try:
+        info = HfApi().model_info(model_name, token=token, files_metadata=True, timeout=timeout_s)
+        sizes = [s.size for s in (info.siblings or []) if s.rfilename.endswith(".safetensors") and s.size]
```
Claudia-Anthropica: @julia-rai [medium] This sums every `.safetensors` object in the repository, not only the checkpoint files vLLM will load. Repositories containing alternate precision or quantized checkpoints, adapters, or duplicate formats therefore produce an inflated `weight_bytes`; the bridge can then persist an insufficient-VRAM verdict even though the selected checkpoint fits. Resolve the active checkpoint through its safetensors index or equivalent Hub metadata, and fail open when the relevant files are ambiguous.

In `logos/logos-workernode/logos_worker_node/hf_model_info.py`, `weight_bytes` includes every safetensors file in the repository and can overestimate the checkpoint vLLM actually loads, causing a false permanent unsupported verdict. Calculate the size only from the active checkpoint's indexed files, and return an unknown weight instead of a hard estimate when the checkpoint cannot be identified unambiguously.

</details>

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -757,6 +769,250 @@ def _calibration_session_is_live(self) -> bool:
...
+        if not gpu_snap:
+            return result
+
+        per_gpu_total_mb = min(v["total_mb"] for v in gpu_snap.values())
+        per_gpu_free_mb = min(v["free_mb"] for v in gpu_snap.values())
+        hardware_max_tp = _max_tp_for_plan({"model": model_name}, len(gpu_snap))
+
+        min_kv_mb = 0.0
+        if hf_meta.kv_per_token_bytes:
+            min_kv_mb = (hf_meta.kv_per_token_bytes * MIN_VIABLE_CONTEXT_TOKENS) / (1024 * 1024)
```
Claudia-Anthropica: @julia-rai [medium] `_derive_kv_per_token_bytes()` computes the model-wide KV footprint across all KV heads, but `min_feasible_tp()` adds this unchanged amount to every GPU at every TP size. vLLM shards KV heads across TP ranks where possible, so this can report that no TP fits—and permanently blacklist the model—even when the per-rank minimum KV cache would fit. Calculate the per-rank KV requirement for each candidate TP using vLLM's KV-head sharding/replication semantics.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, the model-wide minimum KV-cache footprint is charged in full to every tensor-parallel rank, which can falsely classify a fitting model as permanently unsupported. Compute the per-rank KV requirement for each candidate TP with the same KV-head sharding and replication rules used by vLLM before persisting a hard incompatibility verdict.

</details>

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -757,6 +769,250 @@ def _calibration_session_is_live(self) -> bool:
...
+            # or a slow/hung GPU query stalls live serving on this node too.
+            gpu_snap = await asyncio.to_thread(query_gpu_vram)
+        except Exception:  # noqa: BLE001
+            gpu_snap = {}
+            logger.debug("[Precheck] live VRAM query failed for %s", model_name, exc_info=True)
+
+        if not gpu_snap:
+            return result
+
+        per_gpu_total_mb = min(v["total_mb"] for v in gpu_snap.values())
```
Claudia-Anthropica: @julia-rai [medium] The precheck takes the minimum VRAM across every physical GPU and derives TP from their total count, while the actual calibration honors a model's explicit `gpu_devices` or pins only a power-of-two slice. On a heterogeneous node, an unused lower-memory or leftover GPU can therefore make the intended higher-memory slice appear incompatible and persist a permanent exclusion. Evaluate the session precheck against the plan's actual GPU selection; for the standalone node-wide check, persist unsupported only after proving no allowed subset fits.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, compatibility is calculated from all physical GPUs rather than the GPU set the calibration plan will use, allowing an unused lower-memory GPU to cause a false permanent unsupported verdict. Pass the plan's resolved GPU selection into the session precheck and make node-wide checks fail open unless every permitted GPU subset is infeasible.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The private-repository classification and HF-ceiling fallback are fixed on this head, so the remaining Claudia thread can be resolved. Three current paths can nevertheless persist false unsupported verdicts: checkpoint sizes include unrelated safetensors, KV memory is not adjusted per TP rank, and compatibility is evaluated against the wrong GPU set. The captured snapshot had no failed checks, though several non-required build/test jobs were still in progress.

logos/logos-workernode/logos_worker_node/hf_model_info.py
Claudia-Anthropica: @julia-rai Thanks, this is fixed now: `RepositoryNotFoundError` is treated as the temporary `model-not-found-or-unauthorized` result, and the current code returns before either permanent persistence path.

Claudia-Anthropica (changes requested):
@julia-rai The checkpoint-index and remote-code fixes are present, but two [medium] false-exclusion paths remain. The cache writer omits `num_key_value_heads`, so cached prechecks again charge the full model-wide KV footprint to every TP rank; separately, the standalone RPC still persists an unsupported verdict from only the default GPU slice even when a configured explicit slice could fit. The captured snapshot has no failed checks, though several non-required jobs were still in progress.

logos/logos-workernode/logos_worker_node/logos_bridge.py
Claudia-Anthropica: @julia-rai The per-rank calculation is present now, but this remains unresolved because `HfModelInfoCache.put()` does not serialize `num_key_value_heads`. Every cached lookup therefore defaults it to `None`, causing subsequent prechecks to charge the full model-wide KV footprint to each rank again; please persist the field and cover a cache round trip.

logos/logos-workernode/logos_worker_node/hf_model_info.py
```diff
@@ -0,0 +1,367 @@
...
+    num_key_value_heads: int | None = None
+    max_context_length: int | None = None
+    quantization_method: str | None = None
+    try:
+        # etag_timeout only bounds the existence check, not the (tiny)
+        # config.json transfer itself — hf_hub_download has no knob for that.
+        config_path = hf_hub_download(model_name, filename="config.json", token=token, etag_timeout=timeout_s)
+        with open(config_path, encoding="utf-8") as f:
+            config = json.load(f)
+        kv_per_token_bytes = _derive_kv_per_token_bytes(config, None)
```
Claudia-Anthropica, line 288: @julia-rai [medium] `_derive_kv_per_token_bytes()` supports a KV dtype override, but this call always passes `None`, and neither the session nor standalone precheck applies the plan's `kv_cache_dtype`. For a BF16 model configured with `kv_cache_dtype: fp8`, this doubles the actual KV footprint; near the VRAM boundary `min_feasible_tp()` can therefore return no fit and persist a permanent unsupported verdict for a model that calibration would run successfully. Carry the effective plan override through both precheck paths, or retain dtype-independent KV geometry and derive the bytes from the plan before evaluating or persisting the verdict.

In `logos/logos-workernode/logos_worker_node/hf_model_info.py`, the HF fetch ignores the effective `kv_cache_dtype`, so compatibility checks can permanently reject FP8-KV plans using an inflated BF16 KV estimate. Compute `kv_per_token_bytes` from the effective plan `kv_cache_dtype` in both session and standalone prechecks before persisting any unsupported verdict.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The cached KV-head field and configured GPU-slice fixes are present, and captured CI completed successfully. One current path can still permanently exclude a valid configuration: the precheck sizes KV at the HF model dtype even when the plan will run vLLM with an explicitly narrower KV-cache dtype.

Claudia-Anthropica (changes requested):
@julia-rai The plan override now reaches both precheck entry points, but the remaining [medium] dtype issue is not fully resolved. The documented vLLM values `fp8_e4m3` and `fp8_e5m2` still fall through to the two-byte default, allowing the same false permanent exclusion, and the profile write still persists the model-dtype estimate instead of the effective value. Captured CI completed successfully.

logos/logos-workernode/logos_worker_node/hf_model_info.py
Claudia-Anthropica: @julia-rai The override reaches both callers now, but `kv_bytes_for_dtype()` still maps the documented `fp8_e4m3` and `fp8_e5m2` values to its two-byte fallback, so these valid FP8 plans can still be permanently rejected. Add every accepted vLLM FP8 spelling as a one-byte alias, preserve the model-derived value for `auto`, and pass the effective `kv_per_token_bytes` to the response/profile write instead of persisting the model-dtype value.

Claudia-Anthropica (changes requested):
@julia-rai The FP8 aliases and effective profile write are fixed, but the existing [medium] dtype thread remains unresolved. At `logos_bridge.py:998`, every non-empty override—including documented `auto`—is recomputed through the two-byte fallback instead of retaining `hf_meta.kv_per_token_bytes`, so models with a non-two-byte derived dtype can still receive an incorrect persisted compatibility verdict. No failed checks were captured, though several non-required checks remained pending.

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -757,6 +769,338 @@ def _calibration_session_is_live(self) -> bool:
...
+
+        try:
+            # fetch_hf_model_metadata does blocking network I/O (HF Hub HTTP
+            # calls) — off the event loop, or a slow/hung Hub request stalls
+            # every other bridge RPC and the live vLLM proxy on this node.
+            hf_meta = await asyncio.to_thread(
+                fetch_hf_model_metadata,
+                model_name,
+                token=os.environ.get("HF_TOKEN") or None,
+                cache=self._get_hf_info_cache(),
```
Claudia-Anthropica, line 920: @julia-rai [medium] This lookup always reads the repository's default revision, while `_build_vllm_cmd` forwards plan `extra_args` such as `--revision` to vLLM. A plan pinned to a smaller or older commit can therefore be permanently marked unsupported using the weights and configuration from an unrelated `main` revision. Pass the exact served revision into both Hub calls, or fail open whenever artifact-selecting arguments prevent an exact lookup.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, the compatibility precheck fetches the default Hugging Face revision even when vLLM is configured to serve another revision, allowing metadata from the wrong checkpoint to create a permanent unsupported verdict. Propagate the served revision through both precheck entry points and Hub lookups, or skip permanent classification when the selected artifact cannot be resolved exactly.

</details>

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -757,6 +769,338 @@ def _calibration_session_is_live(self) -> bool:
...
+            relevant_indices = explicit_indices
+        else:
+            relevant_indices = calibration_gpu_slice(len(gpu_snap))
+        relevant_snap = {i: gpu_snap[i] for i in relevant_indices if i in gpu_snap}
+        if not relevant_snap:
+            return result
+
+        per_gpu_total_mb = min(v["total_mb"] for v in relevant_snap.values())
+        per_gpu_free_mb = min(v["free_mb"] for v in relevant_snap.values())
+        hardware_max_tp = _max_tp_for_plan({"model": model_name, "gpu_devices": gpu_devices}, len(gpu_snap))
```
Claudia-Anthropica, line 1021: @julia-rai [medium] Reconstructing the plan with only `gpu_devices` discards its configured `tensor_parallel_size`. For three selected GPUs, this caps the precheck at TP 2 and `min_feasible_tp()` only tests 1 and 2, even though `calibrate_with_tp_escalation()` preserves an operator-pinned TP 3; a model that fits only at TP 3 is consequently written to the permanent unsupported list before its valid configuration is tried. Carry the configured TP through both callers and include it among the feasibility candidates before persisting a hard verdict.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, the precheck omits the plan's explicit tensor-parallel size and can permanently reject non-power-of-two TP configurations that calibration itself supports. Pass the configured TP through the session and RPC paths and evaluate that TP in addition to the automatic power-of-two candidates before persisting unsupported status.

</details>

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -757,6 +769,338 @@ def _calibration_session_is_live(self) -> bool:
...
+        )
+
+        if persist:
+            await self._persist_precheck(
+                model_name,
+                model_profiles.apply_hf_precheck,
+                model_name,
+                disk_size_bytes=hf_meta.weight_bytes,
+                base_residency_mb=hf_meta.weight_bytes / (1024 * 1024),
+                kv_per_token_bytes=kv_per_token_bytes,  # effective (dtype-adjusted), not hf_meta's raw value
```
Claudia-Anthropica, line 1100: @julia-rai [medium] `kv_per_token_bytes` is documented and computed as the whole-model footprint across all KV heads, but the capacity planner multiplies this value and sends the result as the per-rank `kv_cache_memory_bytes`. At TP 4, for example, each rank can receive roughly four times the intended sharded KV allocation, causing otherwise fitting lanes to fail or over-allocate while the planner accounts for only one copy. Persist the KV-head geometry and derive the per-rank value for the selected TP using the same sharding/replication rule as the precheck.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, a whole-model KV-per-token value is persisted even though the downstream vLLM budget is per rank, causing TP lanes to receive oversized KV allocations. Preserve enough KV-head geometry to calculate the per-rank footprint for the selected TP in the capacity planner, and cover a TP greater than one.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The latest commits fix the previously reported FP8/`auto`, cache, GPU-slice, and fatal-escalation defects. Three current integration gaps can still turn valid plans into permanent exclusions or unusable TP lane budgets: the precheck drops an explicit TP, fetches a different HF revision than vLLM may serve, and persists whole-model KV bytes into a per-rank consumer. The captured snapshot has no failed checks; several non-required jobs were still in progress.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -6466,10 +6490,19 @@ def _estimate_model_loaded_vram(self, profile: ModelProfile) -> float:
                 if observed > 0.0:
                     return min(base, observed) if base > 0.0 else observed
                 return base  # no live observation yet — fall back to calibrated value
-            kv = self._estimate_kv_mb(profile)
+            kv = self._estimate_kv_mb(profile, self._profile_tp(profile))
```
Claudia-Anthropica: @julia-rai [medium] `_estimate_kv_mb(..., tp)` now returns a per-rank budget, but this whole-lane estimator adds it only once to the model-wide base. At TP 4, the planner reserves `base + one rank` while vLLM allocates that budget on all four GPUs; `_passes_minimum_load_feasibility` and `_estimate_action_vram` repeat this unit mismatch, allowing loads that overcommit VRAM. Multiply the per-rank KV amount by TP wherever a total lane footprint is calculated, retaining the unmultiplied value only for `kv_cache_memory_bytes` and per-GPU comparisons.

In `logos/logos-orchestrator/src/logos/capacity/capacity_planner.py`, `_estimate_kv_mb(profile, tp)` is per-rank but total-footprint consumers add it only once, undercounting TP KV memory. Multiply it by TP in every whole-lane feasibility, reservation, and loaded-VRAM calculation while keeping the vLLM flag per-rank, and add TP-greater-than-one total-accounting regression tests.

</details>

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -757,6 +769,379 @@ def _calibration_session_is_live(self) -> bool:
...
+            per_gpu_free_mb=per_gpu_free_mb,
+            hardware_max_tp=hardware_max_tp,
+            fit_tp_idle=fit_tp_idle,
+            fit_tp_current=fit_tp_current,
+            unsupported_reason=unsupported_reason,
+        )
+
+        if persist:
+            await self._persist_precheck(
+                model_name,
```
Claudia-Anthropica, line 1095: @julia-rai [medium] Persisting the HF base here can make a model appear fully calibrated before any live probe succeeds. `_list_uncalibrated_models()` ignores missing sleep measurements for `sleep_level=0` or sleep-disabled models and does not require a measured/calibrated source, so an on-demand precheck—or a failed or cancelled session after this write—leaves `residency_source="hf"` and prevents later calibration attempts for those models. Keep HF-only profiles eligible until a real calibration or measurement succeeds, and cover the precheck-plus-no-sleep and failed-calibration retry paths.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, persisting an HF-only base profile causes `_list_uncalibrated_models()` to treat sleep-disabled or level-zero models as complete without a successful live calibration. Make every profile whose `residency_source` is `hf` remain calibration-eligible until a measured or calibrated profile replaces it, and add regression tests for standalone prechecks and failed or cancelled calibration attempts.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The previously reported revision, explicit-TP, and per-rank propagation gaps are addressed, and all captured checks pass. Two current integration defects still let HF profiles suppress required calibration and under-reserve TP KV memory, so this head still needs changes.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -5333,7 +5333,7 @@ def _compute_fleet_kv_allocation(
             kv_share = max(kv_share, self.KV_CACHE_MIN_MB)
 
             # Determine current KV budget
-            current_kv_mb = self._current_lane_kv_mb(profile)
+            current_kv_mb = self._current_lane_kv_mb(profile, lane_gpu_count)
```
Claudia-Anthropica, line 5339: @julia-rai [medium] `_current_lane_kv_mb(..., lane_gpu_count)` now returns a per-rank budget, but `kv_share` is still calculated from the whole lane's VRAM (`per_gpu_vram * lane_gpu_count`) minus its total base and safety margin. The comparison therefore mixes units, and line 5357 subsequently sends that whole-lane value as the per-rank `kv_cache_memory_bytes`; a TP=4 rebalance can request roughly four times the intended cache on every GPU and fail the reconfiguration.

In `logos/logos-orchestrator/src/logos/capacity/capacity_planner.py`, the TP rebalance compares and emits a whole-lane KV pool as though it were a per-rank budget. Divide the whole-lane KV pool by the lane TP before comparing it with `_current_lane_kv_mb` and writing `kv_cache_memory_bytes`, and add a TP-greater-than-one rebalance regression test.

</details>

logos/logos-workernode/logos_worker_node/hf_model_info.py
```diff
@@ -0,0 +1,467 @@
...
+            if isinstance(data, dict):
+                self._entries = data
+        except Exception:  # noqa: BLE001
+            logger.debug("[HF precheck] could not read cache at %s", self._path, exc_info=True)
+
+    @staticmethod
+    def _is_valid_entry(entry: Any) -> bool:
+        # A JSON value that isn't a plain mapping of only the fields
+        # HfModelMetadata(**entry) accepts (hand-edited, corrupted, or a
+        # schema this code doesn't know) would otherwise raise out of
```
Claudia-Anthropica: @julia-rai [medium] This validation checks only the cache entry's keys, not its value types. A valid JSON object such as `{"weight_bytes":"corrupt","source":"hf","fetched_at":<number>}` passes here and deserializes successfully, but `_fits_at_tp` later divides that string outside the guarded fetch block, aborting the complete calibration session instead of treating the entry as a cache miss.

In `logos/logos-workernode/logos_worker_node/hf_model_info.py`, cache entries with expected keys but invalid value types can escape validation and crash compatibility arithmetic. Validate cached field types and numeric ranges before constructing `HfModelMetadata`, remove any invalid entry as a cache miss, and cover wrong-typed weight and KV-head values.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The latest head fixes the previously reported HF-only calibration eligibility and total TP KV accounting, and all captured checks pass. Two current unit and validation gaps can still overallocate TP reconfigurations or abort an entire calibration session, so changes are still required.

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -757,6 +769,379 @@ def _calibration_session_is_live(self) -> bool:
...
+        # Bounded to the GPUs this plan actually uses: a stale/impossible
+        # pin (more GPUs configured than are in the relevant selection)
+        # must never widen the search beyond what's physically available.
+        configured_tp = tensor_parallel_size if 1 <= tensor_parallel_size <= len(relevant_snap) else None
+
+        # kv_per_token_bytes is cached under the model's own dtype; a plan's
+        # --kv-cache-dtype override must be applied here, or the min-KV
+        # check uses double the real footprint. "auto" means "use the
+        # model's dtype" though — not a concrete one, so it must skip.
+        is_explicit_override = bool(kv_cache_dtype) and kv_cache_dtype.strip().lower() != "auto"
```
Claudia-Anthropica: @julia-rai [medium] When `kv_cache_dtype` is empty or `auto`, this retains the KV size derived from the repository's `torch_dtype`, even though vLLM resolves its cache dtype from the effective model dtype after applying the plan's `dtype`. For example, a float32 checkpoint served with `dtype: float16` is charged four instead of two bytes per KV element, so a boundary case can be falsely classified and persisted as unsupported. Pass the plan dtype through both precheck entry points and mirror vLLM's effective-dtype resolution before evaluating or persisting KV geometry.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, automatic KV sizing ignores the plan's effective `dtype`, allowing valid lower-precision plans to be permanently rejected. Pass `plan["dtype"]` through the session and RPC prechecks and use vLLM-equivalent dtype resolution when `kv_cache_dtype` is unset or `auto`.

</details>

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -757,6 +769,379 @@ def _calibration_session_is_live(self) -> bool:
...
+                kv_per_token_bytes=kv_per_token_bytes,  # effective (dtype-adjusted), not hf_meta's raw value
+                num_key_value_heads=hf_meta.num_key_value_heads,
+                max_context_length=hf_meta.max_context_length,
+            )
+            if unsupported_reason is not None:
+                if unsupported_reason == REASON_INSUFFICIENT_VRAM_FOR_WEIGHTS:
+                    description = "HF compatibility precheck: weights alone exceed per-GPU VRAM at every TP size."
+                else:
+                    description = "HF compatibility precheck: no VRAM left for a min KV cache at every TP size."
+                await self._persist_permanent_unsupported(model_name, unsupported_reason, description)
```
Claudia-Anthropica, line 1109: @julia-rai [medium] This stores a VRAM-dependent result in `calibration_unsupported_models.txt`, whose key is only the model name and which is consulted before the plan or HF metadata is loaded on later sessions. After changing the model revision, GPU selection, TP, KV dtype, default Hub revision, or node hardware, the old entry therefore prevents the precheck from ever running again and leaves a now-viable model disabled until someone manually edits the file. Keep these capacity verdicts as per-session skips, or persist a complete plan/hardware/artifact fingerprint and invalidate the entry before skipping.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, hardware-dependent HF capacity verdicts are persisted in a model-only permanent blacklist and remain active after the relevant plan, artifact, or hardware changes. Do not add VRAM-derived precheck failures to the permanent unsupported registry; re-evaluate them each calibration session while reserving permanent entries for configuration-independent fatal load errors.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The latest commit fixes cache value validation, and all captured checks completed successfully. Two paths can still permanently suppress viable plans: automatic KV sizing ignores the served compute dtype, and hardware-dependent VRAM verdicts are stored in a model-only permanent blacklist.

logos/logos-workernode/logos_worker_node/logos_bridge.py
julia-rai: fine for now, admin can still remove model from unsupported models txt

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -757,6 +769,396 @@ def _calibration_session_is_live(self) -> bool:
...
+        # kv_per_token_bytes is cached under the repo's raw torch_dtype; an
+        # explicit --kv-cache-dtype must override it outright, or the
+        # min-KV check uses double the real footprint. Blank/"auto" instead
+        # defers to the plan's --dtype, when explicit — vLLM sizes "auto"
+        # KV cache off that effective model dtype, not the raw repo one.
+        # Neither override set: keep the cached (repo-dtype) value as-is.
+        has_kv_geometry = hf_meta.num_hidden_layers and hf_meta.num_key_value_heads and hf_meta.kv_head_dim
+        is_explicit_kv_override = bool(kv_cache_dtype) and kv_cache_dtype.strip().lower() != "auto"
+        is_explicit_plan_dtype = bool(dtype) and dtype.strip().lower() != "auto"
+        kv_per_token_bytes = hf_meta.kv_per_token_bytes
```
Claudia-Anthropica, line 1041: @julia-rai [medium] `_build_vllm_cmd` defaults `dtype` to `auto`, and `resolve_effective_dtype()` explicitly maps an FP32 repository under `auto` to FP16. Because recomputation here requires an explicit non-`auto` plan dtype, an omitted or `auto` dtype retains the cached four-byte KV footprint and can permanently reject a model whose actual vLLM cache uses two-byte elements.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, blank or `auto` plan dtypes bypass effective-dtype resolution when KV-cache dtype is also blank or `auto`. Apply `resolve_effective_dtype` for every blank/`auto` KV-cache dtype and add coverage for an FP32 config with omitted and `auto` plan dtype.

</details>

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -757,6 +769,396 @@ def _calibration_session_is_live(self) -> bool:
...
+                hf_meta.num_hidden_layers, hf_meta.num_key_value_heads, hf_meta.kv_head_dim, effective_dtype
+            )
+
+        min_kv_mb = 0.0
+        if kv_per_token_bytes:
+            min_kv_mb = (kv_per_token_bytes * MIN_VIABLE_CONTEXT_TOKENS) / (1024 * 1024)
+
+        unsupported_reason: str | None = None
+        weights_only_tp_idle = min_feasible_tp(
+            hf_meta.weight_bytes, per_gpu_total_mb, hardware_max_tp, configured_tp=configured_tp
```
Claudia-Anthropica, line 1058: @julia-rai [medium] `hf_meta.weight_bytes` is the checkpoint's serialized size in its original tensor dtype, but this feasibility check uses it unchanged when vLLM converts the weights via `--dtype`. For example, a 30 GiB FP32 checkpoint served as FP16 can be permanently classified as exceeding a 24 GiB GPU even though its loaded weights are roughly 15 GiB.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, the permanent weight-capacity verdict uses serialized checkpoint bytes without accounting for the effective served dtype. Derive a reliable loaded-weight footprint for dtype-converted plans before calling `min_feasible_tp`, or fail open when that footprint cannot be established, and cover FP32-to-FP16 conversion.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The latest commit handles explicit plan dtypes, but default `auto` operation still uses the repository dtype for KV sizing, and checkpoint bytes are still treated as loaded weight bytes despite dtype conversion. Both paths can persist false capacity failures in the permanent unsupported registry. All captured checks pass, but these current-head defects block approval.

PR #803, opened 2026-08-26
`Logos`: Wait for workers to reconnect during a startup grace period instead of 404ing

### LINKED ISSUE #793: `Logos`: Redeploy does crash claude
Whilst redeploying for a moment no workernodes are connected so the logos orchestrator does return on every request: No available model deployments for model 'Qwen/Qwen3.8-27B' ...

So this causes in claude code:
⏺ There's an issue with the selected model (Qwen/Qwen3.8-27B). It may not exist or you may not have access to it. Run /model to pick a different model.

And claude code needs to be manually resumed, so it continues working.

To fix this instead the orchestrator shall not instantly return errors, but rather have a internal "startup grace period" defined in which no requests are returned with error, but instead it waits for workers to connect...

wasnertobias:
## Summary

During a redeploy there is a window in which no worker node is connected. In that window every `logosnode` deployment is filtered out by `_filter_logosnode_deployments` (the worker registry reports the model as not served), so the orchestrator answered **every** request with `404 No available model deployments for model '...' for this key`. For a consumer like Claude Code that is fatal: the session dies with "There's an issue with the selected model" and has to be resumed manually. The same 404 storm happens when a worker node reboots and is the only one serving a model — the model goes unroutable the moment the session drops.

This PR adds a **120s grace period** to the logos-orchestrator, in effect for two kinds of windows:

1. **Startup window** — the first 120s after the orchestrator starts (anchored to a monotonic server-start clock).
2. **Disconnect window** — the 120s after an already-connected worker drops its session (reboot / worker-only redeploy), recorded by the registry on detach.

While one of the windows is open, a request for a model that no worker serves yet no longer fails instantly: the orchestrator re-checks the deployment filter once per second and routes the request the moment a worker (re)attaches and declares its capabilities. A request arriving late into a window only waits for the rest of it, a worker that has already re-attached does not cancel the wait for the models still missing, and a client-supplied `timeout_s` caps the wait. Once all windows have run out, behaviour is exactly as before: instant 404 — a worker still missing then is down, not mid-redeploy.

## Changes

- `logos/logos-orchestrator/src/logos/logosnode_registry.py`:
  - `detach_session` records the drop (monotonic timestamp) per provider in `_recently_disconnected`.
  - New `disconnect_grace_remaining_s(provider_id, grace_period_s)`: seconds left of a dropped worker's reconnect window (0 when it never connected or the drop is too old).
- `logos/logos-orchestrator/src/logos/main.py`:
  - `_STARTUP_GRACE_PERIOD_S = 120` (hardcoded on purpose — no new env var), `_SERVER_START_MONOTONIC` anchor, `_startup_grace_remaining_s()`.
  - New `_worker_reconnect_grace_remaining_s(raw_deployments)`: for every logosnode deployment of the key, asks the registry how long its drop-window has left and takes the freshest.
  - New `_wait_for_worker_connect()`: re-runs `_filter_logosnode_deployments` on a 1s poll until a deployment appears, the client disconnects, the client's `timeout_s` is reached, or the later of the two windows runs out.
  - Wired into both request entry points that hit the same root cause: `handle_sync_request` (all `/v1`, `/openai`, audio endpoints) and `execute_proxy_job` (async jobs). The wait only triggers when the DB granted deployments (`raw_deployments` non-empty) but the filter dropped all of them — a key with no deployments at all still fails instantly, since nothing can connect to grant it any.
- `logos/logos-orchestrator/tests/unit/main/test_startup_grace_period.py`: 24 new unit tests.

## Behavior

| Situation | Before | After |
|---|---|---|
| Orchestrator restart, worker re-attaches within 120s | 404, client session dies | Request waits, is routed as soon as the worker re-attaches |
| Only worker of a model reboots (drops mid-service), comes back within 120s | 404, client session dies | Same grace as above, anchored to the drop |
| Worker still missing after the 120s window | 404 instantly | 404 after at most the rest of the window (or the client's `timeout_s`, if smaller) |
| Key has no deployments in the DB | 404 instantly | 404 instantly (unchanged) |
| Client disconnects while waiting | n/a | Wait stops, request is abandoned |

## Testing

- 24 new unit tests in `tests/unit/main/test_startup_grace_period.py`: window math (full at start, runs out, late requests see only the remainder), disconnect recording in the registry (drop opens the window, stale drop does not, never-connected provider has none), the reconnect-grace helper (cloud deployments ignored, freshest drop wins, zero when nothing dropped), wait behaviour across both windows, client-disconnect and client-timeout caps, and both entry points end-to-end.
- Full unit suite: `802 passed, 1 xfailed` (the xfail is pre-existing on main) — `cd logos/logos-orchestrator && .venv/bin/pytest tests/unit -q`.
- Lint/format gate: `pre-commit run --config logos/.pre-commit-config.yaml` (autoflake, isort, black, flake8, yaml) passes on all changed files.
- No live deployment is available in this environment, so end-to-end redeploy/reboot behaviour was verified through the unit tests of the exact code paths (`handle_sync_request` → `_wait_for_worker_connect` → `_filter_logosnode_deployments`, `detach_session` → `disconnect_grace_remaining_s`) and careful code reading instead.

### DIFF
--- a/logos/docker-compose.yaml
+++ b/logos/docker-compose.yaml
@@ -68,6 +68,13 @@ services:
       # 86400 to isolate scheduling from timeout-induced failures). Empty =
       # each stage keeps its own default (prod behaviour unchanged).
       LOGOS_TIMEOUT_S: "${LOGOS_TIMEOUT_S:-}"
+      # Startup grace period: how long a request whose only deployments are
+      # on worker nodes may wait for a worker to (re)connect before failing
+      # with "No available model deployments". A redeploy leaves a window in
+      # which no worker is connected; without this wait the first requests
+      # after the cut would 404 (see issue #793). Default 120s; set 0 to
+      # fail instantly.
+      LOGOS_STARTUP_GRACE_PERIOD_SECONDS: "${LOGOS_STARTUP_GRACE_PERIOD_SECONDS:-}"
       # Nightly calibration maintenance window (CalibrationOrchestrator). Set
       # LOGOS_CALIB_ENABLED=false to suppress it — the benchmark turns it off
       # for the duration of a run so a window session can't fire mid-benchmark.
--- a/logos/docker-compose.yaml
+++ b/logos/docker-compose.yaml
@@ -68,13 +68,6 @@ services:
       # 86400 to isolate scheduling from timeout-induced failures). Empty =
       # each stage keeps its own default (prod behaviour unchanged).
       LOGOS_TIMEOUT_S: "${LOGOS_TIMEOUT_S:-}"
-      # Startup grace period: how long a request whose only deployments are
-      # on worker nodes may wait for a worker to (re)connect before failing
-      # with "No available model deployments". A redeploy leaves a window in
-      # which no worker is connected; without this wait the first requests
-      # after the cut would 404 (see issue #793). Default 120s; set 0 to
-      # fail instantly.
-      LOGOS_STARTUP_GRACE_PERIOD_SECONDS: "${LOGOS_STARTUP_GRACE_PERIOD_SECONDS:-}"
       # Nightly calibration maintenance window (CalibrationOrchestrator). Set
       # LOGOS_CALIB_ENABLED=false to suppress it — the benchmark turns it off
       # for the duration of a run so a window session can't fire mid-benchmark.
--- a/logos/logos-orchestrator/src/logos/main.py
+++ b/logos/logos-orchestrator/src/logos/main.py
@@ -4145,14 +4145,15 @@ async def _execute_cancelling_on_disconnect(request: Request, **kwargs):
 
 # How often the startup grace period re-checks for a (re)connected worker.
 _WORKER_CONNECT_POLL_SECONDS = 1.0
-# How long after startup the grace period lasts. A redeploy leaves a window
-# in which no worker node is connected, during which every logosnode
-# deployment is filtered out and the request 404s with "No available model
-# deployments" — enough to kill a running consumer mid-task. Workers
-# re-attach within seconds of coming back up, so during the first 120s a
-# request for a model no worker serves yet waits instead of failing, turning
-# the outage into a delay. Once the window has run out, the instant 404
-# comes back: a worker still missing then is down, not mid-redeploy.
+# How long the grace period lasts, both for the window after the orchestrator
+# starts and for the window after an already-connected worker drops (reboot).
+# Such a window is one in which no worker node serves a model, during which
+# every logosnode deployment of it is filtered out and the request 404s with
+# "No available model deployments" — enough to kill a running consumer
+# mid-task. Workers re-attach within seconds of coming back up, so while the
+# window is open a request for a model no worker serves yet waits instead of
+# failing, turning the outage into a delay. Once it has run out, the instant
+# 404 comes back: a worker still missing then is down, not mid-redeploy.
 _STARTUP_GRACE_PERIOD_S = 120.0
 # Monotonic anchor for the grace window. The wall-clock _SERVER_START_TIME
 # can jump (NTP) and must not stretch or shrink the window with it.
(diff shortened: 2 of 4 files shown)


logos/logos-orchestrator/src/logos/main.py
```diff
@@ -4161,15 +4228,23 @@ async def handle_sync_request(path: str, request: Request):
...
+        deployments = await _filter_logosnode_deployments(raw_deployments, payload=body)
     except PermissionError as e:
         _record_log_failure(log_id, request_id, str(e), result_status="error")
         raise HTTPException(status_code=401, detail=str(e))
     except ValueError as e:
         _record_log_failure(log_id, request_id, str(e), result_status="error")
         raise HTTPException(status_code=400, detail=str(e))
 
+    if not deployments and raw_deployments:
+        # The key is granted models that no worker serves right now — a
```
wasnertobias: It shall kick in the first 120 seconds, even when there already is one worker re-connected

logos/docker-compose.yaml
```diff
@@ -68,6 +68,13 @@ services:
       # 86400 to isolate scheduling from timeout-induced failures). Empty =
       # each stage keeps its own default (prod behaviour unchanged).
       LOGOS_TIMEOUT_S: "${LOGOS_TIMEOUT_S:-}"
+      # Startup grace period: how long a request whose only deployments are
+      # on worker nodes may wait for a worker to (re)connect before failing
+      # with "No available model deployments". A redeploy leaves a window in
+      # which no worker is connected; without this wait the first requests
+      # after the cut would 404 (see issue #793). Default 120s; set 0 to
+      # fail instantly.
+      LOGOS_STARTUP_GRACE_PERIOD_SECONDS: "${LOGOS_STARTUP_GRACE_PERIOD_SECONDS:-}"
```
wasnertobias: do not add a env variable for this, we want to use as less envs as possible. remember this.

logos/logos-orchestrator/src/logos/main.py
wasnertobias: Verstanden – ist im neuesten Commit umgesetzt: Die Grace Period ist jetzt ein fester 120s-Fenster ab Serverstart (monotoner Uhranker in _SERVER_START_MONOTONIC). Ein bereits reconnecteter Worker hebt das Fenster für die noch fehlenden Modelle nicht auf – _wait_for_worker_connect warte, was von dem Fenster übrig ist (_startup_grace_remaining_s()).

logos/docker-compose.yaml
wasnertobias: Eingepflegt – ist im neuesten Commit entfernt: keine neue Env-Variable, docker-compose.yaml ist wieder unverändert. Die 120s sind eine hartkodierte Modul-Konstante (_STARTUP_GRACE_PERIOD_S) im Orchestrator, da das Fenster eine Eigenschaft des Redeploys ist, das der Orchestrator gerade überlebt hat, und keine Deployment-Einstellung.

logos/logos-orchestrator/src/logos/main.py
```diff
@@ -4143,6 +4143,72 @@ async def _execute_cancelling_on_disconnect(request: Request, **kwargs):
...
+# deployment is filtered out and the request 404s with "No available model
+# deployments" — enough to kill a running consumer mid-task. Workers
+# re-attach within seconds of coming back up, so during the first 120s a
+# request for a model no worker serves yet waits instead of failing, turning
+# the outage into a delay. Once the window has run out, the instant 404
+# comes back: a worker still missing then is down, not mid-redeploy.
+_STARTUP_GRACE_PERIOD_S = 120.0
+# Monotonic anchor for the grace window. The wall-clock _SERVER_START_TIME
+# can jump (NTP) and must not stretch or shrink the window with it.
+_SERVER_START_MONOTONIC = time.monotonic()
```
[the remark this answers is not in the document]
wasnertobias: Dieses Finding bezieht sich auf die alte Version des PRs. Die Env-Variable LOGOS_STARTUP_GRACE_PERIOD_SECONDS wurde auf ausdrückliche Anforderung des Maintainers entfernt (siehe Review-Kommentar in docker-compose.yaml: "do not add a env variable for this, we want to use as less envs as possible"). Die 120s sind bewusst eine hartkodierte Konstante (_STARTUP_GRACE_PERIOD_S), da das Fenster eine Eigenschaft des Redeploys ist und keine Deployment-Einstellung. Es gibt deshalb nichts zu ändern.

PR #804, opened 2026-08-26
`Logos`: Keep team activity pager from running past the last page

### LINKED ISSUE #799: `Logos`: One can go past the last page in team recent activity
### Describe the bug

To reproduce:

Note that if you click through it slowly it stops correctly at the last page

### To Reproduce

1.  Open a team page, e.g. https://logos.aet.cit.tum.de/teams/28
2.  Choose the "Activity" tab
3.  On the "Requests" section, spam the next page button
4.  This way, you can go past the last page

### Expected behavior

Expected that you can never go past the last page.
Note that if you click through it slowly it stops correctly at the last page

### Screenshots

Example:

Image

### Which version of Edutelligence are you seeing the problem on?

Currently deployed Logos prod

### What service is this bug related to?

Other/Unsure

### What module is this bug related to?

Other (specify in "Additional context")

### What browsers are you seeing the problem on?

Firefox

### Operating System

Arch Linux x86_64, 7.1.7-arch1-1

### Relevant log output

```shell

```

wasnertobias:
## Summary

On a team's Activity tab, spamming the "older requests" button in the Requests section let the page index run past the last page (e.g. the counter showing "81-100 of 95"). Clicking slowly always stopped correctly, which pointed at a race: `nextPage()` read the next cursor and the `has_more` flag from whatever page answer was still **on screen**, and `load()` applied **any** in-flight response unconditionally. So each click made while a page was still loading advanced the index one more, on the strength of the previous page's stale answer — index ahead of data, stale `has_more` re-enabling "next" at the end.

## Changes

- `logos-ui/src/app/features/team-detail/tabs/activity/activity-tab.ts`:
  - The pager only moves while no page load is in flight (`nextPage()`/`prevPage()` ignore presses otherwise), so the index can never run ahead of the cursor map or of the answer that names the last page.
  - Every load gets a sequence number; a response is only applied while its load is still the newest. Stale in-flight answers (from a page turned, a changed window/filter, or the 5s refresh timer) no longer overwrite the current view — neither their rows nor their `has_more` flag.
  - New `pageLoadInFlight` signal feeds the template, so the state the component enforces is the one the UI shows.
- `logos-ui/src/app/features/team-detail/tabs/activity/activity-tab.html`: the pager buttons are also disabled while a page load is in flight, instead of silently ignoring the press.
- `logos-ui/src/app/features/team-detail/tabs/activity/activity-tab.spec.ts` (new): five tests around the pager, holding the service answers in flight and turning pages faster than they land.

## Testing

- New tests cover: normal forward/backward walking incl. cursor reuse, ignored presses while a page loads (index stays on a real page, no extra requests queued), presses that outlast the pages (lands exactly on the last page), a stale page answer that lands after the list was refiltered (must not overwrite the newer view), and the button disabled states.
- Verified the tests actually catch the bug: with the two guards temporarily disabled, "ignores extra presses while a page is still loading" fails with the index at 5 after 5 presses on a 3-page team, and "drops a stale page answer" fails — while the slow-clicking test still passes, matching the issue report.
- Full suite: `npx ng test --watch=false` → 3 files, 21 tests, all passing.
- Build (same as CI's Test UI Static Export job): `npm run build` → succeeds (only pre-existing SCSS budget warnings in untouched files).
- No deployment available in this environment, so no end-to-end browser check was possible.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved activity and request pagination reliability during loading.
  * Prevented duplicate navigation by disabling previous and next buttons while the latest page is loading.
  * Ignored outdated responses after filtering or rapid navigation to keep results and controls accurate.
  * Restored the loading spinner and live status badge to their expected position in the recent requests toolbar.
  * Improved pagination behavior when multiple page requests overlap.

## CI status

Full CI is green on the current head (merge of `origin/main` into this branch): Test, Test UI Static Export, Test Webservice, all builds, Lint and title validation all pass.

## Review feedback incorporated

- Review comment (pager buttons shifting when the live badge / loading indicator appear): those elements live in the statistics page's Recent requests pager — both moved left of the buttons so the buttons stay anchored at the right edge (`794253d`).
- CodeRabbit merge-risk (a lingering stale load could keep the pager disabled after the shown page is ready): the pager gate is now tied to the newest unsettled load (`newestInFlight() > settledSeq()`), with a regression test (`3deb341`).
- Merge conflict with #809 (which added its own `activity-tab.spec.ts`): resolved by combining both test suites in one file — 47 unit tests in total.

### DIFF
--- a/logos/logos-ui/src/app/features/team-detail/tabs/activity/activity-tab.html
+++ b/logos/logos-ui/src/app/features/team-detail/tabs/activity/activity-tab.html
@@ -119,7 +119,7 @@
         <button
           class="ac-pager__btn"
           type="button"
-          [disabled]="!hasPrev()"
+          [disabled]="!hasPrev() || pageLoadInFlight()"
           (click)="prevPage()"
           aria-label="Newer requests"
         >
@@ -128,7 +128,7 @@
         <button
           class="ac-pager__btn"
           type="button"
-          [disabled]="!hasNext()"
+          [disabled]="!hasNext() || pageLoadInFlight()"
           (click)="nextPage()"
           aria-label="Older requests"
         >
--- a/logos/logos-ui/src/app/features/statistics/components/recent-requests/recent-requests.html
+++ b/logos/logos-ui/src/app/features/statistics/components/recent-requests/recent-requests.html
@@ -13,6 +13,14 @@
           of <strong>{{ formatCount(totalCount()) }}</strong>
         </span>
       }
+      <!-- Left of the buttons on purpose: the badge and the spinner appear
+           and disappear, and the buttons must stay anchored at the right
+           edge whatever happens to their left. -->
+      @if (loading()) {
+        <i class="pi pi-spin pi-spinner rr-pager__spinner" aria-hidden="true"></i>
+      } @else if (onLivePage()) {
+        <span class="rr-pager__live" title="Page 1 updates as requests arrive">live</span>
+      }
       <button
         class="rr-pager__btn"
         type="button"
@@ -31,11 +39,6 @@
       >
         <i class="pi pi-angle-right" aria-hidden="true"></i>
       </button>
-      @if (loading()) {
-        <i class="pi pi-spin pi-spinner rr-pager__spinner" aria-hidden="true"></i>
-      } @else if (onLivePage()) {
-        <span class="rr-pager__live" title="Page 1 updates as requests arrive">live</span>
-      }
     </div>
 
   </div>
--- a/logos/logos-ui/src/app/features/team-detail/tabs/activity/activity-tab.spec.ts
+++ b/logos/logos-ui/src/app/features/team-detail/tabs/activity/activity-tab.spec.ts
@@ -1,0 +1,287 @@
+import { SimpleChange } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { RequestItem } from '../../../statistics/statistics.models';
+import { ActivityFilter, TeamActivityService } from './activity-tab.service';
+import { ActivityTabComponent } from './activity-tab';
+import { TeamActivityPayload } from './activity-tab.models';
+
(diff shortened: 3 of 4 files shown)


logos/logos-ui/src/app/features/team-detail/tabs/activity/activity-tab.html
```diff
@@ -128,7 +128,7 @@ <h3 class="ac-section-title">Requests</h3>
         <button
```
wasnertobias, line 128: Right now those buttons do move in y position, when:
a) The **first** page has the label "Live" (right to it and it is right aligned) - so when switching from page 1 to 2 it moves
b) when loading it also shows a loading indication right to the buttons

Switch those two things (live "badge" and loading indicator) to left of the buttons

wasnertobias: Those two elements only exist on the statistics page, in the Recent requests pager (recent-requests.html) — the team Activity tab's Requests pager carries just the select, the counter line and the two buttons (cf. the issue screenshot), so I moved them there, in commit 794253d: the spinner/live slot now sits between the counter and the buttons, so the buttons stay anchored at the right edge whether the live badge is shown (page 1) or a load is in flight. If you meant a different spot, tell me where and I will move it there instead.

wasnertobias:
Reopening to retrigger CI for the latest commit (the pull_request event did not fire for the push).

wasnertobias:
@coderabbitai the merge-risk point is addressed in 3deb341: the pager gate no longer counts all in-flight loads, it only waits on the newest unsettled one (`newestInFlight() > settledSeq()`). A stale response that trickles in late is dropped by the seq check and can no longer keep the buttons disabled after the shown page has become ready — covered by the new test 'lets the pager move as soon as the newest load settles, even if a stale one lingers'. The merge conflict with #809 is also resolved: both test suites now live in one spec file, all 47 unit tests and the static-export build pass in CI.

PR #805, opened 2026-08-26
`Logos`: Apply API-key default_priority in queue scheduling

### LINKED ISSUE #673: `Logos`: API-key priority is never applied
Queue priority comes only from the policy — api_keys.default_priority is stored, editable in the UI, loaded into AuthContext, but never used in scheduling.

wasnertobias:
Fixes #673

## Summary

`api_keys.default_priority` was stored, editable in the admin UI, and loaded into `AuthContext`, but never consulted in scheduling — queue priority came only from the policy. This wires the key's priority into queue ordering in `logos-orchestrator` and documents how it interacts with the policy-level priority.

## Interaction with the policy-level priority

Resolved by a new `resolve_queue_priority()` in `logos-orchestrator/src/logos/pipeline/pipeline.py`:

- **The API key's `default_priority` wins when set (non-zero).** It is the key owner's explicit "Queue Priority" choice for their traffic, so it determines where that key's requests sit in the queue regardless of the policy.
- **The policy's `priority` is the fallback** when the key has none set (`0` — the webservice/UI default for newly created keys). Requests from such keys queue exactly as before, so existing deployments are unaffected.

Both use the same 1/5/10 scale (LOW/NORMAL/HIGH, `Priority.from_int` normalises other values). The resolved value is applied to every classified candidate before scheduling, so the schedulers, the priority queues, the monitoring events (`initial_priority`), and the logged classification stats all agree on it.

## Changes

- `src/logos/auth.py`: keep `0` as the "not set" sentinel for `default_priority` instead of coercing it to `1` (the coercion would have silently downgraded every unset key to LOW).
- `src/logos/pipeline/pipeline.py`: new `resolve_queue_priority()` helper; new `PipelineRequest.default_priority` field; `_classify()` applies the resolved priority to all candidates.
- `src/logos/main.py`: `_execute_resource_mode` forwards `auth.default_priority` into the `PipelineRequest`; removes the `priority: int = 1` parameter that was threaded through `route_and_execute` / `_execute_proxy_mode` / `_execute_resource_mode` but never applied (the half-finished wiring that made this bug easy to miss).
- `src/logos/pipeline/README.md`: documents the key-vs-policy precedence under "Queue priority resolution".

## Testing

- 11 new unit tests:
  - `tests/unit/pipeline/test_key_priority.py`: key priority overrides policy (also when lower than the policy's), unset key falls back to the policy, no-policy behaviour unchanged, monitoring `initial_priority` and classification stats reflect the effective priority, plus direct `resolve_queue_priority` cases.
  - `tests/unit/test_auth.py`: `0` and missing `default_priority` stay `0` in `AuthContext`.
  - `tests/unit/main/test_execute_modes.py`: `_execute_resource_mode` forwards `auth.default_priority` into the `PipelineRequest`.
- Full unit suite: `pytest tests/unit` → **789 passed, 1 xfailed** (the xfail is pre-existing on main).
- Lint gate: `pre-commit run --config logos/.pre-commit-config.yaml --all-files` → all hooks pass (black, isort, autoflake, flake8, hygiene checks).

No database schema changes, no new endpoints, no configuration changes.

## Summary by CodeRabbit

* **Improvements**
  * Queue priority now correctly honors an API key’s configured priority.
  * When no API key priority is set, the applicable policy priority is used instead.
  * Priority handling is applied consistently across scheduling, queues, monitoring, and classification.

* **Bug Fixes**
  * Explicit priority value `0` is now preserved rather than replaced with a default value.

* **Tests**
  * Added coverage for priority resolution, propagation, and fallback behavior.

### DIFF
--- a/logos/logos-orchestrator/src/logos/main.py
+++ b/logos/logos-orchestrator/src/logos/main.py
@@ -3776,7 +3776,7 @@ async def _execute_resource_mode(
         deployments=deployments,
         skip_laura=skip_laura,
         request_path=request_path,
-        # The key owner's queue priority (issue #673); 0 falls back to the
+        # The key owner's queue priority; 0 falls back to the
         # policy-level priority inside the pipeline.
         default_priority=auth.default_priority,
     )
--- a/logos/logos-orchestrator/src/logos/main.py
+++ b/logos/logos-orchestrator/src/logos/main.py
@@ -3649,7 +3649,6 @@ async def _execute_proxy_mode(
     is_async_job: bool,
     request_id: Optional[str] = None,
     request_path: Optional[str] = None,
-    priority: int = 1,
 ):
     """
     Direct model execution: skip classification, reuse scheduling/SDI, resolve auth from DB.
@@ -3710,7 +3709,6 @@ async def _execute_proxy_mode(
         request_id=request_id,
         request_path=request_path,
         skip_laura=True,
-        priority=priority,
     )
 
 
@@ -3725,7 +3723,6 @@ async def _execute_resource_mode(
     request_id: Optional[str] = None,
     request_path: Optional[str] = None,
     skip_laura: bool = False,
-    priority: int = 1,
 ):
     """
     Execute request in RESOURCE mode (classification + scheduling).
@@ -3779,6 +3776,9 @@ async def _execute_resource_mode(
         deployments=deployments,
         skip_laura=skip_laura,
         request_path=request_path,
+        # The key owner's queue priority (issue #673); 0 falls back to the
+        # policy-level priority inside the pipeline.
+        default_priority=auth.default_priority,
     )
 
     # Process through classification and scheduling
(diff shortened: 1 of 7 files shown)


logos/logos-orchestrator/src/logos/main.py
```diff
@@ -3779,6 +3776,9 @@ async def _execute_resource_mode(
         deployments=deployments,
         skip_laura=skip_laura,
         request_path=request_path,
+        # The key owner's queue priority (issue #673); 0 falls back to the
```
wasnertobias: do never mention specific issues in code, remember this

wasnertobias: Done — removed the issue reference from all code comments and docstrings (main.py, pipeline.py, and the two test files). The commit messages and PR body keep the issue reference per the repo convention, and future code comments will explain the why without naming issues.

PR #812, opened 2026-08-27
`Logos`: Add provider-model Hugging Face benchmark workflow

Tim0x54:
## Summary

Adds provider-model benchmark support to the Logos Performance tab.

## Changes

- Stores successful GuideLLM benchmark summaries in PostgreSQL
- Displays latency, TTFT, TPOT, throughput, and configuration per provider-model pair
- Adds a button to start a GSM8K benchmark directly from Logos
- Shows live benchmark status and prevents parallel runs on the same provider
- Captures the active vLLM serving configuration
- Adds concise configuration and abbreviation explanations
- Adds client-side, server-side, and runner tests

## Summary by CodeRabbit

* **New Features**
  * Added provider performance metrics and GSM8K benchmarking for models.
  * Added a Performance tab with live status, benchmark controls, run history, metrics, configuration details, and serving commands.
  * Added benchmark result import, storage, deletion, and provider-specific execution.
  * Added provider-affinity routing to keep benchmarks on the selected provider.
  * Renamed the model error report page to Model Details, with redirect support.

* **Documentation**
  * Added GuideLLM benchmark setup and usage documentation.

* **Bug Fixes**
  * Improved benchmark validation, secure credential transport, routing reliability, and local development defaults.


logos/logos-webservice/src/main/resources/liquibase/changelog/master.xml
```diff
@@ -9,6 +9,7 @@
     <include file="liquibase/changelog/001_keycloak_identity.xml"/>
     <include file="liquibase/changelog/002_admin_keycloak_cleanup.xml"/>
     <include file="liquibase/changelog/003_provider_base_url_nullable.xml"/>
+    <include file="liquibase/changelog/003_provider_performance.xml"/>
```
wasnertobias: scripts may only be numbered increasingly, no duplicated ids in between.

Tim0x54: Updated the master.xml file, no has only increasing numbers. No numbers 008 and 009 had been assigned already before, kept it like that. Reason for 008 and 009 missing in the numbering?

logos/logos-webservice/src/main/resources/liquibase/changelog/014_model_provider_benchmarks.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    logicalFilePath="liquibase/changelog/012_model_provider_benchmarks.xml"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+    <changeSet id="012-model-provider-benchmarks" author="Tim">
```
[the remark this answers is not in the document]
Tim0x54: Addressed in aa169e4e; the master.xml is now strictly ordered and the Liquibase/CI tests pass.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/repository/ModelProviderRepository.java
```diff
@@ -1,20 +1,103 @@
...
+        SELECT mp.id AS modelProviderId,
+               p.id AS providerId,
+               p.name AS providerName,
+               p.provider_type AS providerType,
+               m.id AS modelId,
+               m.name AS modelName,
+               (p.provider_type = 'logosnode' OR
+                COALESCE(NULLIF(mp.endpoint, ''), NULLIF(p.base_url, '')) IS NOT NULL) AS endpointConfigured,
+               (p.provider_type = 'logosnode' OR
+                COALESCE(NULLIF(mp.api_key, ''), NULLIF(p.api_key, '')) IS NOT NULL) AS authenticationConfigured
```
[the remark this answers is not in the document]
Tim0x54: Legacy values are normalized, meaning they are converted to logosnode before reaching the database. For example, a provider submitted as ollama, node, or logos_worker_node is stored as logosnode. The database column only accepts logosnode or cloud, so checking for logosnode already covers these providers.

logos/logos-orchestrator/pyproject.toml
```diff
@@ -32,6 +32,7 @@ matplotlib = "^3.9.2"
 aiohttp = "^3.9.0"
 prometheus-client = "^0.21.0"
 python-multipart = ">=0.0.20"
+guidellm = {version = "0.7.2", extras = ["recommended"]}
```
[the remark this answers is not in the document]
Tim0x54: GuideLLM 0.7.2 was tested through the complete benchmark workflow on Logos Dev. The generated report contains metrics.request_totals, and a 15-sample run was parsed and stored successfully with 15/15 successful requests. Therefore, no parser change is required.

logos/benchmarks/import_guidellm_results.py
```diff
@@ -0,0 +1,143 @@
...
+        return True
+    if parsed.scheme != "http" or not parsed.hostname:
+        return False
+    hostname = parsed.hostname.rstrip(".").lower()
+    if hostname == "localhost":
+        return True
+    try:
+        return ipaddress.ip_address(hostname).is_loopback
+    except ValueError:
+        return False
```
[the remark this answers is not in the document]
Tim0x54: Fixed in b421013b. Credentialed loopback HTTP imports now use a proxy-free opener (`ProxyHandler({})`), and all import requests use a redirect handler that rejects redirects so the bearer token remains bound to the validated origin. Non-loopback credential transport still requires HTTPS. Covered by `test_loopback_import_disables_proxies_and_redirects`.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -1046,6 +1046,61 @@ async def prepare_lane_for_request(
...
+            if target is not None:
+                if target.runtime_state in {"loaded", "running"} and target.sleep_state != "sleeping":
+                    return True
+                if target.runtime_state in {"stopped", "error", "cold", "sleeping"}:
+                    return False
+
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                return False
+            await asyncio.sleep(min(1.0, remaining))
```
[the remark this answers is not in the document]
Tim0x54: Fixed in b421013b. `_wait_for_benchmark_lane_ready` now inspects the raw matching lane list, selects only non-terminal candidates, and returns `False` immediately when the starting lane transitions to `stopped` or `error`. Covered by `test_benchmark_lane_fails_fast_when_an_existing_start_errors`; the full orchestrator unit suite passes (979 passed, 1 expected xfail).

logos/logos-orchestrator/src/logos/main.py
```diff
@@ -2073,6 +2090,222 @@ class _InternalAddLaneRequest(BaseModel):
...
+            raise HTTPException(status_code=409, detail="Provider-model pair has no valid endpoint")
+        api_key = str(target.get("api_key") or "").strip()
+        if provider_type != "logosnode" and api_key and not credential_transport_is_secure(endpoint):
+            raise HTTPException(
+                status_code=409,
+                detail="Provider credentials require HTTPS (plain HTTP is allowed only on loopback)",
+            )
+
+        db.lock_model_benchmark_provider(provider_id)
+        active = db.find_active_model_benchmark_job(provider_id)
```
wasnertobias, line 2384: A benchmark job that dies outside the runner blocks its provider permanently.

The admission check treats any `pending`/`running` row as exclusive, and the query has no staleness bound:

```sql
WHERE environment = 'model-provider-benchmark'
  AND status IN ('pending', 'running')
```

The 7200s timeout only exists while the asyncio task is alive. If the orchestrator restarts, the task is cancelled, or `guidellm` is killed from outside, the row stays `running` forever — and every subsequent click on that provider returns 409 until somebody edits the `jobs` table by hand. Given that a benchmark is expected to occupy the provider for up to two hours, an orchestrator restart during a run is not an unlikely event.

Two changes would close it:

- treat a row whose `updated_at` is older than the timeout budget as stale on admission, so the next request can take over rather than being locked out;
- mark in-flight benchmark jobs failed on orchestrator shutdown — `create_task` currently has no shutdown hook, so a cancelled task never flips its job status.

The per-provider advisory lock around check-and-insert is correct and worth keeping; it just can't help once the process holding the task is gone.

logos/logos-orchestrator/src/logos/main.py
```diff
@@ -2073,6 +2090,222 @@ class _InternalAddLaneRequest(BaseModel):
...
+    """Authenticate a Spring-to-orchestrator administration call."""
+    if not _INTERNAL_SECRET:
+        raise HTTPException(status_code=403, detail="Internal endpoint disabled")
+    auth_header = request.headers.get("authorization", "")
+    token = (
+        auth_header.removeprefix("Bearer ").strip()
+        if auth_header.lower().startswith("bearer ")
+        else auth_header.strip()
+    )
+    if not hmac.compare_digest(token.encode("utf-8"), _INTERNAL_SECRET.encode("utf-8")):
```
wasnertobias: This encoding fix is right, but it was only applied to some of the call sites.

Comparing two `str` values with `hmac.compare_digest` works only while both are pure ASCII — otherwise it raises `TypeError: comparing strings with non-ASCII characters is not supported`. In this file the two forms now coexist:

| encoded (correct) | raw `str` |
|---|---|
| 1944, 2109, 4535 | 1922, 1969, 2020, 2050, 2320, 2371, 2391 |

At the seven remaining sites, a request carrying `Authorization: Bearer ü` produces a **500 instead of a 401**. That is worse than it sounds for endpoints behind a shared secret: it is trivially reachable by an unauthenticated caller, it distinguishes "this endpoint exists and parsed my header" from a flat rejection, and each attempt writes a traceback into the logs.

Most of those sites predate this PR, so this isn't a regression it introduced — but since it already fixes three of them, finishing the set is a two-character change per line and removes the inconsistency rather than leaving the file half-converted.

wasnertobias (commented):
Reviewed with a focus on the security surface and on what happens when things go wrong at runtime — the parts most likely to bite in production.

**The subprocess runner is well built**, which was my main concern going in:

- `asyncio.create_subprocess_exec` with a fixed argument list and no `shell=True` — no injection path from the model or provider names
- the API key travels in a `scenario.json` written with `chmod(0o600)` inside a `TemporaryDirectory`, not on the command line where `ps` would expose it to every user on the box
- the key is redacted from GuideLLM's output before it reaches the exception, and again before the failure message is stored
- a 7200s timeout with `kill()` and a follow-up `communicate()` so the child is reaped

That is the right level of care for something that shells out with credentials in scope.

Two runtime issues and a few smaller notes inline and below.

**Failures leave no trace in the orchestrator log.** The handler catches, redacts, truncates to 1000 characters and writes the result into `jobs.error_message`. The UI gets a truncated last line; the orchestrator log gets nothing at all. A `logger.exception(...)` after redaction — so the key never reaches the log — is the difference between "GuideLLM failed" and being able to find out why.

**Smaller:**

- `env = os.environ.copy()` hands the orchestrator's full environment to the child. This process holds database and registry credentials, and GuideLLM has no need for them. An allowlist of the variables it actually requires would be tighter.
- `scenario_path.chmod(0o600)` runs after `write_text()`. The enclosing `TemporaryDirectory` is 0700, so nothing is exposed in practice, but creating the file with restrictive permissions from the start removes the ordering dependency entirely.
- Dequeue now sorts the whole per-priority list on every pop to honour affinity. Harmless at today's queue depths, but the heap-head fast path is worth keeping for the common case where no affinity is requested — that is every non-benchmark request.

logos/logos-orchestrator/src/logos/main.py
wasnertobias: It would probably make more sense to keep this flag in memory only, also out of performance perspective?

logos/docker-compose.yaml
```diff
@@ -89,6 +89,7 @@ services:
       LOGOS_NODE_DEV_ALLOW_INSECURE_HTTP: "${LOGOS_NODE_DEV_ALLOW_INSECURE_HTTP:-}"
       PROMETHEUS_API_KEY: "${PROMETHEUS_API_KEY:-}"
       LOGOS_INTERNAL_SECRET: "${LOGOS_INTERNAL_SECRET:-}"
+      LOGOS_BENCHMARK_MAX_CONCURRENCY: "${LOGOS_BENCHMARK_MAX_CONCURRENCY:-1}"
```
wasnertobias: This shall rather be a benchmark setting, not a static env variable

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -1092,12 +1092,106 @@ async def prepare_lane_for_request(
...
+        for queued_model in queued_models:
+            if self._facade.get_scheduler_queue_depth_by_model_name(queued_model, provider_id) > 0:
+                return False
+        return True
+
+    def benchmark_lane_is_safe(self, provider_id: int, target: LaneSchedulerSignals) -> bool:
+        """Return whether a benchmark can run without competing with production."""
+        if target.runtime_state not in {"loaded", "running"} or target.sleep_state == "sleeping":
+            return False
+        max_ttft = float(os.getenv("LOGOS_BENCHMARK_MAX_TTFT_SECONDS", "5"))
```
wasnertobias: default values are pretty low

logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
```diff
@@ -494,6 +494,197 @@ def create_job_record(
...
+            .mappings()
+            .first()
+        )
+        return dict(row) if row else None
+
+    def lock_model_benchmark_provider(self, provider_id: int) -> None:
+        """Serialize global and per-provider benchmark admission in this transaction."""
+        self.session.execute(
+            text("SELECT pg_advisory_xact_lock(:namespace, 0)"),
+            {"namespace": 1428947001},
```
wasnertobias: What is this magic value? Rather log in memory, than in db imho.

logos/logos-orchestrator/src/logos/main.py
Tim0x54: Fixed in 24fea4ef. All compare_digest calls now use UTF-8 bytes, so non-ASCII tokens return 401 instead of 500.

logos/logos-orchestrator/src/logos/main.py
Tim0x54: Addressed in 089a26933: active benchmark tasks and provider sessions are now tracked in memory. The DB entry remains only as persistent crash/restart protection.

logos/docker-compose.yaml
Tim0x54: Removed the global environment setting.
Now, only one benchmark at a time can run per provider.
Benchmarks on different providers can run in parallel.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
Tim0x54: Default values removed.
Benchmark admission now only checks that the lane is loaded and awake, and that the provider has no active or queued production requests.

logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
Tim0x54: Magic value removed, now instead:
Replaced the PostgreSQL lock with a short in-memory admission lock.
The DB is updated every 15 seconds; if updates stop for 60 seconds (e.g. during a restart), the benchmark is marked as failed and the provider is released.

PR #817, opened 2026-08-27
`Logos`: Hold overloaded requests at orchestrator level, route prefix-aware, cancel abandoned requests, and fix the monitoring counters

### LINKED ISSUE #735: `Logos`: When connection breaks cancel request

### LINKED ISSUE #790: `Logos`: Monitoring values are capped to 2 minutes
Request latency percentiles values are seemingly capped at 2 minutes. Especially visible for p95 and p99 values. This could either be a reporting fault or a fault in the grafana dashboard, needs to be explored first.

### LINKED ISSUE #814: `Logos`: Do not instantly forward to workernodes, iff. overloaded
Requests shall be kept at the orchestrator level if too many requests come in instead of being fowarded to workernodes and queued up there for several reasons:
1.) If a workernode wants to restart (wait mode) it takes longer, because once forwarded requests cannot be undone
2.) A higher priority tasks cannot be sorted ahead of already forwarded requests on the workernode level
3.) re-scheduling to another worker is not possible
4.) prefix-caching aware forwarding is not easily possible

Prefix-caching aware routing shall be introduced as well:
When a user a* and a user b* both have different coding agent tasks running then all requests of user a shall be forwarded, if cheaply possible, to the same workernode over and over again as that one will then make use of prefix-caching. This is only relevant if the same model is deployed multiple times and used in parallel. Right now it is "random" where the request lands as Logos does not consider prefix-caching aware routing at all. 

*Another thing which has to be taken into account is that multiple "corresponding stream of requests" (sharing a prefix) may exist in parallel from one and the same API key. It is not sufficient to only look at usernames. Look at api keys, and ideally even look at the actual request prefix within API keys requests.

Use the introduced signals from here: https://github.com/ls1intum/edutelligence/pull/781

wasnertobias:
## Closes #790

Three changes to how the orchestrator places requests and how it lets them go, all driven by the live vLLM lane signals introduced in #781. They share one premise: **a request that has been forwarded to a worker is committed** — it sits in the engine's own queue, where Logos can no longer hand it back (a worker asked to drain must first finish everything already forwarded), reorder it (a later high-priority request cannot jump ahead), re-route it (a peer that frees a slot first cannot take over), or revisit its placement. So: forward less, and be able to take back what was forwarded.

---

## 1. Stop forwarding what the worker cannot start (#814)

`AdmissionDecision(can_admit, headroom, reason)` is computed per model from the routable lanes in the latest runtime snapshot:

| Signal | Gate |
|--------|------|
| `queue_waiting` (vLLM `num_requests_waiting`) | anything already waiting inside the engine → hold (`LOGOS_BACKEND_QUEUE_THRESHOLD`, default 0) |
| `gpu_cache_usage_percent` | ≥ 90% → hold; past this point vLLM preempts and recomputes, which also evicts the prefix-cache blocks §2 depends on (`LOGOS_KV_CACHE_PRESSURE_PERCENT`) |
| `requests_running` vs `num_parallel` | lane full → hold |

`headroom` is how many requests the engine could begin right now, summed over the model's lanes — **a busy lane never masks an idle sibling**, which the old single-lane `queue_waiting > 8` check got wrong. When a worker reports nothing usable (older worker, lane still starting) `headroom` is `None` and the capacity gate alone decides, so the gate is never stricter than the signals justify.

Two call sites:
- `try_reserve_capacity` — the arrival path; a refusal routes to the orchestrator queue.
- `reevaluate_model_queues` — the batch dispatcher after a lane loads/wakes. Batch size is now `min(capacity − active, headroom)`; previously a wake event could drain the whole queue onto one worker in a single pass (capacity is 256 for an unreported vLLM lane).

The release path (slot transfer on completion) is deliberately **not** gated — that is what keeps the engine busy while admission holds new arrivals back, so throughput is preserved.

## 2. Prefix-cache-aware placement (#814)

With one model on several workers, two warm workers tie on corrected score and the tie-break is random — the worst outcome for a coding agent, whose every turn re-sends a prompt the *previous* worker already holds in its KV cache. (Prod confirms it: `prefix_hit` sits at 14–18%.)

A stream is neither a user nor an API key: one key can drive many agent loops at once. Identity is `(api_key_id, actual request prefix)`, hashed into a chain of fixed-size blocks the way an engine hashes its own:

```
block₁ = H(api_key_id ‖ text[0:B])
blockᵢ = H(blockᵢ₋₁  ‖ text[(i−1)B : iB])
```

The prompt is serialized **append-only** (one record per message), so turn *n+1* extends turn *n*'s string rather than rewriting it — a plain `json.dumps` of the message list would not qualify, because the closing bracket moves every turn. Only whole blocks are hashed; the trailing partial block is dropped, exactly as an engine drops its own. Lookup walks deepest-first, so two conversations under one key that share only a system prompt match on the early blocks and separate as soon as their first user turns differ.

A hit does not pin the request — it adds `weight_span × CORRECTION_STRENGTH × 0.25` to the corrected score, worth roughly 15s of expected wait. The stream stays put unless a peer is *meaningfully* faster ("if cheaply possible"). The bonus applies to `WARM` candidates only, so affinity never wakes a sleeping worker or triggers a cold load to stay on the familiar one, and only logosnode placements are recorded — cloud upstreams route internally. The table is soft state (in-memory TTL/LRU, 900s / 20k entries); losing it costs one round of cache misses.

## 3. Forward cancellation to the worker (#735)

Holding requests back shrinks the window but does not close it. A request that *was* forwarded and whose client then goes away still occupies a lane. `_execute_cancelling_on_disconnect` already stops the orchestrator from reading, and for the HTTP path that is enough — closing the httpx context closes the connection and vLLM aborts. Worker requests have no such connection: everything is multiplexed over one WebSocket, so `send_stream_command`'s cleanup only popped a local queue. **No cancel frame existed in the protocol at all** — the worker's `hello` never advertised one, and `_command_tasks` was an unkeyed set, so no single request was even addressable.

Prod, last 6 hours: **156 client disconnects, 135 of which had already been forwarded to a worker.** 16 in the last 30 minutes.

```
orchestrator                              worker
     │  infer_stream (cmd_id=X)              │
     │─────────────────────────────────────▶ │  relay task X ──▶ lane (httpx stream)
     │  ◀──── stream_chunk … ─────────────── │
     ▽  client disconnects                   │
     │  cancel_command (target_cmd_id=X)     │
     │─────────────────────────────────────▶ │  task X cancelled
     │                                       │    └─ httpx stream closed ──▶ vLLM aborts,
     │  ◀──── command_result {cancelled}──── │       KV blocks freed, in-flight count released
```

- **Both paths** — `infer` as well as `infer_stream`; worker command tasks are keyed by `cmd_id` so one request is targeted without touching its neighbours.
- **Feature-gated** — only workers advertising `cancel_command` in their hello are sent one; the orchestrator now records each worker's action list. A rolling upgrade keeps the old behaviour instead of collecting "Unsupported bridge command".
- **Fire-and-forget** — dispatched while a cancelled caller unwinds, so it cannot block that unwind. The reply only says whether anything was still running; a cancel racing a completing stream is normal, not an error.
- **Deterministic close** — the response generator wraps the worker stream in `contextlib.aclosing`. A bare `async for` left the inner generator to the async-generator GC hook, deferring the very cleanup that sends the cancellation. `test_closing_the_response_closes_the_worker_stream_at_once` fails without it.

This also keeps §1 honest: a ghost generation inflates `requests_running` and KV usage, so without it the admission gate would throttle on load that no longer exists.

**Second half of #735** — requests in flight when the orchestrator restarts were left with `result_status IS NULL` and no `timestamp_response` forever, so every live-request view kept counting them (prod currently has 11, oldest from 2026-08-24). Startup now finalises them, before the first request is accepted — after that a NULL status means a request that is genuinely still running. A failure there is logged and swallowed: stale rows are a reporting defect, an orchestrator that will not boot is an outage.

## 4. Monitoring that told the truth about neither number (#790)

Both defects surfaced while verifying the above on the production dashboard.

**`logos_requests_in_flight` only ever rose** (the issue's first comment). The gauge was hand-maintained with paired `inc()`/`dec()` and the pairing did not hold in either direction:

- terminal paths that write their own log row — client disconnect, rate-limit and budget rejects — never told the recorder the request had ended, so each left an increment behind forever;
- a request rejected at classification completed *without ever having been enqueued*, decrementing a count it had never added.

Production over 24h: 434 client disconnects + 33 rate limits ≈ **470 leaked increments a day**. The gauge is now **derived** from the map of tracked requests (`set(len(...))`), so neither drift is expressible, and the failure paths release their request through `_record_log_failure` — the funnel they already all pass through. `_settle` is idempotent, so a path that both discards and completes settles once. A sweep drops anything tracked for more than 2h (well past the 1200s queue-wait bound), so a future missed path degrades into a bounded inaccuracy rather than a number that never comes down.

**Latency percentiles were pinned to exactly two minutes** (the issue body). Not a Grafana fault: `histogram_quantile` cannot report above the highest *finite* bucket, and `logos_request_duration_seconds` topped out at `120.0`. Measured against the log table over 24h:

| | value |
|---|---|
| requests > 120s | 1277 / 24510 (**5.2%**) |
| true p95 | 120.8s |
| true p99 | 300.0s |
| max | 1662.3s |

At 5.2% over the top bucket, p95 lands exactly on the boundary and p99 is clamped to it — precisely "capped at 2 minutes, especially visible for p95 and p99". Buckets now run to 3600s, keeping the sub-10s resolution intact.

## Drive-by fix

`ClassificationCorrectingScheduler._estimate_ettft` called `self._logosnode.get_all_lane_signals(...)`, which does not exist on the facade (it is `get_all_provider_lane_signals`). The `except (KeyError, Exception)` swallowed the `AttributeError`, so `all_provider_lanes` was always `None` and reclaim estimation always fell back to the idle-evict constant, never the 30s busy-drain estimate.

## Changes

**Admission + affinity (#814)**
- `sdi/models.py`: new `AdmissionDecision`
- `sdi/providers/logosnode_provider.py`: `evaluate_admission()` / `_routable_lanes()` / `_lane_admission()` replace the single-lane `_backend_queue_exceeds_threshold`; `try_reserve_capacity` gates on it
- `sdi/logosnode_facade.py`: `evaluate_admission(model_id, provider_id)` passthrough
- `pipeline/base_scheduler.py`: `_admission_headroom()` caps the `reevaluate_model_queues` batch
- `pipeline/prefix_affinity.py` (new): `serialize_prefix`, `affinity_keys`, `PrefixAffinityRouter`
- `pipeline/correcting_scheduler.py`: affinity lookup → bounded score bonus → record placement; `get_all_provider_lane_signals` fix
- `pipeline/pipeline.py`, `scheduler_interface.py`, `main.py`: carry `api_key_id` → `affinity_keys` to the scheduler

**Cancellation (#735)**
- `logos-workernode/logos_bridge.py`: `_command_tasks` keyed by `cmd_id`; `cancel_command` action + handler; advertised in `hello`; cancellation-safe cleanup in `_execute_stream_command`
- `logosnode_registry.py`: `ProviderSession.actions`; `_request_command_cancellation()`; sends on early stream unwind and on a cancelled `send_command`
- `main.py`: `aclosing` around the worker stream; `actions` parsed from `hello`; `_close_orphaned_request_logs()` at startup
- `dbutils/dbmanager.py`: `close_orphaned_request_logs()`

**Observability & config**
- `monitoring/recorder.py`: in-flight gauge derived from the tracked-request map; idempotent `_settle`; metrics-only `discard`; stale-request sweep
- `monitoring/prometheus_metrics.py`: duration buckets extended past the 120s ceiling; `logos_admission_holds_total{reason}`, `logos_prefix_affinity_total{result}`, `logos_worker_cancellations_total{result}` — `unsupported` is the number to watch during a rolling worker upgrade, since it counts the nodes still leaking ghost generations
- `main.py`: `/logosdb/scheduler_state` reports the affinity table's hit rate
- `docs/ettft-scheduling.md`: §10 admission + cancellation, §11 prefix affinity, config + metrics tables
- `.env.example`, `docker-compose*.yaml`: one switch, `LOGOS_PREFIX_AFFINITY_ENABLED`. The thresholds and sizing are derived constants next to `CORRECTION_STRENGTH`, not configuration

## Testing

- **104 new tests**: `test_logosnode_admission.py` (18), `test_prefix_affinity.py` (19), `test_correcting_scheduler_prefix_affinity.py` (8), `test_reevaluate_admission.py` (5), `test_pipeline_affinity_keys.py` (5), `test_worker_stream_cancellation.py` (13), `test_close_orphaned_request_logs.py` (7), `test_orphaned_request_logs.py` (12), `test_in_flight_gauge.py` (11), `test_duration_buckets.py` (3) — plus 8 in the worker-node bridge suite (cancel aborts the upstream and frees the lane on both the streaming and the sync path, targets only its own cmd_id, unknown id is not an error, hello advertises the action)
- Orchestrator: **893 passed, 1 xfailed** (baseline on `main`: 792 / 1)
- Worker node: **509 passed, 2 skipped** — including `test_lane_manager.py`, `test_calibration.py` and `test_model_cache.py`, which CI skips
- black / isort (5.13.2, the pinned hook version) / autoflake / flake8 clean on every touched file

## Database Changes

None. The startup sweep is a plain `UPDATE`, no schema change.

## Rollback

`LOGOS_PREFIX_AFFINITY_ENABLED=false` restores random placement. The forwarding gate has no switch by design — its threshold of 0 follows from what the gate is for, and any other value re-introduces the engine-side queueing it removes. Cancellation has no switch — a worker that does not advertise `cancel_command` simply never receives one, so rolling the worker image back disables it for that node.

https://claude.ai/code/session_011QSSHvHpbcm9P7eCYp8vyu

## Summary by CodeRabbit

- **New Features**
  - Added prefix-cache-aware placement to improve routing for related requests.
  - Abandoned or disconnected requests can now be cancelled on supported workers.
  - Worker capabilities and scheduling diagnostics are more visible.

- **Bug Fixes**
  - Improved cleanup of incomplete request logs and in-flight request tracking.
  - Prevented upstream streams from continuing after client disconnects.
  - Improved admission decisions under queue and cache pressure.
  - Requests can wait briefly for workers returning during reconnects.

- **Monitoring**
  - Added worker-cancellation metrics.
  - Extended request-duration tracking for long-running requests.

- **Documentation**
  - Updated scheduling and configuration guidance to reflect derived settings.


wasnertobias:
Thanks — went through all four. Three were valid and are fixed in c2eb0fd; the fourth describes code that has since been replaced.

**Forward all documented prefix-affinity settings** — valid, fixed. `BLOCK_CHARS`, `MAX_BLOCKS` and `MAX_ENTRIES` were documented in `.env.example` and the config table but never reached the container. All three are now in both compose files.

**MD040 on the fenced blocks** — valid, fixed. The three blocks this PR adds now carry `text`. The pre-existing bare fences elsewhere in the file are left alone.

**Bound record construction before serialization (CWE-400)** — valid, fixed. `_canonical` now takes the outstanding budget and applies it to the *input*: a string leaf is sliced before it is rendered, so a 5 MB `content` costs a 1 KB slice rather than a 5 MB copy. `serialize_prefix` also stops before rendering once the budget is spent, instead of rendering and slicing afterwards.

The part worth flagging for the next reviewer: this had to not cost the **append-only** property, which is what the whole feature rests on — turn *n+1* must extend turn *n*'s string, or every long conversation re-keys itself each turn and the affinity map never hits. It holds, because the fragment sequence and the running budget are identical across turns, so each record renders byte-identically. There is now a test that truncates on purpose (`test_append_only_still_holds_when_the_budget_truncates`, with an assertion that the fixture actually exceeds the limit) plus one on the end-to-end consequence (`test_a_truncated_conversation_keeps_its_keys_across_turns`).

One residual, stated plainly: a large *nested* fragment is still `json.dumps`-ed whole, because `json.dumps` has no incremental form. That is bounded by the body the server already parsed and is the same traversal `estimate_prompt_tokens` performs on every request today, so it is not a new exposure.

**Account for pending reservations before admitting another request** — the specific mechanism no longer exists. That review targeted `headroom = num_parallel - requests_running`, which was removed in e8f5132 for an unrelated reason: `num_parallel` is the concurrency vLLM guarantees at *full context*, not a ceiling. Production has a lane reporting `num_parallel=1` while serving 8 concurrent requests at 78 % KV, so reading it as a limit throttled that lane eightfold. `try_reserve_capacity` no longer consumes a headroom count at all — it gates on `queue_waiting` and KV pressure.

The underlying concern is real and unfixable in the direction suggested. Nothing outside the engine can predict whether vLLM will start a request or park it: that depends on whether the new sequence's KV blocks fit against what the running sequences currently occupy, which is not exposed. The engine answers only afterwards, via `num_requests_waiting`. So the gate is sampled by necessity, and several arrivals inside one heartbeat do read the same "nothing waiting".

What bounds that is the local ledger that already exists — `_model_active` against the parallel capacity — which is exact and updates on every reservation. That is the "pending reservation tracking" the finding asks for; it sits one gate earlier. Adding a second ticket system keyed to snapshot freshness would cap throughput at one request per heartbeat, which costs far more than the overshoot it prevents. Both gates and why they differ are now documented on `try_reserve_capacity`, with `test_the_local_ledger_bounds_what_the_sampled_gate_lets_through` pinning it.

---

Two notes on the automated checks:

- **Title check** says request cancellation "is not represented in the changeset". It is — `logos/logos-workernode/logos_worker_node/logos_bridge.py` (the `cancel_command` action), `logosnode_registry.py` (`_request_command_cancellation`), and `tests/unit/main/test_worker_stream_cancellation.py` (13 tests), all of which appear in the review's own file list. The PR closes #735 alongside #814 and #790.
- **Merge risk / "dispatch queued work without consistently enforcing overload limits"**: the release path is ungated on purpose. A completing request hands its slot straight to the next waiter without re-consulting admission, so a lane under KV pressure keeps serving at its completion rate instead of idling the GPU. The admission gate caps *growth* in concurrency, not service.

PR #830, opened 2026-08-28
`Logos`: Display token counts on the K/M/B/T scale

### LINKED ISSUE #816: `Logos`: Minor reformat in displaying token counts
Tokens are currently displayed as "2470.7M". Once the value exceeds the maximum range of a unit, the next unit of magnitude shall be used, with a space between the numeric value and the unit.

The following scale shall be applied:

K = thousand → 1.000
M = million → 1.000.000
B = billion → 1.000.000.000
T = trillion → 1.000.000.000.000

For example, the currently displayed 2470.7M should be displayed as 2.470 B, (not 2.470,7 B) since the value exceeds the million range.

The unit should always be adjusted to the highest applicable magnitude, while preserving the existing decimal notation of the token value.

wasnertobias:
Fixes #816

## What changed

Token counts were formatted with ad-hoc per-site rules: the statistics KPI card capped at millions (`2470.7M`), the team activity tab capped at billions, and the lane health panel only ever showed `k` (a one-million-token context window would have read `1000k`). All three now share one scale.

- `logos-ui/src/app/features/statistics/statistics.utils.ts`: new `formatTokenCount()` — the single source of truth for the issue's scale: K = 1.000, M = 1.000.000, B = 1.000.000.000, T = 1.000.000.000.000. The unit always steps up to the highest applicable magnitude, a space separates the value from the unit, and the dot stays the decimal separator. The value is truncated to one decimal, dropped when zero, so the 2470.7M the KPI card used to show reads `2.4 B`, `1000` reads `1 K`, and `262144` reads `262.1 K`. One digit after the dot keeps the dot unambiguous (a thousands group is three digits, never one) and keeps the abbreviation shorter than the number it replaces. Counts below 1.000 stay plain; input that is not a positive finite number (null, undefined, NaN, ≤ 0) reads `0`, so every caller inherits the guard.
- `logos-ui/src/app/features/statistics/statistics.ts`: the KPI card's `formatTokenCount` now delegates to the shared util.
- `logos-ui/src/app/features/team-detail/tabs/activity/activity-tab.ts`: `formatTokens` delegates to the shared util, including the null/NaN guard.
- `logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.ts`: `formatContextWindow` keeps its own guard (its "nothing" is `null`, not `0`) and delegates the numeric path to the shared util (a context window is a token count, so it follows the same scale).

Per review, this deliberately deviates from the issue's literal `2.470 B` example in favour of `2.4 B`: three fixed decimals is what makes the dot read like a thousands separator (`1.000 K` = a million in German notation). The unit-stepping rule the issue asks for is unchanged.

No server-side change: the orchestrator and webservice return raw numbers, all formatting happens in the UI.

## Verification

- `statistics.utils.spec.ts` pins the scale end to end: non-finite/null/undefined/≤ 0 → `0`, below 1K stays plain, `1000 → "1 K"`, every K/M/B/T boundary (`"1 M"`, `"1 B"`, `"1 T"`), the issue's example (`2_470_700_000 → "2.4 B"`), one-decimal truncation (`1_999_999_999 → "1.9 B"`, `262_144 → "262.1 K"`), and staying on T above the scale top.
- Updated the existing `formatTokens` assertions in `activity-tab.spec.ts` (including a new `NaN → "0"` case) and `formatContextWindow` assertions in `lane-health-panel.spec.ts`.
- `ng test --watch=false`: 5 spec files, 64 tests, all passing.
- `npm run build` (the job CI runs for logos-ui): passes; the only warnings are pre-existing SCSS budget warnings in files this PR does not touch.

### DIFF
--- a/logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.spec.ts
+++ b/logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.spec.ts
@@ -72,16 +72,16 @@ describe('messageIn', () => {
 /**
  * The context window a lane row shows.
  *
- * Abbreviated on the shared K/M/B/T token scale (issue #816), because the row
- * is a dense line of stats read to spot the roomy lane — the exact token
- * count is never what is being asked.
+ * Abbreviated on the shared K/M/B/T token scale, because the row is a dense
+ * line of stats read to spot the roomy lane — the exact token count is never
+ * what is being asked.
  */
 describe('formatContextWindow', () => {
   it('abbreviates on the token scale', () => {
-    expect(formatContextWindow(262144)).toBe('262.144 K');
-    expect(formatContextWindow(111200)).toBe('111.200 K');
-    expect(formatContextWindow(40960)).toBe('40.960 K');
-    expect(formatContextWindow(1000)).toBe('1.000 K');
+    expect(formatContextWindow(262144)).toBe('262.1 K');
+    expect(formatContextWindow(111200)).toBe('111.2 K');
+    expect(formatContextWindow(40960)).toBe('40.9 K');
+    expect(formatContextWindow(1000)).toBe('1 K');
   });
 
   it('leaves small windows alone', () => {
--- a/logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.spec.ts
+++ b/logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.spec.ts
@@ -72,15 +72,16 @@ describe('messageIn', () => {
 /**
  * The context window a lane row shows.
  *
- * Rounded thousands, because the row is a dense line of stats read to spot the
- * roomy lane — the exact token count is never what is being asked.
+ * Abbreviated on the shared K/M/B/T token scale (issue #816), because the row
+ * is a dense line of stats read to spot the roomy lane — the exact token
+ * count is never what is being asked.
  */
 describe('formatContextWindow', () => {
-  it('abbreviates thousands', () => {
-    expect(formatContextWindow(262144)).toBe('262k');
-    expect(formatContextWindow(111200)).toBe('111k');
-    expect(formatContextWindow(40960)).toBe('41k');
-    expect(formatContextWindow(1000)).toBe('1k');
+  it('abbreviates on the token scale', () => {
+    expect(formatContextWindow(262144)).toBe('262.144 K');
+    expect(formatContextWindow(111200)).toBe('111.200 K');
+    expect(formatContextWindow(40960)).toBe('40.960 K');
+    expect(formatContextWindow(1000)).toBe('1.000 K');
   });
 
   it('leaves small windows alone', () => {
--- a/logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.ts
+++ b/logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.ts
@@ -76,7 +76,7 @@ export interface LaneRow {
   kvColor: string | null;
   ttftColor: string | null;
   ttftLabel: string | null;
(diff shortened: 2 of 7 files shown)


wasnertobias:
The K/M/B/T scale and the shared `formatTokenCount` are the right call — the three formatters that existed before could render the same token count three different ways. Keep the unified method and keep using it everywhere, `formatContextWindow` included. But the format it produces needs a fix before this merges.

**The decimal separator is ambiguous.** `formatTokenCount` always emits three decimals with a dot: `262.144 K`, `1.000 K`, `2.470 B`. In German number formatting the dot is the *thousands* separator, so `262.144 K` reads as 262144 K and `1.000 K` reads as a million — a factor of 1000 off in both cases. Three digits after the dot is exactly the shape of a thousands group, which is what makes it misread.

It also defeats the abbreviation. In `lane-health-panel.html:42` the label is `{{ row.contextLabel }} ctx`, so the row goes from `262k ctx` to `262.144 K ctx`. The docstring still argues the short form saves width, but `262,144` is 7 characters and `262.144 K` is 9 — the abbreviation is now longer than the number it abbreviates.

**Suggestion: cap it at one decimal place**, dropped when it is zero. A single digit after the dot can never be read as a thousands group, because those are always three digits:

```
1000        -> 1K
1500        -> 1.5K
40960       -> 40.9K
262144      -> 262.1K
1234567     -> 1.2M
2470700000  -> 2.4B
```

This still satisfies everything the issue asks for — the unit steps up the moment the value leaves its range, always the highest applicable magnitude — and it keeps the truncation behaviour the tests pin down. It does deviate from the `2.470 B` written in the issue, but that exact notation is what causes the ambiguity, so the issue example is worth revising. If the space before the unit should stay, `262.1 K` works the same way.

An alternative, if you prefer the common convention: three significant digits, i.e. no decimal at all once the mantissa reaches 100 (`262K`, `150M`) and one below that (`1.5K`, `2.4B`).

Minor things while you are in there:

- The `?? TOKEN_COUNT_UNITS[0]` fallback in `formatTokenCount` is unreachable — the `< 1_000` early return guarantees `find` matches at least `K`. If it ever did fire it would return `T`, which is the wrong end of the list. Drop it, or make it `K`.
- `NaN` slips through `activity-tab.formatTokens` (`NaN <= 0` is false) and renders as `NaN T`. `formatContextWindow` guards correctly with `Number.isFinite` — that guard belongs inside `formatTokenCount` itself so every caller inherits it.
- Code comments should not carry issue ids; `issue #816` currently appears in six docstrings and test comments.

wasnertobias:
All points addressed in `5f1006a`:

- **One decimal, dropped when zero** (space kept, as the issue asks): `1000 → "1 K"`, `1500 → "1.5 K"`, `40960 → "40.9 K"`, `262144 → "262.1 K"`, `1234567 → "1.2 M"`, `2470700000 → "2.4 B"`. Truncation stays pinned by tests. As you noted, the issue's literal `2.470 B` example is what causes the ambiguity — the issue text in #816 is worth amending to the one-decimal form; say the word and I'll edit it.
- **Guard moved into `formatTokenCount`**: it now accepts `number | null | undefined` and returns `"0"` for non-finite and non-positive input, so `NaN` can no longer reach the scale from the activity tab — `formatTokens` is a plain delegation now, and `formatContextWindow` keeps only its own guard because its "nothing" is `null`, not `"0"`.
- **Dead fallback gone**: the scale is walked K→T with a `reduce` seeded at K, which the `< 1_000` early return guarantees matches — no unreachable branch, and no wrong-end fallback.
- **Issue ids out of the code comments.**

Verification: `statistics.utils.spec.ts` re-pinned to the new shape (boundaries now `"1 K"`/`"1 M"`/`"1 B"`/`"1 T"`, one-decimal truncation, `NaN`/`null`/`undefined` → `"0"`), plus a `NaN → "0"` case in `activity-tab.spec.ts`. `ng test --watch=false`: 64 tests passing; `npm run build`: passing.

PR #846, opened 2026-08-29
`Logos`: Filter recent requests by state in the statistics dashboard

### LINKED ISSUE #820: `Logos`: Recent requests filter by status
In the statistics dashboard I want to be able to filter for different states: queued, running, error and finished

wasnertobias:
## What changed

The recent-requests feed on the statistics dashboard can now be narrowed to one lifecycle state — **queued**, **running**, **error**, **finished** — via a dropdown in the panel header (default: *All states*).

- **Webservice**
  - `POST /logosdb/latest_requests` accepts a `status` field; the bucket predicate is added to both `findLatestRequests` and `countRequestsInRange`, so paged rows and the reported `total` always describe the same set.
  - The stats v2 websocket gains a `set_feed_status` action. The state is deliberately *not* part of the page scope: it only narrows the request feed, while the KPI cards and the volume chart keep summarising the whole team/user selection. Like the scope, the state is carried on `init`, so a reconnect restores the filter instead of quietly widening the list.
  - A status-filtered feed counts itself: the forced feed push and the REST pages report the bucket's `total`, because the page's KPI aggregate is only as narrow as the team/user scope. An unfiltered feed carries no `total` and the page keeps borrowing the aggregate, exactly as before.
- **UI**
  - The dropdown lives in the recent-requests panel header (`panelRight` slot, next to the existing provider badge and the range badge) and sends `set_feed_status` over the websocket; the page stores the state so it survives reconnects and reaches every paged REST fetch, so page 0 and the history pages agree.
  - Only the feed is marked loading while the filter applies; the KPI cards and charts keep their numbers. The "of N" header figure switches from the KPI aggregate to the pushed bucket total, and the empty state names the active state.

The buckets are read off the same `timestamp_forwarding`/`timestamp_response`/`result_status` pair the team activity view already uses: *queued* = not yet forwarded, *running* = forwarded without a response yet, *error* = answered with a failure (timeouts included, matching the row border colouring), *finished* = answered otherwise. Every row falls into exactly one bucket. An unknown state value matches no bucket and fails closed — the same behaviour as an unknown `user_id` — rather than silently returning the unfiltered feed.

## Why

Issue #820 asks for filtering the recent requests by their states, so an operator can look at exactly the queue, the in-flight generations, the failures or the finished work without paging through everything.

## How it was verified

- **Webservice**: full `mvn test` suite — 258 tests, 0 failures — including new coverage:
  - `RequestLogControllerTest.latestRequests_filtersByStatus` exercises each bucket through the endpoint, the combination of `status` with `team_id`, and the fail-closed unknown-state case (seed rows added via `seed-operations-status.sql`, merged over the shared seed, cleaned up by the existing cleanup script).
  - `StatsV2WebSocketHandlerTest.setFeedStatus_narrowsTheRequestPushAndReportsTheBucketTotal` drives the handler through `init` → `set_feed_status` (queued, finished, cleared) and asserts the pushed rows and the bucket total at each step.
- **UI**: `ng test` (63 tests, incl. a new spec for the pure `normalizeFeedStatus` / `resolveFeedTotal` helpers) and the production `ng build` both pass; the only build warnings are pre-existing bundle-budget warnings in files this PR does not touch.
- Read-through of the websocket lifecycle (init/reconnect, scope change, range change, clearing the filter) to confirm the feed, its total and the KPI aggregates stay consistent; a full deployment is not available in this environment.

Fixes #820

## Summary by CodeRabbit

* **New Features**
  * Added status filtering to the Recent Requests panel for queued, running, error, and finished requests.
  * Filtered results and totals update across live updates and pagination.
  * Empty-state messages now reflect the selected status filter.
  * Token usage displays now show prompt estimates and animate streaming updates.
  * Status filtering does not affect statistics, charts, or other page-wide metrics.

* **Bug Fixes**
  * Improved consistency between displayed request results and their totals when filtering.

### DIFF
--- a/logos/logos-ui/src/app/features/statistics/statistics.ts
+++ b/logos/logos-ui/src/app/features/statistics/statistics.ts
@@ -214,6 +214,12 @@ export class Statistics implements OnInit, OnDestroy {
   setFeedStatusFilter(value: string | null): void {
     const next = normalizeFeedStatus(value);
     if (next === this.feedStatus()) return;
+    // The total the last push reported describes the previous bucket — or,
+    // when clearing, there was none — so it is wrong for this selection and
+    // is dropped until the first push for it lands. Without this the header
+    // keeps advertising the old bucket's count, indefinitely if the socket
+    // is down.
+    this.liveFeedTotal.set(null);
     this.feedStatus.set(next);
     // Only the feed changes, so only it is marked loading — the KPI cards and
     // charts keep their current (unaffected) numbers.
@@ -1103,12 +1109,16 @@ export class Statistics implements OnInit, OnDestroy {
       this.latestRequests.set(payload.requests);
       this.requestsPending.set(false);
     }
-    // A status-filtered feed carries its own total (the page's KPI total is only
-    // as narrow as the team/user scope, not the state bucket). An unfiltered
-    // push has no total, so the borrowed aggregate stays in force.
-    this.liveFeedTotal.set(
-      this.feedStatus() && typeof payload.total === 'number' ? payload.total : null,
-    );
+    // A status-filtered feed carries its own total (the page's KPI total is
+    // only as narrow as the team/user scope, not the state bucket). The
+    // server sends it only when the row set it counts has moved, so a push
+    // without a total keeps the last one: the count cannot have changed in
+    // between. Unfiltered pushes have no total, and switching the filter
+    // clears this signal, so the borrowed aggregate is never shown for a
+    // set it does not describe.
+    if (this.feedStatus() && typeof payload.total === 'number') {
+      this.liveFeedTotal.set(payload.total);
+    }
   }
 
   private handleVramWsInitV2(payload: VramV2Payload): void {
--- a/logos/logos-ui/src/app/features/statistics/statistics.ts
+++ b/logos/logos-ui/src/app/features/statistics/statistics.ts
@@ -24,6 +24,9 @@ import {
   chooseDynamicTargetBuckets,
   extractProviderVramMb,
   formatRangeLabel,
+  normalizeFeedStatus,
+  resolveFeedTotal,
+  REQUEST_STATUS_FILTERS,
   BYTES_PER_GIB,
   BYTES_PER_MIB,
   toGb,
@@ -180,6 +183,42 @@ export class Statistics implements OnInit, OnDestroy {
     }
     return parts.join(' · ');
   });
+
+  // ── Request feed state filter ───────────────────────────────────────────────
+  // One lifecycle bucket the recent-requests list is narrowed to; null shows
+  // every state. Deliberately not part of the page scope: a state filter only
+  // makes sense for the list of individual requests, not for the KPI cards and
+  // charts above it, which must keep summarising the whole team/user selection.
+  readonly feedStatus = signal<string | null>(null);
+
+  readonly feedStatusOptions = computed<AppSelectOption[]>(() => [
+    { value: '', label: 'All states' },
+    ...REQUEST_STATUS_FILTERS.map((s) => ({ value: s, label: s[0].toUpperCase() + s.slice(1) })),
+  ]);
+
+  readonly selectedFeedStatusValue = computed(() => this.feedStatus() ?? '');
+
+  /**
+   * The feed's own total, reported by the live push while a state filter is on.
+   * The unfiltered page borrows the statistics aggregate instead, so this stays
+   * null then and `requestFeedTotal` falls back to that total.
+   */
+  readonly liveFeedTotal = signal<number | null>(null);
+
+  /** Total the feed header shows: the status-filtered count, else the KPI total. */
+  readonly requestFeedTotal = computed(() =>
+    resolveFeedTotal(this.feedStatus(), this.liveFeedTotal(), this.totalRequests()),
+  );
+
+  setFeedStatusFilter(value: string | null): void {
+    const next = normalizeFeedStatus(value);
+    if (next === this.feedStatus()) return;
+    this.feedStatus.set(next);
+    // Only the feed changes, so only it is marked loading — the KPI cards and
+    // charts keep their current (unaffected) numbers.
+    this.requestsPending.set(true);
+    this.statsWs.setFeedStatus(next);
+  }
 
   // ── Raw WS signals ────────────────────────────────────────────────────────────
   readonly stats = signal<RequestLogStats | null>(null);
@@ -836,6 +875,7 @@ export class Statistics implements OnInit, OnDestroy {
       // events of the initial load and then never moves again.
       timelineDeltas: true,
       scope: { userId: this.filterUserId(), teamId: this.filterTeamId() },
+      feedStatus: this.feedStatus(),
       handlers: {
         onVramInit: (p) => this.handleVramWsInitV2(p),
         onVramDelta: (p) => this.handleVramWsDeltaV2(p),
@@ -1059,11 +1099,17 @@ export class Statistics implements OnInit, OnDestroy {
 
   // ── WS handlers ───────────────────────────────────────────────────────────────
 
-  private handleRequestsWsData(payload: { requests?: RequestItem[] }): void {
+  private handleRequestsWsData(payload: { requests?: RequestItem[]; total?: number }): void {
     if (payload.requests) {
       this.latestRequests.set(payload.requests);
       this.requestsPending.set(false);
     }
+    // A status-filtered feed carries its own total (the page's KPI total is only
+    // as narrow as the team/user scope, not the state bucket). An unfiltered
+    // push has no total, so the borrowed aggregate stays in force.
+    this.liveFeedTotal.set(
+      this.feedStatus() && typeof payload.total === 'number' ? payload.total : null,
+    );
   }
 
   private handleVramWsInitV2(payload: VramV2Payload): void {
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/operations/controller/RequestLogController.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/operations/controller/RequestLogController.java
@@ -46,9 +46,12 @@ public class RequestLogController {
         String end = body.get("end") instanceof String s ? s : null;
         Integer userId = body.get("user_id") instanceof Number n ? n.intValue() : null;
         Integer teamId = body.get("team_id") instanceof Number n ? n.intValue() : null;
-        // One of queued/running/error/finished; anything else (or nothing)
-        // matches no bucket, which is the same fail-closed answer an unknown
-        // user_id gives.
+        // The feed's state bucket; absent means all states. A supplied value
+        // that names none of the four buckets matches no rows — the same
+        // fail-closed answer an unknown user_id gives. Blank and non-string
+        // values collapse to the all-states sentinel on purpose, mirroring the
+        // client's normalizeFeedStatus: a picker that stopped showing
+        // everything reads as broken, so the feed widens instead.
         String status = body.get("status") instanceof String s && !s.isBlank() ? s : null;
         String cursorTs = body.get("cursor_ts") instanceof String s ? s : null;
         String cursorId = body.get("cursor_id") instanceof String s ? s : null;
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/operations/controller/RequestLogController.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/operations/controller/RequestLogController.java
@@ -46,12 +46,16 @@ public class RequestLogController {
         String end = body.get("end") instanceof String s ? s : null;
         Integer userId = body.get("user_id") instanceof Number n ? n.intValue() : null;
         Integer teamId = body.get("team_id") instanceof Number n ? n.intValue() : null;
+        // One of queued/running/error/finished; anything else (or nothing)
+        // matches no bucket, which is the same fail-closed answer an unknown
+        // user_id gives.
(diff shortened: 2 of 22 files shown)


logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/operations/controller/RequestLogController.java
```diff
@@ -46,12 +46,16 @@ public ResponseEntity<?> latestRequests(@RequestAttribute("authContext") AuthCon
         String end = body.get("end") instanceof String s ? s : null;
         Integer userId = body.get("user_id") instanceof Number n ? n.intValue() : null;
         Integer teamId = body.get("team_id") instanceof Number n ? n.intValue() : null;
+        // One of queued/running/error/finished; anything else (or nothing)
+        // matches no bucket, which is the same fail-closed answer an unknown
+        // user_id gives.
+        String status = body.get("status") instanceof String s && !s.isBlank() ? s : null;
```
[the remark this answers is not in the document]
wasnertobias: Verified against the current code — one correction and one clarification:

- A supplied status that is a non-blank string but names none of the four buckets is already fail-closed: it is passed through to the query predicate (`findLatestRequests`/`countRequestsInRange`), where it matches no rows — the same answer an unknown `user_id` gives.
- The remaining case — a supplied blank or non-string value collapsing to the `null` all-states sentinel — is deliberate and mirrors the client's `normalizeFeedStatus` (its spec asserts "widens an unknown value back to the full feed"): a stale or hand-typed picker value should show the whole feed rather than an unexplained zero rows. Both surfaces are admin-only, so widening never crosses a privilege boundary.

I did not change the coercion itself, because fail-closing only on the server would split the contract from the client's tested behavior. If the team prefers reject/no-match for supplied malformed values, that needs a matching client-side change — happy to pick that up as a follow-up.

What I did fix: the controller comment claimed "anything else (or nothing) matches no bucket", which was wrong for the absent case (the service documents `null` as all states). The comments in `RequestLogController` and `StatsV2WebSocketHandler` now state the actual semantics (3505c24).

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java
```diff
@@ -444,15 +476,26 @@ private void pushRequests(WebSocketSession session, SessionState state, boolean
             String end = state.timelineLive ? Instant.now().toString() : state.timelineEnd;
             Map<String, Object> payload = requestLogService.getLatestRequests(
                 state.timelineStart, end, state.scopeUserId, state.scopeTeamId,
-                null, null, LATEST_REQUESTS_PUSH_SIZE, false);
+                state.feedStatus, null, null, LATEST_REQUESTS_PUSH_SIZE, false);
```
wasnertobias, line 506: This is the blocker. Passing `state.feedStatus` here also silently narrows the change-detection signal that drives the aggregates.

Twelve lines down, `statsDirty` is set from the signature of *this* payload:

```java
String sig = requestsSig(payload);
boolean changed = !sig.equals(state.prevReqSig);
if (changed) {
    // A request arrived, finished, or grew its usage — whatever the
    // aggregates summarise has moved with it.
    state.statsDirty = true;
}
```

That comment was true while the query was unfiltered. It isn't any more: the signature now describes one bucket, while the aggregates still describe the whole user/team scope.

`statsDirty` is set in exactly one place — line 486, this one — and `tick()` at `:315` pushes aggregates only when it's true. So a filtered session can stop refreshing its KPI cards entirely, and the comment right above that check spells out what that means:

> Without this the page's counters never moved after load: stats only ever came with `timeline_init`, i.e. on connect and on a range change.

The realistic case is the common one: an operator filters to `error` to watch for failures. Successful traffic keeps landing, no new errors arrive, so the error page's signature never changes, `statsDirty` stays `false`, and requests / tokens / cost freeze for the rest of the connection — with nothing in the UI indicating the numbers are stale. Filtering to `queued` behaves the same while a running request finishes.

Fix: keep the dirty signal derived from something scope-wide, and use `feedStatus` only for the rows that go out. A second full `getLatestRequests(..., null, ...)` per push would be expensive at this cadence — a cheap scope-wide probe (max `timestamp_request` plus a row count) is enough to answer "has anything moved". A websocket test that filters to `queued`, mutates an out-of-bucket row, and asserts a stats push still happens would lock it down.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java
```diff
@@ -444,15 +476,26 @@ private void pushRequests(WebSocketSession session, SessionState state, boolean
...
             }
-            if (force || !sig.equals(state.prevReqSig)) {
+            if (force || changed) {
+                // A status-filtered feed counts itself: the total an unfiltered
+                // page borrows from the statistics aggregates is only as narrow
+                // as the user/team scope, not the state bucket, so the "of N"
+                // would promise rows the filter has hidden. Counting is a range
+                // scan, so it runs only when a push actually goes out.
+                if (state.feedStatus != null) {
+                    payload.put("total", requestLogService.countFeedRows(
```
wasnertobias, line 549: `requestsSig` includes token fields, so this range COUNT runs far more often than "when a push goes out" suggests.

While any request in the feed is streaming, its token counts grow, the signature changes on every push, `changed` is true, and this `countFeedRows` executes each time — a range scan over `log_entry` (413k rows) repeated for the whole duration of a generation, for every session that has a filter on.

Gating the recount on the row set actually changing — the `request_id` list, not the token values — would keep the "of N" correct while removing the per-tick scans. The count can't change from tokens growing anyway.

logos/logos-ui/src/app/features/statistics/statistics.ts
```diff
@@ -182,6 +185,42 @@ export class Statistics implements OnInit, OnDestroy {
...
+
+  /** Total the feed header shows: the status-filtered count, else the KPI total. */
+  readonly requestFeedTotal = computed(() =>
+    resolveFeedTotal(this.feedStatus(), this.liveFeedTotal(), this.totalRequests()),
+  );
+
+  setFeedStatusFilter(value: string | null): void {
+    const next = normalizeFeedStatus(value);
+    if (next === this.feedStatus()) return;
+    this.feedStatus.set(next);
```
wasnertobias, line 223: `liveFeedTotal` isn't cleared here, so the header keeps advertising the previous bucket's count.

`resolveFeedTotal(status, pushedTotal, aggregateTotal)` returns `pushedTotal ?? aggregateTotal` whenever a status is set, and `liveFeedTotal` is only ever written from the websocket push. So switching from `queued` (total 7) to `finished` leaves the header reading "of 7" until the next push lands — or indefinitely if the socket is down.

The other direction is odd too: going from "all states" to a filter falls back to the KPI aggregate, so the header can briefly read "of 4,312" while the user asked for errors only.

Clearing `liveFeedTotal` to `null` before applying the new status fixes the stale case. For the in-flight window, the list is already marked pending via `requestsPending` — showing `—` there would be more honest than borrowing a number that describes a different set.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/operations/controller/RequestLogController.java
```diff
@@ -46,12 +46,19 @@ public ResponseEntity<?> latestRequests(@RequestAttribute("authContext") AuthCon
         String end = body.get("end") instanceof String s ? s : null;
         Integer userId = body.get("user_id") instanceof Number n ? n.intValue() : null;
         Integer teamId = body.get("team_id") instanceof Number n ? n.intValue() : null;
+        // The feed's state bucket; absent means all states. A supplied value
+        // that names none of the four buckets matches no rows — the same
+        // fail-closed answer an unknown user_id gives. Blank and non-string
+        // values collapse to the all-states sentinel on purpose, mirroring the
+        // client's normalizeFeedStatus: a picker that stopped showing
+        // everything reads as broken, so the feed widens instead.
+        String status = body.get("status") instanceof String s && !s.isBlank() ? s : null;
```
wasnertobias, line 59: Small thing, but the comment above states the opposite of what this line does.

It says the handling mirrors `normalizeFeedStatus`. It doesn't: the client maps *any* value outside the four buckets to `null` and widens back to all states, while this keeps an unknown non-blank string and lets it fall through every SQL branch, returning no rows.

Both behaviours are individually defensible — the picker widening on a tampered value, the API failing closed on a bad one — and I'm not asking to change the behaviour. But the comment asserts they're the same contract, which is exactly the assumption that would bite whoever next touches either side. Either reword it to say the two deliberately differ, or validate against the four buckets here so they genuinely match.

wasnertobias (commented):
The filter wiring is careful work. Things I checked and found solid:

- **No injection surface.** `:status` is a bind parameter compared against literals, never concatenated.
- **The four buckets are disjoint *and* exhaustive.** I walked every combination of `timestamp_forwarding` / `timestamp_response`, including "rejected before forwarding" (forwarding NULL, response NOT NULL) — it lands cleanly in `error`/`finished`. No gap.
- **Count and rows share the predicate**, with a comment saying why they must.
- **Authorisation is untouched** — `scopeUserId`/`scopeTeamId` still bound the query; the status filter can't widen past them.
- **Every caller of the changed signatures was updated**, including `TeamActivityService`.
- The websocket keeping the filter out of `scope` — so it narrows the feed but not the KPI cards and the volume chart — is the right call and well argued.

One blocker inline, then two smaller things.

**On CI:** `Build Logos Worker Node (vLLM)` is failing, but this PR touches no workernode file (18 files, all UI + webservice), and the same job passes on #849, #852 and #858. The failed log shows only teardown steps, no build error — looks flaky. Worth a rerun to confirm rather than investigating, but it should be green before merge.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java
wasnertobias: Fixed. pushRequests now keeps a scope-wide movement probe separate from the filtered page: while feedStatus is on, it calls a new RequestLogService.scopeMovementSig(...), one aggregate over the exact set the aggregates count — same COALESCE window and scope predicates as findTotals, no status narrowing, no joins — returning the row count plus the newest of the three lifecycle timestamps. statsDirty is derived from that probe (ORed with the page's own signature, so a usage rewrite on a visible row still counts), and the probe's baseline is reset wherever timeline_init re-pushes the aggregates, so a reconnect or range change does not add a redundant stats push on top of the one already sent.

One deliberate departure from the sketch: the probe's timestamp is the GREATEST of timestamp_request, timestamp_forwarding and timestamp_response, not just timestamp_request. With only the request timestamp, the second example would still freeze — filtering to queued while a running request settles moves no enqueue timestamp and no count, and that is exactly when the token and cost totals move.

Locked down with a new unit test that runs the scenario as described: filters to queued, leaves the queued page untouched, advances the scope probe, and asserts the tick sends the stats push (and no requests push, since the page did not change), plus a negative case where a stable scope sends nothing. A TestContainers test pins the probe SQL against the real schema, including the completion-without-new-row case.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java
wasnertobias: Done. pushRequests now keeps the page's row set as its own signature (requestIdsSig — request ids only, values excluded) in state.prevFeedIdsSig, and the countFeedRows range scan runs only when that row set actually changes or on a forced push (filter/scope/range change). Token growth on a running request changes the page signature but not the ids, so the figure the last push carried is kept instead of re-scanning. The client side matches: handleRequestsWsData only updates liveFeedTotal when a push actually carries a total, so the "of N" does not flicker on those token-only pushes. The comment above the check now says the row set, not "a push actually goes out".

Pinned in the new unit test: the forced push after a filter change counts once; a token-only change pushes without calling countFeedRows; a second row appearing triggers the recount.

logos/logos-ui/src/app/features/statistics/statistics.ts
wasnertobias: Both directions fixed. setFeedStatusFilter now clears liveFeedTotal before applying the new status, so the header never keeps advertising the previous bucket's count after a switch — or, when clearing the filter, a count that no longer has a bucket. For the in-flight window, resolveFeedTotal no longer borrows the aggregate for a filtered feed: it returns null until the first push of the new bucket lands, and the feed header renders that as "of —" (new totalInRangeLabel, totalInRange widened to number | null). The list is already marked pending there, so the dash reads as "count arriving", which is what it is.

Tests: the in-flight case in statistics.utils.spec.ts now expects null, and recent-requests.spec.ts gains component-level coverage of totalCount() for the unknown-total window.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/operations/controller/RequestLogController.java
wasnertobias: Reworded. The controller comment and the applyFeedStatus javadoc (which made the same claim) now say the two sides deliberately differ: the client widens on any value outside the four buckets, because a picker that stopped showing everything reads as broken; the server keeps an unknown non-blank string and fails closed, so a value aimed straight at this admin-only endpoint cannot silently widen it to the whole platform. Behaviour unchanged, as you said it should stay.

PR #847, opened 2026-08-29
`Logos`: Show current rate limit usage to application developers

### LINKED ISSUE #672: `Logos`: Application developers shall see the rate limit (usage)
Application developers currently have no way to see the currently set and used rate limits.

wasnertobias:
## Summary

Application developers could see the rpm/tpm limits set for their keys in
My Workspace (key settings with team defaults filled in), but had no way to
see how much of those limits they were currently using. This PR adds the
missing half: current usage against the limits.

`GET /me/keys` now returns, per key, a `rate_limit_usage` object with the
requests and tokens the key sent inside the same 60s sliding window the
orchestrator's rate limiter enforces the limits over:

```json
"rate_limit_usage": {
  "window_seconds": 60,
  "cloud_requests": 2, "cloud_tokens": 2500,
  "local_requests": 1, "local_tokens": 700
}
```

The cloud/local split matches the enforcement exactly: the limiter counts
per key and treats `provider_type = 'logosnode'` as local, everything else
as cloud. The My Workspace key cards now render the used figure against the
limit — `12/60 RPM · 5k/50k TPM` when the key or its team set a limit, the
plain used rate (`12 RPM · 5k TPM`) when it is unlimited — so a developer
can see at a glance how close they are to being throttled.

## Changes

- `logos-webservice/.../identity/repository/RateLimitUsageProjection.java`
  (new): projection for the per-key window usage.
- `logos-webservice/.../identity/repository/ApiKeyRepository.java`:
  `findRateLimitUsageForUser` aggregates `log_entry` + `usage_tokens` for
  the caller's own active keys inside the window, split by provider type.
  Queued/rejected requests without a resolved provider are not counted —
  the limiter does not count them either. Token sums read the same
  `total_tokens` rows the limiter records on completion.
- `logos-webservice/.../identity/service/MeKeysService.java`: fetches the
  usage once per request and merges it into each key's map.
- `logos-ui/.../shared/models/my-key.model.ts`: `RateLimitUsage` model.
- `logos-ui/.../features/my-workspace/my-workspace.{ts,html}`: renders
  used/limit in the CLOUD and LOCAL limit groups.
- Tests: two new cases in `MeKeysControllerTest` plus
  `seed-/cleanup-me-keys-rate-limit-usage.sql` fixtures covering traffic
  inside/outside the window, both provider types, queued requests without a
  provider, and another user's key in the same window.

## Authorization

The data comes from the existing `GET /me/keys` endpoint, which only ever
returns the caller's own keys, and the usage query is scoped to
`api_keys.user_id = :userId` — a team member can only ever see the usage
of their own keys (whose limits already include their team's defaults, as
before). No new route, no new privileges.

## Testing

- `mvn -B test` in `logos-webservice`: 311 tests, 0 failures (including the
  2 new `MeKeysControllerTest` cases against a real Postgres 17 via
  Testcontainers).
- `npm run build` and `npx ng test --watch=false` in `logos-ui`: build
  clean, 92 tests passed.
- `pre-commit` (the logos lint gate) passes on all changed files.

Note: the usage window is 60s by design — the same window
`RateLimitConfig.window_seconds` uses in the orchestrator, so the number
shown is directly comparable to the limit it sits next to.

## Summary by CodeRabbit

- **New Features**
  - Workspace key cards now show current cloud and local request/token usage alongside configured rate limits.
  - Added support for synchronizing cloud models and reporting their context-window limits.
  - Added compatibility checks and calibration log access for supported models and providers.
  - Added latency-aware model scheduling for more informed request routing.

- **Bug Fixes**
  - Requests rejected by token limits no longer consume request-rate capacity.
  - Failed requests are excluded from billable usage, improving cost accuracy.
  - Privacy classification now safely rejects unknown privacy levels.

### DIFF
--- a/logos/logos-orchestrator/src/logos/main.py
+++ b/logos/logos-orchestrator/src/logos/main.py
@@ -1131,6 +1131,23 @@ def _discard_in_flight(request_id: Optional[str], result_status: str) -> None:
         logger.debug("Failed to discard in-flight state for %s", request_id, exc_info=True)
 
 
+def _record_rate_limit_admission(request_id: Optional[str], admitted: bool) -> None:
+    """Persist whether the key's rate limiter admitted the request (issue #672).
+
+    Same monitoring semantics as `_discard_in_flight`: failures are logged
+    and never break the request path.
+    """
+    if not request_id:
+        return
+    pipeline = globals().get("_pipeline")
+    if pipeline is None:
+        return
+    try:
+        pipeline.record_rate_limit_admission(request_id, admitted)
+    except Exception:  # noqa: BLE001 — monitoring must never break a request
+        logger.debug("Failed to record rate-limit admission for %s", request_id, exc_info=True)
+
+
 def _record_log_failure(
     log_id: Optional[int],
     request_id: Optional[str],
@@ -4860,6 +4877,12 @@ async def _execute_resource_mode(
         if rl_info:
             rl_cfg = RateLimitConfig(rpm=rl_info.get("rpm"), tpm=rl_info.get("tpm"))
             allowed, reason = get_rate_limiter().check_and_record(rl_key, rl_cfg)
+            # Persist the admission decision before execution: the /me/keys
+            # usage window (issue #672) counts an admitted request while it is
+            # still running, and must skip the ones this check rejects — the
+            # log row already carries timestamp_forwarding from scheduling, so
+            # without the flag a rejected request would show up as usage.
+            _record_rate_limit_admission(result.scheduling_stats.get("request_id") or request_id, allowed)
             if not allowed:
                 try:
                     _pipeline.scheduler.release(
--- a/logos/logos-orchestrator/src/logos/main.py
+++ b/logos/logos-orchestrator/src/logos/main.py
@@ -1132,7 +1132,7 @@ def _discard_in_flight(request_id: Optional[str], result_status: str) -> None:
 
 
 def _record_rate_limit_admission(request_id: Optional[str], admitted: bool) -> None:
-    """Persist whether the key's rate limiter admitted the request (issue #672).
+    """Persist whether the key's rate limiter admitted the request.
 
     Same monitoring semantics as `_discard_in_flight`: failures are logged
     and never break the request path.
@@ -4878,10 +4878,10 @@ async def _execute_resource_mode(
             rl_cfg = RateLimitConfig(rpm=rl_info.get("rpm"), tpm=rl_info.get("tpm"))
             allowed, reason = get_rate_limiter().check_and_record(rl_key, rl_cfg)
             # Persist the admission decision before execution: the /me/keys
-            # usage window (issue #672) counts an admitted request while it is
-            # still running, and must skip the ones this check rejects — the
-            # log row already carries timestamp_forwarding from scheduling, so
-            # without the flag a rejected request would show up as usage.
+            # usage window counts an admitted request while it is still
+            # running, and must skip the ones this check rejects — the log row
+            # already carries timestamp_forwarding from scheduling, so without
+            # the flag a rejected request would show up as usage.
             _record_rate_limit_admission(result.scheduling_stats.get("request_id") or request_id, allowed)
             if not allowed:
                 try:
--- a/logos/logos-orchestrator/src/logos/monitoring/recorder.py
+++ b/logos/logos-orchestrator/src/logos/monitoring/recorder.py
@@ -295,6 +295,20 @@ class MonitoringRecorder:
         """Attach provider_id once it is resolved (after scheduling)."""
         self._write(request_id, provider_id=provider_id)
 
+    def record_rate_limit_admission(self, request_id: str, admitted: bool) -> None:
+        """Persist whether this key's rate limiter admitted the request.
+
+        Written at the admission decision (``check_and_record``), before
+        execution: the per-key usage window (issue #672) must count an
+        admitted request while it is still running, and must skip the ones
+        the limiter rejects after scheduling — the log row already carries
+        ``timestamp_forwarding`` from ``record_scheduled``, so without this
+        flag a rejected request would show up as usage. Keys without a limit
+        are never checked; their rows keep the column NULL, which the usage
+        query treats as "not rejected".
+        """
+        self._write(request_id, rate_limit_admitted=admitted)
+
     def record_provider_metrics(self, request_id: str, provider_metrics: Dict[str, Any]) -> None:
         """
         Update provider metrics (e.g. Azure rate limits) for a request.
--- a/logos/logos-orchestrator/src/logos/monitoring/recorder.py
+++ b/logos/logos-orchestrator/src/logos/monitoring/recorder.py
@@ -299,9 +299,9 @@ class MonitoringRecorder:
         """Persist whether this key's rate limiter admitted the request.
 
         Written at the admission decision (``check_and_record``), before
-        execution: the per-key usage window (issue #672) must count an
-        admitted request while it is still running, and must skip the ones
-        the limiter rejects after scheduling — the log row already carries
+        execution: the per-key usage window must count an admitted request
+        while it is still running, and must skip the ones the limiter rejects
+        after scheduling — the log row already carries
         ``timestamp_forwarding`` from ``record_scheduled``, so without this
         flag a rejected request would show up as usage. Keys without a limit
         are never checked; their rows keep the column NULL, which the usage
--- a/logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
+++ b/logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
@@ -14,7 +14,7 @@ import { IconTileComponent } from '../../shared/components/icon-tile/icon-tile';
 import { MyKeysService } from '../../core/services/my-keys.service';
 import { TeamManagementService } from '../../core/services/team-management.service';
 import { AuthService } from '../../core/auth/services/auth.service';
-import { MyKey, ModelAccess } from '../../shared/models/my-key.model';
+import { MyKey, ModelAccess, RateLimitUsage } from '../../shared/models/my-key.model';
 import { MyTeam } from '../../shared/models/team.model';
 import { formatLastUsed as formatLastUsedLabel } from '../../shared/utils/date';
 import { isInteractiveClick } from '../../shared/utils/interactive-click';
@@ -312,13 +312,31 @@ export class MyWorkspace implements OnInit {
   }
 
   // ── Display ────────────────────────────────────────────────────────────────
-  formatRpm(rpm: number | null): string {
-    return rpm != null ? rpm.toLocaleString() : '∞';
+  /** The key's usage inside the current rate-limit window, zeroed when absent. */
+  usageFor(key: MyKey): RateLimitUsage {
+    return key.rate_limit_usage ?? {
+      window_seconds: 60,
+      cloud_requests: 0,
+      cloud_tokens: 0,
+      local_requests: 0,
+      local_tokens: 0,
+    };
   }
 
-  formatTpm(tpm: number | null): string {
-    if (tpm == null) return '∞';
-    return tpm >= 1000 ? (tpm / 1000).toFixed(0) + 'k' : tpm.toString();
+  /**
+   * Used/limit pair for one side of a rate limit: "12/60" when the key or its
+   * team set a limit, plain "12" when it is unlimited — the used figure reads
+   * directly against the limit it is enforced by.
+   */
+  formatRpm(used: number, limit: number | null): string {
+    return limit != null
+      ? `${used.toLocaleString()}/${limit.toLocaleString()}`
+      : used.toLocaleString();
+  }
+
+  formatTpm(used: number, limit: number | null): string {
+    const compact = (n: number) => (n >= 1000 ? `${(n / 1000).toFixed(0)}k` : n.toString());
+    return limit != null ? `${compact(used)}/${compact(limit)}` : compact(used);
   }
 
   formatLastUsed(iso: string | null): string {
--- a/logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
+++ b/logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
@@ -312,31 +312,40 @@ export class MyWorkspace implements OnInit {
   }
 
   // ── Display ────────────────────────────────────────────────────────────────
-  /** The key's usage inside the current rate-limit window, zeroed when absent. */
-  usageFor(key: MyKey): RateLimitUsage {
-    return key.rate_limit_usage ?? {
-      window_seconds: 60,
-      cloud_requests: 0,
-      cloud_tokens: 0,
-      local_requests: 0,
-      local_tokens: 0,
-    };
+  /**
+   * The key's usage inside the current rate-limit window, or null when the
+   * backend found nothing for the window. Null is kept distinct from zero on
+   * purpose: a rate-limit figure of zero claims the entire budget is
+   * available, so an unknown window must not be rendered as one.
+   */
+  usageFor(key: MyKey): RateLimitUsage | null {
+    return key.rate_limit_usage;
   }
 
   /**
-   * Used/limit pair for one side of a rate limit: "12/60" when the key or its
-   * team set a limit, plain "12" when it is unlimited — the used figure reads
-   * directly against the limit it is enforced by.
+   * Used/limit pair for one side of a rate limit: "12/60" when a limit is
+   * set, "12/∞" when the key or its team set none (the ∞ marker keeps an
+   * unlimited key readable — a bare used figure looks like the limit is the
+   * used figure), "–/60" when the used figure is unknown for the window.
    */
-  formatRpm(used: number, limit: number | null): string {
-    return limit != null
-      ? `${used.toLocaleString()}/${limit.toLocaleString()}`
-      : used.toLocaleString();
+  formatRpm(used: number | null, limit: number | null): string {
+    const u = used == null ? '–' : used.toLocaleString();
+    const l = limit == null ? '∞' : limit.toLocaleString();
+    return `${u}/${l}`;
   }
 
-  formatTpm(used: number, limit: number | null): string {
-    const compact = (n: number) => (n >= 1000 ? `${(n / 1000).toFixed(0)}k` : n.toString());
-    return limit != null ? `${compact(used)}/${compact(limit)}` : compact(used);
+  /**
+   * Like {@link formatRpm} with compact thousand figures. The used side keeps
+   * one decimal ("59.5k/60k"): at 1k granularity a near-exhausted limit
+   * rounds to the limit and reads as fully used, and small used figures
+   * round up. The limit side stays whole ("60k").
+   */
+  formatTpm(used: number | null, limit: number | null): string {
+    const compact = (n: number, decimals: number) =>
+      n >= 1000 ? `${(n / 1000).toFixed(decimals)}k` : n.toString();
+    const u = used == null ? '–' : compact(used, 1);
+    const l = limit == null ? '∞' : compact(limit, 0);
+    return `${u}/${l}`;
   }
 
   formatLastUsed(iso: string | null): string {
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
@@ -115,48 +115,72 @@ public interface ApiKeyRepository extends JpaRepository<ApiKey, Integer> {
      * actually served on ({@code logosnode} is local, everything else
      * cloud).
      *
-     * The window is cut on {@code timestamp_forwarding}, not
-     * {@code timestamp_request}: the limiter charges a request when it is
-     * admitted (i.e. scheduled, see {@code check_and_record} in the
-     * orchestrator's main.py), and {@code timestamp_forwarding} is written
-     * at scheduling. {@code timestamp_request} is when the request
-     * *arrived*, which under queue wait can be long before admission —
-     * cutting on it would drop exactly the requests that waited in the queue
-     * from the figure, so the display would under-report precisely when
-     * developers look at it to understand a 429. Requests that never reached
-     * scheduling (still queued) have no {@code timestamp_forwarding} and are
-     * left out of the window, matching the limiter, which does not count
-     * them either.
+     * Each metric is cut on the timestamp the limiter charges it by:
+     *
+     * <ul>
+     * <li>Requests count when they are *admitted* —
+     * {@code timestamp_forwarding}, written at scheduling, is when
+     * {@code check_and_record} records the RPM slot (the request is charged
+     * while it is still running). Cutting on {@code timestamp_request}
+     * instead would drop exactly the requests that waited in the queue,
+     * under-reporting precisely when developers look at it to understand a
+     * 429.</li>
+     * <li>Tokens count when the request *completes* —
+     * {@code timestamp_response}: the limiter's {@code record_tokens} runs
+     * at completion, so its TPM window holds completed requests. A request
+     * admitted just before the window edge but completing inside it counts
+     * its tokens (and only them); an in-flight request counts its request
+     * and no tokens yet.</li>
+     * </ul>
+     *
+     * The outer predicate is therefore the union of both windows; each
+     * metric applies its own inside the {@code FILTER}. Requests that never
+     * reached scheduling (still queued) have neither timestamp and are left
+     * out, matching the limiter.
+     *
+     * The per-request token figure mirrors the limiter's fallback in the
+     * {@code record_tokens} call sites (main.py):
+     * {@code total_tokens or (prompt_tokens + completion_tokens)} — Python
+     * {@code or}, so a missing *or zero* total falls back to the two parts,
+     * and a non-zero total wins over the parts (which may not add up to it,
+     * e.g. cached tokens). Computed per log row in the LATERAL so one
+     * request contributes exactly one figure.
      *
      * Rate-limiter rejects are excluded by the persisted admission verdict,
      * not by the timestamp: the orchestrator writes
      * {@code timestamp_forwarding} during {@code record_scheduled()} —
      * *before* {@code check_and_record()} runs — so a request the limiter
-     * rejects afterwards (429) already satisfies the window predicate. The
-     * orchestrator persists {@code rate_limit_admitted} at the decision
-     * point (TRUE admitted, FALSE rejected, NULL when no limit is
-     * configured or for pre-migration rows), and the
+     * rejects afterwards (429) already satisfies the admission-window
+     * predicate. The orchestrator persists {@code rate_limit_admitted} at
+     * the decision point (TRUE admitted, FALSE rejected, NULL when no limit
+     * is configured or for pre-migration rows), and the
      * {@code IS DISTINCT FROM FALSE} filter counts everything except the
-     * explicit rejects — unlimited keys keep showing their usage.
-     *
-     * Token sums read the same {@code total_tokens} rows the
-     * limiter records on completion; request counts use DISTINCT ids because
-     * the usage_tokens join fans one request out to one row per token type.
+     * explicit rejects — unlimited keys keep showing their usage. (The
+     * limiter only records an RPM slot after both checks pass, so no
+     * rejected request consumes budget the display would then miss.)
      */
     @Query(value = """
         SELECT le.api_key_id AS keyId,
-               COUNT(DISTINCT le.id) FILTER (WHERE p.provider_type = 'cloud') AS cloudRequests,
-               COALESCE(SUM(ut.token_count) FILTER (WHERE p.provider_type = 'cloud' AND tt.name = 'total_tokens'), 0) AS cloudTokens,
-               COUNT(DISTINCT le.id) FILTER (WHERE p.provider_type = 'logosnode') AS localRequests,
-               COALESCE(SUM(ut.token_count) FILTER (WHERE p.provider_type = 'logosnode' AND tt.name = 'total_tokens'), 0) AS localTokens
+               COUNT(DISTINCT le.id) FILTER (WHERE p.provider_type = 'cloud' AND le.timestamp_forwarding >= :since) AS cloudRequests,
+               COALESCE(SUM(ut.tokens) FILTER (WHERE p.provider_type = 'cloud' AND le.timestamp_response >= :since), 0) AS cloudTokens,
+               COUNT(DISTINCT le.id) FILTER (WHERE p.provider_type = 'logosnode' AND le.timestamp_forwarding >= :since) AS localRequests,
+               COALESCE(SUM(ut.tokens) FILTER (WHERE p.provider_type = 'logosnode' AND le.timestamp_response >= :since), 0) AS localTokens
         FROM log_entry le
         JOIN api_keys ak ON ak.id = le.api_key_id
         LEFT JOIN providers p ON p.id = le.provider_id
-        LEFT JOIN usage_tokens ut ON ut.log_entry_id = le.id
-        LEFT JOIN token_types tt ON tt.id = ut.type_id
+        LEFT JOIN LATERAL (
+            SELECT CASE
+                     WHEN COALESCE(SUM(CASE WHEN tt.name = 'total_tokens' THEN ut2.token_count END), 0) > 0
+                         THEN SUM(CASE WHEN tt.name = 'total_tokens' THEN ut2.token_count END)
+                     ELSE COALESCE(SUM(CASE WHEN tt.name IN ('prompt_tokens', 'completion_tokens') THEN ut2.token_count END), 0)
+                   END AS tokens
+            FROM usage_tokens ut2
+            LEFT JOIN token_types tt ON tt.id = ut2.type_id
+            WHERE ut2.log_entry_id = le.id
+        ) ut ON true
         WHERE ak.user_id = :userId
           AND ak.is_active = true
-          AND le.timestamp_forwarding >= :since
+          AND (le.timestamp_forwarding >= :since OR le.timestamp_response >= :since)
           AND le.rate_limit_admitted IS DISTINCT FROM FALSE
         GROUP BY le.api_key_id
         """, nativeQuery = true)
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
@@ -115,9 +115,20 @@ public interface ApiKeyRepository extends JpaRepository<ApiKey, Integer> {
      * actually served on ({@code logosnode} is local, everything else
      * cloud).
      *
-     * Requests that have no provider yet (still queued, or rejected before
-     * scheduling) are not counted by the limiter either, so they are left
-     * out here. Token sums read the same {@code total_tokens} rows the
+     * The window is cut on {@code timestamp_forwarding}, not
+     * {@code timestamp_request}: the limiter charges a request when it is
+     * admitted (i.e. scheduled, see {@code check_and_record} in the
+     * orchestrator's main.py), and {@code timestamp_forwarding} is written
+     * at exactly that moment. {@code timestamp_request} is when the request
+     * *arrived*, which under queue wait can be long before admission —
+     * cutting on it would drop exactly the requests that waited in the queue
+     * from the figure, so the display would under-report precisely when
+     * developers look at it to understand a 429. Requests that never reached
+     * admission (still queued, or rejected before scheduling) have no
+     * {@code timestamp_forwarding} and are left out of the window, matching
+     * the limiter, which does not count them either.
+     *
+     * Token sums read the same {@code total_tokens} rows the
      * limiter records on completion; request counts use DISTINCT ids because
      * the usage_tokens join fans one request out to one row per token type.
      */
@@ -134,7 +145,7 @@ public interface ApiKeyRepository extends JpaRepository<ApiKey, Integer> {
         LEFT JOIN token_types tt ON tt.id = ut.type_id
         WHERE ak.user_id = :userId
           AND ak.is_active = true
-          AND le.timestamp_request >= :since
+          AND le.timestamp_forwarding >= :since
         GROUP BY le.api_key_id
         """, nativeQuery = true)
     List<RateLimitUsageProjection> findRateLimitUsageForUser(
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
@@ -119,14 +119,25 @@ public interface ApiKeyRepository extends JpaRepository<ApiKey, Integer> {
      * {@code timestamp_request}: the limiter charges a request when it is
      * admitted (i.e. scheduled, see {@code check_and_record} in the
      * orchestrator's main.py), and {@code timestamp_forwarding} is written
-     * at exactly that moment. {@code timestamp_request} is when the request
+     * at scheduling. {@code timestamp_request} is when the request
      * *arrived*, which under queue wait can be long before admission —
      * cutting on it would drop exactly the requests that waited in the queue
      * from the figure, so the display would under-report precisely when
      * developers look at it to understand a 429. Requests that never reached
-     * admission (still queued, or rejected before scheduling) have no
-     * {@code timestamp_forwarding} and are left out of the window, matching
-     * the limiter, which does not count them either.
+     * scheduling (still queued) have no {@code timestamp_forwarding} and are
+     * left out of the window, matching the limiter, which does not count
+     * them either.
+     *
+     * Rate-limiter rejects are excluded by the persisted admission verdict,
+     * not by the timestamp: the orchestrator writes
+     * {@code timestamp_forwarding} during {@code record_scheduled()} —
+     * *before* {@code check_and_record()} runs — so a request the limiter
+     * rejects afterwards (429) already satisfies the window predicate. The
+     * orchestrator persists {@code rate_limit_admitted} at the decision
+     * point (TRUE admitted, FALSE rejected, NULL when no limit is
+     * configured or for pre-migration rows), and the
+     * {@code IS DISTINCT FROM FALSE} filter counts everything except the
+     * explicit rejects — unlimited keys keep showing their usage.
      *
      * Token sums read the same {@code total_tokens} rows the
      * limiter records on completion; request counts use DISTINCT ids because
@@ -146,6 +157,7 @@ public interface ApiKeyRepository extends JpaRepository<ApiKey, Integer> {
         WHERE ak.user_id = :userId
           AND ak.is_active = true
           AND le.timestamp_forwarding >= :since
+          AND le.rate_limit_admitted IS DISTINCT FROM FALSE
         GROUP BY le.api_key_id
         """, nativeQuery = true)
     List<RateLimitUsageProjection> findRateLimitUsageForUser(
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
@@ -109,8 +109,8 @@ public interface ApiKeyRepository extends JpaRepository<ApiKey, Integer> {
     List<MyKeyProjection> findKeysForUser(@Param("userId") int userId);
 
     /**
-     * Per-key request and token counts inside one rate-limit window (issue
-     * #672), split the way the orchestrator's limiter enforces the limits:
+     * Per-key request and token counts inside one rate-limit window, split
+     * the way the orchestrator's limiter enforces the limits:
      * per key, and cloud vs. local by the provider type the request was
      * actually served on ({@code logosnode} is local, everything else
      * cloud).
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
@@ -1,5 +1,6 @@
 package de.tum.cit.aet.logos.logoswebservice.identity.repository;
 
+import java.sql.Timestamp;
 import java.util.List;
 import java.util.Optional;
 
@@ -12,6 +13,7 @@ import de.tum.cit.aet.logos.logoswebservice.identity.entity.ApiKeyType;
 import de.tum.cit.aet.logos.logoswebservice.identity.entity.LogLevel;
 import de.tum.cit.aet.logos.logoswebservice.identity.repository.MyKeyProjection;
 import de.tum.cit.aet.logos.logoswebservice.identity.repository.ModelAccessProjection;
+import de.tum.cit.aet.logos.logoswebservice.identity.repository.RateLimitUsageProjection;
 
 public interface ApiKeyRepository extends JpaRepository<ApiKey, Integer> {
     Optional<ApiKey> findByKeyValueAndIsActiveTrue(String keyValue);
@@ -105,6 +107,39 @@ public interface ApiKeyRepository extends JpaRepository<ApiKey, Integer> {
         ORDER BY ak.id
         """, nativeQuery = true)
     List<MyKeyProjection> findKeysForUser(@Param("userId") int userId);
+
+    /**
+     * Per-key request and token counts inside one rate-limit window (issue
+     * #672), split the way the orchestrator's limiter enforces the limits:
+     * per key, and cloud vs. local by the provider type the request was
+     * actually served on ({@code logosnode} is local, everything else
+     * cloud).
+     *
+     * Requests that have no provider yet (still queued, or rejected before
+     * scheduling) are not counted by the limiter either, so they are left
+     * out here. Token sums read the same {@code total_tokens} rows the
+     * limiter records on completion; request counts use DISTINCT ids because
+     * the usage_tokens join fans one request out to one row per token type.
+     */
+    @Query(value = """
+        SELECT le.api_key_id AS keyId,
+               COUNT(DISTINCT le.id) FILTER (WHERE p.provider_type = 'cloud') AS cloudRequests,
+               COALESCE(SUM(ut.token_count) FILTER (WHERE p.provider_type = 'cloud' AND tt.name = 'total_tokens'), 0) AS cloudTokens,
+               COUNT(DISTINCT le.id) FILTER (WHERE p.provider_type = 'logosnode') AS localRequests,
+               COALESCE(SUM(ut.token_count) FILTER (WHERE p.provider_type = 'logosnode' AND tt.name = 'total_tokens'), 0) AS localTokens
+        FROM log_entry le
+        JOIN api_keys ak ON ak.id = le.api_key_id
+        LEFT JOIN providers p ON p.id = le.provider_id
+        LEFT JOIN usage_tokens ut ON ut.log_entry_id = le.id
+        LEFT JOIN token_types tt ON tt.id = ut.type_id
+        WHERE ak.user_id = :userId
+          AND ak.is_active = true
+          AND le.timestamp_request >= :since
+        GROUP BY le.api_key_id
+        """, nativeQuery = true)
+    List<RateLimitUsageProjection> findRateLimitUsageForUser(
+            @Param("userId") int userId,
+            @Param("since") Timestamp since);
 
     @Query(value = """
         SELECT m.name AS model_name, p.name AS provider_name, p.provider_type::text AS provider_type
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/service/MeKeysService.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/service/MeKeysService.java
@@ -34,6 +34,12 @@ public class MeKeysService {
      * The sliding window the orchestrator's rate limiter enforces rpm/tpm
      * limits over; the usage numbers shown next to a limit must come from the
      * same window to be comparable to it.
+     *
+     * This is a copy of the default of {@code RateLimitConfig.window_seconds}
+     * in {@code logos/logos-orchestrator/src/logos/rate_limiter.py} — the
+     * orchestrator is the source of truth. Keep the two in sync; the test
+     * {@code RateLimitWindowConsistencyTest} fails if they drift, and the
+     * corresponding comment in rate_limiter.py points back here.
      */
     private static final int RATE_LIMIT_WINDOW_SECONDS = 60;
 
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/service/MeKeysService.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/service/MeKeysService.java
@@ -62,8 +62,11 @@ public class MeKeysService {
     /**
      * Rate-limit usage of the user's active keys inside one rate-limiter
      * window (see {@link #RATE_LIMIT_WINDOW_SECONDS}), keyed by key id.
-     * Keys with no traffic in the window are absent; the caller renders them
-     * at zero.
+     * Keys with no traffic in the window are absent — and stay absent: the
+     * caller must not render them at zero, because zero usage is the most
+     * reassuring possible reading for a rate-limit figure ("you have your
+     * entire budget available"). Unknown usage and genuinely-idle usage
+     * deserve different renderings; the UI shows an en dash for the former.
      */
     private Map<Integer, RateLimitUsageProjection> findRateLimitUsage(int userId) {
         Timestamp since = Timestamp.from(Instant.now().minusSeconds(RATE_LIMIT_WINDOW_SECONDS));
@@ -156,7 +159,7 @@ public class MeKeysService {
         m.put("used_micro_cents", p.getUsedMicroCents());
         m.put("settings", resolvedSettings(p));
         m.put("last_used_at", p.getLastUsedAt() != null ? p.getLastUsedAt().toString() : null);
-        m.put("rate_limit_usage", toRateLimitUsage(usage));
+        m.put("rate_limit_usage", usage == null ? null : toRateLimitUsage(usage));
 
         Map<String, Object> team = new LinkedHashMap<>();
         team.put("id", p.getTeamId());
@@ -170,10 +173,10 @@ public class MeKeysService {
     private Map<String, Object> toRateLimitUsage(RateLimitUsageProjection usage) {
         Map<String, Object> u = new LinkedHashMap<>();
         u.put("window_seconds", RATE_LIMIT_WINDOW_SECONDS);
-        u.put("cloud_requests", usage != null ? usage.getCloudRequests() : 0L);
-        u.put("cloud_tokens", usage != null ? usage.getCloudTokens() : 0L);
-        u.put("local_requests", usage != null ? usage.getLocalRequests() : 0L);
-        u.put("local_tokens", usage != null ? usage.getLocalTokens() : 0L);
+        u.put("cloud_requests", usage.getCloudRequests());
+        u.put("cloud_tokens", usage.getCloudTokens());
+        u.put("local_requests", usage.getLocalRequests());
+        u.put("local_tokens", usage.getLocalTokens());
         return u;
     }
 
--- a/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/service/MeKeysService.java
+++ b/logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/service/MeKeysService.java
@@ -1,5 +1,8 @@
 package de.tum.cit.aet.logos.logoswebservice.identity.service;
 
+import java.sql.Timestamp;
+import java.time.Instant;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -17,6 +20,7 @@ import de.tum.cit.aet.logos.logoswebservice.identity.entity.LogLevel;
 import de.tum.cit.aet.logos.logoswebservice.identity.repository.ApiKeyRepository;
 import de.tum.cit.aet.logos.logoswebservice.identity.repository.ModelAccessProjection;
 import de.tum.cit.aet.logos.logoswebservice.identity.repository.MyKeyProjection;
+import de.tum.cit.aet.logos.logoswebservice.identity.repository.RateLimitUsageProjection;
 import de.tum.cit.aet.logos.logoswebservice.orchestrator.OrchestratorModelWindowClient;
 
 @Service
@@ -25,6 +29,13 @@ public class MeKeysService {
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
     private static final TypeReference<Object> OBJ_TYPE = new TypeReference<>() {};
     private static final int API_KEY_TOKEN_LENGTH = 128;
+
+    /**
+     * The sliding window the orchestrator's rate limiter enforces rpm/tpm
+     * limits over; the usage numbers shown next to a limit must come from the
+     * same window to be comparable to it.
+     */
+    private static final int RATE_LIMIT_WINDOW_SECONDS = 60;
 
     private final ApiKeyRepository apiKeyRepository;
     private final OrchestratorModelWindowClient modelWindowClient;
@@ -35,9 +46,26 @@ public class MeKeysService {
     }
 
     public List<Map<String, Object>> getKeysForUser(int userId) {
-        return apiKeyRepository.findKeysForUser(userId).stream()
-            .map(this::toMap)
+        List<MyKeyProjection> keys = apiKeyRepository.findKeysForUser(userId);
+        Map<Integer, RateLimitUsageProjection> usageByKey = findRateLimitUsage(userId);
+        return keys.stream()
+            .map(p -> toMap(p, usageByKey.get(p.getId())))
             .toList();
+    }
+
+    /**
+     * Rate-limit usage of the user's active keys inside one rate-limiter
+     * window (see {@link #RATE_LIMIT_WINDOW_SECONDS}), keyed by key id.
+     * Keys with no traffic in the window are absent; the caller renders them
+     * at zero.
+     */
+    private Map<Integer, RateLimitUsageProjection> findRateLimitUsage(int userId) {
+        Timestamp since = Timestamp.from(Instant.now().minusSeconds(RATE_LIMIT_WINDOW_SECONDS));
+        Map<Integer, RateLimitUsageProjection> byKey = new HashMap<>();
+        for (RateLimitUsageProjection p : apiKeyRepository.findRateLimitUsageForUser(userId, since)) {
+            byKey.put(p.getKeyId(), p);
+        }
+        return byKey;
     }
 
     @Transactional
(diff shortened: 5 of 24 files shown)


logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
```diff
@@ -140,6 +142,39 @@ AND DATE_TRUNC('month', le.timestamp_request)::date = DATE_TRUNC('month', CURREN
...
+               COUNT(DISTINCT le.id) FILTER (WHERE p.provider_type = 'logosnode') AS localRequests,
+               COALESCE(SUM(ut.token_count) FILTER (WHERE p.provider_type = 'logosnode' AND tt.name = 'total_tokens'), 0) AS localTokens
+        FROM log_entry le
+        JOIN api_keys ak ON ak.id = le.api_key_id
+        LEFT JOIN providers p ON p.id = le.provider_id
+        LEFT JOIN usage_tokens ut ON ut.log_entry_id = le.id
+        LEFT JOIN token_types tt ON tt.id = ut.type_id
+        WHERE ak.user_id = :userId
+          AND ak.is_active = true
+          AND le.timestamp_request >= :since
```
wasnertobias: This filters on when the request *arrived*, but the limiter counts it when it is *admitted* — and those are different moments whenever there is queue wait.

`check_and_record` runs at `main.py:4225`, after the pipeline has scheduled the request and resolved `provider_type`. So a request that waited 70s in the queue and then went through charges the limiter's window *now*, while its `log_entry.timestamp_request` is already 70s old and therefore falls outside `now - 60s` here. It never appears in the displayed figure at all, despite currently occupying a slot in the enforced one.

The direction of the error is the unfortunate one: the display under-reports precisely when queues are long, which is exactly when a developer looks at it to understand why they're getting 429s. They see headroom and get rejected anyway.

That this isn't hypothetical is visible in the current PR stack — #851 introduces queue-wait timeouts answered as 429, so measurable queue wait is a normal operating condition here, not an edge case.

`le.timestamp_forwarding` sits much closer to the moment the limiter actually counts, and the seed fixture already populates it. Worth considering as the filter column instead. If you'd rather keep `timestamp_request`, then the tooltip should say the figure is approximate under load rather than describing itself as the limit's own window.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/service/MeKeysService.java
```diff
@@ -26,6 +30,13 @@ public class MeKeysService {
     private static final TypeReference<Object> OBJ_TYPE = new TypeReference<>() {};
     private static final int API_KEY_TOKEN_LENGTH = 128;
 
+    /**
+     * The sliding window the orchestrator's rate limiter enforces rpm/tpm
+     * limits over; the usage numbers shown next to a limit must come from the
+     * same window to be comparable to it.
+     */
+    private static final int RATE_LIMIT_WINDOW_SECONDS = 60;
```
wasnertobias, line 44: This 60 is a second, independent copy of a value that already exists in the orchestrator:

```python
# logos/logos-orchestrator/src/logos/rate_limiter.py
@dataclass
class RateLimitConfig:
    rpm: Optional[int] = None
    tpm: Optional[int] = None
    window_seconds: int = 60
```

Nothing connects the two. If the limiter's window is ever changed, this constant keeps its old value, the SQL keeps querying the wrong range, and no test on either side fails — the drift is completely silent. It also gets worse than a wrong local value: the constant is serialised out to the client as `window_seconds`, so the wrong window is propagated to the UI and shown to users as fact.

The javadoc correctly states that the usage figures are only meaningful if they come from the same window the limiter enforces. That invariant currently rests on two constants in two services and two languages staying manually in sync.

Cleanest fix would be to have the orchestrator expose its window alongside the limits rather than the webservice assuming it. If that's too much for this PR, at minimum a comment on both sides pointing at each other, so whoever changes one has a chance of finding the other.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
```diff
@@ -312,13 +312,31 @@ export class MyWorkspace implements OnInit {
...
+   * team set a limit, plain "12" when it is unlimited — the used figure reads
+   * directly against the limit it is enforced by.
+   */
+  formatRpm(used: number, limit: number | null): string {
+    return limit != null
+      ? `${used.toLocaleString()}/${limit.toLocaleString()}`
+      : used.toLocaleString();
+  }
+
+  formatTpm(used: number, limit: number | null): string {
```
wasnertobias: The rounding here can read as "at the limit" when you aren't.

`compact()` uses `toFixed(0)`, which rounds to the nearest 1k in both operands independently. With a 60000 TPM limit and 59500 used, this renders `60k/60k` — indistinguishable from being fully exhausted, when there is still 500 TPM of headroom. It rounds up in the other direction too: 1500 used shows as `2k`.

For a "how close am I" readout, 1k-granularity is the wrong precision at exactly the point the number matters. Options: one decimal for the used side (`59.5k/60k`), or a percentage next to the pair, or keep the compact form for the limit but show the used value exactly.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
```diff
@@ -312,13 +312,31 @@ export class MyWorkspace implements OnInit {
   }
 
   // ── Display ────────────────────────────────────────────────────────────────
-  formatRpm(rpm: number | null): string {
-    return rpm != null ? rpm.toLocaleString() : '∞';
+  /** The key's usage inside the current rate-limit window, zeroed when absent. */
+  usageFor(key: MyKey): RateLimitUsage {
```
wasnertobias: Both this fallback and the backend's (`usage != null ? usage.getCloudRequests() : 0L`) turn "no data" into "zero used".

For most fields that's a fine default, but for a rate-limit readout zero is the single most dangerous value to display incorrectly — it tells the developer they have their entire budget available. If the usage lookup ever fails or returns nothing for a reason other than genuine idleness, the UI states the most reassuring possible thing.

A key with no traffic and a key whose usage couldn't be determined deserve different renderings. Keeping `rate_limit_usage: null` distinguishable here and showing `–/60` for it would cost little and fail in the safe direction.

wasnertobias (commented):
Useful feature, and the DB side is done carefully — I checked the parts that usually go wrong here and they're fine:

- **No IDOR.** `WHERE ak.user_id = :userId` scopes the query to the caller's own keys, and `getMyKeys_countsOwnTrafficInsideTheRateLimitWindowOnly` explicitly seeds bob's row 9306 to prove it isn't counted. Good test to have written.
- **The query hits an index.** `idx_log_entry_timestamp_request` (changelog 012) covers the 60s window filter, so this doesn't add a sequential scan on the hottest table in the system.
- **`COUNT(DISTINCT le.id)` is the right call** against the `usage_tokens` fan-out, and `provider_type_enum` really only has `('logosnode','cloud')`, so the two FILTER clauses cover the whole space with no silent under-count.
- I also verified the claim in the query's javadoc about unscheduled requests — it holds. `check_and_record` runs at `main.py:4225`, *after* the pipeline resolved a provider, so requests that never got scheduled genuinely don't hit the limiter either. Nice that this was reasoned through rather than assumed.

My concern is with a single theme that the individual comments below all circle: **the number shown is not the number that gets enforced**, and the gap widens exactly under the load where a developer would consult it. The limiter is an in-memory sliding window in the orchestrator process (`rate_limiter.py`); this reads `log_entry` from the database. Those two agree in the quiet case and drift apart in the interesting one. Details inline on the window filter and the duplicated constant.

Two more things beyond the inline comments:

**The displayed figure silently goes stale.** `/me/keys` is fetched on page load only — no polling, and the client cache TTL is 5s — while the value it renders describes a 60-second sliding window. Leave the workspace page open for two minutes and it shows a window that has entirely passed, with nothing in the UI indicating that. For a widget whose whole purpose is "how much headroom do I have right now", a frozen number with no timestamp is worse than no number. Either refresh it periodically while the page is visible, or render an explicit "as of HH:MM:SS" next to it.

**One pre-existing limiter bug worth knowing about, since it feeds the same drift.** In `check_and_record`, a request rejected for TPM has already had its RPM slot consumed:

```python
if len(req_dq) >= config.rpm:
    return False, "RPM limit reached"     # RPM reject: not counted
req_dq.append(now)                        # RPM counted here
if total >= config.tpm:
    return False, "TPM limit reached"     # TPM reject: RPM stays counted
```

Not introduced by this PR, but still something I'd ask you to fix inside it, as it means enforced RPM could exceed what any log-derived view will ever show.

**Minor:** the javadoc on `findRateLimitUsageForUser`, the test section comment in `MeKeysControllerTest`, and the header of `seed-me-keys-rate-limit-usage.sql` each carry an `(issue #672)` reference. Per the project convention those shouldn't go into code, docstrings or tests — the PR body is the right place for the link.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
```diff
@@ -312,13 +312,31 @@ export class MyWorkspace implements OnInit {
...
-    return tpm >= 1000 ? (tpm / 1000).toFixed(0) + 'k' : tpm.toString();
+  /**
+   * Used/limit pair for one side of a rate limit: "12/60" when the key or its
+   * team set a limit, plain "12" when it is unlimited — the used figure reads
+   * directly against the limit it is enforced by.
+   */
+  formatRpm(used: number, limit: number | null): string {
+    return limit != null
+      ? `${used.toLocaleString()}/${limit.toLocaleString()}`
+      : used.toLocaleString();
```
wasnertobias: This branch drops the unlimited marker that the current code has.

On `main`:

```ts
formatRpm(rpm: number | null): string {
  return rpm != null ? rpm.toLocaleString() : '∞';
}
```

A key with no limit rendered as `∞ RPM · ∞ TPM`. With this change the same key renders as just the used figure — so an unlimited key that hasn't been used yet now displays **`0 RPM · 0 TPM`**.

That reads as a cap of zero, which is the opposite of the truth: it suggests the key can't send anything, when in fact it has no ceiling at all. And a used-but-unlimited key showing `12 RPM` is ambiguous in the other direction — it looks like the limit *is* 12. The tooltip is currently the only thing distinguishing the two cases, and tooltips don't help someone scanning the card.

The used figure and the unlimited marker aren't in conflict; both fit: `12/∞`, or `12 · ∞`. `formatTpm` on the line below has exactly the same gap.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
```diff
@@ -140,6 +142,39 @@ AND DATE_TRUNC('month', le.timestamp_request)::date = DATE_TRUNC('month', CURREN
...
+     * Requests that have no provider yet (still queued, or rejected before
+     * scheduling) are not counted by the limiter either, so they are left
+     * out here. Token sums read the same {@code total_tokens} rows the
+     * limiter records on completion; request counts use DISTINCT ids because
+     * the usage_tokens join fans one request out to one row per token type.
+     */
+    @Query(value = """
+        SELECT le.api_key_id AS keyId,
+               COUNT(DISTINCT le.id) FILTER (WHERE p.provider_type = 'cloud') AS cloudRequests,
+               COALESCE(SUM(ut.token_count) FILTER (WHERE p.provider_type = 'cloud' AND tt.name = 'total_tokens'), 0) AS cloudTokens,
```
wasnertobias: This sums only `total_tokens`, but the limiter doesn't.

All three `record_tokens` call sites compute the figure with a fallback:

```python
# main.py:3400, :3569, :3804
total = usage_tokens.get("total_tokens") or (
    usage_tokens.get("prompt_tokens", 0) + usage_tokens.get("completion_tokens", 0)
)
```

And that fallback isn't defensive padding — it's load-bearing. `extract_token_usage` copies provider keys through as they arrive, and `_RESPONSES_USAGE_KEY_MAP` (`responses.py:233-236`) maps only `input_tokens → prompt_tokens` and `output_tokens → completion_tokens`. Nothing anywhere derives a `total_tokens` when the provider doesn't send one. Note also that Python's `or` makes a stored `total_tokens` of 0 fall through to the sum as well.

So for any provider that reports prompt/completion without a total, this query returns **0 tokens** for that traffic while the limiter counts every one of them and throttles on them. The displayed TPM would sit at `0/60000` while requests are being rejected with 429 — the single most confusing state this feature could produce, and precisely the situation it was built to explain.

`COALESCE` over a `total_tokens` sum falling back to `SUM(prompt_tokens) + SUM(completion_tokens)` would mirror what is actually enforced. The same applies to `localTokens` two lines down.

wasnertobias (commented):
Follow-up to my earlier review — two things I missed the first time, both worth acting on, plus a correction to the fix I suggested.

**Correction to my own suggestion.** I proposed filtering on `timestamp_forwarding` instead of `timestamp_request`. That's incomplete: no single column can match the limiter, because the limiter records the two figures at different moments. RPM is counted in `check_and_record` (`main.py:4225`, right after scheduling, before execution), TPM in `record_tokens` at completion (`main.py:3406`, `3572`, `3807`). So `timestamp_forwarding` is the right anchor for the request counts and `timestamp_response` for the token sums — matching both means two different time filters, not one. Whichever way this goes, it's worth saying plainly in the query's javadoc that this is a reconstruction from the log rather than a read of the limiter's own state, so the next person doesn't treat it as authoritative.

**Two smaller notes:**

- `usageFor(key)` is called four times per card in the template, and on the fallback path it allocates a fresh object literal on every call — repeated on every change-detection cycle. The project is on Angular 21.2, so `@let u = usageFor(key)` binds it once and removes both the repetition and the allocations.
- The query's javadoc says `logosnode` is local and "everything else cloud", but the SQL matches `p.provider_type = 'cloud'` positively. Those agree today because the enum only has the two values, so nothing is wrong right now — but the comment describes a rule the code doesn't implement. Either `<> 'logosnode'` or a reworded comment would keep them honest.

Nothing here changes my overall read: the auth scoping, the query shape and the Java tests remain the strongest part of this change.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
wasnertobias: Adopted your suggestion — the window is now cut on `timestamp_forwarding`.

I verified the moment it is written: the pipeline stamps `scheduled_ts` (→ `timestamp_forwarding`) when the request is scheduled (`record_scheduled` in `monitoring/recorder.py`), and `check_and_record` runs immediately after scheduling in `main.py` — so the two moments coincide to the millisecond, unlike arrival, which precedes admission by the full queue wait.

Concrete effects of the change:

- A request that waited in the queue and was admitted inside the window is now counted, as the limiter charges it. The seed fixture pins exactly this case: row 9307 arrived 70s ago but was admitted 10s ago, so the test now expects 3 cloud requests / 2600 cloud tokens instead of 2 / 2500.
- Requests that never reached admission (still queued, or rejected before scheduling) have a NULL `timestamp_forwarding` and drop out of the window automatically — which matches the limiter, which does not count them either. The javadoc on the query now documents both effects and why arrival time was wrong.

The tooltip ("usage is what the last minute (the limit's window) counted") is left as-is: with the filter aligned to the counting moment, that description is now accurate instead of an approximation.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/service/MeKeysService.java
wasnertobias: Took the minimal-but-enforced option you outlined: cross-referencing comments on both sides plus a test that fails on drift.

- `RateLimitConfig.window_seconds` in `rate_limiter.py` now carries a comment marking it the source of truth and pointing at `MeKeysService.RATE_LIMIT_WINDOW_SECONDS`; the javadoc there points back and names the test.
- New `RateLimitWindowConsistencyTest` (plain JUnit, no Spring context) reads `logos-orchestrator/src/logos/rate_limiter.py` from the checkout, parses the `window_seconds: int = <N>` default, and asserts it equals the Java constant. I verified it catches drift: with the Python default temporarily changed to 90 the test fails with an explicit "update both together" message, and changing the Java constant fails the same assertion.

On the cleaner option (orchestrator exposing the window alongside the limits): the limits themselves come from the DB (key settings + team defaults) rather than from an orchestrator endpoint, so wiring the window through would mean new orchestrator surface plus a webservice client with a fallback for when the orchestrator is unreachable — I judged that too much for this PR. The drift test closes the silent-failure gap in the meantime, and exposing the window via the orchestrator is worth a follow-up if the window ever becomes configurable at runtime (today it is only ever the dataclass default — `main.py` also emits it as `Retry-After` from that same default).

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
```diff
@@ -140,6 +142,50 @@ AND DATE_TRUNC('month', le.timestamp_request)::date = DATE_TRUNC('month', CURREN
...
+     * {@code timestamp_forwarding} and are left out of the window, matching
+     * the limiter, which does not count them either.
+     *
+     * Token sums read the same {@code total_tokens} rows the
+     * limiter records on completion; request counts use DISTINCT ids because
+     * the usage_tokens join fans one request out to one row per token type.
+     */
+    @Query(value = """
+        SELECT le.api_key_id AS keyId,
+               COUNT(DISTINCT le.id) FILTER (WHERE p.provider_type = 'cloud') AS cloudRequests,
```
Claudia-Anthropica: @wasnertobias [medium] This counts requests that the limiter never admitted. `_pipeline.process()` writes `timestamp_forwarding` and the provider during `record_scheduled()` before `check_and_record()` runs; therefore an RPM-rejected request already satisfies this query even though it was not appended to the request deque. Repeated 429 attempts can consequently drive the displayed count arbitrarily above the enforced usage.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java`, the cloud and local request counts include scheduled log rows that the limiter subsequently rejects. Persist whether rate-limit admission succeeded and filter both request counts to admitted rows.

</details>

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
```diff
@@ -140,6 +142,50 @@ AND DATE_TRUNC('month', le.timestamp_request)::date = DATE_TRUNC('month', CURREN
...
+               COUNT(DISTINCT le.id) FILTER (WHERE p.provider_type = 'logosnode') AS localRequests,
+               COALESCE(SUM(ut.token_count) FILTER (WHERE p.provider_type = 'logosnode' AND tt.name = 'total_tokens'), 0) AS localTokens
+        FROM log_entry le
+        JOIN api_keys ak ON ak.id = le.api_key_id
+        LEFT JOIN providers p ON p.id = le.provider_id
+        LEFT JOIN usage_tokens ut ON ut.log_entry_id = le.id
+        LEFT JOIN token_types tt ON tt.id = ut.type_id
+        WHERE ak.user_id = :userId
+          AND ak.is_active = true
+          AND le.timestamp_forwarding >= :since
```
Claudia-Anthropica: @wasnertobias [medium] No index supports this new `timestamp_forwarding` range. The schema has `idx_log_entry_api_key_id` and indexes on `timestamp_request`, so PostgreSQL must visit and filter every historical log row for each user's keys before performing the usage joins. For long-lived high-volume keys, every `/me/keys` load becomes progressively more expensive.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java`, the new 60-second predicate has no supporting `timestamp_forwarding` index and scans each key's history. Add a Liquibase migration that creates an index on `(api_key_id, timestamp_forwarding)` concurrently.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The exact-head build, test, and lint checks succeeded, and the query remains scoped to the caller. However, the aggregation counts scheduled requests that never entered the limiter and lacks an index supporting its new timestamp predicate, causing incorrect usage figures and progressively slower key-page loads. Please address the two inline findings before merge.

Claudia-Anthropica (changes requested):
@wasnertobias The persisted admission verdict and composite forwarding index resolve the two prior Claudia findings, so those threads can be closed. However, the current query still selects TPM rows by `timestamp_forwarding`, although the limiter records tokens at completion, and it only sums `total_tokens` despite the limiter's prompt-plus-completion fallback. This retained correctness feedback means the displayed TPM can remain below enforced usage, so the current head still needs changes.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
Claudia-Anthropica: @wasnertobias Thanks, this is fixed now: the orchestrator persists the admission verdict and the usage query excludes explicit rejects.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
Claudia-Anthropica: @wasnertobias Thanks, this is fixed now: migration 020 adds the `(api_key_id, timestamp_forwarding)` index and the baseline test verifies it.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
wasnertobias: Fixed in `71fbbde74` — the admission verdict is now persisted, and the query filters on it.

- New nullable column `log_entry.rate_limit_admitted`: `TRUE` = limiter admitted, `FALSE` = rejected, `NULL` = never checked (no limit configured) or pre-migration rows.
- The orchestrator writes it at the decision point — in `main.py` immediately after `check_and_record(...)` returns, before execution (new `recorder.record_rate_limit_admission`). Writing before execution is deliberate: an admitted request must appear in the window while it is still running (that is the point of the readout), and a rejected one is marked before the 429 goes out. Failures are swallowed like the other monitoring writes.
- The query now filters `rate_limit_admitted IS DISTINCT FROM FALSE`: everything except explicit rejects counts, so unlimited keys (`NULL`) keep showing usage and pre-migration history is unaffected. A plain `= TRUE` would have silently zeroed every unlimited key.
- `result_status` cannot carry this: an in-flight admitted row still has a `NULL` status and must count, while an admitted request that later errors did consume the limiter's slot and counts as well — the verdict at admission is the only faithful signal.

Tests: the seed fixture now contains two in-window rejects (`9308` cloud, `9309` local, both with token usage) that must not count, an admitted-after-queue-wait row (`9307`, `TRUE`) that must, and the `NULL` rows that must keep counting — the existing cloud/local assertions now encode the exclusion. A recorder unit test pins that `False` reaches the DB layer as `False`, not swallowed like an unset value.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
wasnertobias: Fixed in `71fbbde74`: changelog `020` adds `idx_log_entry_api_key_timestamp_forwarding ON log_entry (api_key_id, timestamp_forwarding)` via `CREATE INDEX CONCURRENTLY` in a `runInTransaction="false"` changeset, guarded by an `indexExists` precondition (same house pattern as `012`/`017`), with a `DROP INDEX CONCURRENTLY IF EXISTS` rollback.

One note on the numbering: it is deliberately `020` — `018`/`019` are already taken by the in-flight Ollama-removal branch (#865), and `LiquibaseChangelogOrderTest` requires strictly increasing include numbers, so either merge order stays valid. `LiquibaseBaselineTest.migration020_rateLimitAdmittedColumnAndForwardingIndexExist` pins both the column and the index.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
wasnertobias: Still open — not addressed in the 08-31 fix round (that round covered the two query findings). Keeping this for a follow-up; the one-decimal used-side option looks like the cheapest of the three.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
wasnertobias: Still open — not addressed in the 08-31 fix round. The `–/limit` rendering for undetermined usage is the right direction; it needs the backend to keep `null` distinct from zero, so it goes to the follow-up with the other UI points.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
wasnertobias: Still open — not addressed in the 08-31 fix round. Restoring the unlimited marker for null limits (`used/∞`) goes to the follow-up with the other UI points.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
wasnertobias: Still open — not addressed in the 08-31 fix round. The Responses-API key-mapping point is the decisive one: the query needs the per-request `total_tokens OR (prompt_tokens + completion_tokens)` fallback to mirror what the limiter actually enforces. Going to the follow-up.

wasnertobias:
Both points from the Aug 31 review round are addressed in `71fbbde74`:

1. **Rejected requests no longer count as usage** — the limiter's admission verdict is persisted on `log_entry.rate_limit_admitted` (`TRUE`/`FALSE`, `NULL` = no limit configured / pre-migration) at the `check_and_record` decision point, and the window query filters `IS DISTINCT FROM FALSE`. In-flight admitted requests still count while running; unlimited keys are unaffected.
2. **Index for the window range** — changelog `020` creates `idx_log_entry_api_key_timestamp_forwarding (api_key_id, timestamp_forwarding)` concurrently (numbered `020` because `018`/`019` are taken by #865 and the changelog-order test requires strictly increasing numbers).

Also merged `main` (`474e614e8`). Local verification: orchestrator unit suite 1098 passed (+1 pre-existing xfail), webservice suite 336 passed (incl. the new `migration020` baseline test and the rejected-row seed cases), pre-commit clean.

Claudia-Anthropica (changes requested):
@wasnertobias The persisted verdict and composite index resolve the two replied-to Claudia findings, and exact-head CI succeeded. The aggregation still windows token usage by `timestamp_forwarding` and sums only `total_tokens`, although the limiter records tokens at completion and falls back per request to prompt plus completion; additionally, a TPM rejection is marked `FALSE` only after its RPM timestamp has already been appended, so excluding all `FALSE` rows under-reports enforced RPM. These retained correctness defects can show headroom while the limiter is rejecting traffic, so changes are still required.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
Claudia-Anthropica: @wasnertobias Confirmed on this head: the admission verdict is written before execution and the query excludes explicit rejects, so the original scheduled-reject overcount is fixed.

wasnertobias: The RPM side had a real gap the persisted verdict alone could not close, and it is fixed in `79a8c8e1f`: the limiter checked (and recorded) the RPM slot *before* checking TPM, so a request the TPM limit rejected still consumed an RPM slot. The enforced RPM then included requests the display — admitted rows only — excludes, so the UI could show headroom while 429s kept going out.

`check_and_record` now runs the TPM check first and appends the RPM timestamp only after both checks pass: a rejected request consumes no budget at all, and the displayed RPM equals the enforced one by construction. New unit tests in `tests/unit/test_rate_limiter.py` pin both halves — after TPM rejects, the request window is still empty once the window slides (both RPM slots available), and with both limits exhausted the reject reason is TPM, not RPM.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
wasnertobias: Confirmed resolved — nothing further needed here: changelog 020's `idx_log_entry_api_key_timestamp_forwarding (api_key_id, timestamp_forwarding)` covers the range the query filters on.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
wasnertobias: Both halves are fixed in `79a8c8e1f`:

- **Fallback**: the token figure is now computed per log row in a LATERAL subquery exactly like the limiter's `record_tokens` call sites: `total_tokens` when it is non-zero, else `prompt_tokens + completion_tokens` — the Python `or` semantics, so a stored zero total falls through to the parts, and a non-zero total wins over the parts even when they do not add up to it (e.g. cached tokens).
- **Window**: the token sums are cut on `timestamp_response` (completion), because the limiter's `record_tokens` runs at completion — its TPM window holds requests *completed* inside the window, not requests *admitted* inside it. Request counts stay cut on `timestamp_forwarding` (admission). The outer predicate is the union of both windows, so a request admitted just before the edge but completing inside it contributes its tokens and nothing else.

The seed fixture pins each case: `9310` admitted 120s ago / completed 20s ago (tokens count, request doesn't), `9311` in flight (request counts, no tokens yet), `9313` parts-only (500), `9314` zero total → parts (250), `9315` non-zero total wins (90, not 70), `9316` the local parts-only case (250). The test assertions now read cloud 7 requests / 3840 tokens, local 2 / 950.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
wasnertobias: Fixed in `79a8c8e1f` with the one-decimal used-side option: `formatTpm` now renders the used figure with one decimal when it is compacted (`59.5k/60k`, `1.5k/60k`), while the limit side stays whole (`60k`), so a near-exhausted limit no longer rounds into the limit and small used figures no longer round up. Sub-1k figures remain exact.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
wasnertobias: Fixed in `79a8c8e1f`: the backend no longer collapses a missing usage row into zeros — `rate_limit_usage` is now `null` when the window query found nothing for the key, and the UI renders the used side as `–` for it (`–/60`, `–/∞`), so an unknown window can no longer read as "you have your entire budget available". `MeKeysControllerTest` pins the null in `getMyKeys_keepsUnknownRateLimitUsageDistinctFromZero`.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
wasnertobias: Fixed in `79a8c8e1f`: `formatRpm` and `formatTpm` now render `used/∞` whenever the limit is null (`12/∞` for a used-but-unlimited key, `–/∞` when the window is unknown as well), so an unlimited key can no longer read as a zero cap, and a used figure is no longer ambiguous with a limit. The card tooltip now explains `–` and `∞` as well.

wasnertobias:
All three retained findings from the 03:30 review are fixed in `79a8c8e1f`, together with the four follow-up points from the Aug 29 round that were left open:

1. **Tokens windowed on completion** — token sums are now cut on `timestamp_response` (when the limiter's `record_tokens` runs) while request counts stay cut on `timestamp_forwarding` (admission); the outer predicate is the union of both windows, so a request admitted just before the edge but completing inside it contributes its tokens and nothing else.
2. **Limiter's token fallback mirrored** — per request: `total_tokens` when non-zero, else `prompt_tokens + completion_tokens` — the exact Python `or` semantics from the `record_tokens` call sites (zero total falls through, non-zero total wins over the parts).
3. **TPM rejects no longer consume RPM slots** — `check_and_record` checks TPM before recording the RPM slot, so the displayed RPM equals the enforced RPM by construction; pinned by the new `tests/unit/test_rate_limiter.py` (5 tests).
4. **UI follow-ups** — one-decimal used-side TPM rounding (`59.5k/60k`), `rate_limit_usage` stays `null` when the window has no data (rendered `–/limit` instead of `0/limit`), and `used/∞` restored for null limits.

Local verification: orchestrator unit suite 1103 passed (+1 pre-existing xfail, incl. the 5 new limiter tests), webservice suite 336 passed (incl. the extended window/fallback seed fixture), `ng build` and the UI unit suite green, pre-commit clean.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
```diff
@@ -140,6 +142,86 @@ AND DATE_TRUNC('month', le.timestamp_request)::date = DATE_TRUNC('month', CURREN
...
+                         THEN SUM(CASE WHEN tt.name = 'total_tokens' THEN ut2.token_count END)
+                     ELSE COALESCE(SUM(CASE WHEN tt.name IN ('prompt_tokens', 'completion_tokens') THEN ut2.token_count END), 0)
+                   END AS tokens
+            FROM usage_tokens ut2
+            LEFT JOIN token_types tt ON tt.id = ut2.type_id
+            WHERE ut2.log_entry_id = le.id
+        ) ut ON true
+        WHERE ak.user_id = :userId
+          AND ak.is_active = true
+          AND (le.timestamp_forwarding >= :since OR le.timestamp_response >= :since)
```
Claudia-Anthropica, line 183: @wasnertobias [medium] This new completion-time branch reintroduces the unindexed history scan. Migration 020 only creates `(api_key_id, timestamp_forwarding)`, while this `OR` also selects rows by `timestamp_response`; no index supports that branch, so PostgreSQL must visit and filter historical rows for each key to find recent token completions, making `/me/keys` progressively slower as the history grows.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java`, the completion-window half of the usage predicate has no `(api_key_id, timestamp_response)` index and therefore scans each key's history. Add a Liquibase migration that creates this composite index concurrently and cover it in the baseline migration test.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The current head resolves the retained admission, token-window/fallback, and UI correctness issues; the captured test and lint checks that completed succeeded, while several non-required build/webservice checks remained in progress. One performance defect remains: the newly added completion-time branch lacks a supporting composite index, so `/me/keys` can still scan each key's historical rows.

wasnertobias:
Merged `origin/main` into this branch to clear the reported merge conflict.

The only conflict was in `logos/logos-webservice/src/main/resources/liquibase/changelog/master.xml`: main (#852) added the `018_model_aliases.xml` include and this branch added `020_log_entry_rate_limit_admitted.xml` in the same spot. Both includes were kept (in numeric order, no changeSet id collision), so both migrations run.

All touched suites pass locally:
- Webservice: `mvn test` — 348 tests, 0 failures (includes the Liquibase baseline check that applies the merged changelog)
- Orchestrator: `pytest` — 1122 passed, 1 xfailed
- UI: static export build succeeds
- pre-commit lint (black/isort/autoflake/flake8 + hygiene) — all hooks pass

Claudia-Anthropica (changes requested):
@wasnertobias Exact-head CI completed successfully, and the merge from main preserves the prior correctness fixes. However, the query still selects recent completions through `timestamp_response`, while migration 020 indexes only `(api_key_id, timestamp_forwarding)`; without a corresponding `(api_key_id, timestamp_response)` index, `/me/keys` must filter each key's historical rows, so the existing performance finding remains unresolved.

wasnertobias:
Addressed in `91cd1b158`.

The performance finding was real: the /me/keys usage window
(`ApiKeyRepository.findRateLimitUsageForUser`) filters `log_entry` on
`(timestamp_forwarding >= :since OR timestamp_response >= :since)` grouped by
`api_key_id`, and migration 020 only indexed `(api_key_id, timestamp_forwarding)`.
The completion half of the window — the token metric, cut on `timestamp_response`
— could not be satisfied from that index, so Postgres had to scan each key's full
history to test the OR. That is what made `/me/keys` degrade as a high-volume key
aged.

Migration 021 adds the missing index:

    CREATE INDEX CONCURRENTLY idx_log_entry_api_key_timestamp_response
        ON log_entry (api_key_id, timestamp_response);

The column order matches the predicate — `api_key_id` equality prefix, then the
`timestamp_response` range — so Postgres can now bitmap-OR two index range scans
(`forwarding >= :since` on the 020 index, `response >= :since` on 021) and reach
only each key's recent rows for both the request and the token metric. It is built
`CONCURRENTLY` outside a transaction (same rationale as 012/014/020) so it does not
lock `log_entry` against the orchestrator, and it is idempotent via the same
`indexExists` precondition.

No behavior change — it is index-only. `LiquibaseBaselineTest` now also asserts the
new index exists (`migration021_rateLimitCompletionResponseIndexExists`). The
webservice suite is green (349 tests, including the Liquibase baseline applied
against Postgres 17), the orchestrator suite is green, and pre-commit passes.

Claudia-Anthropica (dismissed):
@wasnertobias Migration 021 now supplies the `(api_key_id, timestamp_response)` index, resolving the remaining completion-window history scan. The master changelog registers both migrations, the baseline tests verify both indexes, and all captured exact-head build, test, UI, and lint checks succeeded. The recorder's best-effort marker failure mode can only conservatively over-report rejected usage within the display window and does not affect enforcement, so it is non-blocking.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
Claudia-Anthropica: @wasnertobias Thanks, this is fixed now: migration 021 adds `(api_key_id, timestamp_response)` concurrently, `master.xml` registers it, and the baseline test verifies the index.

logos/logos-webservice/src/main/resources/liquibase/changelog/027_log_entry_rate_limit_admitted.xml
```diff
@@ -0,0 +1,62 @@
...
+         locking log_entry against the logos-orchestrator, which keeps writing
+         to this table while logos-webservice is redeployed (same rationale as
+         012). -->
+    <changeSet id="020-log-entry-api-key-forwarding-index" author="logos" runInTransaction="false">
+        <preConditions onFail="MARK_RAN">
+            <not><indexExists tableName="log_entry" indexName="idx_log_entry_api_key_timestamp_forwarding"/></not>
+        </preConditions>
+        <sql>
+            CREATE INDEX CONCURRENTLY idx_log_entry_api_key_timestamp_forwarding
+                ON log_entry (api_key_id, timestamp_forwarding);
```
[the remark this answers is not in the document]
Claudia-Anthropica: @coderabbitai Thanks, this is fixed now: migration 021 adds the completion-time composite index and the baseline test verifies it.

logos/logos-orchestrator/src/logos/monitoring/recorder.py
```diff
@@ -295,6 +295,20 @@ def record_provider(self, request_id: str, provider_id: int) -> None:
...
+        Written at the admission decision (``check_and_record``), before
+        execution: the per-key usage window (issue #672) must count an
+        admitted request while it is still running, and must skip the ones
+        the limiter rejects after scheduling — the log row already carries
+        ``timestamp_forwarding`` from ``record_scheduled``, so without this
+        flag a rejected request would show up as usage. Keys without a limit
+        are never checked; their rows keep the column NULL, which the usage
+        query treats as "not rejected".
+        """
+        self._write(request_id, rate_limit_admitted=admitted)
```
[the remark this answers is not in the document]
Claudia-Anthropica: @coderabbitai The recorder intentionally uses best-effort monitoring writes. A failed `FALSE` marker can only conservatively over-report a rejected request during the 60-second display window and cannot affect limiter enforcement; excluding `NULL` would instead hide unlimited-key and pre-migration usage, so no change is needed.

Claudia-Anthropica (dismissed):
@wasnertobias The new commit only normalizes comments and test descriptions; the previously verified behavior remains unchanged. Request admission, token accounting, cloud/local classification, and both timestamp indexes align with the limiter semantics, and the relevant tests cover the edge cases. Captured Helios and label checks succeeded; the pending CodeRabbit and documentation checks are not required.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
wasnertobias: Re-verified on the rebased head (`e0d8cbee2`): `findRateLimitUsageForUser` cuts the request counts on `le.timestamp_forwarding` (admission), not `timestamp_request`, and the javadoc records why arrival time under-reported under queue wait. The queued/never-scheduled rows drop out on their own because both timestamps are NULL. Resolving.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/service/MeKeysService.java
wasnertobias: Re-verified on the rebased head: `RateLimitConfig.window_seconds` in `rate_limiter.py` is commented as the source of truth and names `MeKeysService.RATE_LIMIT_WINDOW_SECONDS`, the javadoc there points back, and `RateLimitWindowConsistencyTest` parses the Python default out of the checkout and fails the build on drift. Having the orchestrator serve the window stays a follow-up, as argued above. Resolving.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
wasnertobias: Re-verified on the rebased head: `formatTpm` renders the used side with one decimal once it is compacted (`59.5k/60k`, `1.5k/60k`) while the limit side stays whole, and sub-1k figures are still exact — so a near-exhausted limit no longer rounds into the limit. Resolving.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
wasnertobias: Re-verified on the rebased head: `MeKeysService.toMap` puts `rate_limit_usage: null` when the window query returned no row for the key (no zero-filling anywhere), and `formatRpm`/`formatTpm` render the used side as `–` for it. `MeKeysControllerTest.getMyKeys_keepsUnknownRateLimitUsageDistinctFromZero` pins it. Resolving.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
wasnertobias: Re-verified on the rebased head: both `formatRpm` and `formatTpm` render `used/∞` whenever the limit is null, so an unlimited key can no longer read as a zero cap and a used figure is no longer mistakable for a limit; the card tooltip explains `–` and `∞`. Resolving.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
wasnertobias: Re-verified on the rebased head: the per-row token figure is computed in a LATERAL as `total_tokens` when non-zero, else `prompt_tokens + completion_tokens` — the same semantics as the limiter’s Python `or` — and the token sums are cut on `timestamp_response`, matching `record_tokens` at completion. The seed fixture pins the parts-only, zero-total and non-zero-total cases (9313/9314/9315/9316). Resolving.

wasnertobias:
Rebased onto current `main` (`78e6e1533`) and force-pushed; the branch was conflicting before.

Two things worth knowing when re-reading the diff:

- `main` has meanwhile taken `026_agent_session_no_push.xml`, so the two migrations added here moved up one slot: `027_log_entry_rate_limit_admitted.xml` and `028_log_entry_api_key_response_index.xml`. Their `logicalFilePath` and `changeSet` ids are unchanged, so environments that already ran them from an earlier numbering do not re-run the changes; `LiquibaseChangelogOrderTest` and `LiquibaseBaselineTest` follow the new numbers.
- The rebase absorbed the `main.py` refactor from `main`; the single `check_and_record` call site now lives in `_execute_resource_mode`, and the admission flag is still persisted right at that decision point.

All six open review threads were the findings from the 08-29 self-review with their own "fixed in …" replies. I re-verified each fix against the rebased head and resolved them; nothing needed a further code change.

Claudia-Anthropica (dismissed):
@wasnertobias The rebased exact head correctly records admission decisions, aggregates request and token usage using their matching 60-second windows, and registers both supporting composite indexes. All retained findings are resolved in the current code, and the captured exact-head webservice, orchestrator, UI, and lint checks succeeded; only non-required build finalizers remained pending.

LogosOSSAgent:
Re-checked the rebased head (`e0d8cbe`) against `main` at `78e6e15` — `git merge-base --is-ancestor` confirms the branch is on top of it. The two renumbered migrations and the `main.py` call site both look right, and I could not find anything the rebase dropped or mis-placed. No code change was needed.

What I verified:

- **Migrations.** `026_agent_session_no_push.xml` is `main`'s; ours are `027_log_entry_rate_limit_admitted.xml` (adds `rate_limit_admitted` + the `(api_key_id, timestamp_forwarding)` index) and `028_log_entry_api_key_response_index.xml` (the `(api_key_id, timestamp_response)` index). `logicalFilePath` is still `020_…`/`021_…` and the changeSet ids are still `020-…`/`021-…`, so a DB that already ran them under the old numbering is not re-run (the recorded FILENAME + id + checksum are unchanged). `master.xml` includes them in strictly-increasing order up to `028`.
- **Tests follow the numbers.** `LiquibaseBaselineTest` has `migration027_rateLimitAdmittedColumnAndForwardingIndexExist` and `migration028_rateLimitCompletionResponseIndexExists` (correct column + index assertions); `LiquibaseChangelogOrderTest` is number-agnostic but enforces the strictly-increasing order that `027/028` satisfies.
- **`main.py` refactor.** Exactly one `check_and_record` call site, now inside `_execute_resource_mode` (main.py:4879). The admission verdict is persisted immediately after it, before execution and before the 429 path, via `_record_rate_limit_admission` (main.py:4885) — the decision point the migration comment describes. The helper guards on a missing `request_id` / unset pipeline and swallows its own errors, so it can't break the request path.
- **Self-review fixes still hold.** I re-confirmed the two load-bearing ones against the code: (1) the TPM check now runs before the RPM slot is recorded, and a TPM reject leaves the RPM slot free — I ran the limiter in isolation (it's stdlib-only) and the TPM-reject-doesn't-consume-RPM behavior holds; (2) `update_log_entry_metrics` filters on `v is not None`, so `admitted=False` is persisted as `False` and only an unset value stays `NULL` — the "persists the flag both ways" recorder test matches the real write path.
- **Whole feature intact.** The orchestrator (limiter/recorder/pipeline/dbmanager/dbmodules), webservice (repository/projection/service), UI (model/component), and the seed/cleanup SQL are all present and mutually consistent after the rebase — the seed rows and the `MeKeysControllerTest` expectations line up with the query and the `rate_limit_admitted` filter.

Two things I could not do from here, said plainly:

- I can't see the six individual review threads (no network), so I re-verified the code the message claims was fixed rather than the thread text itself. On the code, the claimed fixes are in place.
- I could not run the full test harness offline: the orchestrator unit tests need the heavy venv (the `shared` path dependency resolves to a developer's absolute path and `grpc`/`torch` aren't installed), and the Java/Spring tests need a Maven repo that isn't present. I verified the DB-free Java tests statically (changelog order up to `028`; window constant is `60` on both sides, so `RateLimitWindowConsistencyTest` is satisfied) and exercised the rate-limiter logic in isolation instead of via `pytest`. CI's own run of these is the check I'd trust; I couldn't read its status from here.

Net: the rebase is sound, the diff is coherent, and the branch needs no further change from me.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
LogosOSSAgent: Confirmed — I re-read the query, the limiter's `record_tokens` call sites, and the seed fixture on this head, and the three points hold:

- **LATERAL matches the Python `or`.** The per-row figure (ApiKeyRepository.java:171–180) is `total_tokens` when its sum is `> 0`, else `COALESCE(prompt_tokens + completion_tokens, 0)`. The limiter computes `usage_tokens.get("total_tokens") or (prompt_tokens + completion_tokens)` at all three `record_tokens` call sites (main.py:3951, 4160, 4409). Missing total → `None`/`NULL` falls through to the parts in both; zero total is falsy in Python and fails `> 0` in SQL, so it falls through in both; a non-zero total wins in both. The only input where the two differ is a *negative* `total_tokens` (Python would use it, SQL falls through) — token counts don't take that, so it's not reachable in practice.
- **Token sums cut on completion, request counts on admission.** The two token `SUM(ut.tokens) FILTER (…)` are cut on `le.timestamp_response >= :since` (lines 165, 167) — where `record_tokens` runs at completion — while the two `COUNT(DISTINCT le.id)` are cut on `le.timestamp_forwarding >= :since` (lines 164, 166) — where the RPM slot is charged.
- **The fixture pins all three fallback cases.** 9313 is parts-only (300 + 200 → 500, no `total_tokens` row), 9314 is a zero total plus parts (0, 100 + 150 → 250), 9315 is a non-zero total that doesn't equal its parts (90 vs 40 + 30 → 90, total wins), and 9316 is the local parts-only counterpart (120 + 130 → 250). I hand-summed the in-window cloud and local rows against the query's filters: cloud 1000 + 1500 + 100 + 400 + 500 + 250 + 90 = 3840 and local 700 + 250 = 950, matching `MeKeysControllerTest`'s expected `cloud_tokens`/`local_tokens`, with 9308/9309 excluded by the `IS DISTINCT FROM FALSE` filter and 9311 contributing a request but no tokens while in flight.

No change needed; the resolution is accurate. One caveat on how I verified: I couldn't run the Java/Spring + Postgres test harness in this offline environment (no Maven repo to build against), so the token figures are confirmed by reading the query/fixture and arithmetic rather than an executed `MeKeysControllerTest` — CI's run of that test is the one I'd lean on.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
LogosOSSAgent: Confirmed — the formatters and the tooltip do what the thread says, and I checked the behavior, not just the pattern. No change needed.

- **`used/∞` for a null limit, both functions.** Current code (my-workspace.ts:331–335, 343–349): each side renders independently — `used == null ? '–' : …` and `limit == null ? '∞' : …` — joined with `/`, so a null limit yields `12/∞` (or `–/∞` with no traffic) and never `0` or a bare used figure. I extracted the exact method bodies from the file and ran them in node across the case matrix: `(12, null) → "12/∞"`, `(null, 60) → "–/60"`, `(null, null) → "–/∞"`, `(12, 60) → "12/60"`; `formatTpm` likewise, including `formatTpm(59500, 60000) → "59.5k/60k"` (one-decimal used side, whole limit side) and `formatTpm(999, null) → "999/∞"`. All pass.
- **Tooltip explains both markers, on both groups.** The CLOUD and LOCAL `limit-value` spans (my-workspace.html:88, 92) carry `… · – = no usage known for the window · ∞ = no limit set`, and I checked the tooltip glyphs byte-for-byte against the ones the formatters emit: en dash U+2013 and U+221E in both files, so the legend can't be describing different characters than the display.
- **Scope check.** `formatRpm`/`formatTpm` have no other call sites, and the other null-limit renderings in the app (the team-detail tabs) already use the same `∞` convention, so the card now matches existing practice rather than inventing a new one.

One caveat, same as the previous thread: the Angular suite can't run in this offline environment (no `node_modules`, no network to install), and there is no spec covering these two methods — hence the direct function-body run above in place of the test suite.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
LogosOSSAgent: Confirmed — I traced the whole chain on the current head and each link does what the thread says. No change needed.

- **Backend puts `null`, no zero-filling.** `getKeysForUser` builds the usage map only from rows the window query returned (MeKeysService.java:55–58) and `toMap` does `m.put("rate_limit_usage", usage == null ? null : toRateLimitUsage(usage))` (line 162) — an explicit `null` when the lookup misses. The query itself only `GROUP BY`s keys that have matching rows, so a key with no in-window traffic never appears in the map. `toMap` has a single call site (line 58), so no other path can populate the field, and nothing along the way pads zeros.
- **Frontend renders `–`.** `usageFor` (my-workspace.ts:321–323, the cited anchor) returns `key.rate_limit_usage` untouched — its doc comment is exactly the rationale in the thread (null kept distinct from zero, since a zero usage figure claims the whole budget is available). The template consumes all four metrics as `usageFor(key)?.<metric> ?? null`, so a null object yields null used figures, and `formatRpm`/`formatTpm` render the used side as `–` for null — which I exercised in node last round (`(null, 60) → "–/60"`, `(null, null) → "–/∞"`). A grep for `rate_limit_usage` across the UI finds only the model type and `usageFor`; there is no `?? 0` or other zero-fill anywhere.
- **The test pins it.** `getMyKeys_keepsUnknownRateLimitUsageDistinctFromZero` (MeKeysControllerTest.java:74–83) calls `/me/keys` with only the base key seed — no in-window traffic — and asserts `rate_limit_usage.window_seconds` does not exist. A zero-filled object would carry `window_seconds: 60` and fail that assertion, so the test catches exactly the regression the thread is about. (It also holds if Jackson ever dropped the null field rather than serializing it, and the UI renders `–` in either case.)

Same offline caveat as before: I verified this by reading the current code and exercising the formatter bodies in node; I could not run the Spring/MockMvc test here (no Maven repo offline), so `MeKeysControllerTest`'s green status is CI's to confirm.

logos/logos-ui/src/app/features/my-workspace/my-workspace.ts
LogosOSSAgent: Confirmed. I extracted the current `formatTpm` body from the file (my-workspace.ts:343–349) and ran it in node against the thread's examples plus the cases behind the claim — all pass:

- `formatTpm(59500, 60000) → "59.5k/60k"` and `formatTpm(1500, 60000) → "1.5k/60k"` — the two examples in the thread.
- Near-exhausted: `formatTpm(59700, 60000) → "59.7k/60k"`. With the used side at one decimal it can no longer display as the limit itself, which is what would have happened at `0`-decimal compaction (`60k/60k`). Even when the used side lands exactly on a whole thousand — `formatTpm(59000, 60000) → "59.0k/60k"` — the `.0` keeps it visually distinct from the limit side.
- Sub-1k figures stay exact with no suffix: `formatTpm(999, 60000) → "999/60k"`, `formatTpm(999, 100) → "999/100"`.
- The compaction boundary behaves as the threshold in the code: `formatTpm(1000, 60000) → "1.0k/60k"`.

That matches the mechanism the doc comment at the cited line describes: `compact(used, 1)` vs `compact(limit, 0)` with the `n >= 1000` threshold. No change needed.

Same caveat as the earlier formatter thread: no spec covers these methods and the Angular suite can't run offline (no `node_modules`), so this is the direct function-body run standing in for it.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/service/MeKeysService.java
LogosOSSAgent: Confirmed — all three links of the sync mechanism are in place on this head, and I checked one extra thing the thread doesn't mention. No change needed.

- **Python side names the copy.** `rate_limiter.py:14–20`: the `window_seconds: int = 60` default is commented as "the source of truth", names `MeKeysService.java (RATE_LIMIT_WINDOW_SECONDS)` at its path, and says to change both together because `RateLimitWindowConsistencyTest` fails on drift.
- **Java side points back.** The javadoc above the cited line (MeKeysService.java:33–43; line 44 is the `RATE_LIMIT_WINDOW_SECONDS = 60` constant) states it is a copy of `RateLimitConfig.window_seconds` in `rate_limiter.py`, calls the orchestrator the source of truth, and names the same test. Both values are 60 at this head, so the test passes as-is.
- **The test parses the checkout, not a fixture.** `RateLimitWindowConsistencyTest` walks up from the working directory to `logos/logos-orchestrator/src/logos/rate_limiter.py`, extracts the default with `window_seconds\s*:\s*int\s*=\s*(\d+)`, reads the Java field via reflection, and asserts equality — a JUnit failure, i.e. a red build, on drift. One thing I verified beyond the thread: of the file's four `window_seconds` mentions, only the dataclass default (line 20) can match that pattern — the others are `config.window_seconds` in code and f-strings, where `: int =` never follows — so `Matcher.find()` can't latch onto the wrong line.

On "having the orchestrator serve the window stays a follow-up": I can't see the argument from earlier in the thread (the history above this message isn't part of what I was given), so I'm not re-arguing it — but the code is consistent with that scoping. This PR keeps the webservice constant as a guarded copy rather than adding an orchestrator endpoint, and the consistency test is the bridge until the follow-up removes the copy. I'd agree with deferring it: the drift risk is exactly what the test pins, and serving the window changes the orchestrator's API surface, which is a separate decision.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/identity/repository/ApiKeyRepository.java
LogosOSSAgent: Confirmed on this head (`e0d8cbe`, the `e0d8cbee2` in the thread). No change needed.

- **Counts cut on admission.** Both `COUNT(DISTINCT le.id) FILTER (…)` lines (ApiKeyRepository.java:164, 166) use `le.timestamp_forwarding >= :since`; `timestamp_request` appears nowhere in the query, so there is no arrival-time cut to drift back into.
- **The javadoc records the why.** Directly above the query: "Cutting on `timestamp_request` instead would drop exactly the requests that waited in the queue, under-reporting precisely when developers look at it to understand a 429" — the same point the thread makes, anchored to when `check_and_record` charges the RPM slot.
- **Queued rows drop out on their own.** The outer predicate is `(timestamp_forwarding >= :since OR timestamp_response >= :since)` (line 183); a never-scheduled row has both NULL, and `NULL >= :since` is NULL, so the row never enters the query at all. The fixture pins both halves of this: 9305 is queued with both timestamps NULL (and in-window `timestamp_request`) so an arrival-time cut would have counted it, while 9307 arrived 70s ago — outside the window — but was admitted 10s ago and is counted. `cloud_requests = 7` (MeKeysControllerTest.java:100) only holds with 9307 in and 9305 out. One belt-and-braces detail for the record: 9305 also has `provider_id NULL`, so it would fail the `p.provider_type = 'cloud'/'logosnode'` FILTER even if it reached it — the NULL timestamps are the primary mechanism, the provider check is independent of it.

Same offline caveat as the earlier query threads: verified by reading the current query/javadoc/fixture and the arithmetic behind the expected count, not by an executed `MeKeysControllerTest` — no Maven repo offline, so its green status is CI's to confirm.

logos/logos-webservice/src/main/resources/liquibase/changelog/028_log_entry_api_key_response_index.xml
```diff
@@ -0,0 +1,54 @@
...
+         Originally numbered 021; main's later migrations have since taken
+         019-026, so the file moved to 028 (logicalFilePath keeps the old name
+         for environments that already ran it). LiquibaseChangelogOrderTest
+         requires strictly increasing numbers in master.xml.
+
+         Runs outside a transaction so CONCURRENTLY can build the index without
+         locking log_entry against the logos-orchestrator, which keeps writing
+         to this table while logos-webservice is redeployed (same rationale as
+         012/014/027). -->
+    <changeSet id="021-log-entry-api-key-response-index" author="logos" runInTransaction="false">
```
wasnertobias: inline id does not match file name id. fix it!

LogosOSSAgent: Fixed — the `021` id was a leftover from when this branch first numbered the migration, before main took the numbers in between. It kept getting bumped (021 → 027 → 028) with the old id carried along, on the assumption that some environment had already run it. That's not the case: the migration only ever existed on this unmerged branch, and the 021 number was never unique on main to begin with, so no database can hold it. The `logicalFilePath` pinning the old 021 path went with it — with the id aligned, it would only make Liquibase record a file path that no longer exists in the repo.

Changed in `028_log_entry_api_key_response_index.xml`:
- changeSet id `021-log-entry-api-key-response-index` → `028-log-entry-api-key-response-index`
- removed `logicalFilePath="liquibase/changelog/021_log_entry_api_key_response_index.xml"`
- updated the comment block that explained the old numbering

I applied the same fix to its sibling `027_log_entry_rate_limit_admitted.xml`, which had the identical defect (ids `020-log-entry-rate-limit-admitted` / `020-log-entry-api-key-forwarding-index` and a 020 `logicalFilePath`, also introduced by this PR) — fixing only 028 would have left the same mismatch one file up. The `023`/`026` files from main keep their `logicalFilePath` pins: those were genuinely released under those names, so the pin earns its keep there.

Verified: both files are well-formed XML; every changeSet id in the changelog dir now carries its file's number (checked all of them); `master.xml` is untouched and its strictly-increasing include numbers still hold, which is all `LiquibaseChangelogOrderTest` inspects — I replicated that assertion. `LiquibaseBaselineTest` asserts only on the resulting schema and indexes, and nothing about what the changes execute changed here. I could not run the Java suite in this sandbox (no Maven dependency cache, no Docker for the testcontainer), so that one still owes a CI run. One assumption worth stating plainly: the fix is safe because no environment has run either migration — if one somehow had, the index changesets are guarded by their `indexExists` preconditions, but the `addColumn` in 027 is not, and re-running it would fail.

Claudia-Anthropica (approved):
@wasnertobias The exact head aligns the changelog filenames and changeset IDs, while the usage path remains coherent: admission flags exclude limiter rejects, request and token windows follow admission and completion semantics, and both predicates have supporting indexes. The renewed admission-write concern is the same best-effort monitoring case already settled in the retained thread: a failed write can only temporarily over-report displayed usage without affecting enforcement, while NULL must remain countable for unchecked and pre-migration rows. All captured exact-head checks succeeded.

logos/logos-webservice/src/main/resources/liquibase/changelog/028_log_entry_api_key_response_index.xml
wasnertobias: Confirmed on 82665504 — both changeSet ids now match their file names (028-log-entry-api-key-response-index, 027-log-entry-rate-limit-admitted / 027-log-entry-api-key-forwarding-index), the stale 021/020 logicalFilePath pins are gone, and the main-side 020/021/023/026 files keep theirs as intended. Resolving.

logos/logos-orchestrator/src/logos/main.py
```diff
@@ -1131,6 +1131,23 @@ def _discard_in_flight(request_id: Optional[str], result_status: str) -> None:
...
+        return
+    pipeline = globals().get("_pipeline")
+    if pipeline is None:
+        return
+    try:
+        pipeline.record_rate_limit_admission(request_id, admitted)
+    except Exception:  # noqa: BLE001 — monitoring must never break a request
+        logger.debug("Failed to record rate-limit admission for %s", request_id, exc_info=True)
+
+
```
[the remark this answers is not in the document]
wasnertobias: Gap is real, confirming the mechanism: `recorder._write()` swallows all exceptions (recorder.py:335), so a rejected request whose `rate_limit_admitted=FALSE` write fails leaves the column NULL, and `IS DISTINCT FROM FALSE` then counts it.

Two things bound the blast radius, and one thing makes it not a query tweak:

- **Display-only, rare trigger.** It requires the specific best-effort write at main.py:4885 to fail while the request is already going to 429; in-memory limiting is unaffected, so no key actually exceeds its budget — the /me/keys figure is just inflated by the number of rejects whose flag write lost.
- **Indistinguishable at the query level.** The only way a NULL row is a wrongly-counted reject is exactly the failed-write case, which the query cannot tell apart from a legitimate NULL (no limit / pre-migration). So the fix has to land at the write layer, not in `findRateLimitUsageForUser`.
- **It spans both reject paths.** The sync 429 (main.py:4903) funnels into the 5131 handler → `_record_log_failure`, but the async reject (main.py:4896) returns early with no terminal log record. A durable fix has to cover both.

Two shapes, so it's a real choice rather than a one-liner:
(a) record the reject durably in the terminal path(s) and reuse the existing column — needs plumbing through `update_log_entry_metrics`, no migration; or
(b) a dedicated non-NULL "recording failed" state — needs a migration plus a query exclusion.

Given this is a "show usage" PR and the impact is a rare display inaccuracy, I'd lean to tracking (a) as a small follow-up rather than growing this branch — but I can implement (a) here if you'd rather close it on this branch. Your call.

logos/logos-webservice/src/main/resources/liquibase/changelog/028_log_entry_api_key_response_index.xml
LogosOSSAgent: Thanks for verifying. I re-checked the branch head: `027_log_entry_rate_limit_admitted.xml` carries `027-log-entry-rate-limit-admitted` and `027-log-entry-api-key-forwarding-index`, `028_log_entry_api_key_response_index.xml` carries `028-log-entry-api-key-response-index`, and the only `logicalFilePath` pins left in the changelog are main's own in `023_token_cost_decomposition.xml` and `026_agent_session_no_push.xml` — `020`/`021`/`023`/`026` keep their ids untouched, and `master.xml` includes 027/028 in order. Nothing further to change on this point.

logos/logos-orchestrator/src/logos/main.py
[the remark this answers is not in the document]
LogosOSSAgent: Agreed — I'd track (a) as a follow-up rather than grow this branch.

I re-verified the mechanism against the branch head; every line in the comment still holds: `_write` swallows (recorder.py:335), the best-effort write is main.py:4885, the async 429 (4896) returns early with no terminal log record, the sync 429 (4903) funnels through the `except HTTPException` handler at 5131 into `_record_log_failure`, and the query predicate is `IS DISTINCT FROM FALSE` (ApiKeyRepository.java:184).

Why not on this branch: the fix lands in the shared terminal-failure funnel and the generic HTTPException handler on the request hot path, and the payoff is a rare, display-only, self-clearing over-count — the 60s window slides the row out, and in-memory limiting is untouched. I don't want to stretch a near-mergeable show-usage PR for that.

Two things I noticed while reading it, which narrow the follow-up:

- `update_log_entry_metrics` already accepts `rate_limit_admitted` as an allowed field (dbmanager.py:394), so the DB layer is done. Option (a) is really: (1) add a `rate_limit_admitted` kwarg to `_record_log_failure` and forward it into that call; (2) carry the flag from the sync raise site (4904) to the 5131 handler — a marker on the exception or a small subclass; (3) in the async 429 branch, call `_record_log_failure(..., rate_limit_admitted=False)` before returning, mirroring what the budget-reject branch right below already does at 4929. That async 429 return is the only pre-execution reject without a terminal record.
- A failed `admitted=TRUE` write is harmless: NULL is counted, and an admitted request *should* be counted. Only the FALSE write needs durability, so the best-effort write at 4885 stays as is — the follow-up only has to make the reject land in the terminal path. No migration, as you said.

Happy to implement (a) as that follow-up; nothing changes on this branch.

PR #848, opened 2026-08-29
`Logos`: Measure model cold load and wake-from-sleep times in calibration

### LINKED ISSUE #627: `Logos`: Model cold load time shall be measured in calibration
Calibration process shall record model cold load time (and awake from sleep times) so that the ETTFT is more accurate. Also it helps to make adaptive orchestration decisions.

wasnertobias:
## Summary

Calibration now measures and persists two new timings per model per worker node:

- **`cold_load_time_s`** — wall-clock time from the final measurement probe's spawn (the same configuration the serving lane runs with: final KV size, resolved `max_model_len`, ...) to the first request the probe served, i.e. the Phase 2.5 warmup 1-token completion. That is exactly what a client experiences before its first token when a lane has to cold-load for its request: vLLM startup, weight load, and the first-request overhead (CUDA graph capture, JIT, KV page allocation).
- **`wake_from_sleep_time_s`** — wall-clock time from the Phase 5.5 `/wake_up` trigger to the post-wake test request being served. The same "trigger → first served request" interval for the path a lane takes when a request finds it asleep.

Both values are recorded as `None` when the request anchoring each interval did not complete in the run (warmup not served, sleep phases skipped, wake not verified): a value that never actually served a request must not pose as a measurement, following the codebase's established semantics for measured fields. This makes ETTFT estimates more accurate and gives adaptive orchestration a per-model load/wake latency signal. The calibration decision logic itself is unchanged — this is measurement and persistence only.

## Changes

- `logos/logos-workernode/logos_worker_node/calibration.py`
  - `CalibrationResult` gains `cold_load_time_s` / `wake_from_sleep_time_s`.
  - `calibrate_model()` arms a `perf_counter` anchor right before every final-measurement spawn attempt (KV-search and pinned-KV paths) and stops it at the first served request (Phase 2.5 warmup) and the post-wake test request (Phase 5.5); both intervals are logged in the run summary.
  - `result_to_profile_dict()` persists both values into the profile (rounded to 0.1 s like the MB fields). With the existing `merge_profile` semantics a `None` from a run that could not measure keeps the previously calibrated value.
- `logos/logos-workernode/logos_worker_node/model_profiles.py` — `ModelProfileRecord` gains both fields (`to_dict` / `_load_persisted`), so the values round-trip through `model_profiles.yml` and are carried in the runtime status push the server reads.
- `logos/logos-workernode/logos_worker_node/logos_bridge.py` — the `calibration_probe_log` event payload (persisted by the orchestrator into `calibration_probe_logs.summary`) now includes both timings alongside the other measured values.
- `logos/logos-orchestrator/src/logos/sdi/models.py` — server-side `ModelProfile` gains both fields (including `to_dict`).
- `logos/logos-orchestrator/src/logos/sdi/providers/logosnode_provider.py` — `get_model_profiles()` maps both fields from the worker's runtime snapshot, so the capacity planner and any future ETTFT consumer can use them.

Persistence follows the existing calibration-data pattern: the worker's `model_profiles.yml` plus the runtime snapshot are the source of truth (as for `sleep_l1_transient_host_ram_mb`, `host_ram_residual_mb`, `sleep_mode_disabled`, ...), and the orchestrator already persists the full profile dict into `ollama_provider_snapshots.runtime_payload` (JSONB). No DB migration is required.

## Testing

- New tests in `logos/logos-workernode/tests/test_calibration.py`: cold load measured spawn → first served request; `None` when the warmup does not serve; wake time measured `/wake_up` → post-wake request; `None` when the sleep phases are skipped or the wake fails verification; profile-dict mapping (incl. rounding); `merge_profile` keeping prior values; profile-store round trip.
- New tests in `logos/logos-workernode/tests/test_model_profiles.py`: persisted timing fields reload into the record and appear in `get_all_profiles()`; legacy records default to `None`.
- New test in `logos/logos-orchestrator/tests/unit/sdi/test_scheduler_view.py`: `get_model_profiles` maps the new fields from the snapshot into `ModelProfile` (incl. `to_dict`); a profile from a worker that predates the fields stays `None`.

Verified (no GPU deployment available in this environment):
- Orchestrator unit suite: `pytest` in `logos/logos-orchestrator` → **1018 passed, 1 xfailed**.
- Worker suite with the CI exclusions → **340 passed, 2 skipped**; the CI-excluded `test_calibration.py` and `test_lane_manager.py` also pass locally (**161** + **54 passed**).
- Lint: `pre-commit run --config logos/.pre-commit-config.yaml --all-files` → all hooks pass (black, isort, autoflake, flake8, hygiene checks).

## Summary by CodeRabbit

* **New Features**
  * Added cold-load and wake-from-sleep timing measurements to model calibration results.
  * Exposed calibration timings through model profiles and serialized profile data.
  * Preserved timing information across profile storage and reloads.
  * Legacy profiles continue to work with unavailable timings shown as unset.

* **Bug Fixes**
  * Calibration timing values are now retained when new calibration results lack measurements.

### DIFF
--- a/logos/logos-workernode/logos_worker_node/calibration.py
+++ b/logos/logos-workernode/logos_worker_node/calibration.py
@@ -1244,6 +1244,22 @@ class CalibrationResult:
     # makes sleeping and the model cache compete for one pool. None when the
     # run skipped the sleep phases, or when /proc/meminfo was unreadable.
     host_ram_residual_mb: float | None = None
+    # Wall-clock seconds from the final measurement probe's spawn (the same
+    # configuration the serving lane runs with — final KV size, resolved
+    # max_model_len, ...) until the first request it served completed: the
+    # Phase 2.5 warmup 1-token completion. That is exactly what a client
+    # waiting for a cold lane experiences before its first token — vLLM
+    # startup, weight load, and the first-request overhead (CUDA graph
+    # capture, JIT, KV page allocation) all included. None when the warmup
+    # request did not serve: a value that never actually served a request
+    # must not pose as a cold-load measurement.
+    cold_load_time_s: float | None = None
+    # Wall-clock seconds from the Phase 5.5 /wake_up trigger until the
+    # post-wake test request completed — the same "trigger → first served
+    # request" interval as cold_load_time_s, for the path a lane takes when a
+    # request finds it asleep. None when the run skipped the sleep phases,
+    # the wake failed verification, or the post-wake request did not serve.
+    wake_from_sleep_time_s: float | None = None
     # Set when calibrate_model bails because the model itself can never
     # load on this worker (bad repo id, gated repo, unsupported architecture).
     # Caller persists the model into model_profiles.yml so the master's
@@ -1917,6 +1933,12 @@ def calibrate_model(
             return None
 
     proc: subprocess.Popen[str] | None = None
+    # perf_counter anchor for the cold-load measurement. Re-armed right
+    # before every final-measurement spawn attempt; only the attempt that
+    # actually spawns proceeds to Phase 2.5, so whatever is left here belongs
+    # to the live probe. None while no final spawn has been armed yet.
+    final_spawn_at: float | None = None
+    cold_load_time_s: float | None = None
 
     if kv_search:
         search_lo = _KV_CACHE_MIN_STEP_MB  # 1 GB floor
@@ -2247,6 +2269,7 @@ def calibrate_model(
             _final_planned = {**plan, "kv_cache_memory_bytes": _final_kv_str}
             _final_fp = _cmd_fingerprint(_build_vllm_cmd(_final_planned, vllm_binary, host, port, _final_kv_str))
             failed_commands.discard(_final_fp)
+            final_spawn_at = time.perf_counter()
             proc = _try_start(_final_kv, record_blacklist=False, allow_whitelist=False)
             if proc is not None:
                 kv_cache_sent_mb = _final_kv
@@ -2294,6 +2317,7 @@ def calibrate_model(
                 _fixed_kv_str,
             )
         explicit_probe_overrides: dict[str, Any] = {"_max_model_len_retry_count": 0, "_max_num_seqs_retry_count": 0}
+        final_spawn_at = time.perf_counter()
         proc = _try_start(kv_cache_sent_mb, plan_overrides=explicit_probe_overrides)
         if proc is None:
             partial.error = f"Model failed to start with KV cache " f"{_format_kv_mb(kv_cache_sent_mb)} on tp={tp}"
@@ -2356,6 +2380,17 @@ def calibrate_model(
                 "        warmup done in %.1fs — graphs/JIT/KV pools allocated",
                 warmup_dt,
             )
+            # The warmup IS the first request the probe served, so this point
+            # is the end of the cold load a client would experience for a lane
+            # spawned with this exact configuration: final_spawn_at → here
+            # spans vLLM startup, weight load, and the first-request overhead
+            # (CUDA graph capture, JIT, KV page allocation).
+            if final_spawn_at is not None:
+                cold_load_time_s = time.perf_counter() - final_spawn_at
+                logger.info(
+                    "        cold load (spawn → first served request) = %.1fs",
+                    cold_load_time_s,
+                )
         else:
             logger.warning(
                 "        warmup failed (%.1fs) — awake VRAM may underestimate peak",
@@ -2403,6 +2438,7 @@ def calibrate_model(
         sleep_transient_mb: float | None = None
         host_ram_residual_mb: float | None = None
         sleep_mode_disabled: bool | None = None
+        wake_from_sleep_time_s: float | None = None
         if sleep_level <= 0:
             logger.info(
                 "  [4/6 + 5/6 + 5.5/6] Sleep phases skipped (sleep mode disabled for this model) — "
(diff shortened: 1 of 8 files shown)


wasnertobias (commented):
Cleanly built. The None semantics are carried through consistently (no value without a request that actually served), the round trip YAML → snapshot → `ModelProfile` is complete, and there's no behaviour change or migration. The outer retry loop is anchored correctly — `final_spawn_at` sits *inside* `for _attempt in range(_FINAL_MEASUREMENT_RETRIES)` and only the successful attempt survives via `break`. That's the right way round.

Two things inline. Plus:

- **`merge_profile` keeps stale timings.** Correct for VRAM, which is a property of the model. Load time is a property of the *node* — disk speed, page cache state, concurrent load. A value measured months ago on different hardware persists indefinitely.
- **`(issue #627)`** in `test_calibration.py` — per the project convention issue numbers don't go into code or tests.
- **No consumer yet.** I grepped: nothing reads either field. Fine for a preparatory PR, but it does mean a wrong measurement won't surface on its own — which argues for pinning down the semantics now, while the context is fresh. I opened #860 for the estimator wiring, including a warning against substituting the measured value into `OVERHEAD_COLD_S` unexamined.

logos/logos-workernode/logos_worker_node/calibration.py
```diff
@@ -2356,6 +2380,17 @@ def _cancelled() -> bool:
                 "        warmup done in %.1fs — graphs/JIT/KV pools allocated",
                 warmup_dt,
             )
+            # The warmup IS the first request the probe served, so this point
+            # is the end of the cold load a client would experience for a lane
+            # spawned with this exact configuration: final_spawn_at → here
+            # spans vLLM startup, weight load, and the first-request overhead
+            # (CUDA graph capture, JIT, KV page allocation).
+            if final_spawn_at is not None:
+                cold_load_time_s = time.perf_counter() - final_spawn_at
```
wasnertobias, line 2410: This measures a **page-cache-warm** load, not a cold one.

The KV search spawns the same model process repeatedly before this point (`_probe_kv` → `_try_start`, once per KV candidate). By the time the final measurement spawn happens, the weights are already in the OS page cache, so this interval covers a load served from RAM rather than from disk.

That matters because the constant this value is meant to improve on is explicitly documented as the other thing:

```python
# logos/logos-orchestrator/src/logos/pipeline/ettft_estimator.py:39
OVERHEAD_COLD_S = 45.0  # Cold load from disk
```

For a 27B model in fp16 (~54 GB) the difference between the two is on the order of tens of seconds on NVMe, and considerably more on network storage. Substituting this figure for that constant would replace a pessimistic disk assumption with an optimistic cache measurement, under-estimating precisely the case the cold tier exists for.

There are really three states worth distinguishing — not downloaded, on disk but not cached, and cache-warm — and this measures the third while carrying the name of the second. Ideally calibration captures both (2) and (3), e.g. one deliberately cold reading in addition to this one. If that's out of scope here, the docstring should say plainly that this is a warm-cache figure and a lower bound.

Tracked in #860 so the estimator side doesn't consume it under the wrong assumption.

logos/logos-workernode/logos_worker_node/calibration.py
```diff
@@ -2294,6 +2317,7 @@ def _cancelled() -> bool:
                 _fixed_kv_str,
             )
         explicit_probe_overrides: dict[str, Any] = {"_max_model_len_retry_count": 0, "_max_num_seqs_retry_count": 0}
+        final_spawn_at = time.perf_counter()
```
wasnertobias: `_try_start` recurses on itself, so the interval anchored here can include failed attempts.

It retries internally up to `_MAX_MODEL_LEN_RETRIES_PER_KV = 3` (`:1629`) and `_MAX_NUM_SEQS_RETRIES_PER_KV = 2` (`:1636`), returning `_try_start(...)` with incremented counters. The anchor sits before the outer call, so every one of those inner attempts lands inside the measured span — and a start that vLLM rejects for `max_model_len` typically fails *after* loading the weights, so each retry costs close to a full load.

The line right above deliberately resets both counters to 0, so this path expects retries to be available.

The comment at `:1936` ("only the attempt that actually spawns proceeds to Phase 2.5") holds for the outer loop, which is handled correctly — it just doesn't cover the recursion. Anchoring at the point where a spawn actually succeeds, rather than before the first attempt, would close it.

logos/logos-workernode/logos_worker_node/calibration.py
wasnertobias: Addressed in da80305. `cold_load_time_s` now documents plainly that this is a page-cache-**warm** figure and a **lower bound** on a true from-disk cold load: the KV search spawns the same model once per candidate before the final spawn (and prior runs on the host otherwise leave the weights resident), so the load is served from the OS page cache, not disk. The measurement-site comment and both log lines now say "page-cache-warm" as well. A deliberately cold reading (dropping the host page cache) is out of scope for this PR; the docstring also notes that consumers must not substitute this value for a from-disk cold-start constant, and that estimator-side consumption is tracked in #860.

logos/logos-workernode/logos_worker_node/calibration.py
wasnertobias: Fixed in da80305. `_try_start` now captures the timestamp immediately before its own `spawn_vllm` call and publishes it to `final_spawn_at` only after `wait_ready` comes back — i.e. only for the attempt that actually came up. The two pre-call arms (before the kv-search final-measurement loop and before the fixed-KV `_try_start`) are removed, so the recursive `--max-model-len` / `--max-num-seqs` retry attempts — each of which typically fails only after loading the weights — no longer land inside the measured span; the anchor is the successful attempt's spawn. Both final-measurement paths suppress the whitelist (`allow_whitelist=False` / `succeeded_commands.clear()`), so a real spawn is guaranteed and the anchor is always the live probe's. Added a regression test `test_cold_load_anchor_excludes_failed_retry_attempts` that forces one rejected attempt with a simulated ~0.2 s weight-load cost followed by a successful injected-max-model-len retry, and asserts the recorded interval excludes the rejected attempt's cost.

PR #849, opened 2026-08-29
`Logos`: Report the model's historic max context when no workernode reports one

### LINKED ISSUE #829: `Logos`: Claude-Logos context-size fallback is underestimating
The fallback when all workernodes are offline shall be the max context size (historic ever reported value for this model).

Right now it is also not clear why the fallback even jumps in, as workernodes were connected at that time.

wasnertobias:
## Summary

When all workernodes were offline, `/v1/models` carried no context fields for local models at all, so the claude-logos wrapper sized the session from its blind 111,200-token constant — a guess that underestimates a model that has been served wider. This PR makes the orchestrator report the model's **historic maximum reported context** as `max_model_len_overall` in that state, and fixes the faulty detection that made a model look context-unknown while its worker sat connected.

## Root cause analysis — why the fallback triggered with workernodes connected

Two independent mechanisms pushed the client onto its fallback:

1. **All nodes offline → no fields at all.** The context statistics (`_served_context_window_stats`) were derived solely from the live worker runtime snapshots. A disconnected node's session is popped from the registry, so once the last node went offline the stats were empty and every `max_model_len*` field disappeared from `/v1/models`. There was no durable source to consult — the widest window a model had ever been reported at existed only in the (now gone) snapshots.

2. **Node connected, but the model read as context-unknown.** `_profile_native_context_length` only read `max_context_length` (the operator pin — `NULL` for calibrated models) and the widest point of the calibrated KV sweep. When a calibration capped `--max-model-len` to fit the pinned KV budget and recorded no wider KV point, `calibration_max_model_len` was the *only* context the profile reported — and it was ignored. For such a model (nothing loaded yet, or a lane whose window the worker does not report explicitly) no `max_model_len_overall` was produced, so the client fell back to its guess even though the worker was connected and ready to serve the model at exactly that width.

## Changes

- `logos-orchestrator/src/logos/dbutils/dbmanager.py`
  - New `derived_reported_context_length()`: the widest context a worker profile reports — operator pin, calibrated `--max-model-len` cap, or widest KV-sweep point.
  - `upsert_model_profiles()` now maintains `model_profiles.max_reported_context_length` as a high-water mark: `GREATEST` of stored and freshly derived value, so a later, narrower calibration (e.g. on a node with less VRAM) cannot shrink the widest window the model has ever been reported at. `COALESCE` is required because pre-migration rows hold NULL and Postgres `GREATEST(NULL, x)` is NULL.
  - New `get_historic_max_context_by_model()`: per-model maximum across all providers.
- `logos-orchestrator/src/logos/main.py`
  - `_profile_native_context_length()` counts `calibration_max_model_len` via the shared derivation — fixes mechanism 2.
  - `_served_context_window_stats()` tops up each model's `overall` figure from the historic maximum, so `max_model_len_overall` stays known when every workernode is offline (and an orchestrator restart) — fixes mechanism 1. The claude-logos wrapper's existing cascade (`current_max → current_min → overall → constant`) picks the number up without any script change; with nothing running it reports the session as "cold", i.e. sized against a window Logos has not committed to yet.
  - `_historic_max_context_by_model()` caches its lookup for a short TTL (10 s) instead of opening a fresh `DBManager` session on every model-endpoint call — the value only changes when a worker snapshot arrives. A failed lookup is never cached (a broken database recovers on the next call) and now logs a `logger.warning` with trace instead of failing open silently.
  - The `upsert_model_profiles` failure handler in `_capture_logosnode_provider_snapshot()` is raised from `logger.debug` to `logger.warning` and rolls the session back explicitly — the deploy window before the Liquibase migration has created the new column (where the upsert fails wholesale) is loud, not silent.
- `logos-webservice/.../liquibase/changelog/016_model_profiles_max_reported_context_length.xml` (plus `master.xml`): adds the nullable column to `model_profiles`. A second one-shot changeSet backfills the historic mark from the operator pin (`max_context_length > 0`) for rows that pre-date the column, so pinned models are covered the moment the migration runs. Calibrated-only models (pin NULL) cannot be recovered from this table — their mark settles on the next worker snapshot.
- `logos/docs/context-windows.md`: documents the durable fallback and the all-offline case in the field table and troubleshooting section.

The `current_*` figures deliberately stay lane-derived: with nothing loaded there is no window that "always holds", so `max_model_len` remains omitted instead of claiming one — the fallback value is exposed as `max_model_len_overall`, which is what the wrapper's cascade and the "max" context source are built for.

## Verification

- New unit tests: `tests/unit/dbutils/test_model_profile_context.py` (derivation from each profile field, the high-water binding incl. the GREATEST/COALESCE clause, the historic read) and offline/top-up cases in `tests/unit/main/test_v1_models.py` plus `tests/unit/main/test_internal_model_context_windows_endpoint.py`, including a regression test for the capped-calibration case.
- `pytest tests/unit` in `logos/logos-orchestrator`: **1031 passed, 1 xfailed** (pre-existing xfail).
- `pre-commit` on the logos tree (black, isort, autoflake, flake8, hygiene checks): all hooks pass.
- `LiquibaseBaselineTest` (Testcontainers, real Postgres 17, full changelog incl. the new 016 changeSets plus Hibernate `validate`): **4 passed**.


logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
```diff
@@ -1088,6 +1138,7 @@ def upsert_model_profiles(
                 :base_residency_mb, :loaded_vram_mb, :sleeping_residual_mb,
                 :kv_budget_mb, :disk_size_bytes, :engine,
                 :tensor_parallel_size, :kv_per_token_bytes, :max_context_length,
+                :max_reported_context_length,
```
wasnertobias, line 1141: Adding a column to this INSERT makes it fail wholesale until the migration has run — and there is no ordering guarantee that it has.

`logos-orchestrator` and `logos-webservice` both declare only:

```yaml
depends_on:
  logos-db:
    condition: service_healthy
```

Nothing sequences them relative to each other, and Liquibase runs in the webservice. So on every deploy there is a window where the orchestrator is already consuming worker snapshots while `max_reported_context_length` does not exist yet. Every `upsert_model_profiles` in that window raises on an undefined column.

What makes it quiet rather than loud is the existing handler at `main.py:837-844`:

```python
try:
    db.upsert_model_profiles(provider_id, model_profiles)
except Exception:
    logger.debug("Failed to upsert model profiles for provider %s", ..., exc_info=True)
```

`logger.debug` — at production log levels this produces no output at all. And the loss isn't limited to the new column: the whole row fails, so `base_residency_mb`, `loaded_vram_mb`, `kv_budget_mb`, `measurement_count` and `last_measured_at` all stop updating for the duration, with nothing in the logs to say so.

Worth noting too that the `except` sits *inside* the enclosing `with DBManager() as db:` block, so `__exit__` sees `exc_type is None` and skips its `session.rollback()`.

Every prior column addition to this INSERT carried the same hazard; what's different here is that this PR makes the column load-bearing for the very behaviour it introduces. Sequencing the webservice ahead of the orchestrator (or gating the orchestrator on migrations having completed) would close it. Raising that handler above `debug` seems worth doing regardless — a persistent failure to record calibration data should not be invisible.

logos/logos-webservice/src/main/resources/liquibase/changelog/016_model_profiles_max_reported_context_length.xml
```diff
@@ -0,0 +1,36 @@
...
+         /v1/models reports it as max_model_len_overall so the number
+         survives every workernode going offline and an orchestrator
+         restart. Nullable: rows written before this changeSet hold NULL,
+         and the first upsert after it settles the mark. -->
+    <changeSet id="015-model-profiles-max-reported-context-length" author="wasnertobias">
+        <preConditions onFail="MARK_RAN">
+            <not><columnExists tableName="model_profiles" columnName="max_reported_context_length"/></not>
+        </preConditions>
+        <addColumn tableName="model_profiles">
+            <column name="max_reported_context_length" type="INTEGER"/>
```
wasnertobias, line 31: A nullable column with no backfill means the all-offline case this PR targets isn't fixed on deploy.

The scenario in the description is "every workernode is offline, so the live snapshots say nothing". Right after the migration, `max_reported_context_length` is NULL for every existing row, `get_historic_max_context_by_model` filters on `> 0`, and `/v1/models` reports no `max_model_len_overall` — exactly as before. The mark only settles once some worker connects and snapshots at least once. If the fix is deployed *because* the nodes are down, it doesn't help until they come back.

A one-shot backfill in this changeSet would cover the operator-pinned models immediately:

```sql
UPDATE model_profiles
   SET max_reported_context_length = max_context_length
 WHERE max_reported_context_length IS NULL
   AND max_context_length > 0;
```

Calibrated models can't be recovered that way — I checked `000_initial_schema.xml`, and `model_profiles` has `max_context_length` but no `calibration_max_model_len` column and no KV-pair storage, so the two richer sources have never been persisted. Those genuinely need one post-migration snapshot. Still, backfilling what is recoverable narrows the gap, and it's worth stating in the PR that the remainder needs one worker check-in before the fallback works.

logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
```diff
@@ -1102,6 +1153,10 @@ def upsert_model_profiles(
                 tensor_parallel_size = EXCLUDED.tensor_parallel_size,
                 kv_per_token_bytes = EXCLUDED.kv_per_token_bytes,
                 max_context_length = EXCLUDED.max_context_length,
+                max_reported_context_length = GREATEST(
```
wasnertobias, line 1156: This mark only ever grows, and nothing can bring it back down.

`GREATEST(COALESCE(stored, 0), EXCLUDED)` with `MAX(...) GROUP BY model_name` in the reader means the value is the widest context *ever* reported for the model on *any* provider that still has a row. There's no reset path, no TTL, and no tie to whether that hardware still exists. The FK is `ON DELETE CASCADE`, so deleting a provider does clear its rows — but a provider that is merely deactivated, or a node that was re-tasked, keeps its historic mark alive.

The failure mode is concrete:

> Model X calibrated on a 2×H100 node → 262144 recorded.
> That node is retired; X now only runs on a single A6000 at 32768.
> All nodes go offline for maintenance.
> `/v1/models` reports `max_model_len_overall = 262144`.

A client sizes a session from that and gets `maximum context length is 32768` — the exact 400 listed in the troubleshooting table of `context-windows.md`. OpenCode is the sharper case, since it writes `max_model_len_overall` into a startup config that isn't re-read.

It also drifts from what the docs promise the field means: *"what a lane runs at once it gets all the KV cache it asks for"*. After a permanent hardware shrink no lane runs at that width. `test_list_models_historic_max_tops_up_a_narrower_live_overall` encodes the tradeoff deliberately (live 33k, historic 262k wins), so I take it as intentional — but it deserves either a bound (only count providers that still exist and are active) or an operator-facing way to reset a stale mark.

logos/logos-webservice/src/main/resources/liquibase/changelog/master.xml
```diff
@@ -18,5 +18,6 @@
     <include file="liquibase/changelog/012_log_entry_timestamp_index.xml"/>
     <include file="liquibase/changelog/013_drop_models_parallel.xml"/>
     <include file="liquibase/changelog/014_log_entry_model_id_timestamp_index.xml"/>
+    <include file="liquibase/changelog/015_model_profiles_max_reported_context_length.xml"/>
```
wasnertobias: Heads-up on a collision: three open PRs each add a `015_` changelog and each insert an include here.

- this PR — `015_model_profiles_max_reported_context_length.xml`
- #852 — `015_model_aliases.xml`
- #823 — `015_user_passkeys.xml`

`main` currently ends at `014_`. Liquibase keys on changeSet ids rather than filenames, so nothing breaks functionally — but whichever two merge second will conflict here, and the end state is three files numbered `015_` whose numbering no longer reflects the order they actually ran in. Environments that migrate between merges end up with a `DATABASECHANGELOG` order that doesn't match the filenames either.

Cheap to avoid by renumbering on merge; worth coordinating across the three.

logos/logos-orchestrator/src/logos/main.py
```diff
@@ -5588,29 +5588,37 @@ def _as_len(value) -> int:
...
-        return value if value > 0 else 0
 
-    native = _as_len(profile.get("max_context_length"))
-    pairs = profile.get("kv_cache_to_max_model_len_pairs")
-    if isinstance(pairs, list):
-        for item in pairs:
-            if isinstance(item, dict):
-                native = max(native, _as_len(item.get("max_model_len")))
-    return native
+def _historic_max_context_by_model() -> dict[str, int]:
```
wasnertobias, line 5628: This opens a fresh `DBManager` session on every call, and the caller isn't cached.

`_served_context_window_stats()` has no memoisation and is called from five places (`:2174`, `:5684`, `:5743`, `:5796`, `:5853`). `/v1/models` at `:5743` already holds an open session from `get_models_for_api_key` a couple of lines above, so that endpoint now takes two connections where it previously took one — and any path that consults both the stats and the conservative view pays for it again.

The query itself is cheap (`model_profiles` is small, it's a plain GROUP BY); the cost is session setup on an endpoint every client hits at startup. Since the underlying values only change when a worker snapshot arrives, a short TTL cache — or folding the lookup into the session the caller already has — would remove the churn.

Separately, `except Exception: return {}` swallows genuine errors along with connectivity ones. Failing open is right for this feature, but a `logger.warning` would at least make a persistently broken lookup discoverable instead of it looking like "no model has a historic context".

wasnertobias (commented):
The migration itself is done properly — `preConditions onFail="MARK_RAN"` with a `columnExists` guard, a real rollback, nullable so there's no table rewrite. The `COALESCE` rationale (GREATEST with a NULL argument returns NULL in Postgres) is correct and worth having spelled out. `derived_reported_context_length` is defensive in the right places, and turning the registry failure from `return stats` into `provider_ids = []` is the consistent choice.

Five things inline, three of them about behaviour rather than style. The first two are the ones I'd want resolved before merge — one is a deploy-ordering hazard that silently drops data, the other is that the PR doesn't actually fix its own target scenario until a worker reconnects once.

Minor: `(#829)` appears in the `_profile_native_context_length` docstring, twice in `_served_context_window_stats`, and twice in `test_model_profile_context.py`. Per the project convention those belong in the PR description, not in code or tests.

logos/logos-webservice/src/main/resources/liquibase/changelog/master.xml
wasnertobias: Renumbered: this PR's changelog is now `016_model_profiles_max_reported_context_length.xml` (changeSets `016-model-profiles-max-reported-context-length` plus the new backfill changeSet), and `master.xml` includes `016_`. `015_` is free again for #852 and #823 to coordinate — this PR no longer participates in the collision.

logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
wasnertobias: Done in 96113d8: the handler in `main.py` is now a `logger.warning` with `exc_info` and rolls the session back explicitly (as you note, `__exit__` skips its own rollback when the exception is swallowed inside the `with` block). It also states what is actually lost — the entire row update (`base_residency_mb`, `loaded_vram_mb`, `kv_budget_mb`, `measurement_count`, `last_measured_at`) — with the provider name. During the pre-migration window that means one warning per provider per snapshot instead of silence.

Sequencing the webservice ahead of the orchestrator (or gating the orchestrator on Liquibase having completed) is the real fix — I agree it is worth a follow-up, and it would protect any future column addition to this INSERT as well.

logos/logos-webservice/src/main/resources/liquibase/changelog/016_model_profiles_max_reported_context_length.xml
wasnertobias: Added a second one-shot changeSet to the migration (`016_model_profiles_max_reported_context_length.xml`): it backfills `max_reported_context_length` from the operator pin for rows that pre-date the column (`WHERE max_reported_context_length IS NULL AND max_context_length > 0`), so operator-pinned models get the correct all-offline fallback the moment the migration has run.

Calibrated-only models (pin NULL) cannot be recovered from this table — `calibration_max_model_len` and the KV sweep pairs were never persisted in `model_profiles` — so their mark settles on the next worker snapshot; that is now stated in the PR description. The backfill is idempotent (the `WHERE` clause makes it a no-op once the marks have settled), and it is a separate changeSet because the first changeSet's `not(columnExists)` precondition would skip a same-changeSet backfill in environments where the column already exists.

logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
wasnertobias: Intentional for now. The point of the column is that the all-offline fallback never under-reports a model a client has already been served wider, and the widest window is the only figure that always held on at least one node.

There is a removal path: the FK is `ON DELETE CASCADE`, so deleting a provider (and its hardware) also removes its contribution to the mark. What is missing is a reset while the provider still exists, and there is nothing in the schema to key it off — the `providers` table has no active/enabled flag tying a row to hardware that is still around. An explicit reset of the historic mark (or decay when a provider is disabled) is a reasonable follow-up, which I would rather do deliberately than implicitly. The current tradeoff is pinned by `test_list_models_historic_max_tops_up_a_narrower_live_overall`.

logos/logos-orchestrator/src/logos/main.py
wasnertobias: Both addressed. The lookup (`_historic_max_context_by_model`, the one place that opens a `DBManager`) now caches its result for a 10-second TTL — the value only changes when a worker snapshot arrives, so `/v1/models` is back to one connection. A failed lookup is never cached, so a broken database recovers on the very next call. I went with the TTL rather than folding the lookup into the callers' sessions because the five call sites are spread across the internal endpoint, `/v1/models`, `retrieve_model`, and `warmup_model` — one cache covers all of them.

Separately, the `except Exception` still fails open but now logs a `logger.warning` with the traceback, so a persistently broken lookup is discoverable instead of looking like 'no model has a historic context'.

PR #854, opened 2026-08-29
`Logos`: Auto-detect and recover from poisoned vLLM torch.compile / AOT cache

### LINKED ISSUE #579: `Logos`: Auto-detect and recover from poisoned vLLM torch.compile / AOT cache
## Context

On `deioma`, loading `zai-org/GLM-OCR` was failing repeatedly during `add_lane` with vLLM exiting code 1 during startup. The user surface error was uninformative (`RuntimeError: Engine core initialization failed. See root cause above. Failed core proc(s): {}`); the actual root cause was only visible in the saved failure log at `/tmp/logos-vllm-logs/planner-zai-org_GLM-OCR_*_startup_failed.log` inside the container.

## Root cause (this incident)

Not OOM. GPU 0 was at 1.5 / 97 GiB; GPUs 1–2 essentially idle; no `dmesg` OOM kill; no `CUDA out of memory` in worker logs.

The actual exception:

```
AssertionError: Expected tensors only, but got: <class 'int'>
  at torch/_inductor/utils.py:3442 copy_misaligned_inputs
```

Call path:

```
profile_run → glm4_1v.py:1679 forward
            → vllm/compilation/decorators.py:573 aot_compiled_fn
            → torch/_dynamo/aot_compile.py:224
            → vllm/compilation/caching.py:217 optimized_call
            → torch/_inductor/standalone_compile.py:122 CacheCompiledArtifact._compiled_fn
            → copy_misaligned_inputs → assert
```

vLLM was reusing a pre-compiled AOT graph from the on-disk compile cache (configured via `--compilation-config {"cache_dir": "/usr/share/ollama/.ollama/models/.cache/vllm"}`, host path `/mnt/ceph/.cache/vllm`). The cached artifact's expected input signature didn't match what was passed at runtime — a Python `int` arrived where a `torch.Tensor` was expected. Stale / incompatible compile cache, not memory corruption.

Cache layout that was poisoned:

```
/mnt/ceph/.cache/vllm/
├── modelinfos/               # JSON metadata — safe, kept
├── rank_0_0/                 # per-rank backbone cache — wiped
└── torch_compile_cache/
    └── torch_aot_compile/    # AOT graph artifacts — wiped
        ├── 2c21e1d8…/
        ├── f0c45cd7…/
        └── f5096f3c…/
```

Manual fix applied: `rm -rf /mnt/ceph/.cache/vllm/torch_compile_cache /mnt/ceph/.cache/vllm/rank_0_0`. First subsequent start will recompile (slower); subsequent starts rebuild the cache.

## Likely triggers (general)

Any of these can produce a cache that loads but fails at first invocation:
- vLLM / torch wheel upgrade between cache write and reuse (different graph signature)
- Previous startup crashed mid-AOT-write, leaving truncated artifact
- Same `cache_dir` shared across model variants / `--compilation-config` variants with a hash collision

## Proposed work

Goal: no more manual `rm -rf` interventions on the host. Recommended approach is **(A) + (B)** below; (C) is a nice-to-have if upstream cooperates.

### (A) Pre-flight cache version validation

After a healthy first start with a given cache dir, write a `cache_meta.json` containing:
- vLLM version
- torch version
- CUDA runtime version
- container image digest (or `LOGOS_IMAGE_VERSION` env)
- model identifier and `--compilation-config` hash

On every subsequent start, before vLLM is launched: read `cache_meta.json`, compare against current environment, and if any field mismatches, wipe `torch_compile_cache/` and `rank_*_*/` for that model. Keep `modelinfos/`.

This eliminates the most common trigger (image upgrade invalidating the AOT graph) without needing to parse vLLM error logs.

### (B) Detect-and-retry on known cache-poisoning fingerprints

`logos_worker_node.vllm_process` already captures the failure log. Add a classifier over the captured stderr that matches a small allowlist of high-confidence cache-corruption fingerprints:

- `copy_misaligned_inputs … Expected tensors only, but got` (this incident)
- `CacheCompiledArtifact … AssertionError`
- `torch/_inductor … FileNotFoundError` (truncated artifact)
- `vllm/compilation/caching.py … KeyError` (cache-key mismatch)

On match: wipe the cache subdir for that model's compile key only, then re-launch vLLM **once**. Cap at one auto-retry per (model, hour) so we cannot loop forever if the real bug is elsewhere. Log a structured line (e.g. `cache_auto_recovered=true model=… fingerprint=…`) so this is greppable in metrics — silent self-healing is worse than visible self-healing.

### (C) Crash-write protection (stretch)

Wrap the cache directory in atomic write semantics: vLLM writes to `cache_dir.tmp/`, rename to `cache_dir/` only after the engine reports healthy startup. Prevents partial-write corruption from interrupted compiles. Likely requires upstreaming to vLLM or a shim — defer unless we see partial-write incidents.

## Acceptance criteria

- [ ] `cache_meta.json` written after first healthy startup of each (model, cache_dir) pair
- [ ] On startup, mismatched `cache_meta.json` triggers automatic wipe of `torch_compile_cache/` + `rank_*/` with a clear log line
- [ ] vLLM startup failure matching one of the known cache-corruption fingerprints triggers exactly one automatic cache-wipe + retry, with cooldown
- [ ] Auto-recovery events emit a structured log line (greppable: `cache_auto_recovered=true`)
- [ ] `modelinfos/` is never touched by auto-recovery
- [ ] Unit tests covering: version mismatch detection, fingerprint matching, retry cap / cooldown

## Out of scope

- Per-request resilience (this is a startup-time concern)
- Changing vLLM's caching internals (do that via (C) only if needed)

wasnertobias:
## Summary

On `deioma`, loading `zai-org/GLM-OCR` repeatedly failed during `add_lane` because vLLM replayed a stale AOT/torch.compile cache (`AssertionError: Expected tensors only, but got: <class 'int'>` in `copy_misaligned_inputs`). The manual fix was an `rm -rf` of the compile-cache directories on the host. This PR makes the worker detect and recover from exactly this class of failure without manual intervention.

The sibling incident (#580, gemma-4) already landed the core of this feature in `f1816c82a`: a worker-wide version stamp (`.logos_compile_cache_stamp.json`) plus a reactive stack-trace-path detector with a one-shot purge + retry. This PR adds what was still missing against #579's acceptance criteria:

### Changes (`logos/logos-workernode/logos_worker_node/vllm_process.py`)

1. **Per-lane pre-flight `cache_meta.json` (issue approach A).** After a healthy start the worker writes a `cache_meta.json` into each lane's own compile-cache dir (`.cache/vllm/lanes/<model>/`) recording the vLLM version, torch version, CUDA runtime version, `LOGOS_IMAGE_VERSION` (when set), the model identifier, and a hash of the `--compilation-config`. Before every subsequent start of that lane the meta is compared against the current environment; on any mismatch (or a cache without meta, produced by a worker that predates this check) only that lane's `torch_compile_cache/` + `rank_*/` are wiped and vLLM recompiles.

2. **Fingerprint classifier (issue approach B).** `has_poisoned_compile_cache` now also matches a small allowlist of high-confidence cache-corruption message fingerprints — `copy_misaligned_inputs … Expected tensors only`, `CacheCompiledArtifact … AssertionError`, `torch/_inductor … FileNotFoundError` (truncated artifact from a killed mid-AOT-write start), `vllm/compilation/caching.py … KeyError` — in addition to the existing stack-trace-path detector. This catches the #579 incident, where the traceback dies deep inside torch and never names the cache directory.

3. **Cooldown for reactive recovery.** The reactive wipe + retry is capped at **one per (model, hour)** (module-level state, so it survives lane-manager restarts / handle recreation). A genuinely broken model can no longer loop purge → retry → fail.

4. **Narrowed purge — `modelinfos/` is never touched.** The purge no longer removes the whole `.cache/vllm` directory (which contained `modelinfos/`); it removes only the compile artifacts (`torch_compile_cache/`, `rank_*/`, and the shared `torch_inductor/` on a worker-wide wipe). Reactive recovery purges only the failing model's per-lane dir (falling back to the worker-wide set when the artifacts live in the legacy shared location), so a poisoned model does not force every other model on the node to recompile.

5. **Structured, greppable recovery log.** Every auto-recovery event (proactive version stamp, proactive meta mismatch, reactive fingerprint) emits `[lane] cache_auto_recovered=true model=… fingerprint=… purged=…`.

### Tests (`tests/test_vllm_process.py`)

- Updated the existing purge/stamp tests to the narrowed purge (asserting `modelinfos/`, the FlashInfer cache, and sibling lanes' caches survive).
- New: per-lane meta write after a healthy start; version-mismatch / missing-meta / matching-meta / no-artifacts / user-overridden-`--compilation-config` cases; all four fingerprints plus a negative case (fragments must appear together); per-model purge isolation; reactive cooldown across handles (exactly one recovery per (model, hour)); the structured log line.

## Verification

- `pytest` in `logos/logos-workernode`: **541 passed, 2 skipped** — full suite including `test_model_cache.py`, `test_calibration.py` and `test_lane_manager.py`, which CI excludes for GitHub-runner reasons.
- `black`, `isort`, `autoflake`, `flake8` clean on both changed files (versions pinned to `logos/.pre-commit-config.yaml`).
- No live deployment (docker compose / vLLM workers / GPUs) is available in this environment; verification is via the unit tests above and careful reading against the incident's saved failure log.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic recovery from known compilation-cache failures with more precise failure detection.
  * Limited cleanup to affected model artifacts while preserving metadata and unrelated caches.
  * Added validation to refresh stale caches when model, runtime, or cache settings change.
  * Added cooldowns to prevent repeated recovery actions within a short period.

* **Observability**
  * Added structured logging for proactive and automatic cache recovery events.

### DIFF
--- a/logos/logos-workernode/logos_worker_node/vllm_process.py
+++ b/logos/logos-workernode/logos_worker_node/vllm_process.py
@@ -493,11 +493,27 @@ class VllmProcessHandle:
                     raise
                 purged = self._purge_compile_caches(lane_config.model)
                 if not purged:
-                    # This model's own per-lane cache dir holds nothing — its
-                    # artifacts live in the legacy shared location (or a
-                    # user-overridden --compilation-config we can't resolve),
-                    # so fall back to the worker-wide set for the retry.
-                    purged = self._purge_compile_caches()
+                    # The per-lane purge removed nothing, so the poisoned
+                    # artifact (if any) lives in the shared location. Widening
+                    # to the worker-wide set forces every other model on the
+                    # node to recompile, so it is only justified when the
+                    # traceback actually names a file under the compile cache
+                    # (the strong signal). A fingerprint-only match on an
+                    # empty per-lane dir is not: a brand-new lane has no cache
+                    # to be poisoned by, and the fingerprint is a heuristic,
+                    # so widening would risk wiping every model's cache on an
+                    # unrelated startup failure (a mistyped repo id, a gated
+                    # repo). Propagate instead.
+                    if self._logs_reference_compile_cache():
+                        purged = self._purge_compile_caches()
+                    else:
+                        logger.warning(
+                            "[%s] Cache poisoning detected (fingerprint=%s) but the traceback "
+                            "names no compile-cache file and this lane has no per-lane cache — "
+                            "not widening to the worker-wide cache; propagating",
+                            self.lane_id,
+                            fingerprint,
+                        )
                 purged_once = True
                 if not purged:
                     raise
@@ -640,12 +656,16 @@ class VllmProcessHandle:
 
     # Known cache-poisoning failure fingerprints, matched against the captured
     # startup logs. Each fingerprint is a tuple of fragments that must ALL
-    # appear in the recent log buffer; the pair is specific enough that no
-    # other startup failure matches it. They complement the stack-trace
-    # path-fragment detector above: a cached AOT artifact can also die deep
-    # inside torch/vllm library code, where no frame of the traceback points
-    # into the cache directory (the GLM-OCR incident: the artifact asserts
-    # in copy_misaligned_inputs on a stale input signature).
+    # appear in the recent log buffer; each is specific enough that an
+    # unrelated startup failure (a missing HF file, a bad module import) does
+    # not match it. Where a fragment is generic (FileNotFoundError,
+    # AssertionError), the tuple pairs it with a compile-cache path marker so
+    # the match only fires when the failure actually involves the cache. They
+    # complement the stack-trace path-fragment detector above: a cached AOT
+    # artifact can also die deep inside torch/vllm library code, where no
+    # frame of the traceback points into the cache directory (the GLM-OCR
+    # incident: the artifact asserts in copy_misaligned_inputs on a stale
+    # input signature).
     _CACHE_POISONING_FINGERPRINTS: ClassVar[tuple[tuple[str, tuple[str, ...]], ...]] = (
         # Stale AOT graph: a Python int arrives where the cached artifact
         # expects a torch.Tensor.
@@ -653,8 +673,12 @@ class VllmProcessHandle:
         # A cached AOT artifact raises on load/invocation.
         ("aot_artifact_assertion", ("CacheCompiledArtifact", "AssertionError")),
         # A previous startup was killed mid-AOT-write, leaving a truncated
-        # artifact the next startup cannot read.
-        ("inductor_artifact_missing", ("torch/_inductor", "FileNotFoundError")),
+        # artifact the next startup cannot read. The compile-cache path is
+        # required alongside FileNotFoundError so this only matches a file
+        # that is actually under the cache — not an unrelated missing-file
+        # error (a mistyped model id, a gated repo) that merely co-occurs
+        # with ordinary torch/_inductor compilation output in the same log.
+        ("inductor_artifact_missing", ("FileNotFoundError", "torch_compile_cache")),
         # Cache-key lookup hits an entry that no longer matches.
         ("compilation_cache_key_error", ("vllm/compilation/caching.py", "KeyError")),
     )
@@ -709,6 +733,21 @@ class VllmProcessHandle:
             )
         return removed
 
+    def _logs_reference_compile_cache(self) -> bool:
+        """True if the recent logs name a file under the torch.compile cache.
+
+        This is the strong signal of poisoning: the traceback actually reached
+        into a cached artifact, so whatever the originating exception is, the
+        artifact is implicated. Weaker than that is a fingerprint match (a
+        known error signature that may co-occur with an unrelated failure),
+        which ``has_poisoned_compile_cache`` accepts on its own but the
+        worker-wide purge must not be widened on (see ``spawn``).
+        """
+        if not self._recent_logs:
+            return False
+        log_blob = "\n".join(self._recent_logs)
+        return any(frag in log_blob for frag in self._POISONED_COMPILE_CACHE_PATH_FRAGMENTS)
+
     @property
     def has_poisoned_compile_cache(self) -> bool:
         """True if recent logs implicate the on-disk torch.compile cache.
@@ -723,8 +762,7 @@ class VllmProcessHandle:
         """
         if not self._recent_logs:
             return False
-        log_blob = "\n".join(self._recent_logs)
-        if any(frag in log_blob for frag in self._POISONED_COMPILE_CACHE_PATH_FRAGMENTS):
+        if self._logs_reference_compile_cache():
             return True
         return self._matched_cache_poisoning_fingerprint() is not None
 
--- a/logos/logos-workernode/logos_worker_node/vllm_process.py
+++ b/logos/logos-workernode/logos_worker_node/vllm_process.py
@@ -26,6 +26,7 @@ import shutil
 import signal
 import subprocess
 import sys
+import time
 import urllib.parse
 from collections import deque
 from datetime import datetime
@@ -80,6 +81,27 @@ _SCRUBBED_ENV_VARS = (
 
 # Cache for _discover_pip_cuda_lib_dirs() — computed once per process.
 _pip_cuda_lib_dirs: list[str] | None = None
+
+# Wall-clock (time.monotonic) of the last reactive cache recovery per model.
+# Capped at one per (model, hour) so a model whose startup is broken for a
+# non-cache reason (which can still leave a cache path in its traceback)
+# cannot loop purge → retry → fail forever across lane-manager restarts.
+# Module-level because the lane manager recreates handles on restart.
+_last_reactive_cache_recovery: dict[str, float] = {}
+_CACHE_RECOVERY_COOLDOWN_S: float = 3600.0
+
+
+def _reactive_cache_recovery_allowed(model: str, now: float | None = None) -> bool:
+    """True when no reactive cache recovery for this model ran within the cooldown window."""
+    last = _last_reactive_cache_recovery.get(model)
+    if last is None:
+        return True
+    return ((now if now is not None else time.monotonic()) - last) >= _CACHE_RECOVERY_COOLDOWN_S
+
+
+def _note_reactive_cache_recovery(model: str, now: float | None = None) -> None:
+    """Record that a reactive cache recovery ran for this model now."""
+    _last_reactive_cache_recovery[model] = now if now is not None else time.monotonic()
 
 
 def _discover_pip_cuda_lib_dirs() -> list[str]:
@@ -395,13 +417,21 @@ class VllmProcessHandle:
     async def spawn(self, lane_config: LaneConfig) -> ProcessStatus:
         """Spawn the vLLM process for this lane.
 
-        Cache safety: before spawning, the on-disk torch.compile and inductor
-        caches are purged if the recorded (vLLM, torch) versions don't match
-        the venv's current versions — a version bump is the most common cause
-        of cache poisoning. If the spawn still fails with a stack trace
-        pointing inside the compile cache directory (e.g. an AOT-compiled
-        graph that was specialized on a stale shape profile), the caches are
-        purged again and the spawn is retried once.
+        Cache safety: before spawning, two pre-flight checks purge poisoned
+        on-disk torch.compile / inductor caches — the worker-wide version
+        stamp (a vLLM/torch bump is the most common poisoning cause) and the
+        per-lane ``cache_meta.json`` (vLLM + torch + CUDA versions, image
+        version, model and compilation-config hash recorded after the last
+        healthy start of this exact cache dir). If the spawn still fails with
+        a stack trace pointing inside the compile cache directory or a known
+        cache-poisoning error fingerprint (e.g. an AOT-compiled graph that
+        was specialized on a stale shape profile), the caches are purged
+        again and the spawn is retried once. Reactive recovery is capped at
+        one per (model, hour) so a genuinely broken model cannot loop
+        purge → retry → fail forever.
+
+        Every purge-and-retry event logs a greppable
+        ``[lane] cache_auto_recovered=true model=… fingerprint=…`` line.
 
         The same shape of recovery covers a sharded checkpoint the loader
         rejects: the cached conversion is discarded and the lane retried once
@@ -416,7 +446,12 @@ class VllmProcessHandle:
             )
             await self._kill_process()
 
-        self._purge_compile_caches_if_versions_changed()
+        purged = self._purge_compile_caches_if_versions_changed()
+        if purged:
+            self._log_cache_auto_recovered(lane_config.model, "version_stamp_mismatch", purged)
+        purged = self._purge_lane_cache_if_meta_changed(lane_config)
+        if purged:
+            self._log_cache_auto_recovered(lane_config.model, "cache_meta_mismatch", purged)
 
         purged_once = False
         unsharded_once = False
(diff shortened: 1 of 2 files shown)


logos/logos-workernode/logos_worker_node/vllm_process.py
```diff
@@ -567,17 +621,44 @@ def has_fatal_cuda_errors(self) -> bool:
...
+    # in copy_misaligned_inputs on a stale input signature).
+    _CACHE_POISONING_FINGERPRINTS: ClassVar[tuple[tuple[str, tuple[str, ...]], ...]] = (
+        # Stale AOT graph: a Python int arrives where the cached artifact
+        # expects a torch.Tensor.
+        ("copy_misaligned_inputs", ("copy_misaligned_inputs", "Expected tensors only")),
+        # A cached AOT artifact raises on load/invocation.
+        ("aot_artifact_assertion", ("CacheCompiledArtifact", "AssertionError")),
+        # A previous startup was killed mid-AOT-write, leaving a truncated
+        # artifact the next startup cannot read.
+        ("inductor_artifact_missing", ("torch/_inductor", "FileNotFoundError")),
```
wasnertobias: This fingerprint is loose, and it feeds the most destructive path in the file.

Both fragments only have to appear *somewhere* in the log blob. `torch/_inductor` shows up in ordinary compilation output, and `FileNotFoundError` is the single most likely error when a model is first added — a mistyped repo id, a gated repo, a missing file. So an unrelated startup failure on a new model can match.

What makes that serious is what happens next. At `:498-500`:

```python
purged = self._purge_compile_caches(lane_config.model)
if not purged:
    purged = self._purge_compile_caches()   # worker-wide
```

A newly added lane has no per-lane cache dir yet, so `purged` is empty and the worker-wide fallback *always* fires — wiping `torch_compile_cache/` and `rank_*/` for **every** model on the node, plus the shared `torch_inductor/`. One mistyped model id can therefore force a full-node recompile.

Two changes would contain it:

1. Require a cache-path fragment (`torch_compile_cache` / `torch_aot_compile`) alongside the error, so the fingerprint only matches when the failure actually involves the cache.
2. Don't fall back to the worker-wide purge on a fingerprint-only match with an empty lane dir. An empty lane dir means this lane has no cache to be poisoned by — that's evidence *against* cache poisoning, not a reason to widen the purge.

logos/logos-workernode/logos_worker_node/vllm_process.py
```diff
@@ -416,7 +446,12 @@ async def spawn(self, lane_config: LaneConfig) -> ProcessStatus:
             )
             await self._kill_process()
 
-        self._purge_compile_caches_if_versions_changed()
+        purged = self._purge_compile_caches_if_versions_changed()
```
wasnertobias, line 449: This worker-wide check runs per lane spawn, but the fact it tests is global and settled at process start.

`_purge_compile_caches_if_versions_changed()` is only ever called from `spawn()`, so every lane re-answers the same question — "did vLLM/torch/CUDA change?" — and each answer can wipe directories another lane is currently writing into. After an image update with a new vLLM version, two lanes starting concurrently both read the stale stamp and both decide to purge worker-wide; the stamp is only written *after* a successful start, so it doesn't arbitrate the race.

This codebase already solves this exact problem one layer up, in `main.py:327`:

```python
# Pre-warm FlashInfer JIT kernels (single-process, sequential) so that
# subsequent vLLM launches — including TP>1 — find cached .so files and
# skip JIT, avoiding the multi-process compilation race that crashes GPUs.
```

Same shape of hazard, same shape of fix: do the shared-cache work once, sequentially, before any lane exists.

Suggested split:

| dimension | scope | belongs in |
|---|---|---|
| vLLM / torch / CUDA version, `LOGOS_IMAGE_VERSION` | all lanes | `lifespan`, once |
| model id, compilation settings | one lane | `spawn()`, per-lane purge |

A lock would close the race but keep every lane re-checking an immutable fact. Moving it also composes with the fingerprint issue above: with the global purge handled at startup, the reactive path no longer needs a worker-wide fallback at all and can stay strictly scoped to the failing lane.

logos/logos-workernode/logos_worker_node/vllm_process.py
```diff
@@ -748,6 +905,133 @@ def _write_compile_cache_stamp(self) -> None:
...
+            pass
+        image = (os.environ.get("LOGOS_IMAGE_VERSION") or "").strip()
+        if image:
+            meta["image"] = image
+        meta["model"] = lane_config.model
+        import hashlib
+        import json as _json
+
+        meta["compilation_config"] = hashlib.sha256(
+            _json.dumps({"cache_dir": lane_cache_dir}, sort_keys=True).encode("utf-8")
```
wasnertobias, line 994: This hash carries no information.

It's computed over `{"cache_dir": lane_cache_dir}` alone, and `lane_cache_dir` is derived from the model id — which is already stored as its own `model` field in the meta. So `compilation_config` changes exactly when `model` changes, and never otherwise. As a validation input it is redundant with a field that is already checked.

The PR description calls it "a hash of the `--compilation-config`". It isn't: it's a hash of the cache directory path that happens to be passed inside that flag (`:1786`).

The consequence is a real gap. `enforce_eager`, `quantization`, `dtype`, `cuda_graph_sizes` and `extra_args` all change how the model compiles, and none of them invalidate the meta. Flipping `enforce_eager` from `False` to `True` replays the previous AOT graph — the same poisoning class this check exists to prevent, just triggered by a config change instead of a version bump.

Either hash the settings that actually affect compilation, or drop the field and stop describing it as covering configuration changes.

logos/logos-workernode/logos_worker_node/vllm_process.py
```diff
@@ -632,17 +713,34 @@ def _invalidate_sharded_checkpoint(self, lane_config: LaneConfig) -> bool:
...
+        """Name of the first known cache-poisoning fingerprint in the recent logs.
+
+        Returns ``None`` when no fingerprint matches. The returned name is
+        the ``fingerprint=…`` value of the structured auto-recovery log line.
+        """
+        if not self._recent_logs:
+            return None
+        log_blob = "\n".join(self._recent_logs)
+        for name, fragments in self._CACHE_POISONING_FINGERPRINTS:
+            if all(frag in log_blob for frag in fragments):
```
wasnertobias, line 779: Fragments are matched across the whole log buffer rather than within one traceback.

`all(frag in log_blob for frag in fragments)` means the two strings can come from entirely unrelated events in the captured window. The comment above the fingerprint table says *"the pair is specific enough that no other startup failure matches it"*, but nothing binds the fragments to each other — that specificity is an assumption about the log's content, not a property of the check.

`("CacheCompiledArtifact", "AssertionError")` is the clearest example: an informational line mentioning a cached artifact plus an assertion from a different failure is enough.

A false match costs a purge, a retry, and a `cache_auto_recovered=true` line in the log — which then points whoever investigates at the cache instead of the real cause. Matching per line, or within the same traceback block, would keep the intent while removing the coupling to unrelated log content.

logos/logos-workernode/logos_worker_node/vllm_process.py
```diff
@@ -748,6 +905,133 @@ def _write_compile_cache_stamp(self) -> None:
...
+        a meta file, produced by a worker that predates this check) the
+        lane's ``torch_compile_cache/`` and ``rank_*/`` are wiped and vLLM
+        recompiles from scratch. Returns the paths actually removed.
+        """
+        lane_cache_dir = self._lane_compile_cache_dir(lane_config)
+        if lane_cache_dir is None:
+            return []
+        if not self._lane_artifact_paths(lane_cache_dir):
+            return []
+        current = self._current_cache_meta(lane_config)
```
wasnertobias, line 1013: This is fail-open in the direction that matters.

`_current_cache_meta` omits any value it cannot read, `if not current: return []` skips validation entirely when nothing is readable, and the comparison below only iterates keys present in `current`. So if the CUDA runtime version can't be determined, a CUDA upgrade passes validation silently — the validator waves through precisely the case it exists to catch.

For a cache validator, "couldn't establish the environment" is safer treated as "assume it changed" than as "assume it's fine". Recompiling costs minutes; replaying a graph compiled against a different CUDA runtime costs a debugging session.

wasnertobias (commented):
Careful work, and the narrowing of the purge is exactly right — only `torch_compile_cache/` and `rank_*/`, with `modelinfos/`, the FlashInfer JIT cache and the HF weights explicitly spared. I also checked the path construction for traversal: `model.replace("/", "__")` neutralises separators, so a model id like `../..` becomes `..__..` and cannot escape `lanes/`. That's sound.

Three things I'd want resolved before merge, then some smaller ones. The theme running through the first two: the detection is looser than the blast radius it triggers.

**Smaller points beyond the inline comments:**

- **The cooldown doesn't survive a process restart.** `_last_reactive_cache_recovery` is module-level with `time.monotonic()`. The code comment ("the lane manager recreates handles on restart") is accurate, but the stated goal — *"cannot loop purge → retry → fail forever"* — doesn't hold across container restarts, and repeated startup failures are a plausible cause of exactly those. Persisting it next to the version stamp would close the gap.
- **First deploy after this lands purges every warm lane once.** Existing caches have artifacts but no `cache_meta.json`, so the missing-meta branch wipes each one on its first `add_lane`. Intentional, but it's a one-time recompile tax across every model on every node — worth a release note so it isn't mistaken for the bug it's meant to fix.

logos/logos-workernode/logos_worker_node/vllm_process.py
wasnertobias: Both addressed in b00facd0b.

**1. Fingerprint tightened.** `inductor_artifact_missing` no longer matches on `torch/_inductor` + `FileNotFoundError` — that pair co-occurs in an unrelated startup failure (ordinary `torch/_inductor` compilation output plus the most common first-add error: a mistyped repo id, a gated repo, a missing file). It now requires a compile-cache path marker (`torch_compile_cache`) alongside the `FileNotFoundError`, so it only matches a file that is actually under the cache. The table comment no longer overclaims either.

**2. Worker-wide fallback reconsidered and scoped.** I split the two trigger signals rather than dropping the fallback outright, because they have different strength:
- a traceback that **names a file under the compile cache** (the path-fragment detector) is strong evidence — the failure actually reached into a cached artifact, so the worker-wide fallback still fires there. This preserves the legitimate legacy/shared-location recovery (an existing test covers it).
- a **fingerprint-only** match is a heuristic. On an empty per-lane dir it is now evidence *against* poisoning (a brand-new lane has no cache to be poisoned by), so we no longer widen to the worker-wide set — the original failure propagates instead. One mistyped model id can no longer force a full-node recompile.

This also composes with the point about the per-spawn global purge: the reactive path no longer needs an unconditional worker-wide fallback, and the proactive version-stamp check remains the worker-wide mechanism.

Added tests: a positive fingerprint match (a `FileNotFoundError` under a cache path, on a custom cache root so only the fingerprint can match), the false-positive from the comment (ordinary `torch/_inductor` output + an HF-path `FileNotFoundError` → no match), and a spawn-level test that a fingerprint-only match on an empty per-lane dir does not widen to the worker-wide cache (shared artifacts left intact, no retry). The two existing recovery tests (cooldown cap, structured log line) now use a path-fragment trigger so they keep exercising the worker-wide path.

wasnertobias:
One note on CI: the `Build Logos Worker Node (vLLM) / build` job is failing on the image **push**, not the build — the Harbor registry is at its storage cap:

```
failed to push harbor.aet.cit.tum.de/logos/logos-workernode-vllm: denied:
adding 3.6 GiB of storage resource, which when updated to current usage of
199.5 GiB will exceed the configured upper limit of 200.0 GiB.
```

This is a registry-storage / infrastructure issue independent of the change (the image builds successfully; only the push is rejected, and a re-run fails identically). All code-relevant checks — `pre-commit` and `Test` — are green, and the PR is `MERGEABLE`. It should clear once the registry has headroom.

PR #857, opened 2026-08-29
`Logos`: Size GPU auto-placement on the effective vLLM memory reservation

### LINKED ISSUE #570: `Logos`: Capacity planner / vLLM VRAM accounting mismatch on multi-GPU workers
## What

On a multi-GPU worker (`deioma`: 3× RTX PRO 6000 Blackwell Max-Q, ~96 GiB each), a `tp=2` lane for `google/gemma-3-27b-it` fails to start with:

```
ValueError: Free memory on device cuda:0 (79.12/94.97 GiB) on startup is less
than desired GPU memory utilization (0.95, 90.22 GiB).
```

…even though GPUs 1 and 2 were both ~empty (~96 GiB free each) and would have hosted gemma fine. The planner placed gemma on GPUs (0,1) where GPU 0 already had ~16 GiB used by a `deepseek-ai/DeepSeek-OCR-2` lane.

## Why

Two interacting pieces in `logos-workernode/logos_worker_node/lane_manager.py`:

1. **`_pick_best_gpu_subset` (line 1219)** does *best-fit / tight-pack* by `leftover = Σ(free − required)`, score sorted ascending (line 1240). Intentional — keeps wide free slots for future big models.

2. **`_estimate_lane_vram_mb`** returns `base_residency_mb + kv_cache_memory_bytes` (~45 GiB/GPU for gemma). But the actual vLLM reservation is governed by `gpu_memory_utilization × per_gpu_total` (0.95 × 97 GiB ≈ 90 GiB). The planner thinks gemma needs 45 GiB/GPU; vLLM tries to grab 90 GiB/GPU.

Concrete numbers for the failing call (DeepSeek-OCR-2 already on cuda:0):

| combo | per-GPU free | leftover after 45 GiB est. | picked by planner |
|---|---|---|---|
| (0, 1) | 81 / 96 | 36 + 51 = **87** | ✅ (smallest leftover wins) |
| (0, 2) | 81 / 97 | 36 + 52 = 88 | |
| (1, 2) | 96 / 97 | 51 + 52 = 103 | |

Best-fit picks (0,1) honestly given the 45 GiB estimate — but vLLM then needs 90 GiB/GPU, GPU 0 only has 81 GiB free → boom.

## Fix options

**A — reconcile estimate with vLLM's actual reservation.** In `_estimate_lane_vram_mb`, when `kv_cache_memory_bytes` is unset, fall back to `gpu_memory_utilization × per_gpu_total_mb` rather than just `base_residency_mb`. Conservative, matches what vLLM actually grabs.

**B — drop `gpu_memory_utilization` from configs.** Always set `kv_cache_memory_bytes` explicitly in `engines.vllm.model_overrides` (already recommended in `config.example.yml:220-233`). Then planner estimate ≡ vLLM reservation. Less invasive, but relies on every operator following the convention.

Probably do both — A as a defensive fix in the planner, B as a config cleanup on existing worker configs.

## Related — minor

While debugging this, also noticed `sync_logosnode_capabilities` in `dbutils/dbmanager.py:1715-1727` INSERTs into `profile_model_permissions`, a table that was dropped in migration 031/032. The transaction silently rolls back, so worker-announced models never auto-land in `models`. Worth a separate small fix (just delete the dead block). Logged as a footnote here in case it lands in the same area of the code.

## Repro env

- Worker: `ghcr.io/ls1intum/edutelligence/logos-workernode-vllm:latest` (commit d111b290)
- Host: deioma, 3× RTX PRO 6000 Blackwell Max-Q, driver 580.142
- vLLM 0.20.0, CUDA 13.1

wasnertobias:
## Context

On a mixed-occupancy multi-GPU worker, a vLLM lane with an explicit `gpu_memory_utilization` (deioma: `0.95`) could be placed on GPUs where vLLM then refuses to start:

```
ValueError: Free memory on device cuda:0 (79.12/94.97 GiB) on startup is less
than desired GPU memory utilization (0.95, 90.22 GiB).
```

The auto-placement feasibility floor in `LaneManager._auto_place_gpu_devices` used the constant `_VLLM_GMU_FLOOR = 0.5` for every non-kv-pinned lane — correct for vLLM's *auto-derived* GMU, which `VllmProcessHandle._resolve_gmu` clamps to `[0.5, 0.95]`. But `_resolve_gmu` passes an *explicit* operator `gpu_memory_utilization` to vLLM **verbatim, unclamped**, so vLLM reserved `0.95 × total` per GPU while placement only guaranteed `0.5 × total`. Any GPU with 48–90 GiB free on a 96 GiB card passed the placement gate, then vLLM crashed at startup. The planner's numbers and vLLM's reservation disagreed — the exact mismatch #570 describes.

The spread-placement change from #787 already stopped the planner from *preferring* a GPU with an awake lane (deioma's specific (0,1) pick no longer happens), but the accounting mismatch remained: whenever no fully-empty pair is available, placement could still select a subset that fits the 0.5 floor but not the actual reservation.

## Changes

- `logos/logos-workernode/logos_worker_node/lane_manager.py` — the per-GPU feasibility floor for non-kv-pinned lanes now uses the **effective GMU**, i.e. exactly what `VllmProcessHandle._resolve_gmu` hands to vLLM:
  - explicit `vllm_config.gpu_memory_utilization` → that value verbatim (no clamping, mirroring `_resolve_gmu`);
  - otherwise → `_VLLM_GMU_FLOOR` (the minimum the clamped auto-derivation can reach), as before.

  The floor still only *raises* the threshold on top of the calibrated-footprint estimate, and the existing kv-pinned skip (vLLM ignores GMU entirely when `kv_cache_memory_bytes` is pinned) is untouched. An explicit sub-floor value (e.g. 0.25) is honoured as-is instead of being inflated to 0.5 — the same accurate-accounting reasoning as the existing kv-pinned fix.

This is fix option A from the issue in the current architecture: the GMU reservation was deliberately moved out of `_estimate_lane_vram_mb` into the placement threshold (see the comment block there), so that is where the effective GMU is now accounted for. Option B (dropping explicit `gpu_memory_utilization` from worker configs) applies to production deployment configs such as deioma's, which are not in this repo — the planner is now defensive against them either way.

The issue's footnote (dead `profile_model_permissions` INSERT in `sync_logosnode_capabilities`) is intentionally left for a separate fix — it lives in the orchestrator's `dbmanager.py`, outside this area, and the issue itself flags it as "worth a separate small fix".

## Tests

Three new tests in `logos/logos-workernode/tests/test_lane_manager.py` covering the accounting with realistic multi-GPU, mixed-occupancy scenarios (tp=2, deioma-shaped 3× ~96 GiB node):

1. `test_auto_place_accounts_for_explicit_gmu_reservation` — the deioma topology from #570: awake DeepSeek-OCR lane on GPU 0, tp=2 gemma with explicit GMU 0.95. GPU 0 clears the old 0.5 floor (81 GiB free) but not the 0.95 reservation (~93 GiB), so the lane must land on `1,2`.
2. `test_auto_place_rejects_when_no_gpu_meets_explicit_gmu` — only one GPU meets the 0.95 reservation: placement must fail fast with a clear "no feasible GPU subset" error instead of spawning a lane that crashes vLLM at startup. **Fails on the pre-fix code** (it placed the lane).
3. `test_auto_place_explicit_gmu_below_floor_uses_operator_value` — an explicit sub-floor GMU is honoured verbatim (vLLM receives it unclamped), so a small model can place into a tight VRAM gap the old floor wrongly rejected. **Fails on the pre-fix code.**

## Verification

- `pytest tests/` in `logos/logos-workernode`: **529 passed, 2 skipped** (skips pre-existing) — no regressions.
- Verified tests 2–3 against the pre-fix `lane_manager.py` (temporarily restored via `git stash`): both fail there, both pass with the fix.
- Lint clean on the changed files: black 25.1.0, isort (profile=black), autoflake, flake8 with the repo config.
- CI note: the workernode CI job excludes `test_lane_manager.py` (GitHub-runner shutdown-signal issue documented in `logos_test.yml`; "The tests pass locally on macOS and Linux"), so the new tests are covered by the local verification above; the CI-run workernode suite is unaffected.

### DIFF
--- a/logos/logos-workernode/logos_worker_node/lane_manager.py
+++ b/logos/logos-workernode/logos_worker_node/lane_manager.py
@@ -1702,11 +1702,19 @@ class LaneManager:
         per_gpu_required_mb = required_total_mb / float(tp_size)
         per_gpu_threshold_mb = per_gpu_required_mb * _GPU_PLACEMENT_SECURITY_RATIO + _GPU_PLACEMENT_HEADROOM_OFFSET_MB
 
-        # When vLLM sizes its KV cache from gpu_memory_utilization, it clamps a
-        # small model's GMU up to _VLLM_GMU_FLOOR and pre-allocates that fraction
-        # of total GPU memory at startup; checking only the calibrated footprint
-        # would let placement succeed here but vLLM startup fail with
-        # "free memory < desired gpu_memory_utilization".
+        # When vLLM sizes its KV cache from gpu_memory_utilization, it
+        # pre-allocates effective-GMU × total per GPU at startup and refuses
+        # to start if free is below that; checking only the calibrated
+        # footprint would let placement succeed here but vLLM startup fail
+        # with "free memory < desired gpu_memory_utilization".
+        #
+        # The effective GMU is what VllmProcessHandle._resolve_gmu passes to
+        # vLLM: an explicit operator gpu_memory_utilization wins verbatim (no
+        # clamping), while the auto-derived value is clamped to
+        # [_VLLM_GMU_FLOOR, 0.95] — so _VLLM_GMU_FLOOR is the minimum the
+        # auto-derivation can reach. Gating the explicit value on the
+        # derivation floor would understate e.g. a 0.95 reservation as 0.5 and
+        # place the lane on a GPU vLLM cannot start on (#570).
         #
         # BUT when the lane pins kv_cache_memory_bytes, vLLM skips memory
         # profiling and ignores gpu_memory_utilization entirely (it logs exactly
@@ -1719,15 +1727,19 @@ class LaneManager:
         gpu_totals = [float(row.get("total_mb", 0.0)) for row in allowed_rows if row.get("total_mb", 0.0) > 0]
         if gpu_totals and not kv_pinned:
             min_gpu_total_mb = min(gpu_totals)
-            vllm_floor_mb = _VLLM_GMU_FLOOR * min_gpu_total_mb
+            gmu = _VLLM_GMU_FLOOR
+            explicit_gmu = lane_config.vllm_config.gpu_memory_utilization
+            if explicit_gmu is not None:
+                gmu = float(explicit_gmu)
+            vllm_floor_mb = gmu * min_gpu_total_mb
             if vllm_floor_mb > per_gpu_threshold_mb:
                 logger.debug(
                     "Auto-placement lane '%s': raising per-GPU threshold from %.0fMB to %.0fMB "
-                    "(vLLM GMU floor %.2f × %.0fMB GPU total)",
+                    "(vLLM GMU %.2f × %.0fMB GPU total)",
                     lane_id,
                     per_gpu_threshold_mb,
                     vllm_floor_mb,
-                    _VLLM_GMU_FLOOR,
+                    gmu,
                     min_gpu_total_mb,
                 )
                 per_gpu_threshold_mb = vllm_floor_mb
--- a/logos/logos-workernode/logos_worker_node/lane_manager.py
+++ b/logos/logos-workernode/logos_worker_node/lane_manager.py
@@ -27,7 +27,7 @@ from logos_worker_node.models import (
     VllmEngineConfig,
 )
 from logos_worker_node.ollama_process import OllamaProcessHandle
-from logos_worker_node.vllm_process import VllmProcessHandle
+from logos_worker_node.vllm_process import VllmProcessHandle, effective_gmu
 
 logger = logging.getLogger("logos_worker_node.lane_manager")
 
@@ -46,13 +46,23 @@ _HANDLE_CLOSE_TIMEOUT = 10
 # on top of the estimate. OFFSET_MB adds a fixed safety margin on top.
 _GPU_PLACEMENT_SECURITY_RATIO = 1.005
 _GPU_PLACEMENT_HEADROOM_OFFSET_MB = 0.0
-# vLLM clamps auto-derived gpu_memory_utilization to this floor (see VllmProcessHandle).
-# Placement must treat this as the minimum per-GPU reservation when the model's actual
-# footprint is smaller, otherwise placement succeeds but vLLM startup fails with
-# "Free memory ... less than desired GPU memory utilization".
-_VLLM_GMU_FLOOR = 0.5
 _CRASH_RESTART_COOLDOWN_S = 30.0
 _MAX_CRASH_RESTARTS = 5  # per lane; budget resets on confirmed successful restart
+
+
+def _placement_threshold_mb(row: dict[str, float], base_threshold_mb: float, gmu: float | None) -> float:
+    """Per-GPU placement threshold for one snapshot row.
+
+    vLLM's startup gate (vllm/v1/worker/utils.py:request_memory) checks
+    free >= gmu × total on each card individually, so on mixed-size nodes the
+    GMU reservation must be sized against the row's own total rather than a
+    global minimum. Rows with an unknown total — or no GMU bound in play (kv
+    pinned) — fall back to the base (footprint-derived) threshold alone.
+    """
+    total_mb = float(row.get("total_mb", 0.0) or 0.0)
+    if gmu is None or total_mb <= 0:
+        return base_threshold_mb
+    return max(base_threshold_mb, gmu * total_mb)
 
 
 # Bounded drain of in-flight requests before sleeping/stopping a lane, so an
@@ -1590,8 +1600,16 @@ class LaneManager:
         multi_gpu_indices: set[int] | None = None,
         awake_lanes_by_gpu: dict[int, int] | None = None,
         awake_used_mb_by_gpu: dict[int, float] | None = None,
+        gmu: float | None = None,
     ) -> list[int] | None:
-        feasible = [row for row in device_rows if float(row["free_mb"]) >= per_gpu_threshold_mb]
+        # Per-row: vLLM's startup gate is per card (free >= gmu × that card's
+        # total), so on mixed-size nodes each row is checked against its own
+        # total (None gmu = kv-pinned lane, base threshold only).
+        feasible = [
+            row
+            for row in device_rows
+            if float(row["free_mb"]) >= _placement_threshold_mb(row, per_gpu_threshold_mb, gmu)
+        ]
         if len(feasible) < tp_size:
             return None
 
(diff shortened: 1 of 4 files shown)


logos/logos-workernode/logos_worker_node/lane_manager.py
```diff
@@ -1719,15 +1727,19 @@ async def _auto_place_gpu_devices(self, lane_id: str, lane_config: LaneConfig) -
         gpu_totals = [float(row.get("total_mb", 0.0)) for row in allowed_rows if row.get("total_mb", 0.0) > 0]
         if gpu_totals and not kv_pinned:
             min_gpu_total_mb = min(gpu_totals)
-            vllm_floor_mb = _VLLM_GMU_FLOOR * min_gpu_total_mb
+            gmu = _VLLM_GMU_FLOOR
+            explicit_gmu = lane_config.vllm_config.gpu_memory_utilization
+            if explicit_gmu is not None:
+                gmu = float(explicit_gmu)
+            vllm_floor_mb = gmu * min_gpu_total_mb
```
wasnertobias: `min(gpu_totals)` produces a single threshold that is then checked against every candidate GPU's `free_mb` (`:1788` and in the subset picker), but vLLM gates per card on `free >= gmu × that_card_total`.

On mixed-size nodes this under-states the requirement for the larger cards:

> GPU 0 = 24GB, GPU 1 = 48GB, explicit `gmu=0.9`, tp=1.
> `min_gpu_total` = 24GB → threshold 22.1GB. GPU 1 has 25GB free → passes.
> vLLM starts on GPU 1, demands 0.9 × 48GB = 44.2GB → the exact crash this PR fixes.

Homogeneous nodes like the one in the issue aren't affected, but the code is generic. Checking the GMU bound per row against its own `total_mb` rather than against a global minimum would close it. Happy to drop this if mixed-GPU worker nodes aren't a case you support.

logos/logos-workernode/logos_worker_node/lane_manager.py
```diff
@@ -1719,15 +1727,19 @@ async def _auto_place_gpu_devices(self, lane_id: str, lane_config: LaneConfig) -
         gpu_totals = [float(row.get("total_mb", 0.0)) for row in allowed_rows if row.get("total_mb", 0.0) > 0]
         if gpu_totals and not kv_pinned:
             min_gpu_total_mb = min(gpu_totals)
-            vllm_floor_mb = _VLLM_GMU_FLOOR * min_gpu_total_mb
+            gmu = _VLLM_GMU_FLOOR
```
wasnertobias: This branch is a second implementation of `_resolve_gmu`'s explicit-vs-derived logic, and it depends on two separately declared constants agreeing:

- `_GMU_AUTO_FLOOR: ClassVar[float] = 0.5` — `vllm_process.py:1165`
- `_VLLM_GMU_FLOOR = 0.5` — `lane_manager.py:53`

The reason correctness rests on that: for auto-derived values `gmu × total == loaded/tp == per_gpu_required_mb`, so the calibrated threshold already covers the reservation and this floor branch only matters in the clamp-up case. Raise `_GMU_AUTO_FLOOR` to 0.6 and leave `_VLLM_GMU_FLOOR` at 0.5, and vLLM reserves 0.6 × total while placement checks 0.5 × total — the same failure, for every small model, with nothing to catch it.

A shared `effective_gmu(vllm_config)` helper covering the explicit branch plus the floor would make that harder to break. It can't cover the auto-derivation itself (that needs `_model_profiles` and `_per_gpu_total_mb()`), but that part is already implied by the calibrated threshold.

logos/logos-workernode/logos_worker_node/lane_manager.py
```diff
@@ -1702,11 +1702,19 @@ async def _auto_place_gpu_devices(self, lane_id: str, lane_config: LaneConfig) -
...
+        #
+        # The effective GMU is what VllmProcessHandle._resolve_gmu passes to
+        # vLLM: an explicit operator gpu_memory_utilization wins verbatim (no
+        # clamping), while the auto-derived value is clamped to
+        # [_VLLM_GMU_FLOOR, 0.95] — so _VLLM_GMU_FLOOR is the minimum the
+        # auto-derivation can reach. Gating the explicit value on the
+        # derivation floor would understate e.g. a 0.95 reservation as 0.5 and
+        # place the lane on a GPU vLLM cannot start on (#570).
         #
         # BUT when the lane pins kv_cache_memory_bytes, vLLM skips memory
```
wasnertobias: This skip contradicts the field's own documentation. `models.py:83-85`:

> "vLLM uses this as a startup-floor gate: free_per_gpu must be >= gmu * total_per_gpu or the engine refuses to start, **even when kv_cache_memory_bytes is set**."

Both can't be true. And `_build_cmd` (`vllm_process.py:1397-1399`) emits `--gpu-memory-utilization` regardless of pinning — so a lane with explicit GMU *and* a pinned KV gets 0.95 passed to vLLM while placement never checked whether 0.95 × total is free.

If the field text is right, that's the same failure class this PR closes, still open in the pinned branch. One plausible reconciliation: vLLM skips memory *profiling* (KV sizing) but still applies the startup *gate* — in which case both comments are half right and skipping the floor is wrong.

Not introduced here and I wouldn't ask you to change the skip in this PR. But one of the two comments is demonstrably wrong, and it'd be good to know which before this merges.

logos/logos-workernode/logos_worker_node/lane_manager.py
```diff
@@ -1702,11 +1702,19 @@ async def _auto_place_gpu_devices(self, lane_id: str, lane_config: LaneConfig) -
...
+        # footprint would let placement succeed here but vLLM startup fail
+        # with "free memory < desired gpu_memory_utilization".
+        #
+        # The effective GMU is what VllmProcessHandle._resolve_gmu passes to
+        # vLLM: an explicit operator gpu_memory_utilization wins verbatim (no
+        # clamping), while the auto-derived value is clamped to
+        # [_VLLM_GMU_FLOOR, 0.95] — so _VLLM_GMU_FLOOR is the minimum the
+        # auto-derivation can reach. Gating the explicit value on the
+        # derivation floor would understate e.g. a 0.95 reservation as 0.5 and
+        # place the lane on a GPU vLLM cannot start on (#570).
```
wasnertobias: `(#570)` shouldn't go into the code — same for the two test docstrings ("Reproduces #570…", "Same node as #570…"). The PR description is the place for it.

wasnertobias (commented):
Verified the core claim and it holds: `_resolve_gmu:1186-1187` returns an explicit value unclamped while the auto-derivation is clamped, so gating placement on the derivation floor really did understate explicit reservations. `gpu_memory_utilization` is Pydantic-bounded to `[0.1, 1.0]` (`models.py:78-81`), the kv-pinned skip is untouched, and the threshold is still only ever raised. Tests 2 and 3 were checked against pre-fix code — good practice.

Three things inline. Two smaller ones:

- **The failure message now explains a threshold it didn't produce.** `lane_manager.py:1811` renders `per-GPU threshold≈{X}MB (estimate {Y}MB × 1.005 + 0MB)`. When the GMU floor overrode the threshold, that arithmetic visibly doesn't add up, and this is the message an operator reads when a lane won't come up at 3am. The PR makes it more common, since the threshold can now reach 0.95 × total instead of a fixed 0.5 × total. Worth saying "raised to GMU reservation" when that's what happened.
- **`logger.debug` at :1736 is quiet for a value that decides placement.** It fires once per lane start, not in a loop — `info` would make it possible to reconstruct afterwards why the threshold jumped to 93GB.

Also: the PR notes CI excludes `test_lane_manager.py`, so the two regression-catching tests never run automatically. Not an objection, but the exclusion gets more expensive with every fix in this file.

logos/logos-workernode/logos_worker_node/lane_manager.py
wasnertobias: Fixed. The GMU bound is now checked per row against its own `total_mb` via a new `_placement_threshold_mb(row, base, gmu)` helper computing `max(base, gmu × row_total)`. It is used in the subset picker's feasibility filter, the sticky-placement check, and the fail-fast error message, which now prints per-GPU free *and* needed MB so a rejected subset is self-explanatory. Rows with an unknown total fall back to the base (footprint-derived) threshold. Added `test_auto_place_gates_mixed_size_gpus_on_own_card_total` with your 24 GB + 48 GB example (explicit GMU 0.9, 25 GB free on the 48 GB card): it fails against the pre-fix code, which passed the 48 GB card on the 22.1 GB global-min threshold, and passes now. Homogeneous nodes are unaffected — per-card bound equals the global bound there, and all pre-existing placement tests pass unchanged.

logos/logos-workernode/logos_worker_node/lane_manager.py
wasnertobias: Fixed. `vllm_process.py` now defines module-level `GMU_AUTO_FLOOR` / `GMU_AUTO_CEILING` plus a shared `effective_gmu(vllm_config)` helper (explicit operator value verbatim, else the floor). `VllmProcessHandle._GMU_AUTO_FLOOR` / `_GMU_AUTO_CEILING` are now aliases of those constants, and `_resolve_gmu`'s explicit branch calls the shared helper instead of re-implementing it — so the spawner's explicit branch and the placement bound are literally the same code. `lane_manager` drops its own `_VLLM_GMU_FLOOR` and imports `effective_gmu`. Raising the floor to 0.6 in one place now moves both sides, as you described.

logos/logos-workernode/logos_worker_node/lane_manager.py
wasnertobias: Investigated against the vLLM source (v0.28.0, the version the worker Dockerfile pins; cross-checked v0.20.0, the repro version — same behavior in both). Your reconciliation is the right one, and the field doc is the correct of the two:

- `vllm/v1/worker/utils.py:request_memory` computes `requested = ceil(total × gpu_memory_utilization)` and raises exactly the "Free memory ... on startup is less than desired GPU memory utilization" error if `free < requested`. It is called **unconditionally in `GPUWorker.init_device`** — after the NCCL init snapshot, before model load and before `determine_available_memory` — with no knowledge of `kv_cache_memory_bytes`.
- The "skipped memory profiling ... does not respect the gpu_memory_utilization config" message the old comment quoted comes from the kv-pinned branch of `determine_available_memory`, which only governs **KV sizing** (the KV cache is then exactly the pinned bytes). vLLM's own `CacheConfig` docstring says the same thing: `kv_cache_memory_bytes` "ignores gpu_memory_utilization" — in the sizing sense only. vLLM's config also confirms `_build_cmd`'s behavior: it passes `--gpu-memory-utilization` whenever a resolved GMU exists, and that value is what `request_memory` gates on.

So: a pinned lane still hits the startup gate at `free >= gmu × total` (explicit GMU, auto-derived, or vLLM's 0.9 default when neither is present). The old placement comment's "ignores gpu_memory_utilization entirely" was the wrong one.

Changes (no behavior change to the skip itself, per your note):
- Rewrote the placement comment in `lane_manager.py` to state both halves accurately: GMU is not respected for KV sizing of pinned lanes (that part of the old comment was right), **and** the startup gate still applies to pinned lanes — so a pinned lane placed into tight free VRAM can still fail it at spawn. The comment now names that as the residual risk of the skip and defers the trade-off (honouring the gate re-introduces the over-rejection the skip exists to avoid) to a follow-up.
- Fixed the same class of wrong claim in `config.example.yml`: "When set, gpu_memory_utilization is ignored by vLLM (overridden to 0.1)" — there is no such 0.1 override anywhere (repo or vLLM); it now says the KV is sized from the pinned value but the startup gate still applies.

I'd suggest a follow-up issue for the pinned-lane gate itself (either honour it in placement or document the operator expectation), so it isn't lost.

logos/logos-workernode/logos_worker_node/lane_manager.py
wasnertobias: Done — removed the `(#570)` reference from the `lane_manager.py` comment and rewrote the two test docstrings ("Reproduces the deioma scenario..." / "Same node as the deioma scenario..."). The issue linkage now lives only in the PR description (and the commit messages, which follow the repo's `(#NNN)` convention).

PR #859, opened 2026-08-29
`Logos`: Support RoPE scaling techniques (YaRN & friends) for long-context models

### LINKED ISSUE #744: `Logos`: Support RoPE scaling techniques
> For long-horizon tasks where the total length (including both input and output) exceeds this limit, we recommend using RoPE scaling techniques to handle long texts effectively, e.g., YaRN.
https://huggingface.co/Qwen/Qwen3.8-27B

e.g.

> VLLM_ALLOW_LONG_MAX_MODEL_LEN=1 vllm serve ... --hf-overrides '{"text_config": {"rope_parameters": {"mrope_interleaved": true, "mrope_section": [11, 11, 10], "rope_type": "yarn", "rope_theta": 10000000, "partial_rotary_factor": 0.25, "factor": 4.0, "original_max_position_embeddings": 262144}}}' --max-model-len 1000000

wasnertobias:
## Summary

Long-horizon workloads exceed a model's native context window; the usual remedy is a RoPE scaling technique (e.g. YaRN), which vLLM applies by reading the scaling settings from the model's HF config. Operators previously could only reach that knob through the raw `extra_args` pass-through with a hand-written `--hf-overrides` JSON blob. This change exposes `rope_scaling` as a first-class, validated option in the worker node's per-model configuration so it can be enabled per model/provider.

## Changes

- `logos/logos-workernode/logos_worker_node/models.py`
  - New `RopeScalingConfig` schema: `rope_type` (required, validated against the techniques vLLM supports: yarn, linear, dynamic, llama3, longrope, deepseek_yarn, dynamic_yarn, default), `factor` (≥ 1.0), `original_max_position_embeddings` (> 0), `rope_theta`, and `config_path` (dotted path into the model config, default `rope_scaling`). The yarn techniques require `factor` + `original_max_position_embeddings`; unknown `rope_type`s and `config_path` values with path traversal are rejected at config-validation time. Family-specific keys (`mrope_section`, `partial_rotary_factor`, `attention_factor`, …) pass through verbatim (`extra="allow"`).
  - `VllmConfig.rope_scaling: RopeScalingConfig | None = None` — **disabled by default**; an unconfigured lane can never silently serve a scaled context. A model validator makes `rope_scaling` mutually exclusive with a `--hf-overrides` flag in `extra_args` (both would set the same vLLM flag and only the last occurrence would take effect — a silent-wrong-config hazard).
  - `to_hf_overrides()` builds the `--hf-overrides` payload: the scaling dict under `config_path` (default top-level `rope_scaling`; e.g. `text_config.rope_parameters` for Qwen3.5/3.8-style multimodal configs — the exact layout from the issue), with unset optionals omitted so the deep merge never clobbers the model's own values with nulls.
- `logos/logos-workernode/logos_worker_node/vllm_process.py` — `_build_cmd` emits `--hf-overrides <json>` when `rope_scaling` is set (vLLM has no dedicated CLI flag for RoPE scaling).
- `logos/logos-workernode/logos_worker_node/calibration.py` — the calibration probe receives the same setting (`plans_from_config` already carries `model_overrides` keys through, same as `extra_args`/`--hf-overrides`). Without this, calibration would measure the *unscaled* model's context window and the planner would size KV budgets against a window the serving lane never runs. The same conflict rule as `VllmConfig` is enforced here.
- `logos/logos-workernode/logos_worker_node/lane_manager.py` — `_lane_needs_restart` compares `rope_scaling`, so changing the setting (incl. family-specific extras) restarts the lane instead of silently keeping the old engine args.
- `logos/logos-workernode/config.example.yml` — documents the new key (full key reference under `model_overrides`, one-liner in the static-lane field list, worked YaRN example).

Per-model reach for both lane sources: static lanes via `lanes[].vllm_config.rope_scaling`, and — for server-driven lanes — via `engines.vllm.model_overrides.<model>.rope_scaling`, which the worker merges into every incoming lane config at spawn. No DB/orchestrator changes: the orchestrator's `vllm_config` already flows through the same `VllmConfig` validation, and the operator-facing per-model vLLM knobs live in the worker config.

## Testing

- New schema-validation tests in `tests/test_lane_models.py` (defaults, type/requirement validation, factor ≥ 1, unknown types, path traversal, family-key pass-through, `to_hf_overrides()` shape incl. the issue's `text_config.rope_parameters` block, unset keys omitted, extra-args conflict in both flag forms).
- New arg-assembly tests in `tests/test_vllm_process.py` (`_build_cmd` emits no `--hf-overrides` by default; emits the exact JSON for the default and the nested `config_path` case).
- New tests in `tests/test_calibration.py` (plan forwarding of `rope_scaling` from `model_overrides` to the probe command, conflict rule, no flag when unset) and `tests/test_lane_manager.py` (override merge sets `rope_scaling`; `_lane_needs_restart` on change in both directions).
- Verified the issue's Qwen3.8-27B example end-to-end through the schema: it produces exactly the `--hf-overrides` payload quoted in the issue.
- Ran the full worker-node suite: **547 passed, 2 skipped** (was 526/2 before this change). Also ran the CI-equivalent selection (`pytest --timeout=30` excluding the three files CI ignores) and the repo's pre-commit hooks (black, isort, autoflake, flake8, yaml checks) on all touched files.

Note: full deployment (GPU, vLLM) is not available in this environment; vLLM-side behaviour is covered by the schema/assembly tests plus the payload matching the issue's documented example. If a model needs an env-level bypass such as `VLLM_ALLOW_LONG_MAX_MODEL_LEN=1`, that remains available via the existing `env_overrides`.

## Summary by CodeRabbit

* **New Features**
  * Added configurable RoPE scaling for vLLM models, including YaRN and other supported techniques.
  * Supports long-context model configurations and nested model-specific settings.
  * Scaling options are applied automatically when starting model processes.

* **Bug Fixes**
  * Invalid or conflicting scaling configurations now produce clear validation errors.
  * Lane processes automatically restart when RoPE scaling settings change.

* **Documentation**
  * Added configuration guidance and a working long-context model example.

### DIFF
--- a/logos/logos-workernode/config.example.yml
+++ b/logos/logos-workernode/config.example.yml
@@ -307,6 +307,32 @@ engines:
     #     Example: {enable_thinking: false} to opt-out of Gemma 4 thinking mode.
     #     Example: {enable_thinking: false} to disable Qwen3 chain-of-thought.
     #
+    #   rope_scaling: dict        — RoPE scaling for long-context variants,
+    #     e.g. YaRN to extend the context window beyond the model's native
+    #     limit. vLLM reads the technique from the model's HF config, so this
+    #     is passed as an --hf-overrides deep merge into that config (there
+    #     is no dedicated CLI flag). Unset (default) = disabled: the model's
+    #     own rope settings are used as-is.
+    #     Keys:
+    #       rope_type: str (required) — "yarn" | "linear" | "dynamic" |
+    #         "llama3" | "longrope" | "deepseek_yarn" | "dynamic_yarn"
+    #         (| "default" to suppress a scaling the model config ships).
+    #       factor: float — scaling factor, e.g. 4.0 = 4x the native window.
+    #         Required for the yarn techniques.
+    #       original_max_position_embeddings: int — the model's native context
+    #         limit in tokens, before scaling. Required for the yarn
+    #         techniques.
+    #       rope_theta: float — rotary base frequency, e.g. 10000000. Optional.
+    #       config_path: str — where in the model config the dict is written;
+    #         default "rope_scaling". Qwen3.5/3.8-style multimodal configs
+    #         keep their rope settings under "text_config.rope_parameters".
+    #     Any further keys (mrope_section, partial_rotary_factor, ...) are
+    #     passed through to vLLM verbatim. Mutually exclusive with a
+    #     --hf-overrides flag in extra_args. The calibration probe receives
+    #     the same setting, so the profile describes the lane that serves.
+    #     Example: {rope_type: yarn, factor: 4.0,
+    #               original_max_position_embeddings: 32768}
+    #
     #   extra_args: list           — arbitrary extra CLI flags passed verbatim to
     #     "vllm serve". Use for features not yet exposed as first-class fields.
 
@@ -346,6 +372,14 @@ engines:
       some-org/some-qwen3-model:
         chat_template_kwargs:
           enable_thinking: false          # disable chain-of-thought
+
+      # Long-context variant: YaRN scaling 4x on a 32K model (issue #744).
+      # Lane and calibration run with the scaled context window.
+      some-org/some-32k-model-long:
+        rope_scaling:
+          rope_type: yarn
+          factor: 4.0
+          original_max_position_embeddings: 32768
 
       # Custom chat template: the file must exist in the persistent template
       # directory (/opt/logos-workernode/chat-templates on the host, mounted at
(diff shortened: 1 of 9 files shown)


wasnertobias:
Closing: RoPE scaling can already be configured today without this layer.

vLLM has no dedicated CLI flag for it, but `extra_args` / `global_extra_args` are passed verbatim to the vLLM CLI (`vllm_process.py:1528-1529`), so `--hf-overrides '{"rope_scaling": {...}}'` works as-is — and `extra_args` already reaches calibration through `model_overrides` (covered by `test_plans_from_config_forwards_extra_args`), so the profile stays consistent with what the lane serves.

What this PR adds on top is validation, the `config_path` handling for nested configs, and null-safe merging — convenience rather than capability. Not worth ~180 lines of schema and docs for now, especially as it introduces an inconsistency of its own: the collision between `rope_scaling` and an existing `--hf-overrides` raises hard in calibration (`calibration.py:891`) but is unchecked in the serving path (`vllm_process.py:1439`), where `extra_args` is appended later and silently wins.

Reopen if a concrete model turns up where the raw `--hf-overrides` route proves too error-prone in practice.

PR #861, opened 2026-08-29
`Logos`: Use remaining GPUs during calibration on heterogeneous nodes

### LINKED ISSUE #592: `Logos`: Use remaining GPUs during calibration on heterogeneous nodes
## Problem

Calibration today acquires the whole worker node for its run, gated by the orchestrator's `_provider_has_active_requests` check. On nodes with a non-power-of-two GPU count (e.g. 3 GPUs), the tensor-parallel size is rounded down to the largest power of two (tp=2), so the leftover GPU(s) sit idle for the entire calibration window — often several minutes per model.

Background discussion: this came up while implementing the model-level unsupported-list and node-health features. The trigger was the realization that 1 of 3 cards on some workers contributes zero throughput during the nightly maintenance window.

## Proposed change

Allow production lanes to keep serving requests on GPUs the calibration isn't using:

1. **Pin calibration to a specific `gpu_devices` subset** at plan time (today most plans leave it as `"all"`, which makes the worker target every GPU). Pick the largest power-of-two slice — typically GPUs 0..tp-1.
2. **Teach the lane manager** to refuse new lanes that would touch the calibration's GPU subset, but allow lanes on the leftover GPU(s).
3. **Tighten VRAM measurement** to only sum over the calibration's GPUs. `sample_vram_mb(gpu_indices)` already respects this; the gap is making sure the calibration plan passes a concrete `gpu_devices` value (not blank/all) so the measurement set is well-defined. Verify and add a regression test.
4. **Relax `_provider_has_active_requests`** so it only counts requests on the calibration's GPU subset — requests on the leftover GPUs no longer block scheduling.

## Open design choices

- **TP fallback during calibration**: today the planner has a tp-escalation/fallback path. If calibration is using GPUs 0–1 and a request needs Qwen3-Embedding-8B (whose calibrated profile is tp=2), do we let it spawn at tp=1 on GPU 2 as a fallback, or just queue/reject?
  - Recommended: leave the planner's existing tp-fallback alone. Only models whose smallest workable tp fits on the leftover GPU(s) get served during calibration. Everything else queues.
- **NCCL topology**: workers without NVLink already set `NCCL_P2P_DISABLE=1` (visible in calibration logs). Concurrent vLLM on the leftover GPU shouldn't interact with the calibration's NCCL ring, but worth verifying on a real 3-GPU host before flipping the default on.

## Risks

- **VRAM baseline contamination**: if the calibration plan doesn't pin `gpu_devices`, `query_gpu_vram(None)` sums all GPUs and the production lane's footprint corrupts the calibration's measurement (silently). The pinning step (1) is load-bearing.
- **Power draw / thermals**: running calibration + production simultaneously on the same chassis hits PSU/thermal limits harder than nightly-window-only calibration. Worth a watch when first enabled.

## Out of scope

- Cross-node coordination (still one calibration per worker at a time).
- Reordering calibration to prefer the largest workable tp first vs. last (separate optimization).

## Acceptance criteria

- A 3-GPU node calibrating at tp=2 can simultaneously serve requests on the remaining GPU.
- Calibration's `base_residency_mb` measurement is unaffected by what runs on the leftover GPU(s) — regression test that fakes a concurrent lane.
- `_provider_has_active_requests` no longer treats lanes on leftover GPUs as "busy" for the purpose of calibration eligibility.

---

🤖 Filed via Claude Code based on a discussion during feature-#3 implementation. Original session context: deioma 2026-06-04 incident + heterogeneous-GPU optimization brainstorm.

wasnertobias:
## Summary
On a heterogeneous node (e.g. 3 GPUs) the calibration pinned its vLLM probe to *every* GPU — the plan left `gpu_devices` blank/"all" — so the leftover GPUs sat idle for the whole nightly window, and any production lane already running on a leftover GPU silently inflated the measured `base_residency_mb`. This PR pins the calibration to the node's largest power-of-two GPU slice, tears down **only that slice** at session start (leftover lanes keep serving), guards the slice in the lane manager, and relaxes the orchestrator's "is this worker busy" check so leftover-GPU lanes no longer block a session.

All four changes proposed in the issue are implemented; all three acceptance criteria are covered by tests.

## Changes
- `logos-workernode/logos_worker_node/calibration.py`: new `calibration_gpu_slice()`, `_plan_needs_gpu_pin()`, `pin_plan_gpu_devices()`. The pin is applied at the single `calibrate_with_tp_escalation()` choke point, so the probe gets a concrete `gpu_devices` slice — scoping `CUDA_VISIBLE_DEVICES` and making the VRAM baseline a well-defined sum over exactly the GPUs the probe uses. The slice length equals the max tensor-parallel size, so the existing max-first TP escalation is unchanged.
- `logos-workernode/logos_worker_node/lane_manager.py`: `begin_calibration_session()` / `end_calibration_session()` hold the slice; `destroy_lanes_on_gpus()` frees only the slice's vLLM lanes; auto-placement excludes the held slice and fails fast if no leftover subset fits; an explicit lane pin landing on the slice is refused. Leftover GPUs stay placeable.
- `logos-workernode/logos_worker_node/logos_bridge.py`: the calibration session now calls `begin_calibration_session()` + `destroy_lanes_on_gpus()` (instead of `destroy_all()`) so leftover lanes keep serving, and releases the slice in the `finally` block.
- `logos-orchestrator/src/logos/capacity/calibration_orchestrator.py`: `_provider_has_active_requests()` now counts only lanes that touch the calibration slice (new `_calibration_gpu_subset()` + `_lane_touches_subset()` helpers). When the slice cannot be determined (no telemetry / unknown GPU count) the previous conservative behaviour applies — any busy lane blocks.

## New Endpoints
n/a

## Testing
- **22 new tests** in `logos-orchestrator/tests/unit/capacity/test_calibration_orchestrator_gpu_slice.py`: slice derivation from the reported GPU count, selector-vs-slice intersection (blank/"all" span everything, "none" spans nothing, explicit lists intersect), leftover-GPU lanes no longer blocking, and the conservative fallbacks.
- **18 new tests** in `logos-workernode/tests/` across `test_calibration.py` (incl. the VRAM regression — a leftover-GPU lane does **not** inflate `base_residency_mb`; and TP escalation stays unchanged when the plan is pinned to the slice), `test_lane_manager.py` (slice hold/release, slice-only teardown, auto-placement exclusion, explicit-pin refusal), and `test_logos_bridge.py` (the session frees only the slice, not all lanes).
- Full suites green: worker **544 passed, 2 skipped** (pre-existing skips); orchestrator **1039 passed, 1 xfailed** (pre-existing xfail). Linters clean (black, isort, flake8, autoflake).

## CI status
- **All checks passing.** `Logos - Test` (runs both the orchestrator and worker Python suites), `Logos - Lint`, `Logos - Build`, and `Validate PR Title` are green.
- Note for reviewers: the `Logos - Build` worker-vLLM image push **initially** failed with a **shared Harbor registry storage-quota** error (registry at 199.2 GiB of its 200 GiB project cap: `denied: adding 3.6 GiB ... will exceed the configured upper limit of 200.0 GiB`). The failure was at the final **push** step, after every code-copy/dependency step succeeded — i.e. not a code regression. It passed on re-run once the registry had headroom.

## Open questions / current status
- Server-planner-driven **new** placements onto the leftover GPUs during a live session are intentionally **not** enabled here: the capacity planner still excludes calibrating workers via `_is_plannable`, and actively co-locating a new tp=1 lane on the third GPU alongside a tp=2 probe is worth verifying on a real 3-GPU host (NCCL behaviour) before flipping that default. What **is** enabled and verified is the issue's core: the leftover GPUs are no longer torn down, so **existing** leftover lanes keep serving while the probe runs on the slice, and their traffic no longer distorts the calibration.

### DIFF
--- a/logos/logos-workernode/logos_worker_node/calibration.py
+++ b/logos/logos-workernode/logos_worker_node/calibration.py
@@ -2910,6 +2910,53 @@ def _max_tp_for_plan(plan: dict[str, Any], available_gpus: int) -> int:
     return 1 << (n.bit_length() - 1)
 
 
+def calibration_gpu_slice(available_gpus: int) -> list[int]:
+    """Return the GPU indices a calibration run should be pinned to.
+
+    The largest power-of-two slice of the node's GPUs, starting at index 0
+    (3 GPUs → ``[0, 1]``, 5 → ``[0, 1, 2, 3]``, 8 → ``[0..7]``). The slice
+    length equals the maximum tensor-parallel size the node supports, so the
+    existing max-first TP escalation fits inside it unchanged.
+
+    Pinning the calibration to this concrete slice (issue #592) leaves the
+    leftover GPUs (``slice_size..N-1``) free for production lanes, which would
+    otherwise sit idle for the entire calibration window, and makes the VRAM
+    baseline a well-defined sum over exactly the GPUs the probe uses.
+    """
+    n = int(available_gpus) if available_gpus else 0
+    if n < 1:
+        return []
+    slice_size = 1 << (n.bit_length() - 1)
+    return list(range(slice_size))
+
+
+def _plan_needs_gpu_pin(gpu_devices: str) -> bool:
+    """True when a plan's ``gpu_devices`` is still the 'use every GPU' default.
+
+    Blank and "all" mean the plan has not been pinned yet, so the calibration
+    slice may be filled in. A specific set ("0,1") or "none" is an explicit
+    operator choice and is never overridden.
+    """
+    return (gpu_devices or "").strip().lower() in {"", "all"}
+
+
+def pin_plan_gpu_devices(plan: dict[str, Any], available_gpus: int) -> dict[str, Any]:
+    """Return *plan* with a concrete ``gpu_devices`` slice.
+
+    A plan that already names a specific GPU set — or "none" — is returned
+    unchanged: the operator's choice is authoritative. A plan left at
+    "all"/blank is pinned to the node's calibration slice so the probe only
+    touches that slice (``CUDA_VISIBLE_DEVICES``) and its VRAM baseline is
+    measured over a well-defined set of GPUs (issue #592).
+    """
+    if not _plan_needs_gpu_pin(str(plan.get("gpu_devices") or "")):
+        return plan
+    slice_ = calibration_gpu_slice(available_gpus)
+    if not slice_:
+        return plan
+    return {**plan, "gpu_devices": ",".join(str(i) for i in slice_)}
+
+
 def _reset_calibration_log(log_dir: Path, model_name: str) -> None:
     log_path = log_dir / f"{model_name.replace('/', '__')}.log"
     try:
@@ -2995,6 +3042,12 @@ def calibrate_with_tp_escalation(
             available_gpus = len(query_gpu_vram())
         except Exception:
             available_gpus = 1
+
+    # Pin the run to a concrete GPU slice before any probe: with "all"/blank
+    # the probe sees every GPU and its VRAM baseline sums them all, so a
+    # production lane on a leftover GPU silently inflates base_residency_mb.
+    # The slice length equals the max TP below, so the escalation is unchanged.
+    plan = pin_plan_gpu_devices(plan, available_gpus)
 
     original_tp = int(plan.get("tensor_parallel_size", 1))
     max_tp = _max_tp_for_plan(plan, available_gpus)
--- a/logos/logos-orchestrator/src/logos/capacity/calibration_orchestrator.py
+++ b/logos/logos-orchestrator/src/logos/capacity/calibration_orchestrator.py
@@ -289,18 +289,82 @@ class CalibrationOrchestrator:
             return False
         return not bool(node_health.get("healthy", True))
 
+    def _calibration_gpu_subset(self, provider_id: int) -> frozenset[int]:
+        """GPUs the calibration holds on *provider_id* (issue #592).
+
+        Derived from the worker's nvidia GPU count as the largest power-of-two
+        slice — the same slice the worker's lane manager holds and the probe is
+        pinned to (see ``calibration_gpu_slice`` in the worker). Returns an
+        empty set when the count cannot be determined, in which case the caller
+        treats the whole node as busy (conservative, the old behaviour).
+        """
+        try:
+            snap = self._registry.peek_runtime_snapshot(provider_id)
+        except Exception:
+            snap = None
+        if not isinstance(snap, dict):
+            return frozenset()
+        device_list = ((snap.get("runtime") or {}).get("devices") or {}).get("devices") or []
+        n_gpus = 0
+        for dev in device_list:
+            if not isinstance(dev, dict):
+                continue
+            if dev.get("kind") not in (None, "nvidia"):
+                continue
+            try:
+                index = int((dev.get("extra") or {}).get("index"))
+            except (TypeError, ValueError):
+                continue
+            if index >= 0:
+                n_gpus += 1
+        if n_gpus < 1:
+            return frozenset()
+        slice_size = 1 << (n_gpus.bit_length() - 1)
+        return frozenset(range(slice_size))
+
+    @staticmethod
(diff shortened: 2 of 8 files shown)


wasnertobias (commented):
Good, focused change — the conservative fallbacks are right (no telemetry → any busy lane still blocks), `_plan_needs_gpu_pin` correctly leaves explicit operator values and `"none"` alone, the slice is released in a `finally`, explicit lane pins onto the slice are refused, and auto-placement excludes it. Test coverage for the slice derivation and the non-blocking behaviour is solid.

Three things, one of which I think defeats the stated acceptance criterion.

**Waking a leftover lane is refused, so leftover GPUs only serve what happens to be awake already.** `_VRAM_GROWING_ACTIONS` at `logos_bridge.py:48` includes `wake_lane`, and the guard at `:669` tests only `_calibration_session_is_live()` — there is no slice comparison. So a sleeping lane on GPU 2 cannot be woken while the probe runs on GPUs 0–1.

The comment justifying the refusal is about protecting the probe's VRAM, which is exactly right for `add_lane` and `apply_lanes` — a new lane on the slice takes memory the kv-cache search needs. But a wake on a leftover GPU takes nothing from the probe. And during the nightly window leftover lanes are typically in L1 sleep, so acceptance criterion 1 ("a 3-GPU node calibrating at tp=2 can simultaneously serve requests on the remaining GPU") holds only for lanes that were already awake when the session began. `wake_lane` should be allowed when the lane's GPUs don't intersect the held slice — that is serving, not placement.

**An operator pin and the held slice can point at different GPUs.** `pin_plan_gpu_devices` leaves an explicit `gpu_devices` untouched, while `begin_calibration_session` always holds `{0..tp-1}` derived from the node's GPU count. With `gpu_devices: "2"` on a 3-GPU host, the session tears down lanes on `{0,1}`, the probe runs on GPU 2, and whatever production lane sits there contaminates the very baseline this PR exists to clean up. Either hold the set the probe will actually use, or refuse to start a session whose pin falls outside the default slice.

**The slice is derived in three places from three different sources** — `gpu_collector.device_count` for the hold, `len(query_gpu_vram())` for the pin, and the count of snapshot devices with a parseable `extra.index` on the orchestrator side. When those disagree (empty collector, missing `extra.index`, mixed device kinds), the orchestrator can start a session while a slice GPU is still busy, or fall into the empty-subset path and block leftover traffic indefinitely. Having the worker report the slice it actually holds, rather than the orchestrator re-deriving it, would remove the whole class.

logos/logos-workernode/logos_worker_node/calibration.py
```diff
@@ -2910,6 +2910,53 @@ def _max_tp_for_plan(plan: dict[str, Any], available_gpus: int) -> int:
     return 1 << (n.bit_length() - 1)
 
 
+def calibration_gpu_slice(available_gpus: int) -> list[int]:
```
wasnertobias, line 2913: Two things worth putting in this docstring, since both surprise on first read.

**The slice only leaves anything over for non-powers-of-two.** `1 << (n.bit_length() - 1)` returns the whole node for 2, 4 and 8 GPUs — the most common configurations — so there `destroy_lanes_on_gpus(slice)` is exactly the old `destroy_all()` and nothing is freed. The gain applies to 3, 5, 6 and 7. The docstring's examples (3 → `[0,1]`, 5 → `[0,1,2,3]`, 8 → `[0..7]`) show this if you look closely, but don't say that the last case means "no leftover at all". Expected behaviour, just worth stating.

**"Heterogeneous" here means the GPU count, not the hardware.** The slice is purely positional (`range(slice_size)`), with no regard for what those GPUs are. On a node mixing card sizes — say one 24 GB and two 80 GB — the slice is `[0, 1]`, which may straddle both types; a tp=2 probe then measures across unequal cards and the resulting `base_residency_mb` describes neither configuration cleanly. That matters beyond this PR: #857 sizes the placement threshold from `min(gpu_totals)`, so a calibration measured over mixed cards feeds a planner that already treats heterogeneous VRAM conservatively.

If genuinely mixed hardware is in scope, the slice should prefer a homogeneous set rather than starting at index 0. If it isn't, saying so here would set the expectation.

Issue #685, opened 2026-07-16
`Logos`: Rename table "ollama*"


wasnertobias:
Tables such as ollama_provider_snapshot shall be renamed, as there is no relation to ollama existing any more.

bensofficial:
Resolved by #865 (Drop Ollama support) — this also renamed the table `ollama_provider_snapshots` to `provider_snapshots`.

PR #865, opened 2026-08-29
`Logos`: Drop Ollama support

### LINKED ISSUE #778: `Logos`: Drop Ollama support
Ollama is no longer used anywhere in the deployment: every worker serves through vLLM. Keeping the second engine alive costs a branch in every scheduling and capacity path — and each of those branches is a place where the vLLM path has to be reasoned about twice.

It surfaced while documenting context windows in #775: the routing there has to keep any deployment whose context window is unknown, and "Ollama lanes report a configured `context_length` instead of a served window" is one of the reasons that escape hatch exists. Removing the engine makes the remaining rule simpler, not just shorter.

## Scope

~213 references across the codebase. Roughly:

**Orchestrator**
- `sdi/providers/ollama_provider.py` (527 lines) and `sdi/ollama_facade.py` (261 lines)
- The `provider_type == "ollama"` branches in `main.py`, `pipeline/context_resolver.py`, `pipeline/scheduler_interface.py`, `sdi/models.py`
- `_lane_served_context_window()` can drop its non-vLLM path

**Worker node**
- `ollama_process.py` (490 lines)
- `engines.ollama` in the config schema and `config.example.yml`
- The `vllm: false` lane variant

**Webservice / DB**
- `OllamaProviderSnapshotRepository` and the `ollama_provider_snapshot` table
- The `ollama` provider type in `Provider` / `ProviderService`
- A migration for existing rows: any provider still typed `ollama` has to be
  either converted or refused explicitly rather than silently left behind

**UI**
- Provider-type labels and any Ollama-specific columns

## Before starting

Check prod and dev for providers still typed `ollama`. The migration decision
(convert vs. refuse) depends on whether any exist and whether anything routes to
them, and that is worth knowing before the code changes rather than after.

## Not in scope

The `logos-workernode` container currently ships the Ollama binary. Removing it
from the image is a separate, easy win once the code no longer calls it, and
worth doing separately so the image change can be rolled back on its own.

wasnertobias:
## What

Removes Ollama support from all four Logos surfaces, per #778:

**Orchestrator**
- Deleted `sdi/providers/ollama_provider.py` and `sdi/ollama_facade.py`
- Removed the `provider_type == "ollama"` branches and non-vLLM scheduling paths (`runtime_mode` is now `["vllm"]`; `is_vllm` dropped from the lane scheduler signal)
- Deleted Ollama-only benchmarks, test fixtures, and docs
- Renamed the SDI dataclass `OllamaCapacity` to `WorkerCapacity`

**Workernode**
- Deleted `ollama_process.py` and Ollama-only tooling
- Deleted the Ollama binary build stage from the Dockerfile together with `INSTALL_OLLAMA`; removed `LOGOS_WORKER_NODE_HAS_VLLM` / `LOGOS_WORKER_NODE_HAS_OLLAMA` entirely (nothing in the code read them)
- `engines.ollama` removed from the config schema; legacy keys (`models_path`, `gpu_device`) are migrated to `worker.*` with a warning log, so existing deployments keep working one release
- `vllm: false` lanes are rejected with a clear error, as are the Ollama-only lane fields `keep_alive` / `kv_cache_type`
- `WorkerConfig.models_path` / `WorkerConfig.gpu_device` take over the deployment-facing role `OllamaConfig` had
- Model store renamed to `/usr/share/logos/models` (compose volume `logos-models`, env `LOGOS_MODELS_MOUNT`)
- Model volume source is resolved backward-compatibly in compose (`LOGOS_MODELS_MOUNT` → legacy `OLLAMA_MODELS_MOUNT` → the pre-existing `ollama-models` named volume, kept for one release) so in-place upgrades keep their cached weights; a `config.yml` left on the old container path (`/usr/share/ollama/.ollama/models`) is auto-translated to the new mount at load time with a warning
- Ollama scenario, flags, and tag mapping removed from the lane benchmark tooling (`bench_lane_backends.py` is vLLM-only; `ONLY_OLLAMA` / `--only-ollama` / `OLLAMA_MODEL_MAP` are gone from `run_bench.sh` / `benchmark_logos.py`)

**Webservice**
- Provider add/update rejects `provider_type "ollama"` with HTTP 400
- Liquibase `020`: renames the table `ollama_provider_snapshots` to `provider_snapshots` (this also covers the table rename requested in #685); entity/repository renamed to `ProviderSnapshot` / `ProviderSnapshotRepository`
- Liquibase `021`: fails startup with an explicit error while `providers` rows of the dropped type still exist

**UI**
- Dropped the Ollama provider-type label and the vLLM/Ollama lane badge (lane stats are now rendered for every lane)

## Migration strategy: refuse loudly, never silently

Existing `ollama` provider rows are **not** converted — conversion would silently route traffic to servers that no longer exist. Instead the system refuses at four points:

1. Webservice Liquibase `021` — startup aborts while `ollama` rows exist
2. Webservice provider add/update — HTTP 400
3. Orchestrator startup gate — aborts boot while `ollama` rows exist
4. Orchestrator `DBManager.add_provider` — HTTP 400

Note: in the shared schema `providers.provider_type` is a Postgres ENUM `('logosnode', 'cloud')`, so an `ollama` row **cannot exist in a normally provisioned database** — gates 1 and 3 are defense-in-depth against hand-altered or legacy schemas.

### Data check (2026-08-29)

| | logos (prod) | logos-dev |
|---|---|---|
| `providers.provider_type` | 1 cloud, 4 logosnode | 1 logosnode |
| of those, `ollama` | **0** | **0** |
| `model_profiles.engine` | 96 × vllm | 12 × vllm |
| of those, `ollama` | **0** | **0** |
| `ollama_provider_snapshots` | present | present |
| `provider_snapshots` | not yet | not yet |

Both environments are clean, so gates 1 and 3 will pass and the rename has nothing to trip over.

## Out of scope (kept intentionally)

- The public API path `/logosdb/get_ollama_vram_stats` and the legacy column names `ollamaAdminUrl` / `ollama_admin_url` (now carrying the worker base URL) — rename belongs to #685
- The vendored LiteLLM price catalog (`model_prices_and_context_window.json`) — generated upstream data that includes Ollama model entries
- Dated research and benchmark-analysis artifacts (`research/`, dated analysis docs) — historical Ollama-vs-vLLM measurements that the plot tooling still consumes

## Verification

- Orchestrator: 1594 passed (+1 pre-existing xfail)
- Workernode: 816 passed (+2 skipped), incl. 4 new config-path-translation tests
- Webservice: 427 passed (Testcontainers Postgres 17, incl. 029/030 baseline tests)
- UI: `ng build` OK
- Merged `origin/main` (most recently: main.py helper extraction #850, vLLM 0.29.0, idle-lane activity tracking #901); the snapshot-rename and drop-guard migrations are renumbered 027/028 → 029/030 because main took 027/028. The Ollama-removal delta was re-applied on the modules #850 extracted (logosnode_snapshot.py, routers/*), and main's new tests that still referenced the removed OllamaConfig name / is_vllm keyword were updated
- `pre-commit` (black/isort/autoflake/flake8/check-yaml, …): all hooks pass

Fixes #778

## Summary by CodeRabbit

* **New Features**
  * Worker-backed model lanes now run exclusively on vLLM.
  * Capacity planning supports benchmark lanes without reclaiming production capacity.
  * Lane health displays statistics for all lanes, including running request counts.
  * Model profiles include cold-load and wake-from-sleep timing data.

* **Bug Fixes**
  * Improved lane switching, capacity handling, GPU placement, and calibrated profile fallback behavior.

* **Breaking Changes**
  * Ollama providers, configuration, benchmarks, and deployment options are no longer supported. Existing providers must be retyped as LogosNode providers or removed.
  * Model storage configuration now uses the Logos model storage path and environment variable.


logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
```diff
@@ -1005,7 +1017,7 @@ def insert_provider_snapshot(
         """
         sql = text(
             """
-            INSERT INTO ollama_provider_snapshots (
+            INSERT INTO provider_snapshots (
```
wasnertobias, line 1546: Deployment note rather than a code change: this rename needs the webservice to migrate before the orchestrator restarts.

The table is renamed in Liquibase (which runs inside `logos-webservice`) while this same PR moves the orchestrator to the new name — one `INSERT` here plus three `SELECT`s at `:1304`, `:1445`, `:1473`. In `docker-compose.yaml` both services declare only `depends_on: logos-db`, with nothing sequencing them relative to each other, so there are two windows where snapshot capture fails outright:

- **deploy** — the new orchestrator comes up before the migration: `provider_snapshots` doesn't exist yet
- **rollback** — orchestrator reverted, migration kept: `ollama_provider_snapshots` no longer exists

Snapshots are how worker runtime data reaches the database, so during that window the capacity planner is working blind.

#849 introduces the same coupling for a different reason (a column its `INSERT` depends on), so ordering the two services once in compose — or gating the orchestrator on migrations having completed — would cover both PRs rather than being re-litigated per change.

logos/logos-orchestrator/src/logos/main.py
```diff
@@ -934,10 +916,8 @@ def _merge_local_provider_vram_payload(
         entry["last_heartbeat"] = runtime_snapshot.get("last_heartbeat") if runtime_snapshot else None
 
         runtime = runtime_snapshot.get("runtime") if isinstance(runtime_snapshot, dict) else {}
-        lanes = runtime.get("lanes") if isinstance(runtime, dict) and isinstance(runtime.get("lanes"), list) else []
-        runtime_modes = _runtime_modes_for_lanes(lanes)
-        if runtime_modes:
-            entry["runtime_modes"] = runtime_modes
+        runtime.get("lanes") if isinstance(runtime, dict) and isinstance(runtime.get("lanes"), list) else []
```
wasnertobias: Leftover from removing `_runtime_modes_for_lanes` — this expression no longer has a consumer and can go.

wasnertobias (commented):
Good removal. "Refuse loudly, never silently" is the right call — converting `ollama` rows would have routed traffic at servers that no longer exist — and the four gates are consistent about it.

The part I'd single out as well handled is the cache path. `HF_HOME` on the real workers resolves to `/usr/share/ollama/.ollama/models/.hf_cache`, so `engines.ollama.models_path` was load-bearing far beyond Ollama itself: the tmpfs RAM cache, the FlashInfer warmup and the model store all hang off it. Moving that to `cfg.worker.models_path` with the resolved source logged (`main.py:297-304`), plus the legacy-key migration with a warning, is exactly the care that a −6826 line removal needs at that spot.

## The data check the description asks for — done

The PR notes that "a manual data check of the prod/dev databases is still required before deployment". I ran it against both:

| | logos (prod) | logos-dev |
|---|---|---|
| `providers.provider_type` | 1 cloud, 4 logosnode | 1 logosnode |
| of those, `ollama` | **0** | **0** |
| `model_profiles.engine` | 96 × vllm | 12 × vllm |
| of those, `ollama` | **0** | **0** |
| `ollama_provider_snapshots` | present | present |
| `provider_snapshots` | not yet | not yet |

Both environments are clean, so gates 1 and 3 will pass and the rename has nothing to trip over. Worth folding these numbers into the description so the check isn't repeated from scratch at deploy time.

## One residue worth noting, not fixing

`model_profiles.engine` is a separate column from `providers.provider_type`, and nothing rewrites a stored `'ollama'` there. `_build_load_params` would then omit `vllm: true` and `enable_sleep_mode`, and the VRAM estimate would take the non-vLLM branch — `logosnode_provider` only fills `engine` when it is `None`, so an old row survives as-is. Zero such rows exist in either database above, so this is not blocking. It only matters if some other instance predates the vLLM-only era.

## Cleanups

- `main.py:919` — leftover from deleting `_runtime_modes_for_lanes`: the `runtime.get("lanes") if … else []` expression has no remaining consumer.
- `logos/tests/unit/capacity/test_cold_load_placement.py:86` still builds `LaneSchedulerSignals(..., is_vllm=True)`. CI collects `logos-orchestrator`, not `logos/tests/`, so it won't fail the PR — but the helper raises `TypeError` for anyone running it directly.
- `logos/AGENTS.md` still lists `tests/scheduling_data/`, which this PR deletes.
- No orchestrator-side test for `_assert_no_ollama_typed_providers` or `DBManager.add_provider` rejecting `ollama`; the webservice has its 400 test. Given the enum makes the state unreachable, that's a judgement call rather than a gap.
- `OllamaCapacity` in `sdi/models.py` remains the public type name on the hot path. The README already flags it as a legacy name; fine to leave, just the last engine-named type still in the way.

logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
wasnertobias: Acknowledged as a deployment-ordering note: deploy the webservice first. Its Liquibase 018 rename is idempotent (the tableExists precondition MARK_RANs when the old table is already gone), and only then is it safe to restart the orchestrator, whose VRAM stats queries target provider_snapshots. The orchestrator startup gate only reads the providers table, so a partial deployment degrades the stats endpoint at worst, not startup.

logos/logos-orchestrator/src/logos/main.py
wasnertobias: Removed in abef0d2b9 — the orphaned runtime.get("lanes") expression is gone; entry["runtime_modes"] = ["vllm"] now stands alone.

logos/logos-orchestrator/src/logos/sdi/models.py
```diff
@@ -174,15 +174,14 @@ class LaneSchedulerSignals:
...
-    queue_waiting: float  # from backend_metrics (vLLM) or 0 (Ollama)
-    requests_running: float  # from backend_metrics (vLLM) or active_requests (Ollama)
+    queue_waiting: float  # from backend_metrics, 0.0 when the engine reports nothing
+    requests_running: float  # from backend_metrics, or active_requests when no running count is reported
     gpu_cache_usage_percent: Optional[float]  # vLLM only
     ttft_p95_seconds: float  # computed from ttft_histogram, 0.0 if unavailable
     e2e_latency_p50_seconds: float  # p50 end-to-end request latency, 0.0 if unavailable
     effective_vram_mb: float
-    num_parallel: int  # Ollama: explicit; vLLM: worker-reported engine max concurrency (0 until the worker reports it)
+    num_parallel: int  # worker-reported engine max concurrency (0 until the worker reports it)
```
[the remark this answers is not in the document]
wasnertobias: Fixed — the comment now reads "from backend_metrics, 0.0 when no running count is reported". The old active_requests fallback is gone from the construction path: logosnode_provider passes _lane_metric_float(backend_metrics.get("requests_running")) only.

logos/logos-orchestrator/tests/unit/main/test_statistics_endpoints.py
```diff
@@ -40,11 +40,11 @@ def get_local_provider_inventory(self, logos_key):
         assert logos_key == "test-key"
         return self.inventory, self.status
 
-    def get_ollama_vram_stats(self, logos_key, day, bucket_seconds=5):  # noqa: ARG002
+    def get_provider_vram_stats(self, logos_key, day, bucket_seconds=5):  # noqa: ARG002
         assert logos_key == "test-key"
         return self.stats_payload, self.stats_status
 
-    def get_ollama_vram_deltas(self, logos_key, day, after_snapshot_id=0):  # noqa: ARG002
+    def get_provider_vram_deltas(self, logos_key, day, after_snapshot_id=0):  # noqa: ARG002
```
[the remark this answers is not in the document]
wasnertobias: Fixed — the test double now accepts since=None to match the production signature, and a new test (test_initial_all_day_vram_load_is_bounded_by_recent_since_window) exercises the initial day="all" load and asserts the since-bounded get_provider_vram_deltas call.

logos/logos-orchestrator/tests/unit/sdi/test_scheduler_view.py
```diff
@@ -134,16 +133,14 @@ def test_scheduler_view_loaded_vllm_and_sleeping_ollama(monkeypatch):
...
     assert vllm_lane.requests_running == 2.0
     assert vllm_lane.gpu_cache_usage_percent == 45.0
     assert vllm_lane.gpu_memory_utilization == 0.7
     assert vllm_lane.tensor_parallel_size == 2
 
-    ollama_lane = next(l for l in view.lanes if l.lane_id == "ollama-1")
-    assert ollama_lane.is_vllm is False
-    assert ollama_lane.requests_running == 0.0  # Ollama uses active_requests for requests_running
-    assert ollama_lane.gpu_cache_usage_percent is None
+    sleeping_lane = next(l for l in view.lanes if l.lane_id == "sleeping-1")
```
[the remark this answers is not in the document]
wasnertobias: Done — both generator variables in the lane lookups are renamed from l to lane.

logos/logos-workernode/logos_worker_node/lane_manager.py
```diff
@@ -2720,39 +2675,21 @@ async def _build_lane_status(
...
             model = lc.model
-            if lc.vllm:
-                # Use vLLM-reported max concurrency (KV-budget-derived) when available.
-                vllm_max = getattr(handle, "max_concurrency", None)
-                num_parallel = vllm_max if vllm_max and vllm_max > 0 else 0
-            else:
-                num_parallel = lc.num_parallel
+            # Use vLLM-reported max concurrency (KV-budget-derived) when available.
+            vllm_max = getattr(handle, "max_concurrency", None)
+            num_parallel = vllm_max if vllm_max and vllm_max > 0 else 0
```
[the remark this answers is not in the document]
wasnertobias: Done — num_parallel now falls back to the configured lc.num_parallel while the handle's max_concurrency is still None or non-positive, and uses the vLLM-reported value once it goes positive.

logos/logos-workernode/logos_worker_node/main.py
```diff
@@ -348,13 +348,14 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
...
-    # models_path which is the persistent volume in the standard compose.
+    # Honor LOGOS_WORKER_CACHE_ROOT first so deployments with a different
+    # storage layout can point all worker caches at any persistent path;
+    # default to the worker's persistent model store (worker.models_path),
+    # the persistent volume in the standard compose.
     try:
         from logos_worker_node.flashinfer_warmup import warmup as flashinfer_warmup
 
-        workspace_base = os.environ.get("LOGOS_WORKER_CACHE_ROOT", "").strip() or cfg.engines.ollama.models_path
+        workspace_base = os.environ.get("LOGOS_WORKER_CACHE_ROOT", "").strip() or cfg.worker.models_path
```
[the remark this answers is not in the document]
wasnertobias: Done — the FlashInfer workspace_base now resolves through VllmProcessHandle._resolve_persistent_cache_root (LOGOS_WORKER_CACHE_ROOT -> worker.cache_path -> worker.models_path). The shared resolver itself was extended to consult cache_path directly, so it is correct even for configs built without the env-var lift.

logos/logos-workernode/QUICKSTART.md
```diff
@@ -61,12 +61,12 @@ The server decides when to spin up or tear down lanes based on `capabilities_mod
 
 ## 4. Start the Worker
 
-### CPU-only mode (Ollama)
+### Development mode
 ```bash
 docker compose -f docker-compose.dev.yml up --build
```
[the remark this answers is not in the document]
wasnertobias: Done — the Development mode section now documents the vLLM build: INSTALL_VLLM=1 INSTALL_OLLAMA=0 BASE_IMAGE=ubuntu:24.04. The compose file's own defaults are left untouched; dropping the Ollama binary from the image is the separate follow-up.

wasnertobias:
Thank you for the thorough review — every point is addressed in `abef0d2b9`:

- **Data check**: folded your prod/dev numbers into the description (new "Data check (2026-08-29)" section), so the check is not repeated from scratch at deploy time.
- **`main.py:919`**: the orphaned `runtime.get("lanes")` expression is removed.
- **`test_cold_load_placement.py:86`**: the helper no longer passes `is_vllm=True`, so the file collects and runs cleanly against the current `LaneSchedulerSignals`.
- **`logos/AGENTS.md`**: the `tests/scheduling_data/` entry is gone from the tree.
- **`model_profiles.engine` residue**: left as noted — the stored `engine` column is a separate column from `providers.provider_type` and is intentionally not rewritten; with zero `ollama` rows in either environment this is non-blocking, and the four refusal gates remain the defense for instances that predate the vLLM-only era.
- **`OllamaCapacity`**: kept the public type name — the SDI README already flags it as the legacy name, and renaming it belongs with the other engine-name cleanup.
- **No orchestrator-side refusal tests**: agreed this is a judgement call rather than a gap — the `provider_type` ENUM makes the state unreachable in a normally provisioned database, and the webservice 400 test covers the provider-creation side.

Also on this push: merged `main` (which landed the `016_model_profiles_max_reported_context_length` changelog), and per the coordination note the colliding changelogs were renumbered `015`/`016` → `018`/`019` (files, `master.xml` includes, changeSet ids, and the baseline-test method names). Re-verification after the merge: orchestrator 1045 passed (+1 pre-existing xfail), workernode 577 passed (+2 skipped), webservice 317 passed, pre-commit clean.

wasnertobias:
logos/logos-workernode/Dockerfile has flag INSTALL_VLLM which is only not set, if ollama shall be built, which is no longer the case, so this PR shall include removing this flag alltogether

wasnertobias:
Done — `INSTALL_VLLM` is removed in 989a6b97c, so vLLM is now installed unconditionally. Concretely:

- **Dockerfile**: the `INSTALL_VLLM` ARG is gone; the JIT toolchain install, the vLLM install and the CUDA toolkit setup no longer sit behind `if [ "$INSTALL_VLLM" = "1" ]`, and `LOGOS_WORKER_NODE_HAS_VLLM` is pinned to `1`.
- **docker-compose.dev.yml**: `INSTALL_VLLM` build arg dropped, default `BASE_IMAGE` is now `ubuntu:24.04` (the `PYTHON_IMAGE` indirection for the old python:slim CPU build is gone with it).
- **`.env.example`**: `INSTALL_VLLM` / `PYTHON_IMAGE` entries removed.
- **QUICKSTART.md, CI build workflow, bench preflight hint**: flag references dropped.

`INSTALL_OLLAMA` is intentionally kept — removing the Ollama binary from the image is the separate follow-up. A repo-wide grep for `INSTALL_VLLM` comes back clean, `pre-commit` passes, and the Dockerfile validates under BuildKit.

Claudia-Anthropica (dismissed):
@wasnertobias The vLLM-only transition is consistent across the orchestrator, worker, webservice schema/API, and UI; legacy worker paths are migrated and unsupported provider or lane inputs fail explicitly. The complete exact-head CI snapshot passed all tests, builds, and lint checks, and I found no unresolved blocking defect.

wasnertobias:
Do not keep INSTALL_OLLAMA as well, obviously!

Any reference to ollama shall be removed from this repo, basically.

"LOGOS_WORKER_NODE_HAS_VLLM is pinned to 1" do not pin but remove the env as well. We don't need it.

wasnertobias:
Done — `INSTALL_OLLAMA` is gone, `LOGOS_WORKER_NODE_HAS_VLLM` is removed rather than pinned, and every remaining Ollama reference inside `logos/` was cleaned up or consciously kept (reasons below).

**Removed in this round**
- `INSTALL_OLLAMA` (Dockerfile ARG + the whole Ollama binary build stage, compose build-args, `.env.example`, docs)
- `LOGOS_WORKER_NODE_HAS_VLLM` and `LOGOS_WORKER_NODE_HAS_OLLAMA` — deleted outright; nothing in the code ever read them
- Model-store rename: `/usr/share/ollama/.ollama/models` → `/usr/share/logos/models` (HF cache at `/usr/share/logos/models/.hf_cache`), compose volume `ollama-models` → `logos-models`, env `OLLAMA_MODELS_MOUNT` → `LOGOS_MODELS_MOUNT` — across compose files, `.env.example`, config example, docs, workernode source paths, and test fixtures
- `OllamaCapacity` → `WorkerCapacity` in the orchestrator SDI layer
- Benchmark tooling: `bench_lane_backends.py` is now a vLLM-only lane sweep (`--include-ollama` / `--ollama-*` flags and the speedup-vs-Ollama logic dropped), `run_bench.sh` / `benchmark_logos.py` no longer know `ONLY_OLLAMA` / `--only-ollama` / `OLLAMA_MODEL_MAP`, `bench_deepseek_vllm_mem.sh` no longer calls `/admin/ollama/destroy`, `bench_qwen3.py` is vLLM-only, and the benchmark Readme no longer documents the Ollama scenario
- The `responses.py` docstring and the Bruno lane-apply example no longer reference Ollama

**Kept intentionally, with reasons**
1. **Drop-guards that must name the dropped type**: the webservice rejection of `provider_type "ollama"` (HTTP 400), the Liquibase `019` data check, the orchestrator startup gate + `add_provider` rejection, and the workernode rejection of `vllm: false` lanes and legacy Ollama lane fields (plus their tests). These implement the "refuse loudly, never silently" migration strategy this PR builds — removing the word they guard against would defeat them.
2. **Applied Liquibase history** (`000`, `004`, `018`, `019`, `master.xml`) — rewriting already-applied migrations would break production checksums.
3. **Legacy production schema/API names**: the `ollama_admin_url` column (now carrying the worker base URL) and the public endpoint `/logosdb/get_ollama_vram_stats` plus all their consumers (UI statistics service, VramController, benchmark tooling). Renaming both is a schema migration plus a breaking API change — that belongs to #685, not this PR.
4. **Vendored generated data**: the LiteLLM price catalog (`model_prices_and_context_window.json`) contains Ollama model entries because it is generated upstream.
5. **Dated research/analysis artifacts**: `research/` (bench_results data, RESEARCH_SUMMARY, CUDA_BACKEND_SELECTION) and the dated benchmark-analysis docs record historical Ollama-vs-vLLM measurements; the plot tooling consumes those historical rows and would break if stripped.
6. **Other monorepo projects** (`iris/`, `memiris/`, `athena/`, …) use Ollama as a supported LLM backend — out of scope for this Logos PR, untouched.

Verification: case-insensitive greps for `INSTALL_OLLAMA`, `LOGOS_WORKER_NODE_HAS_VLLM`, `LOGOS_WORKER_NODE_HAS_OLLAMA`, `OLLAMA_MODELS_MOUNT`, `OllamaCapacity`, `/usr/share/ollama`, and `ollama-models` return zero hits outside the kept categories above. Workernode suite: 577 passed (+2 skipped); orchestrator unit suite: 1088 passed (+1 pre-existing xfail); `pre-commit` all green. Full details in the PR body.

logos/logos-workernode/docker-compose.yml
```diff
@@ -77,7 +77,7 @@ services:
     volumes:
       - ./config.yml:/app/config.yml:ro
       - ./data:/app/data
-      - ${OLLAMA_MODELS_MOUNT:-ollama-models}:/usr/share/ollama/.ollama/models
+      - ${LOGOS_MODELS_MOUNT:-logos-models}:/usr/share/logos/models
```
Claudia-Anthropica: @wasnertobias [high] This changes the volume variable, default volume name, and container path simultaneously. Existing deployments still have `OLLAMA_MODELS_MOUNT`/`ollama-models`, while the config compatibility validator copies the legacy `/usr/share/ollama/.ollama/models` value unchanged; after this update Compose ignores that variable, creates an empty `logos-models` volume, and mounts only `/usr/share/logos/models`. The worker therefore loses access to all cached model weights and may remain unavailable while every model is downloaded again—or indefinitely when the cache cannot be repopulated.

In `logos/logos-workernode/docker-compose.yml`, changing the volume source and container path simultaneously disconnects existing model storage from migrated legacy configurations. For one release, resolve the mount source in the order `LOGOS_MODELS_MOUNT`, `OLLAMA_MODELS_MOUNT`, then the existing `ollama-models` volume, translate the legacy default `models_path` to `/usr/share/logos/models`, and add an upgrade test using the old config and environment names.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The vLLM-only cleanup is largely consistent, and the captured CI has no failed or pending checks. However, the storage rename breaks in-place worker upgrades by disconnecting the existing model volume from both Compose and the migrated legacy configuration. This needs a backward-compatible transition before merging.

wasnertobias:
One follow-up from making the branch merge-ready: `origin/main` picked up the model-alias feature (#852) with a `018_model_aliases.xml` migration while this PR was open, which collided with this PR's `018_rename_provider_snapshots.xml` (and `LiquibaseChangelogOrderTest` enforces strictly increasing migration numbers, so the duplicate 018 could not stand). Since neither of this PR's migrations has been applied anywhere yet, I merged `origin/main` and renumbered them to `020_rename_provider_snapshots.xml` / `021_refuse_ollama_typed_providers.xml` — the guard is the same drop-guard data check as described above, just as migration 021 instead of 019. The PR body is updated accordingly.

Claudia-Anthropica (changes requested):
@wasnertobias All 29 captured checks succeeded, but the existing model-storage upgrade blocker remains on this exact head. Compose still ignores `OLLAMA_MODELS_MOUNT` and mounts a new volume/path while the compatibility validator preserves the legacy path unchanged, disconnecting upgraded workers from their cached weights.

logos/logos-workernode/docker-compose.yml
Claudia-Anthropica: @wasnertobias This remains unresolved on the current head: `docker-compose.yml` still mounts only `${LOGOS_MODELS_MOUNT:-logos-models}` at `/usr/share/logos/models`, while the config migrator preserves the legacy `/usr/share/ollama/.ollama/models` value. Existing `OLLAMA_MODELS_MOUNT`/`ollama-models` deployments are therefore still disconnected from their cached weights.

wasnertobias: Resolved in d11f21e06 — this is the one-release backward-compat path the comment describes.

Compose now resolves the model-volume source in the order `LOGOS_MODELS_MOUNT` → legacy `OLLAMA_MODELS_MOUNT` → the pre-existing `ollama-models` named volume, all mounted at `/usr/share/logos/models` (both the worker and the calibrate service):

    - ${LOGOS_MODELS_MOUNT:-${OLLAMA_MODELS_MOUNT:-ollama-models}}:/usr/share/logos/models

Both `logos-models` and `ollama-models` are declared as named volumes. So an existing deployment that still sets `OLLAMA_MODELS_MOUNT` — or has nothing set, which falls back to the pre-existing `ollama-models` volume — keeps pointing at the volume that already holds its cached weights: no empty `logos-models` volume is created for it and nothing is re-downloaded. Fresh deployments set `LOGOS_MODELS_MOUNT` (as `.env.example` now documents).

On the config side, the `WorkerConfig.models_path` field validator translates the Ollama-era container path at load time: a config left at the legacy default `/usr/share/ollama/.ollama/models` (or any subpath under it) is rewritten to `/usr/share/logos/models`, with a warning telling the operator to update the file. So an Ansible-managed config not yet bumped also lands on the correctly-mounted path.

Coverage: new tests in `tests/test_lane_models.py` cover the exact path, a subpath, a non-legacy passthrough, and the full `engines.ollama` migrator + validator chain. (The compose env-name fallback is a plain shell default; I verified it resolves to `ollama-models` via `docker compose config`. I can add a committed regression test for that if you want one.)

wasnertobias: This comment was reviewed against a pre-d11f21e06 head — on the current head `docker-compose.yml` no longer mounts `${LOGOS_MODELS_MOUNT:-logos-models}`. Both the worker and the calibrate service now mount:

    - ${LOGOS_MODELS_MOUNT:-${OLLAMA_MODELS_MOUNT:-ollama-models}}:/usr/share/logos/models

with both `logos-models` and `ollama-models` declared as named volumes, so `OLLAMA_MODELS_MOUNT`/`ollama-models` deployments resolve to the volume that already holds their cached weights. The `WorkerConfig.models_path` validator additionally translates the legacy `/usr/share/ollama/.ollama/models` config default (and its subpaths) to `/usr/share/logos/models` at load time, with a warning. Both sides are covered by the exact / subpath / passthrough / migrator-chain tests in `tests/test_lane_models.py`.

Claudia-Anthropica (dismissed):
@wasnertobias The model-storage upgrade blocker is resolved: Compose preserves the legacy environment/volume source while mounting it at the new path, and legacy configured paths are translated with focused migration tests. All 29 captured exact-head checks succeeded, and I found no remaining blocking defect.

Claudia-Anthropica (changes requested):
@wasnertobias [medium] The existing deployment-ordering thread remains blocking on this head. `_capture_logosnode_provider_snapshot()` performs the new-table insert synchronously inside the worker WebSocket loop, so starting the orchestrator before Liquibase 020 raises through the handler and detaches every worker; rolling back the orchestrator after migration has the inverse incompatibility. Preserve the old table name through a one-release compatibility view or otherwise make both rollout directions safe; the storage-volume compatibility and 020/021 ID corrections themselves are sound.

LogosOSSAgent:
# Reply to review — PR #865 (deployment ordering)

**Point (medium):** `_capture_logosnode_provider_snapshot()` performs the
new-table insert synchronously inside the worker WebSocket loop, so (a) starting
the orchestrator before Liquibase 020 raises through the handler and detaches
every worker, and (b) rolling back the orchestrator after migration has the
inverse incompatibility. *"Preserve the old table name through a one-release
compatibility view or otherwise make both rollout directions safe."*

Checked against the current code first: **both directions were real.** I made
both safe — the non-fatal capture fixes (a), a one-release compatibility view in
migration 020 fixes (b).

## (a) Deploy direction — new orchestrator + old schema: fixed

`_capture_logosnode_provider_snapshot()` (src/logos/main.py:848) called
`db.insert_provider_snapshot(...)` with no guard. The `status` handler
(src/logos/main.py:5658) invokes it synchronously inside the worker WS `while`
loop, whose only `except` is `WebSocketDisconnect` and whose `finally` calls
`detach_session`. So a missing `provider_snapshots` table (020 not yet run) —
or any transient DB error — propagated out and dropped the worker.

The DB block is now wrapped in `try/except`. On failure it logs a warning and
`return`s (skipping that one sample), so the exception no longer reaches the
handler and the worker stays connected until the next status message retries.
This mirrors the already-guarded sibling `calibration_probe_log` branch
(src/logos/main.py:5666) and the `upsert_model_profiles` guard one level down.
Non-fatal capture is consistent with the PR's intent: the refusal strategy is
about not routing traffic to a provider type nothing serves, not about dropping
a VRAM telemetry row.

## (b) Rollback direction — rolled-back orchestrator + new schema: fixed

A rolled-back orchestrator (the previous release) still issues
`INSERT ... INTO ollama_provider_snapshots`; once 020 renames the table that
fails. That old orchestrator's code is not in this branch, so the (a) fix cannot
reach it — the schema itself must keep the old name writable.

Migration 020 now also creates a **one-release compatibility view**:
`CREATE VIEW ollama_provider_snapshots AS SELECT * FROM provider_snapshots`.
It is a simple single-table view, so it accepts the previous release's INSERT
(the omitted `id` falls back to the `provider_snapshots` serial default) and its
SELECTs, and the data lands in the renamed table. The comment marks it to be
dropped next release.

Two supporting changes to 020:
- The precondition tightened from `tableExists("ollama_provider_snapshots")` to
  a `sqlCheck` on `table_type = 'BASE TABLE'`. Without this, once the name is a
  view a re-evaluated run would try to `RENAME` a view and fail; with it, the
  run is `MARK_RAN`. This matches the existing `sqlCheck` idiom in 000/001.
- The `<rollback>` now `DROP VIEW`s before renaming the table back, so the two
  statements don't collide on the shared name.

I did not need a round-trip test to prove the view is INSERT-able: a plain
`SELECT *` over a single base table is unconditionally INSERT-able in Postgres,
and every `ollama_provider_snapshots` column was only ever *added* across
migrations (023/024) — never renamed or dropped — so the `SELECT *` view covers
every column name any prior-release statement could reference.

## Regression coverage

Orchestrator — new `tests/unit/main/test_provider_snapshot_capture.py`:
- a snapshot-insert failure does not raise and logs the warning (**fails on the
  old code, passes now**);
- a failure is not half-recorded in memory (no `record_runtime_sample` task is
  scheduled);
- a successful insert still persists and schedules the in-memory recording
  (guards the happy path against the refactor).

Webservice — `LiquibaseBaselineTest`:
- `migration020_providerSnapshotsTableRenamed` now asserts
  `provider_snapshots` is a `BASE TABLE` and `ollama_provider_snapshots` is a
  `VIEW` (the old `tableExists(...).isFalse()` assertion is now false by design,
  so it was replaced, not just appended);
- new `migration020_compatibilityViewMirrorsTheRenamedTable` asserts the view
  exposes the columns the previous release wrote (`id`, `provider_id`,
  `snapshot_ts`, `loaded_models`, `runtime_payload`, `scheduler_signals`).

## What I could not run, and what I did instead

This sandbox has no network (PyPI is blocked), no Maven, no Docker (so the
Testcontainers Postgres suite cannot run), and no ruff/black. So I **could not**
execute the pytest suite, the Maven/Spring (Liquibase) suite, or the Python/Java
linters. Instead I verified statically:
- `python3 -m py_compile` on `main.py` and the new test — both compile; the
  longest line in the touched regions is 119 chars, under the ruff
  `line-length` of 120 (the `runtime_payload=(...)` argument is split to stay
  under it after re-indenting).
- `xml.etree` parse of the modified migration 020 — well-formed.
- A brace/paren balance check on the edited Java test.
- Read the fixed function and the whole `status` WS handler: the snapshot
  insert was the only unguarded DB write in that path (the sibling
  `calibration_probe_log` is already guarded; the other registry calls are
  in-memory). The `except` path `return`s before `snapshot_id` is read (so no
  `UnboundLocalError`) and before `record_runtime_sample` is scheduled, and
  `DBManager.__exit__` still rolls back and closes the session on the failure
  path.
- Confirmed the current orchestrator references only `provider_snapshots` (no
  lingering `ollama_provider_snapshots`), and that the JPA entity maps to
  `@Table(name = "provider_snapshots")`, so Hibernate `ddl-auto=validate` is not
  affected by the extra view (a view is not an entity).

Please run in CI to confirm: the pytest suite (at minimum the new
`test_provider_snapshot_capture.py`) and the Maven build (which runs the
Liquibase Testcontainers tests).

## No change needed

Storage-volume compatibility and the 020/021 changeset-ID corrections from the
prior round were endorsed as sound; left as-is.

Claudia-Anthropica (dismissed):
@wasnertobias The deployment-ordering blocker is resolved on this exact head: snapshot writes now fail non-fatally before migration 020, while migration 020 preserves the old table name as a writable single-table compatibility view for rollback. Focused tests cover the non-fatal capture path and compatibility-view shape, the snapshot has no failed or pending checks, and I found no remaining blocking defect.

logos/logos-orchestrator/src/logos/sdi/models.py
```diff
@@ -174,15 +174,14 @@ class LaneSchedulerSignals:
     model_name: str
     runtime_state: str  # cold|starting|loaded|running|sleeping|stopped|error
     sleep_state: str  # unsupported|unknown|awake|sleeping
-    is_vllm: bool
     active_requests: int
-    queue_waiting: float  # from backend_metrics (vLLM) or 0 (Ollama)
-    requests_running: float  # from backend_metrics (vLLM) or active_requests (Ollama)
+    queue_waiting: float  # from backend_metrics, 0.0 when the engine reports nothing
```
Claudia-Anthropica, line 178: @wasnertobias [medium] Removing `is_vllm` from `LaneSchedulerSignals` left two unit-test constructors passing the deleted keyword: `tests/unit/pipeline/test_ettft_estimator.py:425` and `tests/unit/capacity/test_kv_per_rank_sharding.py:188`. Both now fail during setup with an unexpected-keyword `TypeError`; remove those arguments so the full unit suite can run.

In `logos/logos-orchestrator/src/logos/sdi/models.py`, the `is_vllm` field was removed but two unit-test constructors still pass that keyword, causing their tests to fail before assertions. Remove `is_vllm=True` from both remaining `LaneSchedulerSignals(...)` test constructors.

</details>

logos/benchmarks/benchmark_logos.py
```diff
@@ -3049,8 +3012,8 @@ def _apply_benchmark_workernode_config_via_ssh(
                 new_lines.append("LOGOS_TMPFS_CACHE_PATH=")
             elif stripped.startswith("TMPFS_SIZE="):
                 new_lines.append("TMPFS_SIZE=0")
-            elif local_cache_path and stripped.startswith("OLLAMA_MODELS_MOUNT="):
-                new_lines.append(f"OLLAMA_MODELS_MOUNT={local_cache_path}")
+            elif local_cache_path and stripped.startswith("LOGOS_MODELS_MOUNT="):
```
Claudia-Anthropica: @wasnertobias [medium] The one-release Compose fallback supports upgraded `.env` files containing only `OLLAMA_MODELS_MOUNT`, but this helper now updates only an existing `LOGOS_MODELS_MOUNT` line and never appends one. On those upgrades `--benchmark-local-cache` silently leaves the old mount active while claiming it changed it; the reset helper at line 5083 also reads only the new name and therefore skips wiping legacy-mounted weights, invalidating a requested from-scratch benchmark.

In `logos/benchmarks/benchmark_logos.py`, the benchmark configuration and reset helpers ignore legacy-only `OLLAMA_MODELS_MOUNT` files even though Compose preserves that upgrade path. Resolve model storage using `LOGOS_MODELS_MOUNT` first and `OLLAMA_MODELS_MOUNT` second, append or replace the new variable when applying an override, and add coverage for a legacy-only `.env`.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias Two current-head regressions remain despite the Ollama-removal and deployment-compatibility paths otherwise being consistent. The orchestrator unit suite still constructs `LaneSchedulerSignals` with the removed `is_vllm` keyword in two tests, and the benchmark tooling no longer honors or resets legacy-only `OLLAMA_MODELS_MOUNT` environments that Compose explicitly supports during the one-release transition.

logos/logos-orchestrator/src/logos/sdi/models.py
wasnertobias: Fixed on 35a76c252 — dropped the `is_vllm` kwarg from both remaining `LaneSchedulerSignals` test constructors (`test_ettft_estimator.py`, `test_kv_per_rank_sharding.py`). Orchestrator unit subset green (59 passed).

logos/logos-workernode/tools/bench_qwen3.py
```diff
@@ -161,27 +155,13 @@ async def warmup(url: str, model: str):
...
-    if len(sys.argv) < 2 or sys.argv[1] not in ("ollama", "vllm"):
-        print("Usage: python bench_qwen3.py [ollama|vllm]")
-        print("Env overrides: OLLAMA_URL, VLLM_URL")
-        sys.exit(1)
-
-    backend = sys.argv[1]
+    url = os.environ.get("VLLM_URL", "http://localhost:8000/v1/chat/completions")
+    model = "/tmp/qwen3-30b-a3b.gguf"
+    print(f"\n{'='*60}")
+    print(f"BENCHMARK: qwen3:30b-a3b on vLLM (GGUF)")
```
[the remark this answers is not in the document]
wasnertobias: Fixed on 35a76c252 — dropped the last redundant f-string prefix (the "Warming up..." print, F541).

logos/logos-orchestrator/src/logos/main.py
wasnertobias: Already gone on this head — the bare `runtime.get("lanes") if ...` expression was removed; only the two `lanes = ...` assignments (main.py:502, :733) remain, and both are consumed. Resolving.

logos/logos-workernode/docker-compose.yml
wasnertobias: Confirmed on the current head (35a76c252): worker and calibrate both mount `${LOGOS_MODELS_MOUNT:-${OLLAMA_MODELS_MOUNT:-ollama-models}}` with both named volumes declared, and `WorkerConfig` rewrites the legacy `/usr/share/ollama/.ollama/models` default (and subpaths) to `/usr/share/logos/models` at load. Covered by the legacy migrator/validator tests in `tests/test_lane_models.py`. Resolving.

logos/benchmarks/benchmark_logos.py
wasnertobias: Fixed on 35a76c252 — the local-cache override now sets `LOGOS_MODELS_MOUNT` even for a legacy-only `.env` (append when absent, replace when present), and the from-scratch weight wipe reads `LOGOS_MODELS_MOUNT` with an `OLLAMA_MODELS_MOUNT` fallback, mirroring the compose one-release fallback. Added tests for the legacy-only .env, the in-place replace, and the wipe fallback.

logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
wasnertobias: The deployment-ordering conclusion is documented in-thread (deploy the webservice first — its Liquibase rename is idempotent; the orchestrator startup gate only reads `providers`, so a partial deploy degrades stats at worst, not startup). No code change on this branch. Resolving — reopen if you would rather keep it as a deployment checklist item.

LogosOSSAgent: Agreed — no need to reopen. I checked both halves of the conclusion against the current branch and they hold:

- **Startup gate reads only `providers`.** `_assert_no_ollama_typed_providers()` (`main.py:1802`, invoked from `lifespan` at `main.py:1868`) calls `find_ollama_typed_providers()`, which is a single `SELECT id, name, provider_type FROM providers` (`dbmanager.py:2810-2817`). The only other `provider_snapshots` references in the orchestrator are this insert (`dbmanager.py:1545`) and the three stats/deltas selects, and all of their call sites are runtime paths (the WS status handler and the VRAM-payload endpoint path), never startup.
- **The rename is safe to run first.** The 027 changeSet guards the `RENAME` with a `MARK_RAN` precondition on the old name being a base table, so a re-evaluated run marks itself applied instead of failing on the view; and the compat view keeps a previous-release orchestrator's `INSERT ... INTO ollama_provider_snapshots` working through the renamed table. `LiquibaseBaselineTest` pins exactly that end state (base table + view exposing the old release's columns).

So in the reverse order (orchestrator ahead of the migration) the insert fails on the missing relation and is caught by the non-fatal handler in `_capture_logosnode_provider_snapshot` (`main.py:862`, covered by `tests/unit/main/test_provider_snapshot_capture.py`), and the stats readers fall back to an empty payload — telemetry degrades, nothing at startup touches the table.

One nuance for the record, not a blocker: the gate does still abort startup, but only when stale `ollama`-typed rows exist — and 028 refuses to run for the same rows on the webservice side. That's the intentional data gate, so the only hard precondition of this deployment is "fix the data first", in either order. The ordering itself needs no checklist entry.

No code change.

Claudia-Anthropica (dismissed):
@wasnertobias The two regressions from the previous review are resolved: the remaining `is_vllm` test arguments are gone, and the benchmark helpers now handle `LOGOS_MODELS_MOUNT` with the legacy fallback and focused coverage. The broader exact-head diff preserves model-volume and snapshot-table rollout compatibility, and I found no unresolved blocking defect; the snapshot contains no failed checks, although several non-required Logos jobs were still unfinished.

logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
LogosOSSAgent: Both failing checks are fixed — the fixes are in the latest commit on this branch (`5dd7931`). I'm working from an offline checkout, so I can't watch the CI re-run here; instead I re-ran both jobs' commands locally, and I re-ran the full test surface too. Details below.

**pre-commit (failure) → fixed**

`logos/benchmarks/tests/test_run_all_helpers.py` had been committed unformatted, so the pinned `black` hook (25.1.0) failed on it. I applied the hook's formatting (4 lines rewrapped, no logic change). Re-running the job's exact command from the repo root:

```
pre-commit run --config logos/.pre-commit-config.yaml --all-files --show-diff-on-failure --color always
```

all 13 hooks now pass.

**Test (failure) → fixed**

`logos/logos-orchestrator/tests/unit/main/test_request_logging.py`, `test_http_ndjson_response_preserves_content_type_and_does_not_append_sse_on_failure`: the test's `context` `SimpleNamespace` was missing `anthropic_dialect`, but `main._streaming_response` reads `context.anthropic_dialect` on every streaming path, so the test hit `AttributeError`. Added `anthropic_dialect=None` to the fixture namespace. The file's remaining delta vs. the merge base is only the intended `http://ollama:11434/api/chat` → `http://worker:8000/api/chat` URL change.

**Verification done in this environment (offline)**

- Orchestrator unit tests (the job's `testpaths = tests/unit` scope, asyncio auto mode): **1420 passed**, including every PR-touched test file (request logging, executor, sdi providers, capacity, main, dbutils, benchmarks helpers). The only failures are 12 tests in 5 pre-existing files that need real third-party SDKs which don't exist in this offline image — real FastAPI `TestClient` (`test_audio_api.py` ×4, `test_error_handlers_integration.py` and `test_rest_conformance.py` at module import), the openai SDK (`test_request_content.py` ×6, two tests in `test_executor_streaming.py`), and a real starlette `Request(scope, receive)`. In `test_executor_streaming.py` the PR only changed worker URLs in two other tests in that file, and those pass. None of these failures is caused by this branch's changes.
- Workernode suite (the job's `testpaths = tests`, all 28 files): **778 passed, 0 failed, 1 skipped**. The skip is `TestAgainstRealVllmMetal`, which the test file itself marks `skipif` when the vllm-metal venv isn't on the host — same as on a CI runner.
- Benchmarks tests: **20/20 passed**, including the reformatted `test_run_all_helpers.py`.
- pre-commit: all 13 hooks pass (command above).

**What I could not verify here, stated plainly**

- The webservice Maven job: no network means no Maven dependency resolution, so the Java changes (Ollama-snapshot → provider-snapshot rename, the `ProviderService` rejection of the removed provider type, the new `LiquibaseBaselineTest`, the added `ProviderControllerTest` case) are not compiled or executed here. Sanity checks I did perform: both new changelog XMLs (027/028) are well-formed and included in `master.xml`, and the remaining "ollama" strings in the touched webservice files are the intentional rejection logic and its tests.
- The UI npm job: the `lane-health-panel` changes (html/scss/ts/spec and `statistics.models.ts`) are not test-run here for the same reason.
- The five orchestrator SDK-dependent files listed above will run in CI where the real dependencies are installed; I could only confirm the PR touches none of their code paths (one exception above, which passes).

logos/benchmarks/benchmark_logos.py
LogosOSSAgent: Verified against the current head — the fix addresses both points of the review. No further changes were needed; here is what I checked and how.

**Append/replace.** The override helper now tracks whether a `LOGOS_MODELS_MOUNT=` line exists (`benchmark_logos.py:3016`): with a local cache path given, an existing line is replaced in place, and when no such line exists — e.g. a legacy-only `.env` carrying only `OLLAMA_MODELS_MOUNT` — one is appended (`:3028-3032`). The legacy line stays in the file but is shadowed, because Compose resolves `${LOGOS_MODELS_MOUNT:-${OLLAMA_MODELS_MOUNT:-ollama-models}}` (`docker-compose.yml:85,163`), so the appended line is the mount the benchmark actually gets.

**Wipe fallback.** `_wipe_calibration_and_weights_via_ssh`, when no explicit cache path is passed, now reads `.env` preferring `LOGOS_MODELS_MOUNT` and falling back to `OLLAMA_MODELS_MOUNT` (`:5096-5107`) — the same one-release fallback the compose file uses.

**Tests.** All three claimed tests exist in `tests/test_run_all_helpers.py` and exercise the real helpers with faked subprocesses:
- `test_apply_config_appends_logos_models_mount_for_legacy_only_env`
- `test_apply_config_replaces_existing_logos_models_mount_without_duplicating`
- `test_wipe_weights_falls_back_to_ollama_models_mount`

To confirm they are genuine regression tests, I ran them against the file state immediately before the fix (the direct parent of the fix commit — it only handled a pre-existing `LOGOS_MODELS_MOUNT` line at the old line 3015, and the wipe grepped only the new name at the old line 5083, exactly as the review described): the legacy-only append test and the wipe-fallback test fail on that state, while the in-place replace test passes, since that behavior predates the fix and the test pins it. Against the current head, all 11 tests in the file pass.

**Beyond the unit tests.** The wipe test asserts the emitted script contains both greps but not their order, so I also executed the actual emitted one-liner locally (captured from the helper with a stubbed `subprocess.run`, run with `sh` against a sandbox workernode dir) across five `.env` shapes:
- legacy-only → only the legacy mount is wiped (fallback works)
- both variables → only the `LOGOS_MODELS_MOUNT` path is wiped (precedence correct)
- neither variable → no mount wiped, exit 0
- `LOGOS_MODELS_MOUNT=/` → the `wp != "/"` guard prevents wiping the root
- quoted value → quotes stripped, correct path wiped

All exited 0, and the calibration state (`model_profiles.yml`, `calibration_logs/`) was wiped in every case.

One non-blocking note: if you want the precedence pinned in the unit test itself (not just the emitted-script text), asserting that the `LOGOS_MODELS_MOUNT` grep appears before the `OLLAMA_MODELS_MOUNT` grep in the script would close that small gap — happy to take it, or leave it as is.

LogosOSSAgent: Verified against the current head — the fix addresses both points of the review. No further changes were needed; here is what I checked and how.

**Append/replace.** The override helper now tracks whether a `LOGOS_MODELS_MOUNT=` line exists (`benchmark_logos.py:3016`): with a local cache path given, an existing line is replaced in place, and when no such line exists — e.g. a legacy-only `.env` carrying only `OLLAMA_MODELS_MOUNT` — one is appended (`:3028-3032`). The legacy line stays in the file but is shadowed, because Compose resolves `${LOGOS_MODELS_MOUNT:-${OLLAMA_MODELS_MOUNT:-ollama-models}}` (`docker-compose.yml:85,163`), so the appended line is the mount the benchmark actually gets.

**Wipe fallback.** `_wipe_calibration_and_weights_via_ssh`, when no explicit cache path is passed, now reads `.env` preferring `LOGOS_MODELS_MOUNT` and falling back to `OLLAMA_MODELS_MOUNT` (`:5096-5107`) — the same one-release fallback the compose file uses.

**Tests.** All three claimed tests exist in `tests/test_run_all_helpers.py` and exercise the real helpers with faked subprocesses:
- `test_apply_config_appends_logos_models_mount_for_legacy_only_env`
- `test_apply_config_replaces_existing_logos_models_mount_without_duplicating`
- `test_wipe_weights_falls_back_to_ollama_models_mount`

To confirm they are genuine regression tests, I ran them against the file state immediately before the fix (the direct parent of the fix commit — it only handled a pre-existing `LOGOS_MODELS_MOUNT` line at the old line 3015, and the wipe grepped only the new name at the old line 5083, exactly as the review described): the legacy-only append test and the wipe-fallback test fail on that state, while the in-place replace test passes, since that behavior predates the fix and the test pins it. Against the current head, all 11 tests in the file pass.

**Beyond the unit tests.** The wipe test asserts the emitted script contains both greps but not their order, so I also executed the actual emitted one-liner locally (captured from the helper with a stubbed `subprocess.run`, run with `sh` against a sandbox workernode dir) across five `.env` shapes:
- legacy-only → only the legacy mount is wiped (fallback works)
- both variables → only the `LOGOS_MODELS_MOUNT` path is wiped (precedence correct)
- neither variable → no mount wiped, exit 0
- `LOGOS_MODELS_MOUNT=/` → the `wp != "/"` guard prevents wiping the root
- quoted value → quotes stripped, correct path wiped

All exited 0, and the calibration state (`model_profiles.yml`, `calibration_logs/`) was wiped in every case.

One non-blocking note: if you want the precedence pinned in the unit test itself (not just the emitted-script text), asserting that the `LOGOS_MODELS_MOUNT` grep appears before the `OLLAMA_MODELS_MOUNT` grep in the script would close that small gap — happy to take it, or leave it as is.

logos/logos-workernode/docker-compose.yml
LogosOSSAgent: Confirmed against the current head (`5dd7931`); since the commit you verified, only two unrelated test files changed, so everything below is as you saw it:

- `docker-compose.yml`: `logos-workernode` (line 85) and `logos-workernode-calibrate` (line 163) both mount `${LOGOS_MODELS_MOUNT:-${OLLAMA_MODELS_MOUNT:-ollama-models}}:/usr/share/logos/models`, and both named volumes are declared — `logos-models` and `ollama-models` (lines 176-177). `logos-models` is the fresh-deployment default via `.env.example`, `ollama-models` covers the pre-existing-volume fallback.
- `WorkerConfig.models_path` defaults to `/usr/share/logos/models`, and `_translate_legacy_ollama_models_path` (`models.py:495`) rewrites the exact legacy container path and subpaths under it — with a warning telling operators to update their config — while passing any other path through untouched.
- All 15 tests in `tests/test_lane_models.py` pass at head, including the migrator/validator set: legacy default translation, subpath translation, non-legacy passthrough, `engines.ollama` → `worker.*` migration, explicit-`worker` preference over the legacy section, and the full migrate-then-translate chain.

Agreed — resolved.

logos/logos-orchestrator/src/logos/main.py
LogosOSSAgent: Confirmed on the current head (`5dd7931`):

- The conditional form `runtime.get("lanes") if isinstance(...) else []` occurs in exactly two places, both as `lanes = ...` assignments: line 502 in `_build_logosnode_scheduler_signals` (consumed by the `lane_count` fallback and the lane loop that follows) and line 733 in `_build_live_local_provider_sample` (consumed by `_normalize_loaded_models(lanes)` at line 758, with no reassignment in between). No bare/dropped conditional of that form remains.
- This branch is what removed the third occurrence of that form — the merge base carried one more at old line 1030, whose `lanes` became unused once `_runtime_modes_for_lanes` was dropped, so the assignment went with it.
- The other `runtime.get("lanes")` sites (lines 2132, 2961, 3175, 6495) predate the branch and are each consumed inline or behind an `isinstance` guard — outside the pattern the review flagged.

Agreed — resolved.

logos/logos-workernode/tools/bench_qwen3.py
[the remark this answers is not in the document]
LogosOSSAgent: Confirmed on the current head (`5dd7931`):

- The print is now `print("\nWarming up...")` at `bench_qwen3.py:169` — no f-prefix. The diff of the fix commit shows exactly that one change (`print(f"\nWarming up...")` → `print("\nWarming up...")`), and no later commit touched the file, so head is the fixed state.
- "Last" checks out: an AST scan of the whole file finds 21 f-strings and none without a placeholder, so no other F541 remains.
- For context: the flake8 hook deliberately excludes `logos-workernode/tools/` (benchmark scripts use script-style imports), so this fix was for review cleanliness rather than something the hook would have caught — the hooks that do apply to the file (ast check, autoflake, isort, black, whitespace, end-of-file) all pass on it.

Agreed — resolved.

logos/logos-orchestrator/src/logos/sdi/models.py
LogosOSSAgent: Confirmed on the current head (`5dd7931`):

- `35a76c2` removes exactly the two `is_vllm=True` lines the review cites — `tests/unit/pipeline/test_ettft_estimator.py:425` (in `_make_lane`) and `tests/unit/capacity/test_kv_per_rank_sharding.py:188` — one line per file, nothing else touched.
- No `is_vllm` remains anywhere in `logos-orchestrator/src` or `logos-orchestrator/tests` (grep over both trees is empty), and `LaneSchedulerSignals` (`src/logos/sdi/models.py:168`) has no such field, so the setup-time unexpected-keyword `TypeError` can no longer occur. Every keyword the two constructors now pass maps to an existing dataclass field.
- The two files hold 49 + 10 = 59 tests, matching the "59 passed" subset claim. (I could not re-execute the suite in this sandbox — no pytest/pydantic/fastapi/grpc available offline — so that number is reconciled by test count, not a re-run.)

Agreed — resolved.

PR #772, opened 2026-08-24
`Logos`: Add an agent runner that develops Logos on its own spare capacity

wasnertobias:
## Motivation and Context

Logos knows at every moment how much of its local serving fleet is in use. Outside working hours that number is usually small and the GPUs idle anyway. This PR spends that idle capacity on our own backlog — coding agents, driven by Logos-served models, working unattended in isolated containers, producing draft pull requests a human reviews like any other.

Half of all Logos commits already carry a `Co-Authored-By: Claude` trailer, and `Claudia-Anthropica` has been opening PRs from issues since August. Both of those run on someone's laptop or an external VM. This makes it part of the platform: governed, accounted, visible, and — crucially — throttled by the same capacity signal that governs user traffic.

## Summary

New service **`logos-agent`** (Python/FastAPI, `/api/agent/*`), plus an **Agents** page in the UI.

- **Sessions** run one task each in their own container: clone → work → test → commit → push → draft PR → optionally deploy to dev → screenshot.
- **Workspaces** are the unit of parallelism. One workspace = one working copy = one session at a time. Want more parallel work? Add workspaces.
- **The UI** shows live capacity, every session, live transcripts, screenshots, and PR links.

## The capacity gate — this is the part worth reviewing

The runner reads `/logosdb/scheduler_state` every 15 s and computes the busy share of *loaded* serving slots.

| Condition | Behaviour |
|---|---|
| load < 60 % and nobody queueing | queued sessions may start |
| load ≥ 85 %, **or any user queueing at all** | running sessions pause |
| load falls back below 60 % | paused sessions resume mid-task |
| orchestrator unreachable | nothing starts, everything running pauses |

Design decisions behind that table:

- **Pausing, not killing.** The cgroup freezer keeps the process tree and working copy, so a resumed session continues rather than restarting. Losing an hour of agent work to a traffic spike would make the whole thing not worth running.
- **Two thresholds, not one.** A session resumed at exactly the load that paused it pauses again next tick. Resume uses the *start* threshold.
- **Yield beats admit.** A pass that pauses does not also start something.
- **Unknown load counts as busy.** If the orchestrator is unreachable, something is wrong, and agent work is the cheapest thing in the system to interrupt.
- **Unloaded models contribute no capacity.** Counting a cold model's slots would make a node with nothing resident look mostly idle.

## Isolation

The runner holds the Docker socket. The sessions it creates deliberately do not:

- no Docker socket, no root (`User: agent`), `CapDrop: ALL`, `no-new-privileges`
[...]

PR #884, opened 2026-09-02
`Logos`: Fix the agent gateway, the network check, and the agents page styling

wasnertobias:
Three faults from #772 that showed up on its first real deployment. All three are needed for the agent runner to work at all on a deployed environment.

## 1. The gateway had no configuration

A deployment copies exactly one file to its VM — `docker-compose.yaml` — so the bind mount of `./agent-gateway/nginx.conf` resolved to a path that does not exist there. Docker does not fail on a missing bind source: it creates an empty **directory** in its place. nginx therefore started with its stock configuration, and the health check probed a route that configuration does not have:

```
"GET /healthz HTTP/1.1" 404 — open() "/usr/share/nginx/html/healthz" failed
FailingStreak: 15 → unhealthy
Container logos-agent-gateway Error dependency failed to start: container logos-agent-gateway is unhealthy
```

The configuration now ships inside a `logos-agent-gateway` image, built alongside the other two agent images. It stays a *template*, so the entrypoint still runs envsubst at container start: the Logos key enters the running configuration, never an image layer. Verified by running the built image — `/healthz` answers 200, `Authorization: Bearer <key>` is rendered, and nginx's own `$host` survives envsubst untouched.

## 2. The network check read a response as a dict

`ensure_network()`'s new `Internal`-flag verification called `.get()` on the return of `_request()`, which is an `httpx.Response`. The runner raised `AttributeError` inside `manager.start()` and exited before serving anything:

```
File "/srv/app/docker_engine.py", line 120, in ensure_network
    actual = bool(existing.get("Internal", False))
AttributeError: 'Response' object has no attribute 'get'
```

The tests passed because they stubbed `_request` with plain dicts. They now answer with a response object — the shape the function really has — so a dict stub fails the check instead of hiding this class of bug.

## 3. Every colour on the agents page was invalid CSS

This design system stores colour tokens as raw RGB channels (`--color-background-50: 18 8 43`), so a declaration has to wrap them: `rgb(var(--color-background-50))`. The agents stylesheet used all 62 of them bare. The browser drops those declarations, so the page kept its spacing, padding and radii and lost every background, border and text colour — cards, buttons, inputs and the error banner rendered as plain text on the page background.

Confirmed against the deployed page, whose live rules read `color: var(--color-typography-700)`, and the fix verified there by re-applying the same rules through the wrapper: buttons, banner and typography come back in the app's own palette.

## Also

A note in the runner's README: anything the gateway or the runner reads at runtime belongs in an image, because a bind mount of a file the deployment does not copy becomes an empty directory rather than an error.

## Verification

- 209 agent tests and 107 UI tests pass; pre-commit clean; UI production build clean.
- The gateway image built and run locally: health endpoint, credential injection, nginx variable handling.
- Prod was unblocked by hand in the meantime (the real `nginx.conf` placed on the host, the runner's `docker_engine.py` patched in its container) — both become unnecessary with this change, and `/opt/logos/agent-gateway/` can be removed after it deploys.

https://claude.ai/code/session_01FLK1R5FwY2LnG5LeHnnWAb

## Summary by CodeRabbit

* **New Features**
  * Added automated container image builds and publishing for the Agent Gateway.
  * Agent Gateway deployments now use a packaged nginx configuration for more reliable startup.

* **Bug Fixes**
  * Corrected network status handling when checking existing Docker networks.

* **Documentation**
  * Added deployment guidance explaining that runtime-read files must be included in container images.

* **Style**
  * Updated agent interface color styling for consistent theme rendering.

* **Tests**
  * Updated Docker Engine test fixtures to match response handling.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -25,6 +25,9 @@
 !logos/logos-agent/app
 !logos/logos-agent/workspace/run_session.py
 !logos/logos-agent/workspace/screenshot.js
+# The model gateway's nginx configuration, shipped in its image rather than
+# bind-mounted: a deployment copies only the compose file to its VM.
+!logos/agent-gateway/nginx.conf
 !iris/pyproject.toml
 !iris/poetry.lock
 !iris/src/iris
--- a/.github/workflows/logos_build-and-push-docker.yml
+++ b/.github/workflows/logos_build-and-push-docker.yml
@@ -239,6 +239,37 @@ jobs:
           username: ${{ vars.LOGOS_HARBOR_USER }}
           password: ${{ secrets.LOGOS_HARBOR_PASSWORD }}
 
+  build-agent-gateway:
+    name: Build Logos Agent Gateway
+    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == 'ls1intum/edutelligence' }}
+    uses: docker/github-builder/.github/workflows/build.yml@v1
+    permissions:
+      contents: read
+      id-token: write
+    with:
+      output: image
+      push: true
+      # The gateway's own directory is the context: it needs nothing but its
+      # nginx configuration, and a root context would put the whole
+      # repository through the .dockerignore whitelist for one small file.
+      context: ./logos/agent-gateway
+      file: logos/agent-gateway/Dockerfile
+      platforms: linux/amd64
+      runner: ubuntu-latest
+      sign: false
+      meta-images: ${{ vars.LOGOS_HARBOR_REGISTRY }}/logos/logos-agent-gateway
+      meta-tags: |
+        type=ref,event=pr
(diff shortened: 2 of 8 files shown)


Claudia-Anthropica (dismissed):
@wasnertobias The gateway image correctly includes the existing nginx template through the root-context whitelist, and the exact-head gateway build succeeded. The network check now follows the `httpx.Response` contract, with response-shaped tests covering the relevant isolation cases.

Claudia-Anthropica (approved):
@wasnertobias The gateway image and compose wiring are consistent, and the network check now correctly handles the `httpx.Response` contract with response-shaped tests. The RGB wrappers match the UI token definitions, while the captured image builds, lint, tests, and UI static export succeeded at this exact head.

Issue #886, opened 2026-09-03
`Logos`: Statistic page visual bug


wasnertobias:
<img width="239" height="340" alt="Image" src="https://github.com/user-attachments/assets/69276878-e522-4e8a-bf5f-61fbadb04b43" />

LogosOSSAgent:
# Statistic page visual bug

## What the bug is

On the statistics page the KPI cards sit six across at the default width, so each
card is close to its minimum. In the right slot of a card the sparkline is a fixed
92px drawing, and that slot is a flex item of the value row that would not give
up its width. When the card is too narrow for value + gap + a full 92px slot, the
slot kept the sparkline's width and the line ran into the card's right border —
exactly what the screenshot shows (the line's last point sits on the card edge).

## The change

Three stylesheets, the smallest set that keeps a card's right slot inside the
card's padding:

- `sparkline.scss` — the host now declares its 92px width only as an upper bound
  (`width: 92px; min-width: 0`), and the drawing is fluid
  (`width: 100%; max-width: 92px; height: auto`). The existing `viewBox` carries
  the ratio, so a tight card scales the line down instead of overflowing, and a
  roomy card still draws the identical 92×28 — no visible change where there is
  room.
- `stat-kpi-card.scss` — the projected right slot is a flex item of `.kpi-row-2`;
  it now gets `min-width: 0` so it can shrink below whatever it draws and yield
  its width to the value. This is the load-bearing rule for projected content of
  either kind.
- `statistics.scss` — the lane micro-bars are fixed-width bars that cannot
  compress, so the slot that holds them clips at the card padding
  (`overflow: hidden`) instead of reaching the border. Same class of bug, one
  line; it is the other fixed-width drawing that lives in this slot, so I fixed
  it in the same pass.

## Tests

- `sparkline.spec.ts` (new) — the point-math (full-width span, maximum at the
  top, single-point center, nothing for an empty or all-zero series),
  `shouldRender`, a render test that the svg carries no fixed width/height
  attributes and keeps its `viewBox`, and the stylesheet contract: the host may
  shrink (`min-width: 0`) and the drawing is fluid with no fixed 92×28 box.
- `stat-kpi-card.spec.ts` (new) — the slot projection still lands where the page
  puts it (sparkline and value in the value row, hint in its own row), and the
  stylesheet contract: the right slot has `min-width: 0` and the lane bars clip.
- `raw-assets.d.ts` (new) — the `declare module '*.scss?raw'` shim so the
[...]

PR #889, opened 2026-09-03
`Logos`: Make a running session legible, and stop losing what it was asked

wasnertobias:
Six things a day of production found, all of them in the way of using the runner at all, plus two about telling the agent the truth.

## The session limit could never be set

```
column "max_parallel" is of type integer but expression is of type text
```

`CASE WHEN :clear THEN NULL ELSE :max_parallel END` is a CASE over two untyped parameters: PostgreSQL resolves the whole expression to `text` and refuses to write it into an integer column. Every value, every time — the interface only ever said "could not change the session limit". Both branches are cast now, reproduced and verified against a real database:

```
-- before
ERROR:  column "max_parallel" is of type integer but expression is of type text
-- after
INSERT 0 1   max_parallel = 6
```

The unit tests could not see this because they never speak to a database, so there is a guard on the statement itself.

## Output arrived minutes late

The transcript was flushed every twenty lines and nothing else. An agent reading code and running tests prints a handful of lines a minute, so a session that was working perfectly well looked like a session that had hung. Flushed by time as well: whatever has been printed is on the page within two seconds.

## A running session showed no spending

The authoritative token counts arrive with the result file at the end, so the row said zero for the whole run. The agent phase now reports what it has spent on its own transcript — its only channel out of a sandbox that holds no credential — and the runner carries it into the row as it goes. Settlement still overwrites both with the totals from the result file.

## A launch that never started consumed its request

A trigger reference is remembered forever on purpose: an assignment must not produce a second pull request next week. But a session that failed *before its agent ever ran* did no work and answered nothing. A host missing the session image failed sixteen launches in as many seconds, and every one of those assignments and questions was then permanently invisible to the poller — the only way back was editing rows by hand. Such a request is taken up again, at most three times, so a launch that can never work stops rather than loops.

## The session image never reached the host

It is not a compose service — the runner starts containers from it through the Docker socket — so `docker compose pull` did not fetch it and every launch failed with `404: No such image`. The deploy pulls it where the registry login already is, and only where the agent profile is on.

## A question with no answer was asked forever

A settled session's artefacts are final: if it wrote no reply, no later pass will find one. The sweep asked every few seconds anyway, for the life of the deployment. It lets go now; the 😕 the settlement leaves is what the thread gets.

## Telling the agent the truth

A handover carries the pull request's review conversation in its task. The sandbox cannot fetch it, and an agent asked to "start from whatever is still open in the review" with no review in front of it reconstructs one from the diff — guesswork dressed up as work. Observed in production, verbatim:

> I don't have access to GitHub in this sandbox, so I'll rebuild the review status from the branch itself.

The house rules now say plainly that everything from GitHub is already in the task, that `gh` and `git fetch` will not answer, and that a genuinely missing piece is worth reporting rather than reconstructing.

## And the dial is a slider

Which is what a number with a floor, a ceiling and no useful precision should have been. It shows the limit in force, and offers a reset to the configured value when an override is set.

---

313 tests green, pre-commit clean, UI builds. The database work is verified against the production database; the deploy change is verified against the `.env` it greps.

## Summary by CodeRabbit

- **New Features**
  - Finished or failed sessions can be queued again with **Run again**.
  - Interrupted agent runs automatically resume up to three times while preserving conversation context.
  - Session activity updates appear in real time.
  - Request images are downloaded for agent sessions.
  - The concurrent session limit can be adjusted with a range slider and reset option.

- **Bug Fixes**
  - Improved session shutdown, retry handling, and log delivery.
  - Fixed saving concurrency settings and deployment image availability.

- **Improvements**
  - Capacity reporting reflects locally available models.
  - Incomplete or unauthorized repository context is clearly identified.
  - Pull request discussions now indicate unavailable or truncated content.
  - Sessions provide fallback responses when no answer is produced.

### DIFF
--- a/.github/workflows/logos_deploy-dev.yml
+++ b/.github/workflows/logos_deploy-dev.yml
@@ -106,6 +106,12 @@ jobs:
         uses: appleboy/ssh-action@v1.2.5
         env:
           LOGOS_HARBOR_PASSWORD: ${{ secrets.LOGOS_HARBOR_PASSWORD }}
+          LOGOS_HARBOR_REGISTRY: ${{ vars.LOGOS_HARBOR_REGISTRY }}
+          # Through the environment rather than into the script text: this
+          # value comes from whoever dispatched the workflow, and a shell
+          # command built by string substitution is a command they get to
+          # help write.
+          WORKSPACE_IMAGE_TAG: ${{ inputs.image-tag || 'latest' }}
         with:
           host: ${{ vars.VM_HOST }}
           username: ${{ vars.VM_USERNAME }}
@@ -115,7 +121,7 @@ jobs:
           proxy_key: ${{ secrets.DEPLOYMENT_GATEWAY_SSH_KEY }}
           proxy_port: ${{ vars.DEPLOYMENT_GATEWAY_PORT }}
           command_timeout: 30m
-          envs: LOGOS_HARBOR_PASSWORD
+          envs: LOGOS_HARBOR_PASSWORD,LOGOS_HARBOR_REGISTRY,WORKSPACE_IMAGE_TAG
           script: |
             printf '%s' "$LOGOS_HARBOR_PASSWORD" | docker login ${{ vars.LOGOS_HARBOR_REGISTRY }} -u '${{ vars.LOGOS_HARBOR_USER }}' --password-stdin
             docker compose -f ${{ matrix.base-path }}/${{ steps.extract-basename.outputs.basename }} --env-file=${{ matrix.base-path }}/.env pull
@@ -125,7 +131,7 @@ jobs:
             # launch with a 404. Pulled here, where the registry login is,
             # and only where sessions actually run.
             if grep -q '^COMPOSE_PROFILES=.*agent' ${{ matrix.base-path }}/.env; then
-              docker pull ${{ vars.LOGOS_HARBOR_REGISTRY }}/logos/logos-agent-workspace:${{ inputs.image-tag || 'latest' }}
+              docker pull "$LOGOS_HARBOR_REGISTRY/logos/logos-agent-workspace:$WORKSPACE_IMAGE_TAG"
             fi
 
       - name: Deploy with docker compose
--- a/.github/workflows/logos_deploy-dev.yml
+++ b/.github/workflows/logos_deploy-dev.yml
@@ -125,7 +125,7 @@ jobs:
             # launch with a 404. Pulled here, where the registry login is,
             # and only where sessions actually run.
             if grep -q '^COMPOSE_PROFILES=.*agent' ${{ matrix.base-path }}/.env; then
-              docker pull ${{ vars.LOGOS_HARBOR_REGISTRY }}/logos/logos-agent-workspace:${{ inputs.image-tag }}
+              docker pull ${{ vars.LOGOS_HARBOR_REGISTRY }}/logos/logos-agent-workspace:${{ inputs.image-tag || 'latest' }}
             fi
 
       - name: Deploy with docker compose
--- a/.github/workflows/logos_deploy-dev.yml
+++ b/.github/workflows/logos_deploy-dev.yml
@@ -125,14 +125,6 @@ jobs:
           script: |
             printf '%s' "$LOGOS_HARBOR_PASSWORD" | docker login ${{ vars.LOGOS_HARBOR_REGISTRY }} -u '${{ vars.LOGOS_HARBOR_USER }}' --password-stdin
             docker compose -f ${{ matrix.base-path }}/${{ steps.extract-basename.outputs.basename }} --env-file=${{ matrix.base-path }}/.env pull
-            # The session image is not a compose service — the runner starts
-            # containers from it through the Docker socket — so `compose pull`
-            # does not fetch it and a host that has never seen it fails every
-            # launch with a 404. Pulled here, where the registry login is,
-            # and only where sessions actually run.
-            if grep -q '^COMPOSE_PROFILES=.*agent' ${{ matrix.base-path }}/.env; then
-              docker pull "$LOGOS_HARBOR_REGISTRY/logos/logos-agent-workspace:$WORKSPACE_IMAGE_TAG"
-            fi
 
       - name: Deploy with docker compose
         uses: appleboy/ssh-action@v1.2.5
@@ -149,6 +141,14 @@ jobs:
 
       - name: Clean up old images
         uses: appleboy/ssh-action@v1.2.5
+        env:
+          LOGOS_HARBOR_PASSWORD: ${{ secrets.LOGOS_HARBOR_PASSWORD }}
+          LOGOS_HARBOR_REGISTRY: ${{ vars.LOGOS_HARBOR_REGISTRY }}
+          # Through the environment rather than into the script text: this
+          # value comes from whoever dispatched the workflow, and a shell
+          # command built by string substitution is a command they get to
+          # help write.
+          WORKSPACE_IMAGE_TAG: ${{ inputs.image-tag || 'latest' }}
         with:
           host: ${{ vars.VM_HOST }}
           username: ${{ vars.VM_USERNAME }}
@@ -157,5 +157,15 @@ jobs:
           proxy_username: ${{ vars.DEPLOYMENT_GATEWAY_USER }}
           proxy_key: ${{ secrets.DEPLOYMENT_GATEWAY_SSH_KEY }}
           proxy_port: ${{ vars.DEPLOYMENT_GATEWAY_PORT }}
+          envs: LOGOS_HARBOR_PASSWORD,LOGOS_HARBOR_REGISTRY,WORKSPACE_IMAGE_TAG
           script: |
             docker image prune -af
+            # After the prune, not before it. The session image is not a
+            # compose service — the runner starts containers from it through
+            # the Docker socket — so nothing references it between sessions
+            # and `prune -af` deletes it. That is how a host ends up failing
+            # every launch with a 404 for an image the deploy just pulled.
+            if grep -q '^COMPOSE_PROFILES=.*agent' ${{ matrix.base-path }}/.env; then
+              printf '%s' "$LOGOS_HARBOR_PASSWORD" | docker login "$LOGOS_HARBOR_REGISTRY" -u '${{ vars.LOGOS_HARBOR_USER }}' --password-stdin
+              docker pull "$LOGOS_HARBOR_REGISTRY/logos/logos-agent-workspace:$WORKSPACE_IMAGE_TAG"
+            fi
--- a/.github/workflows/logos_deploy-dev.yml
+++ b/.github/workflows/logos_deploy-dev.yml
@@ -119,6 +119,14 @@ jobs:
           script: |
             printf '%s' "$LOGOS_HARBOR_PASSWORD" | docker login ${{ vars.LOGOS_HARBOR_REGISTRY }} -u '${{ vars.LOGOS_HARBOR_USER }}' --password-stdin
             docker compose -f ${{ matrix.base-path }}/${{ steps.extract-basename.outputs.basename }} --env-file=${{ matrix.base-path }}/.env pull
+            # The session image is not a compose service — the runner starts
+            # containers from it through the Docker socket — so `compose pull`
+            # does not fetch it and a host that has never seen it fails every
+            # launch with a 404. Pulled here, where the registry login is,
+            # and only where sessions actually run.
+            if grep -q '^COMPOSE_PROFILES=.*agent' ${{ matrix.base-path }}/.env; then
+              docker pull ${{ vars.LOGOS_HARBOR_REGISTRY }}/logos/logos-agent-workspace:${{ inputs.image-tag }}
+            fi
 
       - name: Deploy with docker compose
         uses: appleboy/ssh-action@v1.2.5
--- a/logos/docker-compose.yaml
+++ b/logos/docker-compose.yaml
@@ -342,6 +342,11 @@ services:
   logos-agent:
     image: ${REGISTRY:-ghcr.io/ls1intum/edutelligence}/logos-agent:${IMAGE_TAG:-latest}
     container_name: logos-agent
+    # Long enough to freeze every running session before the process is
+    # killed. A deploy restarts this service and the model gateway together;
+    # a session frozen first survives that, one left talking to a gateway
+    # being replaced loses the turn it was in the middle of.
+    stop_grace_period: 30s
     restart: unless-stopped
     # Opt-in: the service does nothing until a profile enables it, so an
     # existing deployment that does not want agent sessions is unchanged by
--- a/logos/logos-agent/app/attachments.py
+++ b/logos/logos-agent/app/attachments.py
@@ -18,9 +18,11 @@ write, and this is the one place where its text turns into fetches.
 
 from __future__ import annotations
 
+import ipaddress
 import logging
 import re
 from pathlib import Path
+from urllib.parse import urlsplit
 
 logger = logging.getLogger(__name__)
 
@@ -45,14 +47,70 @@ _IMAGE_TYPES = {
 }
 
 
+# Where a request's images may come from. This is not a nicety: the text is
+# written by whoever opened the issue, and every URL in it is a URL the
+# runner would otherwise connect to while holding a GitHub token — at a host
+# of the author's choosing, from inside the network the orchestrator and the
+# database live on. So only GitHub's own attachment origins are followed,
+# and the token only ever goes to GitHub itself.
+_ATTACHMENT_HOSTS = frozenset({"user-images.githubusercontent.com", "raw.githubusercontent.com"})
+_ATTACHMENT_PATHS = ("/user-attachments/assets/",)
+_GITHUB_HOSTS = frozenset({"github.com", "www.github.com", "api.github.com"})
+_ASSET_PATH = re.compile(r"^/[^/]+/[^/]+/assets/", re.IGNORECASE)
+
+
+def from_github(url: str) -> bool:
+    """Whether this is an image GitHub itself is serving.
+
+    Anything else is a stranger's host named in a stranger's issue.
+    """
+    parsed = urlsplit(url)
+    if parsed.scheme != "https" or not parsed.hostname:
+        return False
+    host = parsed.hostname.lower()
+    if host in _ATTACHMENT_HOSTS or host.endswith(".githubusercontent.com"):
+        return True
+    if host in _GITHUB_HOSTS:
+        return parsed.path.startswith(_ATTACHMENT_PATHS) or bool(_ASSET_PATH.match(parsed.path))
+    return False
+
+
+def may_carry_the_token(url: str) -> bool:
+    """Whether our credential may be sent to this hop.
+
+    GitHub answers an attachment link with a redirect to signed storage, and
+    a signed URL carries its own authorisation. Ours has no business there —
+    and on any host that is not GitHub's, no business at all.
+    """
+    parsed = urlsplit(url)
+    return parsed.scheme == "https" and (parsed.hostname or "").lower() in _GITHUB_HOSTS
+
+
+def is_public(url: str) -> bool:
+    """Whether a redirect may be followed at all.
+
+    A hop that names a bare host or a private address is not a picture on
+    the internet; it is the runner being pointed at its own network.
+    """
+    parsed = urlsplit(url)
+    if parsed.scheme != "https" or not parsed.hostname:
+        return False
+    host = parsed.hostname.lower()
+    try:
+        return not ipaddress.ip_address(host).is_private
+    except ValueError:
+        return "." in host and not host.endswith(".local")
+
+
 def urls_in(text: str) -> list[str]:
-    """Every image URL a request refers to, in the order it mentions them."""
+    """Every image URL a request refers to that GitHub itself serves."""
     found: list[str] = []
     for pattern in (_MARKDOWN, _HTML, _BARE):
         for url in pattern.findall(text or ""):
             cleaned = url.rstrip(".,)")
-            if cleaned not in found:
-                found.append(cleaned)
+            if cleaned in found or not from_github(cleaned):
+                continue
+            found.append(cleaned)
     return found[:MAX_IMAGES]
 
 
@@ -70,4 +128,14 @@ def directory(artifacts: Path) -> Path:
     return artifacts / "attachments"
 
 
-__all__ = ["MAX_BYTES", "MAX_IMAGES", "directory", "is_image", "name_for", "urls_in"]
+__all__ = [
+    "MAX_BYTES",
+    "MAX_IMAGES",
+    "directory",
+    "from_github",
+    "is_image",
+    "is_public",
+    "may_carry_the_token",
+    "name_for",
+    "urls_in",
+]
--- a/logos/logos-agent/app/attachments.py
+++ b/logos/logos-agent/app/attachments.py
@@ -1,0 +1,73 @@
+"""The images a request carries, fetched so the agent can look at them.
+
+An issue that says "the statistics page looks wrong" and attaches a
+screenshot *is* that screenshot. The agent phase has no network, so it met
+one of those with `WebFetch`, was refused, read code for an hour and changed
+nothing — the request had no text to work from and the picture that was the
+whole description stayed on GitHub.
+
+So the runner fetches them. It holds the token and the egress the sandbox
+deliberately does not, writes the images into the session's own artefact
+directory, and tells the agent where they are. Claude Code reads a local
+image perfectly well; it just cannot go and get one.
+
+Bounded on purpose: a handful of images, a few megabytes each, and only what
+looks like an image when it arrives. A request is a thing a stranger can
+write, and this is the one place where its text turns into fetches.
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+# How much of a request's imagery is worth carrying. Beyond this a session is
+# reading pictures instead of code.
+MAX_IMAGES = 5
+MAX_BYTES = 8 * 1024 * 1024
+
+# Markdown images, HTML `<img src>`, and the bare attachment links GitHub
+# produces when somebody pastes a screenshot into a comment.
+_MARKDOWN = re.compile(r"!\[[^\]]*\]\((https?://[^\s)]+)\)")
+_HTML = re.compile(r"<img[^>]+src=[\"'](https?://[^\"']+)[\"']", re.IGNORECASE)
+_BARE = re.compile(r"https?://(?:user-images\.githubusercontent\.com|github\.com/user-attachments/assets)/[^\s\"'<>)]+")
+
+# What a fetched attachment may be. The runner is asking on behalf of an
+# unattended agent, so anything that is not an image is not worth having.
+_IMAGE_TYPES = {
+    "image/png": ".png",
+    "image/jpeg": ".jpg",
+    "image/gif": ".gif",
+    "image/webp": ".webp",
+}
+
+
+def urls_in(text: str) -> list[str]:
+    """Every image URL a request refers to, in the order it mentions them."""
+    found: list[str] = []
+    for pattern in (_MARKDOWN, _HTML, _BARE):
+        for url in pattern.findall(text or ""):
+            cleaned = url.rstrip(".,)")
+            if cleaned not in found:
+                found.append(cleaned)
+    return found[:MAX_IMAGES]
+
+
+def name_for(index: int, content_type: str) -> str:
+    """What a fetched image is called in the artefact directory."""
+    suffix = _IMAGE_TYPES.get(content_type.split(";")[0].strip().lower(), "")
+    return f"{index:02d}{suffix}" if suffix else ""
+
+
+def is_image(content_type: str) -> bool:
+    return content_type.split(";")[0].strip().lower() in _IMAGE_TYPES
+
+
+def directory(artifacts: Path) -> Path:
+    return artifacts / "attachments"
+
+
+__all__ = ["MAX_BYTES", "MAX_IMAGES", "directory", "is_image", "name_for", "urls_in"]
--- a/logos/logos-agent/app/attachments.py
+++ b/logos/logos-agent/app/attachments.py
@@ -103,15 +103,26 @@ def is_public(url: str) -> bool:
 
 
 def urls_in(text: str) -> list[str]:
-    """Every image URL a request refers to that GitHub itself serves."""
-    found: list[str] = []
+    """Every image URL a request refers to that GitHub itself serves.
+
+    In the order the request mentions them, not in the order the patterns
+    happen to run: the files are numbered from this list and the prompt
+    lists them by that number, so "the first screenshot" has to mean the
+    first one in the text — even in a comment that mixes an `<img>` tag with
+    a markdown image.
+    """
+    seen: set[str] = set()
+    found: list[tuple[int, str]] = []
     for pattern in (_MARKDOWN, _HTML, _BARE):
-        for url in pattern.findall(text or ""):
-            cleaned = url.rstrip(".,)")
-            if cleaned in found or not from_github(cleaned):
+        for match in pattern.finditer(text or ""):
+            url = (match.group(1) if match.groups() else match.group(0)).rstrip(".,)")
+            if url in seen or not from_github(url):
                 continue
-            found.append(cleaned)
-    return found[:MAX_IMAGES]
+            seen.add(url)
+            found.append((match.start(), url))
+    # Cut after ordering rather than during: which five a request gets is
+    # its first five, whichever notation they were written in.
+    return [url for _, url in sorted(found)[:MAX_IMAGES]]
 
 
 def name_for(index: int, content_type: str) -> str:
--- a/logos/logos-agent/app/capacity.py
+++ b/logos/logos-agent/app/capacity.py
@@ -15,11 +15,16 @@ spare to give away.
 fleet answers a question nobody asked: an embedding model, a reranker and a
 120B chat model share nothing but a building, and "1 of 60 slots busy" says
 only that most of the fleet is asleep. What decides whether another agent
-session is safe to start is the model that session will actually be served
-by — so the reading is filtered to the models the runner's own key can
-reach, and falls back to the fleet-wide figure only when none of them is
-resident (there is nothing of ours to measure, and the old signal is the
-better of the two available answers).
+session is safe to start is the deployment that session will actually be
+served by — so the reading counts exactly the deployments the runner's key
+can reach, by provider and model id rather than by name. A name is not
+enough: the same model is served by providers this key has no permission
+for, and their idle slots would make a busy lane look free.
+
+It falls back to the fleet-wide figure only when none of those deployments
+is resident — there is nothing of ours to measure then, and the older signal
+is the better of the two available answers. A key that reaches *nothing*
+is a different thing entirely, and fails closed.
 """
 
 from __future__ import annotations
@@ -60,12 +65,13 @@ class Reading:
 UNKNOWN = Reading(load=1.0, busy_slots=0, total_slots=0, queue_total=0, ok=False, detail="orchestrator unreachable")
 
 
-async def read_load(timeout_s: float = 5.0, models: frozenset[str] | None = None) -> Reading:
+async def read_load(timeout_s: float = 5.0, lane: frozenset[tuple[str, str]] | None = None) -> Reading:
     """Ask the orchestrator how busy the serving lane we would use is.
 
-    ``models`` are the names the runner's key can be served by; without them
-    the reading is fleet-wide, which is what it was before and what it falls
-    back to when none of ours is loaded.
+    ``lane`` holds the (provider id, model id) pairs the runner's key can be
+    served by. ``None`` asks for the fleet-wide figure, which is what this
+    answered before it knew about lanes; an empty set says the key reaches
+    nothing at all, which is not the same question and is refused.
     """
     if not settings.agent_api_key:
         return Reading(
@@ -95,22 +101,37 @@ async def read_load(timeout_s: float = 5.0, models: frozenset[str] | None = None
         logger.warning("capacity read failed: %s", exc)
         return UNKNOWN
 
-    return parse_scheduler_state(payload, models=models)
+    return parse_scheduler_state(payload, lane=lane)
 
 
-def parse_scheduler_state(payload: dict, models: frozenset[str] | None = None) -> Reading:
+def parse_scheduler_state(payload: dict, lane: frozenset[tuple[str, str]] | None = None) -> Reading:
     """Turn the orchestrator's debug payload into a single load figure.
 
     Kept separate from the HTTP call so it can be tested against recorded
     payloads, and so a change in the orchestrator's shape surfaces as a test
     failure rather than as a runner that silently believes the fleet is idle.
 
-    ``models`` narrows the ratio to the lane the runner's sessions are served
-    by. The queue is deliberately *not* narrowed: models share GPUs, so a
-    person waiting on any of them is a person this runner should get out of
-    the way of.
+    ``lane`` narrows the ratio to the deployments the runner's sessions are
+    served by, matched on the payload's own provider and model ids. The
+    queue is deliberately *not* narrowed: models share GPUs, so a person
+    waiting on any of them is a person this runner should get out of the way
+    of.
     """
-    wanted = {name.strip().lower() for name in (models or frozenset()) if name and name.strip()}
+    if lane is not None and not lane:
+        # A key that reaches no local deployment has no lane to measure and
+        # nothing it could legitimately run on. Refusing here rather than
+        # measuring the fleet keeps a paused session from being resumed into
+        # a permission it no longer has.
+        return Reading(
+            load=1.0,
+            busy_slots=0,
+            total_slots=0,
+            queue_total=int(payload.get("queue_total") or 0),
+            ok=True,
+            detail="the runner's key reaches no local deployment",
+            reclaimable=False,
+        )
+    wanted = set(lane or ())
     queue_total = int(payload.get("queue_total") or 0)
     providers = ((payload.get("logosnode") or {}).get("providers")) or {}
 
@@ -118,9 +139,9 @@ def parse_scheduler_state(payload: dict, models: frozenset[str] | None = None) -
     total = 0
     fleet_busy = 0
     fleet_total = 0
-    for provider in providers.values():
+    for provider_id, provider in providers.items():
         deployments = (provider or {}).get("models") or {}
-        for model in deployments.values():
+        for model_id, model in deployments.items():
             if not isinstance(model, dict):
                 continue
             # Only loaded models hold capacity. An unloaded one contributes
@@ -136,7 +157,7 @@ def parse_scheduler_state(payload: dict, models: frozenset[str] | None = None) -
             queue_total += int(model.get("queue_depth") or 0)
             fleet_total += capacity
             fleet_busy += active
-            if wanted and str(model.get("model_name") or "").strip().lower() not in wanted:
+            if wanted and (str(provider_id), str(model_id)) not in wanted:
                 continue
             total += capacity
             busy += active
--- a/logos/logos-agent/app/capacity.py
+++ b/logos/logos-agent/app/capacity.py
@@ -133,6 +133,10 @@ def parse_scheduler_state(payload: dict, lane: frozenset[tuple[str, str]] | None
         )
     wanted = set(lane or ())
     queue_total = int(payload.get("queue_total") or 0)
+    # Counted per model as well as in total: a lane holding a saturated
+    # model and an idle one is not half busy — a session bound for the
+    # saturated one has nowhere to go, and the average would hide that.
+    per_model: dict[str, list[int]] = {}
     providers = ((payload.get("logosnode") or {}).get("providers")) or {}
 
     busy = 0
@@ -161,6 +165,10 @@ def parse_scheduler_state(payload: dict, lane: frozenset[tuple[str, str]] | None
                 continue
             total += capacity
             busy += active
+            name = str(model.get("model_name") or model_id)
+            slots = per_model.setdefault(name, [0, 0])
+            slots[0] += active
+            slots[1] += capacity
 
     fell_back = False
     if wanted and total == 0 and fleet_total > 0:
@@ -184,19 +192,35 @@ def parse_scheduler_state(payload: dict, lane: frozenset[tuple[str, str]] | None
             reclaimable=False,
         )
 
+    if wanted and not fell_back and len(per_model) > 1:
+        # The busiest of them decides. Being kept out of an idle model
+        # because another is full costs this runner some capacity; letting a
+        # session into a full one costs a user their turn.
+        name, (model_busy, model_total) = max(
+            per_model.items(), key=lambda item: (item[1][0] / item[1][1]) if item[1][1] else 0.0
+        )
+        return Reading(
+            load=model_busy / model_total,
+            busy_slots=model_busy,
+            total_slots=model_total,
+            queue_total=queue_total,
+            ok=True,
+            detail=f"{model_busy}/{model_total} slots busy on {name}, the busiest model this runner may use",
+        )
+
     if not wanted:
-        lane = ""
+        lane_note = ""
     elif fell_back:
-        lane = " across the fleet (none of the runner's own models is resident)"
+        lane_note = " across the fleet (none of the runner's own models is resident)"
     else:
-        lane = " on the models the runner uses"
+        lane_note = " on the model this runner uses"
     return Reading(
         load=busy / total,
         busy_slots=busy,
         total_slots=total,
         queue_total=queue_total,
         ok=True,
-        detail=f"{busy}/{total} slots busy{lane}",
+        detail=f"{busy}/{total} slots busy{lane_note}",
     )
 
 
--- a/logos/logos-agent/app/capacity.py
+++ b/logos/logos-agent/app/capacity.py
@@ -10,6 +10,16 @@ serve concurrently (``max_capacity``), plus the depth of the queue waiting for
 it. We read that, and treat a non-empty queue as saturation regardless of the
 ratio — a user waiting is the strongest possible signal that there is nothing
 spare to give away.
+
+**Measured on the agent's own lane.** Summing every resident model in the
+fleet answers a question nobody asked: an embedding model, a reranker and a
+120B chat model share nothing but a building, and "1 of 60 slots busy" says
+only that most of the fleet is asleep. What decides whether another agent
+session is safe to start is the model that session will actually be served
+by — so the reading is filtered to the models the runner's own key can
+reach, and falls back to the fleet-wide figure only when none of them is
+resident (there is nothing of ours to measure, and the old signal is the
+better of the two available answers).
 """
 
 from __future__ import annotations
@@ -50,8 +60,13 @@ class Reading:
 UNKNOWN = Reading(load=1.0, busy_slots=0, total_slots=0, queue_total=0, ok=False, detail="orchestrator unreachable")
 
 
-async def read_load(timeout_s: float = 5.0) -> Reading:
-    """Ask the orchestrator how busy the local serving fleet is."""
+async def read_load(timeout_s: float = 5.0, models: frozenset[str] | None = None) -> Reading:
+    """Ask the orchestrator how busy the serving lane we would use is.
+
+    ``models`` are the names the runner's key can be served by; without them
+    the reading is fleet-wide, which is what it was before and what it falls
+    back to when none of ours is loaded.
+    """
     if not settings.agent_api_key:
         return Reading(
             load=1.0,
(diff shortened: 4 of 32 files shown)


logos/logos-agent/app/triggers.py
```diff
@@ -533,7 +573,13 @@ async def _candidates(self, now: datetime) -> list[dict[str, Any]]:
                     {
                         "ref": f"pr-{number}-assigned",
                         "kind": "takeover",
-                        "task": takeover_task(number, pull["title"], pull["body"], branch),
+                        "task": takeover_task(
+                            number,
+                            pull["title"],
+                            pull["body"],
+                            branch,
+                            conversation_block(await self._conversation(number)),
```
Claudia-Anthropica: @wasnertobias [high] Unlike `_comment_candidates`, this takeover path gives every collected public comment to a session that can modify and push the PR branch without checking whether its author may direct changes. An arbitrary commenter can therefore steer the eventual takeover session once the PR is assigned; filter the conversation through the existing writer authorization policy before building this branch-writing task.

In `logos/logos-agent/app/triggers.py`, assigned takeovers pass comments from unauthorized authors into a session allowed to push the PR branch. Retain only conversation entries whose authors pass the existing writer authorization policy before calling `takeover_task`.

</details>

logos/logos-agent/app/github.py
```diff
@@ -449,6 +449,59 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+        _get_all(f"/repos/{settings.repo_slug}/pulls/{number}/comments"),
+        _get_all(f"/repos/{settings.repo_slug}/issues/{number}/comments"),
+        return_exceptions=True,
+    )
+    entries: list[dict[str, Any]] = []
+
+    def add(items: object, kind: str) -> None:
+        if not isinstance(items, list):
+            # One source failing is not worth losing the other two over: a
+            # partial conversation still beats none.
```
Claudia-Anthropica: @wasnertobias [medium] A failed reviews, inline-comments, or discussion request is silently collapsed into the same plain list returned by a complete read, while `takeover_task` later states that this is the entire conversation. A transient failure can therefore hide requested feedback while telling the agent nothing is missing; return missing-source metadata and render an explicit incomplete-context warning in the task.

In `logos/logos-agent/app/github.py`, partial conversation acquisition is indistinguishable from a complete result even though the consumer claims completeness. Return the missing source names alongside the entries and make the takeover task explicitly disclose that incomplete context.

</details>

logos/logos-ui/src/app/features/agents/agents.ts
```diff
@@ -78,6 +78,28 @@ export class Agents implements OnInit {
...
+    return state?.max_parallel_override ?? state?.max_parallel_configured ?? 0;
+  });
+
+  /**
+   * How far the slider goes. The configured value is the normal working
+   * point, so the scale reaches past it without running to the schema's
+   * hundred, which no deployment has the capacity for.
+   */
+  readonly limitCeiling = computed(() => {
+    const configured = this.controls()?.max_parallel_configured ?? 10;
```
Claudia-Anthropica: @wasnertobias [medium] This ceiling ignores the current override. For example, a valid persisted override of 80 with a configured value of 10 makes the label show 80 while the browser clamps the range control to its maximum of 20, and the next interaction unexpectedly lowers the limit; include the effective override in the computed ceiling.

In `logos/logos-ui/src/app/features/agents/agents.ts`, `limitCeiling` can be lower than the valid limit currently in force, causing the range control to display a clamped value and jump downward on interaction. Compute the ceiling as at least the current override, while retaining the schema maximum of 100.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias This head has two medium correctness issues in conversation capture and slider state, plus a high authorization gap: takeover sessions receive arbitrary public comments as branch-writing instructions. The captured snapshot had no failed checks, but several non-required checks were still in progress.

logos/logos-ui/src/app/features/agents/agents.ts
```diff
@@ -264,16 +287,60 @@ export class Agents implements OnInit {
...
+      try {
+        for await (const event of this.agentService.streamEvents(
+          sessionId,
+          this.lastEventId,
+          controller.signal,
+        )) {
+          if (this.selectedId() !== sessionId) return;
+          if (event.id <= this.lastEventId) continue;
+          this.lastEventId = event.id;
+          this.events.update((existing) => [...existing, event]);
```
Claudia-Anthropica, line 357: @wasnertobias [medium] The stream and the four-second `loadEvents()` poll append independently. If polling requests events after N and the stream receives N+1 before that response returns, this path appends N+1 first, then `loadEvents()` appends the same event because it never re-filters its response against the advanced `lastEventId`, duplicating transcript output. Merge both sources through one ID-deduplicating update path or suspend polling while the stream is active.

In `logos/logos-ui/src/app/features/agents/agents.ts`, concurrent SSE and polling responses can append the same event ID twice. Deduplicate incoming events against the current `lastEventId` or stored event IDs before appending from either source.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The three existing findings remain reproducible: takeover still forwards unfiltered conversation authors into a branch-writing task, partial acquisition is still presented as complete, and `limitCeiling` still ignores the persisted override. The new SSE client also races the retained poller, allowing the same event to be appended twice. The captured agent test, pre-commit, title-validation, and UI-export checks passed; several non-required build and documentation checks remained in progress.

logos/logos-agent/app/triggers.py
wasnertobias: Fixed in a2624bbb3, and you are right about the asymmetry: the review path already refuses a change request from somebody who cannot write to the repository, and this path handed the same kind of instruction to a session that pushes to the branch.

`_conversation` now filters every entry through `_writer()` before it reaches `takeover_task`. What is left out is counted and named in the task rather than silently removed — the task tells the agent the conversation is complete, so a silent drop would be a lie in the other direction.

Covered by `test_a_stranger_cannot_direct_a_session_that_may_push`: a comment from an account without write access asking to delete an auth check does not appear in the task, and the task says comments were withheld.

logos/logos-agent/app/github.py
wasnertobias: Fixed in a2624bbb3. `pull_request_conversation` returns `(entries, missing)`, where `missing` names the sources that failed — reviews, inline comments, comments — and `conversation_block` renders them:

> Some of the conversation could not be read (reviews); what follows is incomplete. Treat a point you cannot see as unresolved rather than as absent, and say so in your final message.

A failure to reach the whole conversation reports as `the conversation`, and comments withheld for lack of write access are reported the same way, since both are the same thing from the agent's side: context it does not have. `test_a_conversation_that_could_not_be_read_says_so` covers it.

logos/logos-ui/src/app/features/agents/agents.ts
wasnertobias: Fixed in a2624bbb3:

```ts
const inForce = Math.max(state?.max_parallel_override ?? 0, this.limitShown());
return Math.min(100, Math.max(20, configured * 2, inForce));
```

The ceiling is never below the value the control is showing, so a stored 80 against a configured 10 renders at 80 instead of clamping to 20 and lowering the limit on the first drag.

wasnertobias:
All nine review points are addressed in a2624bbb3 — three from @Claudia-Anthropica, six from CodeRabbit (three of which were the same findings).

**[high] A stranger could direct a takeover session.** Correct, and it was an asymmetry with the review path, which already refuses a change request from an account without write access. The conversation is now filtered through the same `_writer()` policy before it reaches the task, and what is withheld is counted and disclosed rather than silently dropped.

**[medium] A failed source read like an empty one.** `pull_request_conversation` returns the missing source names alongside the entries, and the task opens with an explicit warning telling the agent to treat what it cannot see as unresolved.

**[medium] The slider ceiling ignored the override.** It is now at least the value in force, capped at the schema's 100.

**Reply files.** Only `FileNotFoundError` and a genuinely empty file are final; any other `OSError` records a failed attempt and stays eligible for retry, so a mount that is not there yet no longer discards an answer.

**Transcript queue.** Bounded at 2000 lines, with the end-of-stream marker placed without blocking, and a test that drives twenty lines through a queue of three against a deliberately slow `add_event`.

**Dev and test deploys.** `${{ inputs.image-tag || 'latest' }}`, as in production — an empty tag built an invalid `image:` reference.

Two things this push also carries, from operating the runner today:

**The load figure measured the wrong thing.** It summed every resident model in the fleet, so a platform where all six busy slots were on the agent's own model reported "6/120 busy" — 5 %, when its own lane was at 50 % of one node's worth. It counts the models the runner's key can reach now, falling back to the fleet-wide figure only when none is resident. The queue stays unfiltered: models share GPUs.

**A deploy no longer costs a turn.** Session containers survive a redeploy and are re-adopted, but the model gateway is replaced alongside the runner. Shutdown freezes running sessions first — the same pause the capacity logic uses, which ends the upstream generation cleanly — and the restarted runner resumes them mid-task.

And one about the tests themselves: the launch checks whether the session image is present, and unstubbed that question went to whatever Docker daemon happened to be running on the developer's machine. The suite hung or passed depending on whether Docker Desktop was open; it is stubbed now.

331 tests green, pre-commit clean, UI builds.

logos/logos-ui/src/app/features/agents/agents.ts
wasnertobias: Fixed in 9b850abab. Real race, and the dedup was only on one side of it: the stream skipped an event at or below the cursor, the poll did not re-check its answer against a cursor the stream had advanced in the meantime.

`loadEvents()` now filters what comes back against the cursor as it is when the response arrives, rather than trusting the `after_id` it sent, and drops the answer entirely if the selection changed while it was in flight. Polling stays alongside the stream on purpose — it is the fallback if a proxy will not hold a long response — so deduplicating is the fix rather than suspending it.

logos/logos-agent/app/github.py
```diff
@@ -449,6 +449,59 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+            logger.info("could not read the %s of #%s: %s", kind, number, items)
+            return
```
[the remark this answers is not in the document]
wasnertobias: Fixed across a2624bbb3 and 9b850abab, on all four points:

- `pull_request_conversation` returns `(entries, missing)`; a source that fails is named rather than collapsing into an empty list.
- `conversation_block` renders an explicit warning when anything is missing: "Some of the conversation could not be read (...); what follows is incomplete. Treat a point you cannot see as unresolved rather than as absent."
- `_conversation` propagates the status instead of turning a failure into an empty conversation — including a total failure, which reports as `the conversation`.
- The house rule in `conventions.py` is qualified: "If the task says part of it could not be read, that part is missing rather than empty — treat what you cannot see as unresolved."

`test_a_conversation_that_could_not_be_read_says_so` covers the queue-anyway-and-disclose behaviour.

logos/logos-ui/src/app/features/agents/agents.ts
```diff
@@ -264,21 +311,70 @@ export class Agents implements OnInit {
...
       this.resetScreenshots();
       return;
     }
     this.selectedId.set(session.id);
     this.events.set([]);
     this.lastEventId = 0;
+    this.stopStream();
     this.resetScreenshots();
     await this.loadEvents();
+    this.startStream(session.id);
```
Claudia-Anthropica, line 329: @wasnertobias [medium] `select()` awaits the initial event request and then unconditionally starts a stream for its original argument. If the user deselects that session or selects another while the request is in flight, the stale call still opens an SSE connection; overlapping completions can also overwrite `this.stream`, leaving the selected session's connection unabortable and the server polling a stale connection for up to 20 minutes. Re-check `selectedId()` after the await before starting the stream.

In `logos/logos-ui/src/app/features/agents/agents.ts`, `select()` starts an SSE stream after an asynchronous load even when the selection changed, allowing stale untracked streams. After `await this.loadEvents()`, return unless `this.selectedId() === session.id` before calling `startStream`.

</details>

logos/logos-agent/app/pulse.py
```diff
@@ -0,0 +1,53 @@
...
+
+import asyncio
+
+_bells: dict[int, asyncio.Event] = {}
+
+
+def _bell(session_id: int) -> asyncio.Event:
+    bell = _bells.get(session_id)
+    if bell is None:
+        bell = _bells[session_id] = asyncio.Event()
```
Claudia-Anthropica: @wasnertobias [low] Every `db.add_event()` calls `ring()`, and this insertion creates a retained dictionary entry even when no SSE client has ever watched the session. `forget()` is only reached when a watcher observes a terminal state, so unwatched sessions and clients that disconnect early accumulate bells for the lifetime of the runner. Register bells only while streams are subscribed and release them when the last subscriber exits.

In `logos/logos-agent/app/pulse.py`, event writes permanently retain an `asyncio.Event` for sessions without an active watcher. Add subscriber lifetime tracking so `ring()` does not create entries without subscribers, and remove each entry in the stream generator's `finally` block when its last subscriber exits.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The earlier authorization, incomplete-context, slider-ceiling, and stream/poll deduplication findings are resolved, and the captured CI is complete and successful. Two current-head lifecycle gaps remain: a stale asynchronous selection can open an untracked SSE stream, and the pulse registry retains entries for sessions that no client watches.

logos/logos-ui/src/app/features/agents/agents.ts
wasnertobias: Fixed in fdf5c6870. Both halves of it:

`select()` re-checks `selectedId()` after `await this.loadEvents()` and returns if the selection moved on, so a stale call opens nothing. And `startStream()` aborts whatever was streaming before it takes `this.stream`, so two overlapping completions cannot leave the older connection with no handle to abort it — the guard covers the ordinary case, this covers the interleaving one.

logos/logos-agent/app/pulse.py
wasnertobias: Fixed in fdf5c6870, and the write side is the wrong place to decide it: `add_event` has no idea whether anybody is watching, and almost nobody is.

`ring()` now only sets a bell that already exists — a no-op for an unwatched session. Bells are created by `pulse.watching(session_id)`, an async context manager the stream generator wraps its whole loop in, and removed when the last watcher leaves. Counted rather than flagged, since two people may watch the same session, and it releases on the ordinary ending too: a browser that disconnects mid-stream unwinds the generator through the same `finally`.

`pulse.wait()` on a session with no bell waits out its timeout, so a waiter that arrives without registering degrades to the polling this replaced rather than spinning.

Covered by `TestWhoIsListening`: an unwatched write leaves nothing behind, a watcher registers and lets go, a cancelled stream lets go, and the last of several watchers is the one that clears it.

logos/logos-agent/app/capacity.py
```diff
@@ -109,9 +132,21 @@ def parse_scheduler_state(payload: dict) -> Reading:
             capacity = int(model.get("max_capacity") or 0)
             if capacity <= 0:
                 continue
-            total += capacity
-            busy += min(int(model.get("active") or 0), capacity)
+            active = min(int(model.get("active") or 0), capacity)
             queue_total += int(model.get("queue_depth") or 0)
+            fleet_total += capacity
+            fleet_busy += active
+            if wanted and str(model.get("model_name") or "").strip().lower() not in wanted:
```
Claudia-Anthropica: @wasnertobias [medium] Filtering solely by model name does not restrict this reading to deployments the runner's key can reach. The scheduler payload is separated by provider and model IDs, while `reachable_deployments` already returns those IDs; if an inaccessible idle provider serves the same model as the permitted busy provider, both capacities are counted and the runner may start work against a falsely low load. Preserve the permitted deployment IDs in `ModelPolicy` and filter the scheduler payload by the exact provider/model pairs.

In `logos/logos-agent/app/capacity.py`, name-only filtering includes deployments on providers the runner key cannot access and can understate its actual lane load. Carry the reachable provider/model ID pairs from `ModelPolicy` and count only those exact deployments.

</details>

logos/logos-agent/app/sessions.py
```diff
@@ -450,7 +515,7 @@ async def scheduler_pass(self) -> None:
         #    under the lock makes every claim pay for its own observation;
         #    the backlog drains at one fresh reading per admission.
         async with self._admission_lock:
-            reading = await capacity.read_load()
+            reading = await capacity.read_load(models=model_policy.current().local_models)
```
Claudia-Anthropica: @wasnertobias [medium] This capacity sample uses `model_policy.current()` before line 524 refreshes permissions. If the key moves from local model A to local model B, the refreshed policy permits a launch on B while `start_decision` still uses A's load, allowing one session to enter a saturated lane. Refresh the policy first under the admission lock, reject an unsafe policy, and then read capacity using that returned policy.

In `logos/logos-agent/app/sessions.py`, admission samples capacity from the stale model policy and refreshes permissions only afterward. Refresh first and take the admission capacity reading from the returned current policy.

</details>

logos/logos-agent/app/github.py
```diff
@@ -449,6 +449,66 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+                    "body": body,
+                }
+            )
+
+    add(reviews, "reviews")
+    add(inline, "inline comments")
+    add(discussion, "comments")
+    entries.sort(key=lambda entry: entry["at"] or datetime.min.replace(tzinfo=timezone.utc))
+    # The newest are the ones still open; an old conversation is history.
+    return entries[-limit:], missing
```
Claudia-Anthropica, line 609: @wasnertobias [medium] This slice silently drops every conversation entry older than the newest 40 while leaving `missing` empty, after which `takeover_task` tells the agent that the supplied conversation is complete. An unresolved early review comment can therefore disappear merely because later discussion exceeded the limit. Record the omitted count as incomplete context—or explicitly retain all unresolved review threads—before returning the bounded conversation.

In `logos/logos-agent/app/github.py`, the 40-entry slice silently omits older review context while the generated task claims completeness. Preserve unresolved threads and report any deliberately omitted history through the existing missing-context metadata.

</details>

logos/logos-agent/workspace/run_session.py
```diff
@@ -442,15 +442,48 @@ def run_agent(task: str) -> dict[str, object]:
...
+    if not isinstance(usage, dict):
+        return
+    context = sum(
+        value
+        for key in ("input_tokens", "cache_creation_input_tokens", "cache_read_input_tokens")
+        if isinstance(value := usage.get(key), int)
+    )
+    written = usage.get("output_tokens")
+    before = dict(_spent)
+    _spent["in"] = max(_spent["in"], context)
```
Claudia-Anthropica: @wasnertobias [medium] Each assistant event reports usage for that model call, so taking only the maximum input context discards the input and cache tokens consumed by earlier tool turns. Output is accumulated, and settlement later writes the aggregate result, but during a long-running session `tokens_in` can substantially understate what was actually spent. Accumulate each assistant event's input/cache usage just as output usage is accumulated.

In `logos/logos-agent/workspace/run_session.py`, live input-token accounting keeps only the largest turn and drops usage from earlier model calls. Sum each assistant event's input, cache-creation, and cache-read tokens into the running total.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The two lifecycle findings that triggered this pass are fixed on the exact head, and their retained threads are resolved. Four medium correctness issues remain: lane load includes inaccessible deployments, admission samples the pre-refresh policy, conversation truncation is presented as complete, and live input-token usage discards earlier turns. The captured CI has no failed or required-pending checks; CodeRabbit itself was still pending.

logos/logos-agent/app/capacity.py
wasnertobias: Fixed in 9235bbeba. You are right that a name is not a lane: `Qwen/Qwen3.8-27B` is resident on three providers here, and a key permitted on one of them was reading the other two's idle slots as its own headroom.

`ModelPolicy` now carries `local_deployments` — the `(provider_id, model_id)` pairs from the same rows `classify()` reads, restricted to the locally served names — and `policy.lane()` hands them to `read_load`. `parse_scheduler_state` matches on the payload's own provider and model keys, so a deployment outside the lane contributes to neither side of the ratio.

`test_the_same_model_on_a_provider_we_cannot_reach_does_not_count` pins it: the same model on provider 61 at 0/20 no longer halves a 20/20 lane on provider 15.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 9235bbeba. The refresh moved above the reading, and the reading is taken on the lane the refreshed policy returns:

```python
policy = await model_policy.refresh()
if not policy.ok:
    return
reading = await capacity.read_load(lane=policy.lane())
```

Both under the admission lock, so the sample a claim pays for is the one its own permissions describe. The unsafe-policy rejection was already there and stays where it is — before the reading, now, rather than after it.

logos/logos-agent/app/github.py
wasnertobias: Fixed in 9235bbeba. The slice reports what it dropped:

```python
if len(entries) > limit:
    missing = [*missing, f"{len(entries) - limit} older comment(s), beyond what fits in a task"]
```

It travels through the same `missing` channel as a failed source, so the task carries the same warning — treat what you cannot see as unresolved. Keeping unresolved threads preferentially would be better still, but the REST listing does not carry resolution state; that needs the GraphQL API and is worth doing on its own rather than folded in here.

`test_older_comments_that_do_not_fit_are_declared` covers it.

logos/logos-agent/workspace/run_session.py
wasnertobias: Fixed in 9235bbeba — accumulating both sides now.

The reasoning behind the maximum was that a turn's input is mostly the same conversation being re-read, so summing looks like double counting. But the authority is `result.usage` at the end, which aggregates the same way, and consistency with the number that lands on the row matters more than a defensible-but-different definition: a figure that jumps when the session settles is worse than one that is merely large.

.github/workflows/logos_deploy-dev.yml
```diff
@@ -119,6 +119,14 @@ jobs:
           script: |
             printf '%s' "$LOGOS_HARBOR_PASSWORD" | docker login ${{ vars.LOGOS_HARBOR_REGISTRY }} -u '${{ vars.LOGOS_HARBOR_USER }}' --password-stdin
             docker compose -f ${{ matrix.base-path }}/${{ steps.extract-basename.outputs.basename }} --env-file=${{ matrix.base-path }}/.env pull
+            # The session image is not a compose service — the runner starts
+            # containers from it through the Docker socket — so `compose pull`
+            # does not fetch it and a host that has never seen it fails every
+            # launch with a 404. Pulled here, where the registry login is,
+            # and only where sessions actually run.
+            if grep -q '^COMPOSE_PROFILES=.*agent' ${{ matrix.base-path }}/.env; then
+              docker pull ${{ vars.LOGOS_HARBOR_REGISTRY }}/logos/logos-agent-workspace:${{ inputs.image-tag || 'latest' }}
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 9235bbeba. The tag travels through `env` (`WORKSPACE_IMAGE_TAG`, forwarded with `envs:`) and the script quotes it:

```bash
docker pull "$LOGOS_HARBOR_REGISTRY/logos/logos-agent-workspace:$WORKSPACE_IMAGE_TAG"
```

Same in dev, test and production, and the `|| 'latest'` fallback is applied where the variable is declared, so an empty dispatch input no longer builds a reference ending in a colon either.

logos/docker-compose.yaml
```diff
@@ -342,6 +342,11 @@ services:
   logos-agent:
     image: ${REGISTRY:-ghcr.io/ls1intum/edutelligence}/logos-agent:${IMAGE_TAG:-latest}
     container_name: logos-agent
+    # Long enough to freeze every running session before the process is
+    # killed. A deploy restarts this service and the model gateway together;
+    # a session frozen first survives that, one left talking to a gateway
+    # being replaced loses the turn it was in the middle of.
+    stop_grace_period: 30s
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 9235bbeba. A launch stores the container id as part of the move to running, so a session started in the shutdown window has a live agent and a row that knows nothing about it — and `_pause` cannot help, because `starting → paused` is not a legal edge.

`_freeze_starting()` therefore freezes the container without touching the row, matched by the `logos.agent.session` label against the rows still in `starting`. Restart reconciliation already handles exactly that shape — a paused container under a starting row — and normalizes it through running, so the session resumes instead of losing its turn.

`test_a_container_whose_row_is_still_starting_is_frozen_too` covers it, including that helper containers and exited ones are left alone.

logos/logos-agent/app/capacity.py
```diff
@@ -80,24 +95,32 @@ async def read_load(timeout_s: float = 5.0) -> Reading:
...
     Kept separate from the HTTP call so it can be tested against recorded
     payloads, and so a change in the orchestrator's shape surfaces as a test
     failure rather than as a runner that silently believes the fleet is idle.
+
+    ``models`` narrows the ratio to the lane the runner's sessions are served
+    by. The queue is deliberately *not* narrowed: models share GPUs, so a
+    person waiting on any of them is a person this runner should get out of
+    the way of.
     """
+    wanted = {name.strip().lower() for name in (models or frozenset()) if name and name.strip()}
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 9235bbeba, and the distinction you drew is the right one: `models=frozenset()` from a policy that reaches nothing was being read as "no filter" and answered with a fleet-wide measurement, which `resume_decision` would then act on.

`read_load(lane=...)` now separates the two. `None` means fleet-wide, which only a caller with no policy has any business asking for. An empty lane is refused outright — `reclaimable=False`, `load=1.0` — so nothing starts and anything paused stays paused. `policy.lane()` returns the empty set for a policy that is not `ok`, including one that could not be established at all.

logos/logos-agent/app/sessions.py
```diff
@@ -1015,22 +1080,94 @@ async def _supervise_session(self, session_id: int, container_id: str) -> None:
...
+                raise
+            except Exception as exc:
+                logger.warning("log collection for session %s stopped: %s", session_id, exc)
+            finally:
+                # Never blocking on a full queue: the end-of-stream marker
+                # is what lets the writer finish, and waiting for room in a
+                # queue nobody is draining any more would hang the task
+                # that is trying to end.
+                try:
+                    lines.put_nowait(None)
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 9235bbeba, with your reader-completion approach: `reader_done` is set in the reader's `finally`, before the sentinel is attempted, and the writer returns on a timeout when the reader is done and the queue is empty. The sentinel stays as the fast path.

`test_the_end_of_the_stream_is_not_lost_to_a_full_queue` drives one line through a queue of one, which is the case that hung.

logos/logos-ui/src/app/core/services/agent.service.ts
```diff
@@ -64,6 +66,58 @@ export class AgentService {
...
+    signal: AbortSignal,
+  ): AsyncGenerator<AgentEvent> {
+    const kc = this.keycloak;
+    // Same refresh window the interceptor uses: a stream opened with a token
+    // about to expire would be cut off mid-session.
+    await kc.updateToken(30).catch(() => undefined);
+    const response = await fetch(
+      `${AgentService.BASE}/sessions/${sessionId}/stream?after_id=${afterId}`,
+      { headers: { Authorization: `Bearer ${kc.token ?? ''}` }, signal },
+    );
```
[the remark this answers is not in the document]
wasnertobias: Both fixed in 9235bbeba.

`cache: 'no-store'` on the request. And the frame parser normalizes `\r\n`/`\r` to `\n` before splitting — a proxy rewriting line endings would otherwise leave every frame in the buffer, which is no events plus a string that grows for as long as the session runs — then joins a frame's `data:` lines with newlines before parsing, per the specification, and skips a frame it cannot parse rather than ending the stream over it.

logos/logos-ui/src/app/core/services/agent.service.ts
```diff
@@ -64,6 +66,58 @@ export class AgentService {
...
+      let boundary = buffer.indexOf('\n\n');
+      while (boundary >= 0) {
+        const frame = buffer.slice(0, boundary);
+        buffer = buffer.slice(boundary + 2);
+        for (const line of frame.split('\n')) {
+          if (!line.startsWith('data:')) continue;
+          const data = line.slice(5).trim();
+          // The keep-alives and the closing frame carry nothing to show.
+          if (!data || data === '{}') continue;
+          yield JSON.parse(data) as AgentEvent;
```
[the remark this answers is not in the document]
wasnertobias: Both fixed in 9235bbeba.

`cache: 'no-store'` on the request. And the frame parser normalizes `\r\n`/`\r` to `\n` before splitting — a proxy rewriting line endings would otherwise leave every frame in the buffer, which is no events plus a string that grows for as long as the session runs — then joins a frame's `data:` lines with newlines before parsing, per the specification, and skips a frame it cannot parse rather than ending the stream over it.

logos/logos-agent/app/sessions.py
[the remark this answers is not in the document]
wasnertobias: Fixed in 9235bbeba. The refresh moved above the reading, and the reading is taken on the lane the refreshed policy returns:

```python
policy = await model_policy.refresh()
if not policy.ok:
    return
reading = await capacity.read_load(lane=policy.lane())
```

Both under the admission lock, so the sample a claim pays for is the one its own permissions describe. The unsafe-policy rejection was already there and stays where it is — before the reading, now, rather than after it.

wasnertobias:
All eleven points from this round are addressed in 9235bbeba — four from @Claudia-Anthropica, seven from CodeRabbit — and each thread has its own answer. The theme running through most of them: a number is only useful if it is about the thing it claims to be about.

**The lane was matched by name, not by permission.** `Qwen/Qwen3.8-27B` is resident on three providers here, and a key permitted on one was counting the other two's idle slots as its own headroom — a saturated lane reading as half free. `ModelPolicy` carries the reachable `(provider_id, model_id)` pairs now, which is exactly how the scheduler payload is keyed, and nothing outside them touches either side of the ratio.

**An empty lane is not the absence of a lane.** A key reaching nothing produced a fleet-wide measurement that `resume_decision` would act on. It is refused outright now, distinctly from `None`.

**Admission sampled the policy it was about to replace.** Refresh first, read the lane it returns, both under the admission lock.

**A conversation cut to the newest forty claimed to be complete.** The dropped count travels through the same `missing` channel as a failed source. Preferentially keeping *unresolved* threads would be better still, but resolution state is not in the REST listing — that needs GraphQL and deserves its own change.

**Live token counts understated what was spent.** Input was the largest turn rather than the sum, so the number jumped when settlement wrote the real total. Both sides accumulate now.

**A session starting during shutdown kept talking to the gateway.** Its container id has not reached the row yet and `starting → paused` is not a legal edge, so the container is frozen without touching the row — the shape restart reconciliation already normalizes through running.

**The end-of-stream marker could be dropped by a full queue**, leaving the writer waiting for it forever. It goes by the reader being finished instead.

And the smaller ones: the workspace image tag travels to the deploy host through `env` rather than being pasted into a remote shell command, the frame parser handles CRLF and multi-line `data:` payloads, and the transcript fetch does not leave a session's output in the browser cache.

345 tests green, pre-commit clean, UI builds.

logos/logos-agent/app/main.py
```diff
@@ -118,7 +118,7 @@ async def health() -> dict[str, object]:
 
 @app.get("/capacity", response_model=CapacityState, tags=["capacity"])
 async def get_capacity(_: Principal = Depends(require_agent_operator)) -> CapacityState:
-    reading = await capacity.read_load()
+    reading = await capacity.read_load(models=model_policy.current().local_models)
```
Claudia-Anthropica: @wasnertobias [medium] `capacity.read_load()` now accepts `lane`, not `models`, so every authenticated `/capacity` request raises an unexpected-keyword `TypeError` before the response is assembled. Pass `lane=model_policy.current().lane()` and add an endpoint regression test.

In `logos/logos-agent/app/main.py`, the `/capacity` endpoint calls `capacity.read_load()` with the removed `models` keyword and returns an internal error. Pass the current policy through the supported `lane` keyword and add a test that exercises the endpoint.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias Two medium issues remain on the exact head. The `/capacity` endpoint calls `read_load()` with the removed `models` keyword and therefore returns 500; separately, `_stand_down()` returns whenever no supervisor is registered, so the new STARTING-container freeze is skipped in the start/shutdown race it is intended to cover. The lane, admission, conversation, and token-accounting fixes are present, and captured CI has no failed or required-pending checks.

logos/docker-compose.yaml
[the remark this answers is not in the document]
Claudia-Anthropica: @wasnertobias The current head still has `if not self._supervisors: return` before `_freeze_starting()`. If shutdown cancels a launch after Docker starts it but before `_supervise()`, cancellation unregisters the launch and leaves `_supervisors` empty, so the STARTING container is never frozen; please remove this guard or make it account for in-flight/STARTING containers, and cover the no-supervisor race.

logos/logos-agent/app/main.py
wasnertobias: Correct, and it was broken exactly as described — fixed in 86b8769c3. The rename to `lane=` reached the three call sites in the scheduler and missed this one, so every authenticated `/capacity` request raised `TypeError: read_load() got an unexpected keyword argument 'models'` before assembling anything.

More useful than the one-line fix is why nothing caught it: the tests around `main` check that the app builds its routes and never call a handler, so a call with an argument the callee no longer takes passes route assembly and fails on the first request. `TestTheEndpointsThatCombineModules` now calls `/capacity` and `/triggers` with their dependencies stubbed — not to pin the numbers, which the modules' own tests do, but so the seam between them fails in the suite. The capacity test also asserts the reading is taken on `policy.lane()` rather than on nothing.

347 tests green.

Claudia-Anthropica (changes requested):
@wasnertobias The `/capacity` call and regression coverage are fixed on this head, and captured CI has no failures or pending required checks. One medium shutdown-race issue remains: `_stand_down()` still returns when `_supervisors` is empty, so a launch cancelled after starting Docker but before registering `_supervise()` leaves its live `STARTING` container unfrozen through the gateway restart. Remove or relax that guard and add coverage for a STARTING container with no registered supervisor.

Claudia-Anthropica (changes requested):
@wasnertobias The shutdown-race defect remains on this exact head: `_stand_down()` returns when `_supervisors` is empty, so scheduler cancellation after Docker starts a container but before `_supervise()` skips `_freeze_starting()` and leaves that STARTING container live through the gateway restart. Remove or relax this guard and add regression coverage for a running STARTING container with no registered supervisor. The `/capacity` fix and continuation/retry changes are present; captured CI has no failures or required-pending checks, although optional build jobs were still running.

logos/docker-compose.yaml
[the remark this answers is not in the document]
wasnertobias: You are right, and thank you for coming back to it after I resolved the thread — the guard defeated exactly the case the starting half was written for. Fixed in bdeec6ffa: it is gone, and what is running is asked of the database rather than of this process's own bookkeeping.

The guard existed to keep the tests from asking a database that is not there, which belongs in the tests: `no_sessions_by_default` answers "nothing is running" unless a test says otherwise. `test_a_launch_with_no_supervisor_yet_is_still_frozen` asserts an empty `_supervisors` and still expects the starting container to be frozen.

logos/logos-agent/app/model_policy.py
```diff
@@ -85,13 +85,29 @@ class ModelPolicy:
...
+    def lane(self) -> frozenset[tuple[str, str]] | None:
+        """The deployments a session of this runner would be served by.
+
+        An empty set means the key reaches nothing, which capacity reads as
+        "no lane" and fails closed on — the same answer as a policy that
+        could not be established at all. ``None`` is reserved for "do not
+        filter", which only a caller without a policy has any business
+        asking for.
+        """
+        return self.local_deployments if self.ok else frozenset()
```
Claudia-Anthropica: @wasnertobias [medium] `lane()` returns every local deployment granted to the key, although each session resolves to one model. If permitted model A is saturated at 20/20 while unrelated permitted model B is idle at 0/100, `parse_scheduler_state()` reports only 17% load and can admit another session targeting A despite A having no spare slot. Keep deployment pairs per model and sample the lane for the queued or running session's resolved model rather than the union of all permitted models.

In `logos/logos-agent/app/model_policy.py`, `ModelPolicy.lane()` combines unrelated permitted models, allowing idle deployments to dilute the load of the model a session will use. Store deployment pairs per resolved model and update scheduler capacity reads to use the affected session's actual model lane.

</details>

logos/logos-agent/workspace/run_session.py
```diff
@@ -434,23 +461,98 @@ def run_agent(task: str) -> dict[str, object]:
...
+    usage: dict[str, object] = {}
+    for attempt in range(_MAX_CONTINUATIONS + 1):
+        resuming = attempt > 0
+        if resuming:
+            log(f"the agent's connection was cut; continuing where it left off ({attempt}/{_MAX_CONTINUATIONS})")
+        code, run_usage, interrupted = _drive_agent(
+            _agent_command(CONTINUE_PROMPT if resuming else prompt, resuming=resuming)
+        )
+        if run_usage:
+            usage = run_usage
```
Claudia-Anthropica: @wasnertobias [medium] `_drive_agent()` returns usage for one invocation, but this assignment discards any result usage and cost captured before an interruption. Live output may show cumulative tokens, yet settlement later overwrites the row from the returned final-invocation result, so a paused and continued session underreports its total spend. Accumulate token and cost totals across invocations, preserving assistant-event totals when an interrupted invocation emits no result, and cover a two-invocation run in the tests.

In `logos/logos-agent/workspace/run_session.py`, `run_agent()` replaces prior invocation usage whenever a continuation produces a result, causing final token and cost accounting to omit work before the interruption. Accumulate usage across every invocation and add a regression test asserting combined totals after one interrupted run and one successful continuation.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The no-supervisor shutdown race is fixed on this head, and the captured test/build checks succeeded; the optional CodeRabbit review remains pending. Two medium issues remain: capacity still aggregates unrelated permitted models, and a continuation replaces rather than accumulates earlier invocation usage. These can admit work against a diluted load signal and understate final accounting after a pause.

.github/workflows/logos_deploy-dev.yml
```diff
@@ -115,10 +121,18 @@ jobs:
...
           script: |
             printf '%s' "$LOGOS_HARBOR_PASSWORD" | docker login ${{ vars.LOGOS_HARBOR_REGISTRY }} -u '${{ vars.LOGOS_HARBOR_USER }}' --password-stdin
             docker compose -f ${{ matrix.base-path }}/${{ steps.extract-basename.outputs.basename }} --env-file=${{ matrix.base-path }}/.env pull
+            # The session image is not a compose service — the runner starts
+            # containers from it through the Docker socket — so `compose pull`
+            # does not fetch it and a host that has never seen it fails every
+            # launch with a 404. Pulled here, where the registry login is,
+            # and only where sessions actually run.
+            if grep -q '^COMPOSE_PROFILES=.*agent' ${{ matrix.base-path }}/.env; then
+              docker pull "$LOGOS_HARBOR_REGISTRY/logos/logos-agent-workspace:$WORKSPACE_IMAGE_TAG"
```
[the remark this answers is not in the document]
wasnertobias: Fixed in dc9dac33e — and this one had already happened: production spent this morning failing every launch with `No such image: logos-agent-workspace`, and the pull I added would have been undone by the very next deploy.

The pull moved into the cleanup step, after `docker image prune -af`, with its own registry login. Nothing references the session image between sessions, so `prune -af` is exactly the thing that deletes it, and pulling before it was pointless.

logos/logos-agent/app/github.py
```diff
@@ -449,6 +449,71 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+    add(reviews, "reviews")
+    add(inline, "inline comments")
+    add(discussion, "comments")
+    entries.sort(key=lambda entry: entry["at"] or datetime.min.replace(tzinfo=timezone.utc))
+    # The newest are the ones still open; an old conversation is history.
+    # Said rather than silently dropped, though: an early review comment
+    # nobody answered is exactly the kind of thing that falls off the end,
+    # and the task built from this claims to be complete.
+    if len(entries) > limit:
+        missing = [*missing, f"{len(entries) - limit} older comment(s), beyond what fits in a task"]
```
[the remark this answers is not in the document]
wasnertobias: Fixed in dc9dac33e. `_get_all_bounded` returns `(items, truncated)`, and `pull_request_conversation` adds `"<source> beyond the first N"` to `missing` when the ceiling bit. `_get_all` keeps its old signature for the callers that only want the items.

Worth naming which half is lost: these listings answer oldest-first, so a truncated read keeps the *oldest* entries and drops the newest — the part still open. The task's warning says to treat what it cannot see as unresolved.

`test_a_listing_that_hit_its_page_ceiling_is_named` covers it.

logos/logos-agent/app/sessions.py
```diff
@@ -384,7 +483,10 @@ async def _scheduler_loop(self) -> None:
                 pass
 
     async def scheduler_pass(self) -> None:
-        reading = await capacity.read_load()
+        # Measured on the lane these sessions are served by, not on the
+        # whole fleet: an embedding model being idle says nothing about
+        # whether another agent session is safe to start.
+        reading = await capacity.read_load(lane=model_policy.current().lane())
```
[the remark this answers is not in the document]
wasnertobias: Fixed in dc9dac33e. The scheduler pass refreshes the policy before its first capacity read, and pause and resume decisions use that reading:

```python
policy = await model_policy.refresh()
reading = await capacity.read_load(lane=policy.lane())
```

The admission block keeps its own refresh under the lock — it needs a reading taken inside the lock — but it no longer re-reads a lane the pass has already established.

logos/logos-ui/src/app/features/agents/agents.ts
```diff
@@ -264,21 +311,102 @@ export class Agents implements OnInit {
...
+   * either: the trigger counts as handled the moment a session exists for
+   * it, so without this the request was simply gone.
+   */
+  async retry(session: AgentSession): Promise<void> {
+    if (this.retrying() !== null) return;
+    this.retrying.set(session.id);
+    try {
+      const fresh = await this.agentService.retrySession(session.id);
+      await this.refresh({ quiet: true });
+      await this.select(fresh);
```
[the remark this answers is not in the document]
wasnertobias: Fixed in dc9dac33e: `if (this.selectedId() === session.id) await this.select(fresh);`. Same class as the stale-stream finding — a person who moved on while the request was in flight keeps what they moved to.

logos/logos-agent/app/model_policy.py
wasnertobias: Fixed in dc9dac33e, both halves of it.

`ModelPolicy` keeps the pairs per model as well as in total, and `lane(model)` answers for one model — resolving the deployment's default when none is named, which on a one-model key is exact. A model the key cannot reach gets an empty lane rather than the fleet.

And the reading itself no longer averages: within a lane, `parse_scheduler_state` counts per model and reports the busiest one. Your example — A at 20/20, B at 0/100 — now reads 100%, not 17%. That is deliberately conservative in the direction that costs us capacity rather than a user their turn: being kept out of an idle model because another is full costs this runner some idle GPU; letting a session into a full one costs somebody waiting.

`test_a_saturated_model_is_not_averaged_away` and `TestTheLane` cover both.

logos/logos-agent/workspace/run_session.py
wasnertobias: Fixed in dc9dac33e. `run_agent` accumulates tokens and cost across invocations and returns one usage record standing for all of them, so settlement writes the total rather than the last invocation's.

An invocation cut off before its result event contributes what the assistant events showed — the running totals the live reporting already tracks — so it counts for something rather than nothing.

`test_what_an_interrupted_run_spent_still_counts` (two invocations, 1300/250/0.7) and `test_an_invocation_that_reported_nothing_still_counts` cover it.

wasnertobias:
All six points from this round are in `dc9dac33e`, each answered in its own thread. One of them had already cost a morning:

**`docker image prune -af` ran after the workspace image was pulled.** Nothing references that image between sessions, so every deploy deleted what it had just fetched — which is precisely why production spent this morning failing every launch with `No such image: logos-agent-workspace`, and why the pull I added would have been undone by the next deploy. It is pulled after the prune now.

The others: a listing that hit its page ceiling reported as complete (it now travels through the same missing-context channel, and the truncated half is the *newest* — these endpoints answer oldest-first); pause and resume decisions sampling the cached policy; a lane averaging a saturated model with an idle one; an interrupted invocation's tokens dropped from the total; a retry opening its new session over a selection made in the meantime.

---

This push also carries the feature this pull request was missing.

**A pull request is one piece of work, not one session per round.** An issue becomes a change, a review comes back, then another — and every round met the repository as a stranger: new container, new conversation, the whole checkout read again. The sessions that ran today spent five to six million input tokens each, most of it re-reading what the previous round had already understood.

A session that continues what its workspace was doing keeps the conversation. Three things had to hold:

- **The wipe stays, the transcripts cross it.** The agent's home is replaced between sessions on purpose — `settings.json` hooks, `core.hooksPath`, a global `CLAUDE.md`, shell profiles, build caches: executable configuration written by a session that held a push token, which the next one must not inherit. Only `*.jsonl` under the CLI's project directory is carried, and that is data the same agent authored. There is a test asserting the dangerous half stays wiped.
- **A workspace follows the branch its session pushed.** That is what lets a review find the working copy that produced the change instead of starting from main — the issue-to-pull-request step that otherwise broke the chain on its first link.
- **A workspace with an open pull request is never swept**, however idle, because removing the volume takes the conversation with it. A branch whose pull request cannot be looked up is kept too: losing a working copy to a brief GitHub outage is the more expensive mistake.

The task of a continued session is what is *new* — the review that just came in — introduced as one more round rather than as a fresh assignment.

373 tests green, pre-commit clean, UI builds.

logos/logos-agent/app/sessions.py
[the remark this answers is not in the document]
wasnertobias: This was true of `bdeec6ffa`, the head you reviewed, and is fixed in `dc9dac33e`: the refresh is the first thing `scheduler_pass` does, before the reading the pause and resume decisions are made from.

```python
policy = await model_policy.refresh()
reading = await capacity.read_load(lane=policy.lane())
```

Failing closed then follows from two things you cannot see from the pass alone: `lane()` returns an empty set for a policy that is not `ok`, and an empty lane is *refused* by `parse_scheduler_state` rather than measured — `reclaimable=False`, `load=1.0`. So a revoked permission pauses what is running and resumes nothing, rather than leaving sessions on a permission that no longer exists.

That was an implication with no test behind it, which is a fair thing to have caught. `84b1d74df` adds `TestPermissionsRevokedMidFlight`: with the policy invalid, a running session is handed back and a paused one is not resumed.

logos/logos-agent/app/sessions.py
```diff
@@ -450,15 +578,18 @@ async def scheduler_pass(self) -> None:
...
-            policy = await model_policy.refresh()
-            if not policy.ok:
+            # once at startup. Re-established *before* the load is read,
+            # because the policy is what says which lane to read: a key
+            # moved from one local model to another would otherwise be
+            # admitted against the load of the model it no longer uses.
+            admission_policy = await model_policy.refresh()
+            if not admission_policy.ok:
                 return
+            reading = await capacity.read_load(lane=admission_policy.lane())
```
Claudia-Anthropica: @wasnertobias [medium] This still samples the deployment default because `lane()` receives no model, while the claimed session below can explicitly name another allowed model. With default A idle and queued B saturated, `start_decision()` admits B from A's reading; `_launch()` later checks only permission and never rechecks B's capacity. Select the candidate before the capacity decision and use its model-specific lane, applying the same correction to pause and resume decisions.

In `logos/logos-agent/app/sessions.py`, scheduler capacity reads omit each session's explicit model and can make start or resume decisions from the default model's load. Take admission, pause, and resume readings with `policy.lane(session.get("model"))`, selecting the queued candidate before its capacity check, and add a regression with idle default model A and saturated explicit model B.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The shutdown, pagination, accounting, and revoked-permission fixes are present; captured CI has no failed or pending required checks, although optional jobs remain unfinished. One medium scheduler defect remains: every load read calls `policy.lane()` without the affected session's model, so an explicit non-default session can still be admitted or resumed against the default model's load. Wire the model-specific lane into each decision before approval.

logos/logos-agent/app/github.py
```diff
@@ -449,6 +462,146 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+    """Download an image a request refers to, or None if it is not one.
+
+    Authenticated, because an attachment on a private repository needs the
+    token; bounded, because this is a fetch of a URL that came out of text a
+    stranger can write. Redirects are followed — GitHub answers attachment
+    links with one — and anything that is not an image, or is too big, is
+    left alone.
+    """
+    async with httpx.AsyncClient(timeout=30.0, follow_redirects=False) as client:
+        target, headers = url, _headers()
```
Claudia-Anthropica: @wasnertobias [high] `url` comes from task Markdown/HTML and may name any HTTP(S) host, but the first request always receives `_headers()`. A request containing `x` therefore sends the runner's GitHub token to that server; arbitrary URLs can also target services on the runner's internal network before the response type is checked. Restrict source URLs to expected GitHub attachment origins and paths before connecting, and derive headers for every hop so authorization is sent only to exact GitHub hosts.

In `logos/logos-agent/app/github.py`, `fetch_image()` sends `_headers()` to an unvalidated initial URL even though task Markdown and HTML can name any HTTP(S) host, leaking the runner token and enabling SSRF. Reject sources outside the expected GitHub attachment origins and paths before connecting, and compute headers per hop so authorization is sent only to exact GitHub hosts.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The explicit-model capacity defect remains on this head: `scheduler_pass()` still calls `lane()` without each running, paused, or queued session's model, so a saturated non-default model can be admitted or resumed based on the default model's load. The new attachment fetch also sends the GitHub authorization header on its first request to any Markdown or HTML image URL, allowing task text to exfiltrate the runner token and reach arbitrary internal hosts before the content type is checked. Captured required CI has no pending or failed checks; the successful test and lint checks do not cover these defects.

Claudia-Anthropica (changes requested):
@wasnertobias The latest transcript-detail change is present with focused unit coverage, and captured CI has no failed or required-pending checks, although optional jobs remain unfinished. The existing [medium] scheduler defect remains because pause, resume, and admission readings still call `policy.lane()` without the affected session's explicit model. The existing [high] attachment issue also remains because `fetch_image()` sends GitHub authorization on its first request before validating the Markdown or HTML URL, permitting token disclosure and SSRF; both issues are already recorded in unresolved threads, so I have not duplicated them inline.

logos/logos-agent/app/github.py
wasnertobias: You are right, and it is worse than a leaked header: the same URL could have named a service on the runner's own network — the orchestrator, the database, the Docker socket — and the fetch would have happened before anything looked at what came back. That was mine, from two commits ago. Fixed in 6ae0ff643.

Three rules now, all in `attachments`:

- `from_github(url)` gates the *source*: https, and either a `githubusercontent` host or `github.com` on an attachment path (`/user-attachments/assets/…` or `/<owner>/<repo>/assets/…`). `urls_in` drops everything else before anything is fetched, so an issue naming `attacker.example` produces no request at all.
- `may_carry_the_token(url)` is evaluated *per hop* rather than once: the credential goes to `github.com`/`api.github.com` or to nobody. GitHub redirects an attachment to signed S3 storage, and a signed URL carries its own authorisation.
- `is_public(url)` gates each redirect: https, not a private address, not a bare container name. A hop into our own network is refused rather than followed.

Verified against the real attachment on issue #886 — which redirects to `github-production-user-asset-….s3.amazonaws.com` — it still arrives (55 KB PNG), and `https://attacker.example/x.png` returns `None` without a connection. `tests/test_attachments.py` covers the stranger's host, the internal name, plain http, and the `github.com.attacker.example` lookalike.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 6ae0ff643. Admission looks at the session it would claim before it measures:

```python
candidate = await db.next_queued_session()
wanted = admission_policy.resolve(candidate.get("model"))
reading = await capacity.read_load(lane=admission_policy.lane(wanted))
```

`next_queued_session` mirrors the claim's ordering, and both run inside the admission lock, so what is measured is what is then claimed.

I left the pass-level reading on the whole lane on purpose: pausing and resuming are not decisions about one session, and models share GPUs — so `lane()` without a model means every deployment the key can reach, and the reading over several models reports its *busiest*. That keeps your earlier point (a saturated model is never averaged away) while making the per-session decision exact.

`test_the_lane_measured_is_the_candidate_s` pins both halves: with model A and B permitted, the pass measures the union and admission measures B.

logos/logos-agent/app/attachments.py
```diff
@@ -0,0 +1,73 @@
...
+
+def urls_in(text: str) -> list[str]:
+    """Every image URL a request refers to, in the order it mentions them."""
+    found: list[str] = []
+    for pattern in (_MARKDOWN, _HTML, _BARE):
+        for url in pattern.findall(text or ""):
+            cleaned = url.rstrip(".,)")
+            if cleaned not in found:
+                found.append(cleaned)
+    return found[:MAX_IMAGES]
```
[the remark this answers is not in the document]
wasnertobias: Fixed in daa2aa0ad, though not quite as suggested — your two findings pull against each other. Stopping collection at five would keep whichever five the *patterns* reach first, which is the ordering problem in your other comment.

So: matches are collected with their position, ordered by it, and cut afterwards. The quadratic part was the `in found` list scan, and that is a set now. What remains is a bounded walk over the task text, which the task builders already cap at a few thousand characters.

logos/logos-agent/app/attachments.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def urls_in(text: str) -> list[str]:
+    """Every image URL a request refers to, in the order it mentions them."""
+    found: list[str] = []
+    for pattern in (_MARKDOWN, _HTML, _BARE):
+        for url in pattern.findall(text or ""):
+            cleaned = url.rstrip(".,)")
+            if cleaned not in found:
+                found.append(cleaned)
```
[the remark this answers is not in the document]
wasnertobias: Fixed in daa2aa0ad — `finditer` with the match offset, ordered by position before the cap:

```python
found.append((match.start(), url))
...
return [url for _, url in sorted(found)[:MAX_IMAGES]]
```

`test_a_markdown_image_before_a_tag_stays_first` and `test_the_request_s_own_order_is_kept` pin it from both sides, since the bug only shows when the notations are interleaved.

logos/logos-agent/app/sessions.py
```diff
@@ -450,15 +609,18 @@ async def scheduler_pass(self) -> None:
...
```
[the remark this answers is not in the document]
wasnertobias: Already fixed in 6ae0ff643, one commit before the head you reviewed. Admission looks at the session it would claim, and measures that model's lane:

```python
candidate = await db.next_queued_session()
wanted = admission_policy.resolve(candidate.get("model"))
reading = await capacity.read_load(lane=admission_policy.lane(wanted))
```

`next_queued_session` mirrors the claim's ordering and both run inside the admission lock, so what is measured is what is claimed.

The pause and resume readings stay on the model-free lane deliberately, and that is not the same gap: handing capacity back is not a decision about one session, and models share GPUs, so the honest lane there is every deployment the key can reach. A reading over several models reports its *busiest* rather than their average, so a saturated model is never hidden by an idle one — which is the property the per-session lane gives admission, arrived at from the other side.

`test_the_lane_measured_is_the_candidate_s` asserts both: the pass measures the union, admission measures the candidate's model.

PR #890, opened 2026-09-03
`Logos`: Stop the runner from pausing itself, and leave it more room

wasnertobias:
Two things production showed within a minute of each other, both visible in the runner's own log.

## It paused itself for its own load

```
paused session 31: users are queueing (4 waiting)
paused session 32: users are queueing (4 waiting)
paused session 33: users are queueing (4 waiting)
```

Those four waiting were its own sessions. The orchestrator reports how busy a model is and never who is keeping it busy — I checked the payload: `active`, `queue_depth`, `scheduler_signals`, and nothing that attributes any of it to a caller. So a runner with four sessions in flight reads four users, pauses everything, watches the queue empty along with its own requests, resumes, and does it again.

It subtracts itself now. A running session has at most one request outstanding — being served, or waiting for a slot — so that many come off the figures before any decision is made, taken from the queue first because a request that is waiting is not being served. A real user waiting still shows and still stops the runner, which is the property that mattered.

## It kept half the fleet idle for nobody

The trigger quota was half the parallel ceiling. Ten pieces of re-triggered work today produced four running sessions and an empty queue: the other six were simply left in the repository, correctly but pointlessly, while five slots sat reserved for an operator who might turn up.

The reservation is a couple of places now — a fifth of the ceiling, at least one. Enough that a session queued by hand always finds room next to the automation, without holding half a platform idle whose whole point is spending capacity that would otherwise go to waste.

---

420 tests green, pre-commit clean. `TestNotReactingToItself` covers the discount from both sides: our own queueing is not a user waiting, and a user waiting behind us still is.

## Summary by CodeRabbit

- **New Features**
  - Operators can reorder queued sessions, including moving them up, down, or to the front.
  - Pull request tasks now include review and discussion context, with supported image attachments.
  - Scheduling accounts for deployment-specific capacity, queue activity, and cache pressure.
  - Model aliases are recognized consistently across deployments.

- **Bug Fixes**
  - Failed session replies retry safely with bounded attempts.
  - Resuming already-running or exited containers no longer blocks scheduling.
  - Deployment image tags are validated before use.
  - Cache-read tokens are excluded from usage totals for more accurate reporting.

- **Documentation**
  - Updated capacity and queueing behavior documentation.

### DIFF
--- a/.github/workflows/logos_deploy-dev.yml
+++ b/.github/workflows/logos_deploy-dev.yml
@@ -55,10 +55,23 @@ jobs:
 
       - name: Create .env file
         run: |
+          # Checked before it is written anywhere: this comes from whoever
+          # dispatched the workflow, and a value carrying a newline would
+          # add lines of its own to a file every later step reads as
+          # configuration. Matched with `case` rather than a regular
+          # expression because that is line-based, and a newline is exactly
+          # what has to be caught.
+          case "$WORKSPACE_IMAGE_TAG" in
+            "" | *[!A-Za-z0-9._-]* | [!A-Za-z0-9_]* )
+              echo "image-tag is not a Docker tag" >&2
+              exit 1
+              ;;
+          esac
+          if [ "${#WORKSPACE_IMAGE_TAG}" -gt 128 ]; then
+            echo "image-tag is too long for a Docker tag" >&2
+            exit 1
+          fi
           echo "ENVIRONMENT=${{ matrix.environment }}" > .env
-          # Through the environment and quoted: the tag comes from whoever
-          # dispatched the workflow, and a shell line built by substitution
-          # is a shell line they get to help write.
           echo "IMAGE_TAG=$WORKSPACE_IMAGE_TAG" >> .env
           echo "REGISTRY=${{ vars.LOGOS_HARBOR_REGISTRY }}/logos" >> .env
 
--- a/.github/workflows/logos_deploy-dev.yml
+++ b/.github/workflows/logos_deploy-dev.yml
@@ -56,7 +56,10 @@ jobs:
       - name: Create .env file
         run: |
           echo "ENVIRONMENT=${{ matrix.environment }}" > .env
-          echo "IMAGE_TAG=${{ inputs.image-tag }}" >> .env
+          # Through the environment and quoted: the tag comes from whoever
+          # dispatched the workflow, and a shell line built by substitution
+          # is a shell line they get to help write.
+          echo "IMAGE_TAG=$WORKSPACE_IMAGE_TAG" >> .env
           echo "REGISTRY=${{ vars.LOGOS_HARBOR_REGISTRY }}/logos" >> .env
 
           echo "$SECRETS_CONTEXT" | jq -r '
@@ -88,6 +91,7 @@ jobs:
         env:
           VARS_CONTEXT: ${{ toJson(vars) }}
           SECRETS_CONTEXT: ${{ toJson(secrets) }}
+          WORKSPACE_IMAGE_TAG: ${{ inputs.image-tag || 'latest' }}
 
       - name: Copy .env file to VM
         uses: appleboy/scp-action@v1.0.0
--- a/logos/logos-agent/README.md
+++ b/logos/logos-agent/README.md
@@ -63,6 +63,17 @@ session is never resumed into a permission it no longer has. The **queue** is
 deliberately not filtered — models share GPUs, so a person waiting on any of
 them is a person this runner gets out of the way of.
 
+**Minus its own share.** The orchestrator reports how busy a model is; it
+does not report *who* is keeping it busy, and nothing in its payload could
+say. So the runner subtracts itself: a running session has at most one
+request outstanding, either being served or waiting for a slot, and that
+many are taken off the figures before anything is decided. Without it a
+runner reads its own sessions as user traffic — it pauses itself for them,
+the load it reacted to leaves with them, it resumes, and it does it again —
+and its own sessions queueing read as "users are queueing", which is the
+signal that means stop everything. A real user waiting still shows, and
+still stops it.
+
 | Condition | What happens |
 |---|---|
 | load < `START_BELOW_LOAD` (default 60 %) and no queue | queued sessions may start |
@@ -349,8 +360,12 @@ answered twice. Only the *newest* changes-requested review of a pull request
 counts as work; the older ones were answered by it. A session that failed is
 re-queued by a person, who can see why it failed.
 
-**Bounded.** At most half the parallel ceiling may be self-queued sessions,
-so an operator queueing work by hand always finds room. Bots are ignored:
+**Bounded.** Self-queued sessions may use the parallel ceiling less a couple
+of places kept for people, so an operator queueing work by hand always finds
+room — triggered sessions carry higher priorities and would otherwise win
+every slot. Keeping half the fleet idle for a session nobody has asked for
+is the wrong trade on a platform whose point is spending what would go to
+waste. Bots are ignored:
 their findings reach the agent through the next review, not as a session per
 note. Workspaces are created on demand up to the ceiling, since a session the
 runner queued has nobody to prepare a working copy for it.
--- a/logos/logos-agent/README.md
+++ b/logos/logos-agent/README.md
@@ -360,12 +360,13 @@ answered twice. Only the *newest* changes-requested review of a pull request
 counts as work; the older ones were answered by it. A session that failed is
 re-queued by a person, who can see why it failed.
 
-**Bounded.** Self-queued sessions may use the parallel ceiling less a couple
-of places kept for people, so an operator queueing work by hand always finds
-room — triggered sessions carry higher priorities and would otherwise win
-every slot. Keeping half the fleet idle for a session nobody has asked for
-is the wrong trade on a platform whose point is spending what would go to
-waste. Bots are ignored:
+**Bounded.** Self-queued sessions may *run* up to the parallel ceiling less a
+fifth of it, at least one place, kept for people — triggered sessions carry
+higher priorities and would otherwise win every slot. Keeping half the fleet
+idle for a session nobody has asked for is the wrong trade on a platform
+whose point is spending what would go to waste. What does not fit runs
+later: it is queued, in priority order, and visible as work waiting rather
+than left in the repository where nobody sees it. Bots are ignored:
 their findings reach the agent through the next review, not as a session per
 note. Workspaces are created on demand up to the ceiling, since a session the
 runner queued has nobody to prepare a working copy for it.
--- a/logos/logos-agent/README.md
+++ b/logos/logos-agent/README.md
@@ -63,16 +63,24 @@ session is never resumed into a permission it no longer has. The **queue** is
 deliberately not filtered — models share GPUs, so a person waiting on any of
 them is a person this runner gets out of the way of.
 
-**Minus its own share.** The orchestrator reports how busy a model is; it
-does not report *who* is keeping it busy, and nothing in its payload could
-say. So the runner subtracts itself: a running session has at most one
-request outstanding, either being served or waiting for a slot, and that
-many are taken off the figures before anything is decided. Without it a
-runner reads its own sessions as user traffic — it pauses itself for them,
-the load it reacted to leaves with them, it resumes, and it does it again —
-and its own sessions queueing read as "users are queueing", which is the
-signal that means stop everything. A real user waiting still shows, and
-still stops it.
+**Minus its own share, where that is the right question.** The orchestrator
+reports how busy a model is; it does not report *who* is keeping it busy,
+and nothing in its payload could say. So the runner estimates: a running
+session has at most one request outstanding, per model, and that many come
+off the figures.
+
+Whether to *hand capacity back* is a question about other people, so it is
+decided on that adjusted figure. Without it a runner reads its own sessions
+as user traffic — it pauses itself for them, the load it reacted to leaves
+with them, it resumes, and it does it again — and its own sessions queueing
+read as "users are queueing", the signal that means stop everything. A real
+user waiting still shows, and still stops it.
+
+Whether to *take more on* is a question about the model, and the estimate is
+an upper bound: a running session may be between turns, running tests,
+making no request at all. Subtracting too much there would add work to a
+lane that is genuinely busy, so admission uses the figure as measured. What
+bounds the runner's own concurrency is the parallel ceiling, not the load.
 
 | Condition | What happens |
 |---|---|
--- a/logos/logos-agent/README.md
+++ b/logos/logos-agent/README.md
@@ -181,6 +181,14 @@ survive a restart, and both are visible to whoever finds the runner stopped.
 
 ## What gets worked on first
 
+An operator can overrule it. The queue on the page carries three controls on
+every session waiting in it — to the front, up one, down one — because the
+order is what the runner works through while the platform is busy, and
+which review is holding up a release is something the person watching knows
+and the rules do not. A move past a session of equal priority goes one above
+it: ties are broken by age, and nothing can make a session older.
+
+
 A session is admitted per capacity reading, so the order of the queue is what
 the platform actually works on while it is busy. That order is urgency, not
 arrival: a review (which holds a pull request shut) outranks a question,
--- a/logos/logos-agent/app/capacity.py
+++ b/logos/logos-agent/app/capacity.py
@@ -31,7 +31,7 @@ from __future__ import annotations
 
 import logging
 from collections.abc import Mapping
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 
 import httpx
 
@@ -179,7 +179,7 @@ def parse_scheduler_state(
     another, and subtracting them there would turn somebody else's busy
     lane into an idle-looking one.
     """
-    mine = {name.strip().lower(): count for name, count in (ours or {}).items()}
+    mine = {str(name).strip().lower(): int(count) for name, count in (ours or {}).items()}
     if lane is not None and not lane:
         # A key that reaches no local deployment has no lane to measure and
         # nothing it could legitimately run on. Refusing here rather than
@@ -310,6 +310,35 @@ def parse_scheduler_state(
         queue_total=queue_total,
         ok=True,
         detail=f"{busy}/{total} slots busy{lane_note}",
+    )
+
+
+def without_our_own(reading: Reading, ours: Mapping[str, int]) -> Reading:
+    """The reading again, with this runner's own requests taken out.
+
+    A convenience over re-reading the platform: the per-model discount lives
+    in :func:`parse_scheduler_state`, and this applies it to a reading that
+    was already taken — used where reacting to our own load is the mistake
+    (pausing, resuming) and not where it would flatter us (admitting).
+
+    The count is an estimate. A running session may be between turns with no
+    request outstanding at all, so this is an upper bound on what is ours:
+    right for deciding not to pause, wrong for deciding to add more.
+    """
+    total_ours = sum(max(0, count) for count in (ours or {}).values())
+    if total_ours <= 0 or not reading.ok:
+        return reading
+    serving = min(total_ours, reading.busy_slots)
+    waiting = min(total_ours - serving, reading.queue_total)
+    if serving == 0 and waiting == 0:
+        return reading
+    busy = reading.busy_slots - serving
+    return replace(
+        reading,
+        load=(busy / reading.total_slots) if reading.total_slots else reading.load,
+        busy_slots=busy,
+        queue_total=reading.queue_total - waiting,
+        detail=f"{busy}/{reading.total_slots} in flight besides this runner's {total_ours}",
     )
 
 
--- a/logos/logos-agent/app/capacity.py
+++ b/logos/logos-agent/app/capacity.py
@@ -31,7 +31,7 @@ from __future__ import annotations
 
 import logging
 from collections.abc import Mapping
-from dataclasses import dataclass, replace
+from dataclasses import dataclass
 
 import httpx
 
@@ -310,35 +310,6 @@ def parse_scheduler_state(
         queue_total=queue_total,
         ok=True,
         detail=f"{busy}/{total} slots busy{lane_note}",
-    )
-
-
-def without_our_own(reading: Reading, ours: Mapping[str, int]) -> Reading:
-    """The reading again, with this runner's own requests taken out.
-
-    A convenience over re-reading the platform: the per-model discount lives
-    in :func:`parse_scheduler_state`, and this applies it to a reading that
-    was already taken — used where reacting to our own load is the mistake
-    (pausing, resuming) and not where it would flatter us (admitting).
-
-    The count is an estimate. A running session may be between turns with no
-    request outstanding at all, so this is an upper bound on what is ours:
-    right for deciding not to pause, wrong for deciding to add more.
-    """
-    total_ours = sum(max(0, count) for count in (ours or {}).values())
-    if total_ours <= 0 or not reading.ok:
-        return reading
-    serving = min(total_ours, reading.busy_slots)
-    waiting = min(total_ours - serving, reading.queue_total)
-    if serving == 0 and waiting == 0:
-        return reading
-    busy = reading.busy_slots - serving
-    return replace(
-        reading,
-        load=(busy / reading.total_slots) if reading.total_slots else reading.load,
-        busy_slots=busy,
-        queue_total=reading.queue_total - waiting,
-        detail=f"{busy}/{reading.total_slots} in flight besides this runner's {total_ours}",
     )
 
 
--- a/logos/logos-agent/app/capacity.py
+++ b/logos/logos-agent/app/capacity.py
@@ -199,7 +199,8 @@ def parse_scheduler_state(
     # Counted per model as well as in total: a lane holding a saturated
     # model and an idle one is not half busy — a session bound for the
     # saturated one has nowhere to go, and the average would hide that.
-    per_model: dict[str, list[int]] = {}
+    # Per model, not per deployment: busy, waiting, capacity, cache.
+    per_model: dict[str, list[float]] = {}
     providers = ((payload.get("logosnode") or {}).get("providers")) or {}
 
     busy = 0
@@ -222,28 +223,35 @@ def parse_scheduler_state(
                 continue
             name = str(model.get("model_name") or model_id)
             active, waiting, cache = _live(model, capacity)
-            # Ours come off this model's figures, from what is *running*
-            # first. The other way round empties the queue on the assumption
-            # that our sessions are the ones waiting — and a queue that
-            # reads as empty is the signal that a user is not waiting, which
-            # is the one thing worth being wrong about in the safe
-            # direction.
-            ours_here = mine.get(name.strip().lower(), 0)
-            ours_serving = min(ours_here, active)
-            ours_waiting = min(ours_here - ours_serving, waiting)
-            active -= ours_serving
-            waiting -= ours_waiting
-            queue_total += waiting
             fleet_total += capacity
             fleet_busy += active
             if wanted and (str(provider_id), str(model_id)) not in wanted:
+                # Still counted towards the queue: somebody waiting on a
+                # deployment we cannot reach is somebody waiting.
+                queue_total += waiting
                 continue
             total += capacity
-            busy += active
-            slots = per_model.setdefault(name, [0, 0, 0.0])
+            slots = per_model.setdefault(name, [0, 0, 0, 0.0])
             slots[0] += active
-            slots[1] += capacity
-            slots[2] = max(slots[2], cache)
+            slots[1] += waiting
+            slots[2] += capacity
+            slots[3] = max(slots[3], cache)
+
+    # Ours come off each model *once*, after its deployments are added up:
+    # the same model served by three providers is one lane, and subtracting
+    # the same sessions from each of them would erase three times what this
+    # runner is doing. From what is *running* before what is waiting, too —
+    # the other way round empties the queue on the assumption that our
+    # sessions are the ones waiting, and a queue that reads as empty is the
+    # signal that no user is waiting.
+    for name, slots in per_model.items():
+        ours_here = mine.get(name.strip().lower(), 0)
+        ours_serving = min(ours_here, slots[0])
+        ours_waiting = min(ours_here - ours_serving, slots[1])
+        slots[0] -= ours_serving
+        slots[1] -= ours_waiting
+        busy += slots[0]
+        queue_total += slots[1]
 
     fell_back = False
     if wanted and total == 0 and fleet_total > 0:
@@ -271,9 +279,9 @@ def parse_scheduler_state(
         # The busiest of them decides. Being kept out of an idle model
         # because another is full costs this runner some capacity; letting a
         # session into a full one costs a user their turn.
-        name, (model_busy, model_total, cache) = max(
+        name, (model_busy, _model_waiting, model_total, cache) = max(
             per_model.items(),
-            key=lambda item: (max((item[1][0] / item[1][1]) if item[1][1] else 0.0, item[1][2])),
+            key=lambda item: (max((item[1][0] / item[1][2]) if item[1][2] else 0.0, item[1][3])),
         )
         where = f" on {name}" if len(per_model) > 1 else ""
         detail = f"{model_busy}/{model_total} requests in flight{where}"
--- a/logos/logos-agent/app/capacity.py
+++ b/logos/logos-agent/app/capacity.py
@@ -104,6 +104,44 @@ async def read_load(timeout_s: float = 5.0, lane: frozenset[tuple[str, str]] | N
     return parse_scheduler_state(payload, lane=lane)
 
 
+def _live(model: dict, capacity: int) -> tuple[int, int, float]:
+    """What a model is really doing: in flight, waiting, and how full it is.
+
+    The engine's own numbers where it reports them, the orchestrator's
+    ledger only as a fallback. They disagree in production — a lane the
+    ledger counted as serving two requests reported none running and an
+    empty KV cache — and the ledger is the one that cannot see inside vLLM.
+
+    The third figure is the one that matters most and has no slot count in
+    it at all: the KV cache. A model can be three requests into a
+    twenty-request allowance and still have no room for a fourth, because
+    concurrency there is bounded by cache, not by a number in a
+    configuration file. Whichever of the two is worse is the pressure.
+    """
+    signals = model.get("scheduler_signals") or {}
+    if not isinstance(signals, dict):
+        signals = {}
+
+    def number(*keys: str) -> float | None:
+        for key in keys:
+            value = signals.get(key)
+            if isinstance(value, (int, float)):
+                return float(value)
+        return None
+
+    running = number("requests_running_current")
+    ledger_active = float(model.get("active") or 0)
+    active = int(min(running if running is not None else ledger_active, capacity))
+
+    waiting = number("queue_waiting_current")
+    queue = int(waiting if waiting is not None else float(model.get("queue_depth") or 0))
+
+    cache = number("gpu_cache_usage_percent_max", "gpu_cache_usage_percent_avg")
+    by_slots = (active / capacity) if capacity else 0.0
+    pressure = max(by_slots, (cache or 0.0) / 100.0)
+    return active, queue, pressure
+
+
 def parse_scheduler_state(payload: dict, lane: frozenset[tuple[str, str]] | None = None) -> Reading:
     """Turn the orchestrator's debug payload into a single load figure.
 
@@ -157,8 +195,8 @@ def parse_scheduler_state(payload: dict, lane: frozenset[tuple[str, str]] | None
             capacity = int(model.get("max_capacity") or 0)
             if capacity <= 0:
                 continue
-            active = min(int(model.get("active") or 0), capacity)
-            queue_total += int(model.get("queue_depth") or 0)
+            active, waiting, pressure = _live(model, capacity)
+            queue_total += waiting
             fleet_total += capacity
             fleet_busy += active
             if wanted and (str(provider_id), str(model_id)) not in wanted:
@@ -166,9 +204,10 @@ def parse_scheduler_state(payload: dict, lane: frozenset[tuple[str, str]] | None
             total += capacity
             busy += active
             name = str(model.get("model_name") or model_id)
-            slots = per_model.setdefault(name, [0, 0])
+            slots = per_model.setdefault(name, [0, 0, 0.0])
             slots[0] += active
             slots[1] += capacity
+            slots[2] = max(slots[2], pressure)
 
     fell_back = False
     if wanted and total == 0 and fleet_total > 0:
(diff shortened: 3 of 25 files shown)


logos/logos-agent/app/sessions.py
```diff
@@ -384,15 +535,29 @@ async def _scheduler_loop(self) -> None:
...
         control = await controls.current()
 
         running = await db.sessions_in_status(SessionStatus.RUNNING)
         paused = await db.sessions_in_status(SessionStatus.PAUSED)
 
+        # The platform's load, minus this runner's own share of it. Without
+        # this the runner reads its own sessions as user traffic: it pauses
+        # itself, the load it reacted to vanishes with them, it resumes, and
+        # it does it again.
+        reading = capacity.without_our_own(measured, len(running))
```
Claudia-Anthropica: @wasnertobias [high] `measured` describes the busiest reachable model, but `len(running)` counts sessions on every model. If model A has 18 user requests while five agent sessions run on model B, this subtracts those five from A and turns 90% user load into 65%; during candidate admission, the same mismatch can also erase a global queue entry. Scope the self discount by each session's resolved model/deployment and apply it before selecting the busiest adjusted lane.

In `logos/logos-agent/app/sessions.py`, all running sessions are subtracted from a lane-specific reading, which can hide unrelated user load and queues on multi-model deployments. Count running sessions by resolved model/deployment and subtract each count only from its matching lane before selecting the busiest adjusted reading.

</details>

logos/logos-agent/app/sessions.py
```diff
@@ -433,7 +598,7 @@ async def scheduler_pass(self) -> None:
                         break
                     await self._resume(session, why)
                     live += 1
-                    reading = await capacity.read_load()
+                    reading = await capacity.read_load(lane=model_policy.current().lane())
```
Claudia-Anthropica: @wasnertobias [medium] This post-resume reading is not passed through the self-load adjustment. On an otherwise idle two-slot lane, resuming two sessions makes the raw reading 100%, so the loop stops and leaves additional paused work waiting for later scheduler passes even though the only load is its own. Apply the corrected lane-scoped self discount to every reading inside this loop before calling `resume_decision`.

In `logos/logos-agent/app/sessions.py`, the resume loop evaluates raw capacity after each resumed session and therefore reacts to its own load. Adjust each post-resume reading for the resumed sessions in the matching lane before evaluating `resume_decision`.

</details>

logos/logos-agent/app/sessions.py
```diff
@@ -450,17 +615,29 @@ async def scheduler_pass(self) -> None:
...
+            # moved from one local model to another would otherwise be
+            # admitted against the load of the model it no longer uses.
+            admission_policy = await model_policy.refresh()
+            if not admission_policy.ok:
+                return
+            # Whose lane to measure: the session that would be claimed next.
+            # Admitting it against another model's load is how a queued
+            # session enters a full lane on an idle one's figure — the
+            # launch checks permission afterwards, never capacity.
+            candidate = await db.next_queued_session()
```
Claudia-Anthropica: @wasnertobias [medium] The peek and claim are separate database operations, and `_admission_lock` only serializes this process. While `read_load` awaits, the candidate can be cancelled or a higher-priority session on another model can be queued; `claim_queued_sessions(1)` then claims that different row using the first candidate's lane reading. Claim the exact peeked ID (returning nothing if it is no longer eligible), so the row launched is always the row whose lane was measured.

In `logos/logos-agent/app/sessions.py`, admission measures one queued session but may subsequently claim a different session after the queue changes. Add an exact-ID conditional claim and launch only the candidate whose model lane was measured.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias I found three scheduler correctness gaps in the new lane-aware admission path. The current accounting can hide real user load across models, post-resume checks still include the runner's own request, and the measured candidate can change before it is claimed.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 88f259152. The discount moved inside the measurement, where the per-model numbers still exist:

```python
ours_here = mine.get(name.strip().lower(), 0)
ours_serving = min(ours_here, active)
ours_waiting = min(ours_here - ours_serving, waiting)
```

`_ours_by_model` resolves each running session's model through the policy and counts per name, so five sessions on model B come off model B and nothing else. The busiest model is then chosen from the adjusted figures, not before them.

One deliberate change of direction while I was in there: ours come off what is *running* before what is waiting. The other way round empties the queue on the assumption that our sessions are the ones waiting — and an empty queue is the signal that no user is waiting, which is the one thing worth being wrong about in the safe direction.

`test_sessions_on_another_model_are_not_subtracted_here` is your example, at nine of ten on the measured lane with five of ours on another.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 88f259152 — the loop re-reads through the same discount now, counting the sessions as they stand after each resume:

```python
resumed = await db.sessions_in_status(SessionStatus.RUNNING)
reading = await capacity.read_load(lane=policy.lane(), ours=_ours_by_model(resumed, policy))
```

So `resume_decision` sees what everybody else is doing, and `_last_reading` publishes the same adjusted figure the rest of the pass does.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 88f259152. `db.claim_session(id)` takes the exact row that was measured — under `FOR UPDATE SKIP LOCKED`, still queued, its workspace still free — and returns nothing if it is no longer eligible. Admission claims that or nothing:

```python
taken = await db.claim_session(int(candidate["id"]))
claimed = [taken] if taken else []
```

A candidate that was cancelled, or overtaken by something more urgent on another model, is simply not claimed; the next pass measures whatever is next then.

logos/logos-agent/app/github.py
```diff
@@ -315,21 +316,34 @@ async def _get_all(path: str, params: dict[str, Any] | None = None) -> list[Any]
...
+    it.
+    """
     collected: list[Any] = []
-    for page in range(1, _MAX_PAGES + 1):
+    for _ in range(_MAX_PAGES):
+        page = len(collected) // _PAGE_SIZE + 1
         payload = await _get(path, {**(params or {}), "per_page": _PAGE_SIZE, "page": page})
         if not isinstance(payload, list):
-            break
+            return collected, False
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 88f259152: a 200 that is not a list returns `(collected, True)` and is logged with the type that came back, so `pull_request_conversation` reports that source as missing rather than as complete. You are right that "not a list" is not "nothing more to read" — something answered, and it was not this endpoint.

logos/logos-agent/app/sessions.py
```diff
@@ -68,6 +89,58 @@ def container_name(session_id: int) -> str:
...
+    elif status == SessionStatus.SUCCEEDED.value:
+        lines.append(
+            "I worked through this and did not change anything — and I did not "
+            "leave a note saying why, which I should have. Nothing here is a "
+            "verdict on the request: read it as 'not done', not as 'nothing to do'."
+        )
+    else:
+        lines.append(f"I did not get through this one: the session {status}.")
+        if error:
+            lines.append(f"\n> {error[:500]}")
```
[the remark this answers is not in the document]
wasnertobias: Removed entirely in 88f259152, which is the better answer than redaction: the note itself was wrong. "I did not get through this one: the session failed" on somebody's pull request is the runner talking about itself in front of people who asked about their work, and the exception text you flagged was the worst part of it.

What that note *meant* is that the request was not dealt with — so the request is taken up again instead. Same task, workspace, branch, thread and urgency, a new row so the failed attempt keeps its transcript, bounded to three attempts per request. The thread hears from whichever attempt has something to report; nothing about the runner's internals reaches it either way.

logos/logos-agent/app/sessions.py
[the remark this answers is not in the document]
wasnertobias: Fixed in 88f259152 — the loop re-reads through the same discount now, counting the sessions as they stand after each resume:

```python
resumed = await db.sessions_in_status(SessionStatus.RUNNING)
reading = await capacity.read_load(lane=policy.lane(), ours=_ours_by_model(resumed, policy))
```

So `resume_decision` sees what everybody else is doing, and `_last_reading` publishes the same adjusted figure the rest of the pass does.

logos/logos-agent/README.md
```diff
@@ -319,8 +360,12 @@ answered twice. Only the *newest* changes-requested review of a pull request
 counts as work; the older ones were answered by it. A session that failed is
 re-queued by a person, who can see why it failed.
 
-**Bounded.** At most half the parallel ceiling may be self-queued sessions,
-so an operator queueing work by hand always finds room. Bots are ignored:
+**Bounded.** Self-queued sessions may use the parallel ceiling less a couple
+of places kept for people, so an operator queueing work by hand always finds
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 88f259152 — the README and the module docstring both say "the parallel ceiling less a fifth of it, at least one place" now. You are right that "a couple" is only true at ten.

wasnertobias:
The outside-diff finding on the `.env` step is fixed in ab614a33e. You were right that it is the same thing twice: the pull step already took the tag through the environment, while the step that writes `.env` still pasted it into an `echo`. Same input, same person dispatching the workflow.

All three deployments now declare `WORKSPACE_IMAGE_TAG` on the step and write `"$WORKSPACE_IMAGE_TAG"`, with the `latest` fallback where the variable is declared rather than in the shell line. `inputs.image-tag` appears only in `env:` blocks now, nowhere in a `run:` or `script:` body.

The five inline comments from this pass are answered in their own threads.

logos/logos-agent/app/sessions.py
```diff
@@ -623,14 +637,23 @@ async def scheduler_pass(self) -> None:
             # Admitting it against another model's load is how a queued
             # session enters a full lane on an idle one's figure — the
             # launch checks permission afterwards, never capacity.
-            candidate = await db.next_queued_session()
+            # The automation may fill the platform but not the last places
+            # in it: with its quota used up, only work a person queued is
```
Claudia-Anthropica, line 659: @wasnertobias [medium] The triggered-session count and the later claim occur in separate transactions, while `_admission_lock` only serializes this process. With a quota of three and two triggered sessions active, two replicas can both observe room and claim different triggered rows, consuming four slots and eliminating the promised manual reservation. Enforce the trigger quota atomically in the claim transaction under a database-level lock.

In `logos/logos-agent/app/sessions.py`, concurrent replicas can independently pass the triggered-session quota check and exceed the quota. Enforce the quota atomically in the database transaction that claims a triggered session.

</details>

logos/logos-agent/app/sessions.py
```diff
@@ -1634,12 +1736,17 @@ async def _post_reply(self, session_id: int) -> None:
...
-            logger.info("session %s left no answer; reporting what the runner knows", session_id)
+            # A session that wrote nothing has nothing to say, and a thread
+            # is the wrong place to say so: "the session failed, run it
+            # again from the page" is the runner talking about itself in
+            # front of people who asked about their pull request. What that
+            # means is that the request was not dealt with — so it is dealt
+            # with again, quietly, and the thread hears from the attempt
+            # that has something to report.
+            logger.info("session %s left no answer; taking the work up again instead of saying so", session_id)
+            await db.abandon_reply(session_id, attempts=_MAX_REPLY_ATTEMPTS)
```
Claudia-Anthropica, line 1825: @wasnertobias [medium] This marks the original reply permanently abandoned before `take_up_again` has created its replacement. If counting attempts or creating the new session fails transiently, `take_up_again` returns `None`, the original no longer appears in `sessions_owing_a_reply`, and the request is silently lost. Make replacement creation and abandonment an atomic, idempotent handoff, leaving the original retryable until a replacement is durable.

In `logos/logos-agent/app/sessions.py`, a failed replacement creation after `abandon_reply` permanently loses the request. Create the replacement durably and idempotently before marking the original reply abandoned.

</details>

logos/logos-agent/app/db.py
```diff
@@ -832,7 +836,83 @@ async def list_sessions(
...
+            .mappings()
+            .all()
+        )
+        order = [dict(row) for row in rows]
+        at = next((index for index, row in enumerate(order) if row["id"] == session_id), None)
+        if at is None:
+            await db.rollback()
+            return None
+        if move == "first":
+            target = max((row["priority"] for row in order), default=50) + 1 if at > 0 else order[at]["priority"]
```
Claudia-Anthropica: @wasnertobias [medium] Clamping makes valid queue moves no-ops at the priority boundaries. For example, when the first row already has priority 100, moving a later row to `first` computes 101, clamps it to 100, and the older row remains first because `created_at` breaks the tie; the equivalent failure occurs when moving below priority 0. Represent or rebalance the affected ordering so every enabled move changes the actual `ORDER BY`, and cover both boundaries.

In `logos/logos-agent/app/db.py`, clamped priorities can leave `first`, `up`, or `down` moves unchanged at priorities 100 and 0. Implement a stable mutable queue ordering that can perform every requested move and add boundary regression tests.

</details>

logos/logos-ui/src/app/features/agents/agents.ts
```diff
@@ -373,6 +373,31 @@ export class Agents implements OnInit {
 
   private stream: AbortController | null = null;
   retrying = signal<number | null>(null);
+  moving = signal<number | null>(null);
+
+  /** The queue, in the order the runner will work through it. */
+  readonly queuedSessions = computed(() => this.sessions().filter((s) => s.status === 'queued'));
```
Claudia-Anthropica: @wasnertobias [medium] `getSessions()` is ordered by `created_at DESC`, so filtering it does not produce the scheduler's queue order (`priority DESC, created_at ASC`). Consequently the UI disables the up/front/down controls on the wrong rows and continues displaying the queue newest-first after a move, making the new controls misleading. Sort and render queued sessions using the scheduler's exact ordering.

In `logos/logos-ui/src/app/features/agents/agents.ts`, `queuedSessions` preserves the API's newest-first order instead of the scheduler's priority queue order. Sort and render queued sessions by priority descending, creation time ascending, then ID ascending.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The exact head still contains the retained unresolved findings for deployment-keyed capacity accounting, the one-slot trigger quota, and successful issue sessions lacking a required reply artifact. I found four additional correctness gaps in cross-replica quota enforcement, retry durability, and the queue controls. CI captured for this head is successful, but these issues still require changes.

logos/logos-agent/app/capacity.py
```diff
@@ -157,18 +220,30 @@ def parse_scheduler_state(payload: dict, lane: frozenset[tuple[str, str]] | None
...
-            queue_total += int(model.get("queue_depth") or 0)
+            name = str(model.get("model_name") or model_id)
+            active, waiting, cache = _live(model, capacity)
+            # Ours come off this model's figures, from what is *running*
+            # first. The other way round empties the queue on the assumption
+            # that our sessions are the ones waiting — and a queue that
+            # reads as empty is the signal that a user is not waiting, which
+            # is the one thing worth being wrong about in the safe
+            # direction.
+            ours_here = mine.get(name.strip().lower(), 0)
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 0dddb4535, and it was worse than averaging: `ours` was subtracted *inside* the per-deployment loop, so on this platform — where the same model is resident on three providers — three sessions came off three times.

Deployments are now added up per model first (busy, waiting, capacity, and the worst cache among them), and the runner's share comes off that total once. `per_model` merging same-named deployments is deliberate and stays: a session can be routed to any of them, so "how busy is my model" is the sum across its providers, not one provider's view of it.

logos/logos-agent/app/triggers.py
```diff
@@ -124,16 +127,18 @@ def _next_auto_name(existing: set[str]) -> str:
...
+    higher priorities and would otherwise win every slot — but keeping half
+    the fleet idle for a session nobody has asked for is a poor trade on a
+    platform whose whole point is spending capacity that would go to waste.
+
+    It follows the *current* ceiling rather than the configured one, so
+    lowering the limit during an incident lowers this with it.
     """
     limit = settings.max_parallel_sessions if ceiling is None else ceiling
-    return max(1, limit // 2)
+    return max(1, limit - max(1, limit // 5))
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 0dddb4535 — `max(0, limit - max(1, limit // 5))`, exactly as suggested. At a ceiling of one the single place is a person's, and at zero the automation reports zero rather than one.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 0dddb4535. The count moved into the claim, evaluated in the same statement that takes the row:

```sql
AND ( s.trigger_ref IS NULL
      OR CAST(:trigger_quota AS INTEGER) IS NULL
      OR (SELECT COUNT(*) FROM agent_sessions mine
           WHERE mine.trigger_ref IS NOT NULL AND mine.status = ANY(:occupying)) < CAST(:trigger_quota AS INTEGER) )
```

The count before the capacity read stays as a cheap filter for *which* candidate to measure; the claim is what decides. Under READ COMMITTED two transactions can still read the same count in the same instant, so this narrows the window rather than closing it absolutely — the row lock is on the session, not on the quota. Worth saying plainly rather than implying more.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 0dddb4535 — the order is reversed. `take_up_again` runs first, and only a replacement that exists (or a request that has used its three attempts) leads to the original being abandoned. A transient failure leaves it owing an answer, so the next sweep tries again rather than losing the request between two failures.

logos/logos-agent/app/db.py
wasnertobias: Fixed in 0dddb4535, and thank you for pinning it down at both ends — the top boundary is the one this deployment actually reaches, since a security-labelled issue sits at 100.

Moves are expressed as positions now: the row is taken out of the ordered list and put back at the wanted index, then the queue is renumbered across the range with an even step. So every enabled control changes what the scheduler does next, whatever the numbers were before. The reason is recorded only on the row somebody moved — the others were renumbered to preserve an order nobody decided to change.

`test_a_full_priority_at_the_top_is_no_obstacle` and `test_the_bottom_of_the_range_is_no_obstacle_either` are your two boundaries, both asserting the resulting *order* rather than the numbers.

logos/logos-ui/src/app/features/agents/agents.ts
wasnertobias: Fixed in 0dddb4535: `queuedSessions` sorts by `priority DESC, created_at ASC, id` — the scheduler's own order — so the list reads as the queue and the arrows are disabled on the actual first and last rows.

logos/logos-agent/app/sessions.py
```diff
@@ -89,35 +94,20 @@ def container_name(session_id: int) -> str:
...
-        if error:
-            lines.append(f"\n> {error[:500]}")
-    lines.append(
-        "\nThe transcript is in the Logos agent page under session "
-        f"{session.get('id', '?')}, and this can be run again from there."
-    )
-    return "\n".join(lines)
+    counted: dict[str, int] = {}
+    for session in running:
+        name = policy.resolve(session.get("model"))
```
Claudia-Anthropica: @wasnertobias [medium] `ModelPolicy.resolve()` preserves the requested or configured alias, while `parse_scheduler_state()` keys its aggregates by the scheduler's canonical `model_name`. If the default or an explicit session model is an alias, this count never matches its lane, so the runner still treats its own requests as user load and can resume the pause cycle. Canonicalize aliases before building this mapping.

In `logos/logos-agent/app/sessions.py`, `_ours_by_model` keeps alias spellings although capacity aggregation uses canonical scheduler model names, so alias-selected sessions are not discounted. Canonicalize every resolved alias to its primary model name before building the mapping and add explicit-alias and default-alias regression tests.

</details>

logos/logos-agent/app/sessions.py
```diff
@@ -89,35 +94,20 @@ def container_name(session_id: int) -> str:
...
+        if name:
+            counted[name] = counted.get(name, 0) + 1
```
Claudia-Anthropica, line 112: @wasnertobias [medium] A `RUNNING` session is not necessarily making a model request—it can be executing local tools between turns. Deducting one request for every such row can erase unrelated user activity; for example, five agents running tests make eight real user requests appear as only three, allowing more agent work onto an already busy lane. Count only runner requests known to be outstanding rather than inferring ownership from session status.

In `logos/logos-agent/app/sessions.py`, every running database session is treated as an outstanding model request, so sessions doing local work can subtract unrelated user requests. Track attributable outstanding gateway requests per model and pass only those counts to `read_load`; do not infer request ownership from session status alone.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias [medium] `claim_session` still checks the trigger count without locking a shared quota row, so two replicas that selected different candidates can both pass under READ COMMITTED; serialize the quota decision and claim. [medium] The reply handoff remains non-idempotent: concurrent sweeps can create duplicate replacements, and repeated attempt-count failures cause the original to be abandoned; create or identify the replacement and abandon the original atomically. [medium] `issue_task` still omits the `reply.md` requirement, the UI renders queued rows through newest-first `activeSessions()` instead of `queuedSessions()`, and attachment redirects still accept arbitrary public DNS names without resolution or an approved-host check. Captured CI is successful, but these defects and the two inline self-load issues remain.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 064755908. The policy now carries every name a deployment answers to against the one the platform uses — built from the same rows `classify()` reads, so it cannot drift from them — and `_ours_by_model` counts under `policy.canonical(policy.resolve(...))`.

Worth naming what this would have looked like: nothing. The count would have matched no lane, the discount would have subtracted zero, and the runner would have gone back to pausing itself with no error anywhere. `TestAliases` covers the alias, the canonical name, and an unknown name being left alone.

logos/logos-agent/app/sessions.py
wasnertobias: You are right, and the fix is to stop pretending the estimate is a measurement. It is an upper bound — a running session may be between turns, running tests, reading files — and the two decisions do not want the same mistake:

- **Handing capacity back** is a question about other people. Counting our own sessions there is the bug this whole change exists to remove, and over-subtracting is harmless: at worst the runner keeps working while the platform is quiet. That decision keeps the adjusted figure.
- **Taking more on** is a question about the model. It does not matter who filled it, and subtracting five agents that are all running tests would put new work on a lane that is genuinely busy — your example exactly. Admission uses the figure as measured now.

What bounds the runner's own concurrency is the parallel ceiling and the trigger quota, not the load reading, so nothing is lost by admitting against the raw number.

Counting *known outstanding requests* would be better than either, and the place that knows is the gateway — every session's model traffic passes through it. That is a real change (a counter in nginx, or the runner proxying its own sessions' requests) and worth doing on its own rather than folded in here.

logos/logos-agent/app/capacity.py
```diff
@@ -224,6 +313,35 @@ def parse_scheduler_state(payload: dict, lane: frozenset[tuple[str, str]] | None
...
+    A convenience over re-reading the platform: the per-model discount lives
+    in :func:`parse_scheduler_state`, and this applies it to a reading that
+    was already taken — used where reacting to our own load is the mistake
+    (pausing, resuming) and not where it would flatter us (admitting).
+
+    The count is an estimate. A running session may be between turns with no
+    request outstanding at all, so this is an upper bound on what is ours:
+    right for deciding not to pause, wrong for deciding to add more.
+    """
+    total_ours = sum(max(0, count) for count in (ours or {}).values())
```
Claudia-Anthropica: @wasnertobias [high] `reading` contains only the busiest model selected by `parse_scheduler_state`, but this sums runner sessions across every model. With nine user requests on model A and five runner sessions on model B, those five are subtracted from A, turning 90% load into 40% and preventing the runner from yielding. Apply the canonicalized ownership map during per-model parsing, and use that adjusted reading only for pause/resume decisions.

In `logos/logos-agent/app/capacity.py`, `without_our_own` sums ownership across models and subtracts it from a single-model reading, allowing runner sessions on one model to mask user load on another. Apply canonicalized ownership during per-model parsing and use the adjusted result only for pause and resume decisions.

</details>

logos/logos-ui/src/app/features/agents/agents.ts
```diff
@@ -373,6 +373,44 @@ export class Agents implements OnInit {
...
+  moving = signal<number | null>(null);
+
+  /**
+   * The queue, in the order the runner will work through it.
+   *
+   * Not the order the list arrives in: sessions come newest-first, and the
+   * scheduler takes the most urgent, oldest among equals. Showing one and
+   * moving by the other would grey out the arrows on the wrong rows.
+   */
+  readonly queuedSessions = computed(() =>
```
Claudia-Anthropica, line 393: @wasnertobias [medium] This sorted projection only drives the button boundary checks; the template still iterates `activeSessions()`, which preserves the API's `created_at DESC` order. After a move, queued rows therefore remain visibly newest-first while the controls operate on priority order. Render queued rows from `queuedSessions()` separately from the other active states.

In `logos/logos-ui/src/app/features/agents/agents.ts`, `queuedSessions` is sorted into scheduler order but is not used to render the queue, so the displayed order disagrees with the movement controls. Render queued sessions from `queuedSessions()` and keep non-queued active sessions in their existing view.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias Alias canonicalization is fixed, but the new post-read discount again lets sessions on one model conceal user load on another. The queue projection is sorted correctly but is not used for rendering, so the displayed order still disagrees with the controls. The cross-replica trigger quota remains a check-then-claim race, while reply replacement remains non-atomic and `issue_task` still does not require the artifact that the retry path expects. The captured snapshot has no failed or required checks, although several non-required jobs were still running.

logos/logos-agent/app/capacity.py
wasnertobias: Fixed in 3ae562505 — and you caught the same mistake I had just fixed one level down: the convenience that adjusted a finished reading summed our sessions across every model, while the reading described only the busiest one.

There is no adjusting a reading after the fact, because the per-model figures are gone by then. So the pass takes two readings of the same moment instead — the platform's, and the platform's with `ours` applied during parsing, while the models are still separate. Pausing and resuming decide on the second, admission on the first. The convenience is gone.

`test_our_sessions_on_one_model_do_not_empty_another` is your example: nine on A, five of ours on B, and the adjusted figure still reads 90% on A.

logos/logos-ui/src/app/features/agents/agents.ts
wasnertobias: Fixed in 3ae562505. The queue is its own group above the active ones, rendered from `queuedSessions()`, and `activeSessions()` no longer includes queued rows. Sorting only the projection that drives the arrows was exactly half a fix: the controls moved things the list did not show.

logos/logos-agent/app/capacity.py
```diff
@@ -192,20 +275,26 @@ def parse_scheduler_state(payload: dict, lane: frozenset[tuple[str, str]] | None
...
-    if wanted and not fell_back and len(per_model) > 1:
+    if wanted and not fell_back and per_model:
         # The busiest of them decides. Being kept out of an idle model
         # because another is full costs this runner some capacity; letting a
         # session into a full one costs a user their turn.
-        name, (model_busy, model_total) = max(
-            per_model.items(), key=lambda item: (item[1][0] / item[1][1]) if item[1][1] else 0.0
+        name, (model_busy, _model_waiting, model_total, cache) = max(
+            per_model.items(),
+            key=lambda item: (max((item[1][0] / item[1][2]) if item[1][2] else 0.0, item[1][3])),
```
Claudia-Anthropica: @wasnertobias [high] Including cache pressure in this `max()` chooses one model and then exposes only that model's request load. With model A at 0% load and 95% cache while model B has 90% user load and no cache pressure, this returns A's 0% load; because pause/resume deliberately ignore cache pressure, the runner neither yields to B nor keeps paused work asleep. Compute the worst request load and worst cache pressure independently, or use decision-specific readings, and cover this two-model case.

In `logos/logos-agent/app/capacity.py`, selecting one model by the maximum of load and cache pressure lets an idle high-cache model hide another model's high request load from pause and resume decisions. Preserve independent worst-case request-load and cache-pressure values, and add a regression test with an idle 95%-cache model beside a 90%-loaded model.

</details>

logos/logos-agent/app/sessions.py
```diff
@@ -690,27 +739,62 @@ async def _detach_from_model_gateway(self, session_id: int, container_id: str) -
...
+        Unknown counts as present: a daemon that cannot be asked is not
+        evidence that a session's container is gone, and settling a live
+        session on a failed question would throw its work away.
+        """
+        try:
+            state, _ = await docker_engine.container_state(container_id)
+        except Exception as exc:
+            logger.info("could not ask about container %s: %s", container_id, exc)
+            return True
+        return state != "gone"
```
Claudia-Anthropica: @wasnertobias [medium] An `exited` container satisfies this predicate, but Docker cannot unpause it, so `_resume` returns `False` on every pass and leaves the database row paused forever. That state is reachable when a clean settlement observes that the pauser won the row—the new settlement branch deliberately retains the exited, non-auto-removed container. Distinguish resumable states from `exited`, carry its exit code into a valid settlement/finalization path, and add coverage for a paused row whose container already exited.

In `logos/logos-agent/app/sessions.py`, `_container_exists` treats an exited container as resumable, leaving its paused session permanently stuck when unpause fails. Detect exited containers, transition the paused row through a valid settlement/finalization path using the recorded exit code, and add a regression test.

</details>

logos/logos-ui/src/app/features/agents/agents.ts
```diff
@@ -137,7 +137,10 @@ export class Agents implements OnInit {
   );
 
   // ── grouped view ─────────────────────────────────────────────────────────
-  activeSessions = computed(() => this.sessions().filter((s) => isActive(s.status)));
+  /** What is under way: everything active that is not still waiting. */
+  activeSessions = computed(() =>
+    this.sessions().filter((s) => isActive(s.status) && s.status !== 'queued'),
```
Claudia-Anthropica, line 142: @wasnertobias [medium] `activeSessions()` also drives `tick()`'s decision to refresh session data. When the page contains only queued rows and nothing is selected, excluding those rows makes every tick take the capacity-only branch; later starts, cancellations, and queue changes never reach the UI because its stale state continues to report zero active sessions. Keep this rendering split, but base the polling gate on all non-terminal sessions—including queued ones—and add a queued-only polling test.

In `logos/logos-ui/src/app/features/agents/agents.ts`, excluding queued rows from `activeSessions` causes `tick` to stop refreshing sessions when only queued work exists. Use a separate non-terminal-session predicate for polling that includes queued rows, while retaining the separate queue rendering, and add a queued-only polling regression test.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The two current-trigger replies are corroborated: self-load is now discounted per model before selection, and the queue renders from `queuedSessions()`. Three additional current-head regressions still block approval: cache pressure can hide a busier model from yielding, an exited paused container remains permanently stuck, and a queued-only UI stops polling session state. The earlier unresolved comments remain applicable; captured CI had no failed or required checks, while two non-required build jobs were still running.

logos/logos-agent/app/capacity.py
wasnertobias: Fixed in 2eff2b7d9, and this was mine from an hour ago: I made "busiest" mean whichever of load and cache was worse, then reported that model's *request* load. A model at 0% with a 95% cache therefore represented a lane whose other model was at 90% — and since pause and resume deliberately ignore cache pressure, the runner neither yielded to B nor kept its paused work asleep.

They are chosen apart now: the busiest lane by request load decides `load`, `busy_slots` and the queue; the fullest cache across the lane decides `cache_pressure`, which only gates starting new work. `TestLoadAndCacheAreChosenApart` is your case — A at 0%/95%, B at 90%/0% — asserting 0.9 for the load, `model-b` in the detail, and that the 95% cache still refuses a start.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 2eff2b7d9. `_container_now` returns the state and the exit code, and an exited container goes through the ordinary settlement on its own exit code rather than being unpaused forever. As you say, the reachable path is a settlement that lost the row to the pauser and — since the last change — correctly left the container alone; that container then exits on its own, and the row had no way back.

logos/logos-ui/src/app/features/agents/agents.ts
wasnertobias: Fixed in 2eff2b7d9. The rendering split stays, and the polling gate counts every non-terminal session — `this.sessions().filter((s) => isActive(s.status))` — so a page showing only a queue keeps refreshing. Otherwise a start, a cancellation or a reorder would never have reached it.

.github/workflows/logos_deploy-dev.yml
```diff
@@ -56,7 +56,10 @@ jobs:
       - name: Create .env file
         run: |
           echo "ENVIRONMENT=${{ matrix.environment }}" > .env
-          echo "IMAGE_TAG=${{ inputs.image-tag }}" >> .env
+          # Through the environment and quoted: the tag comes from whoever
+          # dispatched the workflow, and a shell line built by substitution
+          # is a shell line they get to help write.
+          echo "IMAGE_TAG=$WORKSPACE_IMAGE_TAG" >> .env
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 2eff2b7d9 — validated before anything is written, in all three deployments. Not with a regular expression: `grep` is line-based, so `^…$` happily matches the first line of a value that carries a newline, which is the case worth catching. A `case` glob sees the whole value:

```sh
case "$WORKSPACE_IMAGE_TAG" in
  "" | *[!A-Za-z0-9._-]* | [!A-Za-z0-9_]* ) exit 1 ;;
esac
```

plus a length check. Verified against `latest`, `pr-889-abc123`, a value containing a newline, an empty one, a slash, a leading dot and a space.

logos/logos-agent/app/capacity.py
```diff
@@ -157,18 +221,37 @@ def parse_scheduler_state(payload: dict, lane: frozenset[tuple[str, str]] | None
...
             if wanted and (str(provider_id), str(model_id)) not in wanted:
+                # Still counted towards the queue: somebody waiting on a
+                # deployment we cannot reach is somebody waiting.
+                queue_total += waiting
                 continue
             total += capacity
-            busy += active
-            name = str(model.get("model_name") or model_id)
-            slots = per_model.setdefault(name, [0, 0])
+            slots = per_model.setdefault(name, [0, 0, 0, 0.0])
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 2eff2b7d9: the key is `name.strip().lower()`. On this platform the same model is reported by three providers, so a spelling difference between them would have split one lane into two pools and let "the busiest" mean one provider's view of a model. `test_a_case_difference_is_not_a_second_model` covers it.

logos/logos-agent/app/capacity.py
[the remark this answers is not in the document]
wasnertobias: This one is already gone, in 3ae562505: there is no adjusting a finished reading any more. The discount happens inside `parse_scheduler_state`, per model, while the figures are still separate, and a pass takes two readings of the same moment — the platform's, and the platform's minus this runner's share. Pause and resume use the second, admission the first.

`test_our_sessions_on_one_model_do_not_empty_another` is the regression case you asked for: nine user requests on model A, five runner sessions on model B, and the adjusted reading still says 90%.

logos/logos-agent/app/db.py
```diff
@@ -832,7 +836,111 @@ async def list_sessions(
...
+    a queue whose top row is already at the maximum has no room above it,
+    and a move that cannot change the order is a move that did not happen.
+    Only rows whose number actually changes are written.
+    """
+    count = len(order)
+    if count == 0:
+        return []
+    step = max(1, 100 // (count + 1))
+    for index, row in enumerate(order):
+        row["priority"] = max(0, min(100, 100 - (index + 1) * step))
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 2eff2b7d9, by refusing rather than pretending. Beyond a hundred queued rows two of them must share a number and the tie falls to `created_at` — which is exactly what a move overrules, so the move would silently not hold. `move_in_queue` raises, and the endpoint answers 400 with the reason.

A dedicated rank column is the better answer if the queue ever gets that long. It cannot today: a queued session occupies a workspace, and the workspace pool is the parallel ceiling. Worth doing when that stops being true rather than now.

logos/logos-agent/app/db.py
```diff
@@ -875,7 +985,73 @@ async def next_queued_session() -> dict[str, Any] | None:
...
+                       AND (
+                             s.trigger_ref IS NULL
+                             OR CAST(:trigger_quota AS INTEGER) IS NULL
+                             OR (
+                                  SELECT COUNT(*) FROM agent_sessions mine
+                                   WHERE mine.trigger_ref IS NOT NULL
+                                     AND mine.status = ANY(:occupying)
+                                ) < CAST(:trigger_quota AS INTEGER)
+                           )
+                     FOR UPDATE OF s SKIP LOCKED
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 2eff2b7d9. The claim takes `pg_advisory_xact_lock(hashtext('logos-agent-admission'))` before it counts, held to the end of the transaction that does the claiming, so two replicas serialise rather than both reading the same count in the same instant.

logos/logos-agent/app/sessions.py
```diff
@@ -1379,6 +1458,69 @@ async def _note_usage(self, session_id: int, lines: list[str]) -> None:
...
+        Distinguishes "the replacement failed to be created" from "this
+        request has had its three goes": the first is worth waiting for, the
+        second would keep a settled session owing an answer forever.
+        """
+        ref = str(session.get("trigger_ref") or "")
+        if not ref:
+            return False
+        try:
+            return await db.attempts_for_trigger(ref) < _MAX_ATTEMPTS_PER_REQUEST
+        except Exception:
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 2eff2b7d9 — a read failure returns `True` now, so the answer stays owed and a later sweep decides once the database can be asked. You are right that the two `False`s met: no attempts left *and* no replacement created, and the request went quietly.

logos/logos-agent/tests/test_db.py
```diff
@@ -188,3 +188,145 @@ def test_the_retry_bound_is_written_as_a_number(self):
...
+                return list(self._values)
+
+            def first(self):
+                return self._values[0] if self._values else None
+
+        class _Session:
+            async def execute(self, statement, params=None):
+                sql = str(statement)
+                if "ORDER BY priority DESC" in sql:
+                    return _Rows(sorted(state, key=lambda row: (-row["priority"], row["id"])))
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 2eff2b7d9: both helpers sort by `(-priority, created_at, id)`, the key the statement uses. `TestMovingWhenAgeAndIdDisagree` adds the inverse case — the older row carries the higher id — and asserts a move to the front holds against the age tie-break, plus the over-long queue being refused.

PR #858, opened 2026-08-29
`Logos`: Manage host RAM dynamically in the RAM cache and surface it in the statistics

wasnertobias:
## What

Host RAM is now managed dynamically the same way VRAM already is, and is displayed on the statistics page alongside VRAM.

1. **Workernode — re-plan of the RAM cache against live host RAM.** The tmpfs RAM cache was planned once at startup, a static snapshot: after boot, lanes sleep (weights move into host RAM), lanes stop (RAM frees up), and `MemAvailable` shifts, but the cache kept its boot-time share. The re-plan now re-runs the same candidate arithmetic as at startup (factored into `_build_ram_cache_candidates` so both call sites agree) on a 60 s tick *and* right after any lane goes to sleep — the reactive path, because lanes can sleep in quick succession and, with compose having no `mem_limit`, the host OOM killer would otherwise be the only other reactor (and it picks vLLM). Each pass re-derives the host-RAM floor from the plan — sleep reserve plus a safety margin that scales with host size (5% of total RAM, clamped to 8–32 GiB), excluding models that are already asleep since their RAM is already out of `MemAvailable` — reclaims models that no longer fit, and re-caches them in the background (after a short hold-down) when RAM frees up again. The floor itself is checked against live `MemAvailable` on every copy, so a lane that sleeps shrinks the cache's room immediately, between re-plans; the re-plan's own job is the reclaim. Models a lane with a live process still reads are never evicted — and when that protection is exactly what blocks every eviction, the re-plan says so in a warning instead of staying silent.
2. **Orchestrator — host RAM in the provider scheduler signals.** `host_ram_total_mb` / `host_ram_used_mb` / `host_ram_available_mb` are read from the worker runtime `host_memory` summary and emitted in the provider `scheduler_signals` — the established path that flows live (WS), persists in the snapshot JSONB, and passes through the Spring webservice unchanged. Without a summary the fields stay `null`, so "not reported" is distinguishable from a measured 0 MB.
3. **UI — host RAM displayed like VRAM.** A Host RAM card in the Workers & GPUs panel (used/total with its own bar, free amount in the footer) plus a cluster-wide used/total RAM line in the active-lanes KPI card. Both stay out when a worker does not report host RAM (older worker, non-Linux host).

## Why

#797: VRAM is managed dynamically but RAM is not — the memcache is statically configured, and sleeping models moving from VRAM to RAM need both sides handled; the host RAM shall be displayed like the VRAM in the statistics page.

[...]

PR #893, opened 2026-09-03
`Logos`: Let the agent lint, learn from red checks, and be re-instructed

wasnertobias:
Three things that came out of reading what `LogosOSSAgent` actually did today, and one that was asked for.

## It could not run the linters it was told to run

The house rules say to run the repository's hooks. The session image has no `black`, no `flake8`, no `pre-commit` — and the sandbox has no network, so it cannot install them, and `pre-commit` cannot fetch its hook environments either. The instruction was one the agent had no way to follow.

What that looked like in production: a session pushed a 111-line test to #858, `pre-commit` went red on `black`, and nothing about it reached anyone. The next session on that pull request happened to delete the file while answering a review — so the red check went away by accident.

The hook environments are installed into the image at build time now, from the repository's own `.pre-commit-config.yaml`, keyed by the same pinned revisions CI uses. `pre-commit run --files …` works offline, and both the prompt and the house rules say so.

## It never learned what its change did to the checks

A session ends minutes before its pull request's checks conclude, so the one verdict a person would notice within a minute of pushing was the one thing the agent could never see. The runner follows the checks of what a session pushed; a red one becomes another attempt at the same request, with the failure in its task and a reminder that the hooks now run in the sandbox. Bounded by the same three-attempts-per-request rule as every other retry, and only for work the runner queued itself — a session a person created is that person's to follow up.

## What every session is told can be changed without a release

The two standing blocks — the **house rules** and the **environment notes** — ship as defaults in code and are now overridable at runtime, edited on the agents page, in force from the next session.

- A cleared half goes back to what the code ships with.
- An empty half is an operator saying "say nothing here", which is a decision and is kept apart from "unset".
- A database that cannot be read keeps the last known text rather than silently reverting a session mid-flight to something else.

These are prompts, and prompts are the part of an unattended agent most worth adjusting after watching it work: every line in the defaults is there because a session went wrong without it.

## And each session shows the exact task it was given

Under every session, the full text that was handed over — the request, the house rules appended to it, and the environment notes the sandbox adds. A session that surprises somebody can now be read instead of guessed at.

---

464 tests green, pre-commit clean, UI builds. The migration is `022_agent_instructions.xml`: one row, two nullable columns, null meaning "use the default".

## Summary by CodeRabbit

* **New Features**
  * Added an Agents page panel for viewing and editing the house rules and environment notes shared with every session.
  * Added controls to save changes or restore either instruction set to its default.
  * Added expandable session details showing the exact task and environment notes provided.
  * Issue tasks now include relevant conversation context from the issue reporter and repository contributors.
  * Sessions retry work when submitted changes fail required checks or requests fail, including failure details.

* **Bug Fixes**
  * Sessions can run configured linting checks offline.
  * Timeout failures now include a clear explanation.
  * Sessions can continue until the agent finishes when no timeout is configured.

### DIFF
--- a/logos/logos-agent/README.md
+++ b/logos/logos-agent/README.md
@@ -515,6 +515,13 @@ an `Authorization` header, and the token is what authorises the read.
   shell profiles and build caches are executable configuration written by a
   session that held a push token, and the next session must not inherit
   them. Transcripts are data the same agent already authored.
+- **A request the runner could not finish comes back by itself.** A trigger
+  reference counts as handled the moment a session exists for it — that is
+  what keeps an assignment from producing a pull request a week — so a
+  failed session used to take its request with it, permanently invisible to
+  every later pass. A failure now takes the work up again with the reason
+  attached, bounded by the same three attempts as everything else, so a task
+  that cannot be done stops rather than loops.
 - **Work can be run again.** A failed session keeps its task, its workspace,
   its branch and the thread it came from, and *Run again* queues all of it as
   a new session. This is the only way back for work the runner took on
--- a/logos/logos-agent/README.md
+++ b/logos/logos-agent/README.md
@@ -187,6 +187,31 @@ survive a restart, and both are visible to whoever finds the runner stopped.
 | **Pause everything** | Running sessions are paused on the next pass and the capacity goes back to the platform. Nothing is cancelled: resuming picks the work up mid-task. |
 | **Sessions at once** | A ceiling for now, overriding the configured one. Zero drains without pausing. |
 
+## What the agent is told, and changing it
+
+Every task carries two standing blocks: the **house rules** — the conventions
+the agent works to — and the **environment notes**, which describe the
+container it works in. Both ship as defaults in code and can be edited on the
+page, taking effect on the next session rather than the next deployment.
+They are prompts, and prompts are the part of an unattended agent most worth
+adjusting after watching a few sessions: every line in the defaults is there
+because a session went wrong without it.
+
+Each session shows the exact text it was handed, so a surprising session can
+be read rather than guessed at.
+
+**It can run the repository's own hooks.** `pre-commit`'s hook environments
+are installed into the session image at build time, where there is a network
+to install them with — a session has none, and without this the instruction
+to lint before finishing was one the agent could not follow. Sessions were
+pushing unformatted code and learning nothing about it.
+
+**And it finds out when its change turned the checks red.** A session ends
+minutes before its pull request's checks conclude, so it never saw the one
+verdict a person would notice immediately. The runner follows the checks of
+what the session pushed, and a red one becomes another attempt with the
+failure in its task.
+
 ## What gets worked on first
 
 An operator can overrule it. The queue on the page carries three controls on
--- a/logos/logos-agent/README.md
+++ b/logos/logos-agent/README.md
@@ -250,7 +250,7 @@ has a default that is right for this deployment.
 | `LOGOS_AGENT_PAUSE_ABOVE_LOAD` | `0.85` | Pause at or above this load |
 | `LOGOS_AGENT_SESSION_MEMORY_MB` | `4096` | Per-session memory ceiling |
 | `LOGOS_AGENT_SESSION_CPUS` | `2` | Per-session CPU ceiling |
-| `LOGOS_AGENT_SESSION_TIMEOUT_S` | `10800` | Wall-clock ceiling per session (paused time does not count) |
+| `LOGOS_AGENT_SESSION_TIMEOUT_S` | `0` | Wall-clock ceiling per session; `0` is none, which is the default |
 | `LOGOS_AGENT_SESSION_MODEL_URL` | `http://logos-agent-gateway` | Where sessions send model traffic — a gateway that exposes only the orchestrator's `/v1` model surface, so a session never reaches the rest of the internal network |
 | `LOGOS_AGENT_SESSION_GITHUB_TOKEN` | falls back to the token above | Given to containers; best without `workflow` scope |
 | `LOGOS_AGENT_DEPLOY_ENABLED` | `false` | Whether dev deploys may be dispatched at all |
@@ -490,8 +490,17 @@ an `Authorization` header, and the token is what authorises the read.
   scheduler resumes them mid-task. `stop_grace_period` is 30 s for this, and
   whatever cannot be frozen in time simply runs through the deploy as it did
   before.
-- **A stuck session is capped**, not left to burn capacity — `SESSION_TIMEOUT_S`
-  stops it and records the reason.
+- **A session ends when it is done, not when a clock says so.** There is no
+  wall-clock limit unless a deployment sets one. A clock is the wrong
+  instrument: a session that has read the repository for two hours and is
+  halfway through a change is not stuck, and stopping it throws away
+  everything it has done — uncommitted, in a checkout the next session
+  resets. One went that way: ninety minutes and thirty-eight million tokens,
+  killed at the deadline with nothing to show. What this runner protects is
+  capacity, and capacity is protected by the things that measure it — the
+  pause, the parallel ceiling, the trigger quota — none of which care how
+  long a session has been at it. A session that really is stuck is visible
+  on the page and can be cancelled there.
 - **Yielding does not cost the work.** Freezing a session and cutting it off
   the model network — which is how capacity is handed back — ends the answer
   it was reading, and the agent meets a dead connection when it thaws. That
--- a/logos/logos-agent/README.md
+++ b/logos/logos-agent/README.md
@@ -4,7 +4,7 @@ Runs coding agents in isolated containers, on serving capacity Logos is not
 otherwise using, and gives that capacity back the moment a user needs it.
 
 A session is one agent run: it gets a working copy, a task, and a Logos key.
-It works unattended, and what it produces arrives as a draft pull request that
+It works unattended, and what it produces arrives as a pull request that
 a human reviews like any other. Sessions can be told to deploy their result to
 the **dev** environment and screenshot the pages they changed.
 
@@ -160,7 +160,7 @@ session's behaviour drift between builds.
 ## What a session may and may not do
 
 **May:** read and change the working copy, run tests and linters, push a branch
-under `logos/agent/`, open a draft pull request, and — if enabled — have its result
+under `logos/agent/`, open a pull request, and — if enabled — have its result
 deployed to dev and screenshotted.
 
 **May not:** reach the Docker daemon, run as root, push to `main` or any
@@ -324,7 +324,7 @@ use it the way you use a person's:
 
 | What you do | What it does |
 |---|---|
-| assign it an issue | works on it and opens a draft pull request |
+| assign it an issue | works on it and opens a pull request |
 | assign it a pull request | takes it over, on that pull request's own branch |
 | request changes on one of its pull requests | addresses that review on its branch |
 | comment on a pull request it is responsible for | reads the thread and answers (within a day of writing) |
@@ -536,7 +536,7 @@ an `Authorization` header, and the token is what authorises the read.
   a new session. This is the only way back for work the runner took on
   itself: a trigger counts as handled the moment a session exists for it, so
   no later pass finds that issue, review or question again.
-- **Nothing merges itself.** Pull requests are opened as drafts, and a person
+- **Nothing merges itself.** A person
   approves and merges exactly as they do for human work.
 - **Screenshots follow the deploy.** A session that asks for dev screenshots
   gets them only after the runner has dispatched its dev deploy and watched
--- a/logos/logos-agent/README.md
+++ b/logos/logos-agent/README.md
@@ -330,6 +330,21 @@ use it the way you use a person's:
 | comment on a pull request it is responsible for | reads the thread and answers (within a day of writing) |
 | mention it anywhere by name | answers there (within a day); changes code only if that is what was asked |
 
+**Whose word counts.** A session pushes branches and answers in this
+repository's name, so what may direct it is a question about people:
+membership of the `logos-developers` or `logos-maintainers` team
+(`LOGOS_AGENT_TRUSTED_TEAMS`). Where the runner cannot ask — a token without
+`read:org` — it falls back to the coarser rule of write permission on the
+repository. Comments from anybody else are left out of the task and the
+omission is disclosed, the reporter's own included: anybody can open an
+issue on a public repository, and a maintainer can repeat what matters in
+their own words.
+
+**It does not take over its own work.** This repository assigns every pull
+request to its author, so one the agent opens comes back a moment later as
+one assigned to it. Reviews and questions on it still reach the agent; a
+handover of what it has just written does not.
+
 No labels and no separate vocabulary — the ordinary gestures. **Consent is
 per item:** nothing is picked up because it exists, only because somebody
 assigned it, reviewed its work, or asked it something by name. That is why
--- a/logos/logos-agent/README.md
+++ b/logos/logos-agent/README.md
@@ -340,6 +340,16 @@ omission is disclosed, the reporter's own included: anybody can open an
 issue on a public repository, and a maintainer can repeat what matters in
 their own words.
 
+One narrow exception, and only to *reading*: the review apps this
+repository runs on its own pull requests (`LOGOS_AGENT_REVIEW_BOTS`,
+`coderabbitai[bot]` and `Claudia-Anthropica` by default). What they wrote
+travels with a task somebody trusted has already directed, because a
+handover exists to answer a review and dropping the review left the agent
+reconstructing one from the diff — on production, every handover was losing
+between six and seventeen comments that way. They direct nothing: no review
+of theirs starts a session and no comment of theirs steers one. Setting the
+variable to nothing removes the exception.
+
 **It does not take over its own work.** This repository assigns every pull
 request to its author, so one the agent opens comes back a moment later as
 one assigned to it. Reviews and questions on it still reach the agent; a
--- a/logos/logos-agent/README.md
+++ b/logos/logos-agent/README.md
@@ -385,10 +385,13 @@ hourglass, so these three are the states it can show.
 whole report in the first comment is an ordinary way to file one, and a
 session handed the title alone can only say so — which is exactly what
 happened on an issue whose description was a maintainer's comment. The
-comments travel with the task now: the people who may write to the
-repository, and the person who opened it, who is usually the one with the
-details. A passer-by on a public issue is left out and counted, because the
-session that reads this will push a branch.
+comments travel with the task now — the ones from the trusted teams
+(`LOGOS_AGENT_TRUSTED_TEAMS`, `logos-developers` and `logos-maintainers` by
+default). Everybody else is left out and counted, the person who opened the
+issue included: anybody can open one, and the session that reads it will
+push a branch. Where the teams cannot be read at all — a token without
+`read:org` — the coarser rule stands in for them: whoever may write to the
+repository.
 
 **It can see what you attached.** An issue whose whole description is a
 screenshot is unreadable to a sandbox with no network — the agent met one of
--- a/logos/logos-agent/README.md
+++ b/logos/logos-agent/README.md
@@ -366,6 +366,15 @@ that is not queued; and because the row is what the queue is, a restart
 changes nothing about it. GitHub's reaction palette is fixed and has no
 hourglass, so these three are the states it can show.
 
+**An issue is its thread, not its body.** A title, an empty body and the
+whole report in the first comment is an ordinary way to file one, and a
+session handed the title alone can only say so — which is exactly what
+happened on an issue whose description was a maintainer's comment. The
+comments travel with the task now: the people who may write to the
+repository, and the person who opened it, who is usually the one with the
+details. A passer-by on a public issue is left out and counted, because the
+session that reads this will push a branch.
+
 **It can see what you attached.** An issue whose whole description is a
 screenshot is unreadable to a sandbox with no network — the agent met one of
 those with `WebFetch`, was refused, read code for an hour and changed
--- a/logos/logos-agent/app/conventions.py
+++ b/logos/logos-agent/app/conventions.py
@@ -98,10 +98,23 @@ your assumptions in the final summary.
 - Do not run git commit, git push, or gh: the harness commits and opens the
   pull request for you after you finish.
 - Run the project's tests for the code you touch, and fix what you break.
-- Run `pre-commit run --files <the files you changed>` before you finish and
-  fix what it reports. The hooks are installed in this image, so it works
-  without a network; CI runs the same ones, and a session that skips this
-  hands somebody a red pull request.
+- Linting is on you, and you can run it: from the top of the checkout,
+  `pre-commit run --files <the files you changed>`. That is the same command
+  CI runs and the same pinned hook versions — for every service, not only
+  Logos. They are installed in this image, so it works with no network.
+  Some of it reformats in place (black, isort, autoflake, end-of-file-fixer):
+  a hook that reports "files were modified by this hook" has already fixed
+  it, so run the command again and expect it to pass the second time. What
+  flake8 reports you fix yourself.
+  To chase one hook rather than all of them, name it:
+  `pre-commit run flake8 --files <files>`. To see every hook a file goes
+  through, `pre-commit run --files <file> --verbose`.
+  Two hooks cannot work in here and are not your fault: `pylint` and `mypy`
+  under iris/ and memiris/ run through `poetry`, and there is no network to
+  install those environments with. If one of those is what failed, say so in
+  your summary and move on.
+  A session that skips linting hands somebody a red pull request over a
+  blank line, so do not skip it.
 - If the task turns out to be impossible or already done, say so plainly
   instead of inventing changes.
 """.strip()
@@ -125,6 +138,11 @@ DEFAULTS = Instructions()
 
 _cached: Instructions = DEFAULTS
 _read_at: float = 0.0
+# Bumped whenever the stored text changes. A read that was already in
+# flight when somebody saved has read the old row, and must not put it back
+# in the cache: without this the page shows the new text, the next five
+# seconds of sessions are built from the old one, and nothing looks wrong.
+_generation: int = 0
 # Long enough that building a task costs no query, short enough that an
 # edit reaches the next session rather than the next restart.
 _CACHE_S = 5.0
@@ -136,13 +154,20 @@ async def current() -> Instructions:
     now = time.monotonic()
     if now - _read_at < _CACHE_S:
         return _cached
+    started_at = _generation
     try:
         row = await db.get_instructions()
     except Exception as exc:
         logger.warning("could not read the agent instructions; using the last known text: %s", exc)
         return _cached
+    fresh = _from_row(row)
+    if started_at != _generation:
+        # Overtaken while we were reading. What we have is the text as it
+        # was before the save, so it is returned to this caller and thrown
+        # away rather than cached.
+        return fresh
     _read_at = now
-    _cached = _from_row(row)
+    _cached = fresh
     return _cached
 
 
@@ -166,10 +191,16 @@ def _from_row(row: dict[str, Any] | None) -> Instructions:
     )
 
 
-async def set_instructions(*, house_rules: str | None, environment_notes: str | None, by: str) -> Instructions:
-    """Store an override, or clear one by passing None."""
-    global _read_at
+async def set_instructions(
+    *,
+    house_rules: str | None | db.Unchanged = db.UNCHANGED,
+    environment_notes: str | None | db.Unchanged = db.UNCHANGED,
+    by: str,
+) -> Instructions:
+    """Store an override, clear one by passing None, keep one by leaving it out."""
+    global _generation, _read_at
     await db.set_instructions(house_rules=house_rules, environment_notes=environment_notes, updated_by=by)
+    _generation += 1
     _read_at = 0.0
     return await current()
 
--- a/logos/logos-agent/app/conventions.py
+++ b/logos/logos-agent/app/conventions.py
@@ -11,6 +11,15 @@ none of them can quietly drift from it.
 
 from __future__ import annotations
 
+import logging
+import time
+from dataclasses import dataclass
+from typing import Any
+
+from . import db
+
+logger = logging.getLogger(__name__)
+
 # Conventions every session is held to, whatever kind of work it is.
 HOUSE_RULES = """
 --- How work is done here ---
@@ -35,8 +44,11 @@ Verifying before you claim:
   addressed already — say so instead of changing it twice.
 - A regression test that passes before your fix is not a regression test.
   Confirm it fails on the unfixed code, then make it pass.
-- Run the tests and linters of the part of the repository you touched, and
-  the repository's pre-commit hooks. Report what you ran.
+- Run the tests of the part of the repository you touched, and
+  `pre-commit run --files <what you changed>`. Both work offline: the hook
+  environments are installed in this image. CI runs the same hooks, so
+  skipping them means handing somebody a red pull request. Report what you
+  ran.
 - Before you finish, check that the pull request's checks are green and that
   it is mergeable. If a check is red for a reason your change did not cause,
   say which and why rather than leaving it unexplained.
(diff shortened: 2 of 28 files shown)


logos/logos-agent/workspace/Dockerfile
```diff
@@ -75,6 +75,22 @@ RUN npm install -g \
...
+# first time, `pip install black` reaches PyPI, and the sandbox reaches
+# neither. Sessions were pushing unformatted code and finding out from a red
+# check nobody told them about.
+#
+# The environments are keyed by hook repository and revision, so they are
+# reused as long as the pinned revisions in the config match. When they stop
+# matching, the session's `pre-commit run` says so rather than silently
+# skipping — which is why the config is copied in and resolved here, not
+# guessed.
+ENV PRE_COMMIT_HOME=/opt/pre-commit
```
Claudia-Anthropica: @wasnertobias [high] `PRE_COMMIT_HOME` points to `/opt/pre-commit`, but session containers run as `agent` with `ReadonlyRootfs=true`. Pre-commit writes its store lock and configuration-usage database during a run, so the advertised offline command will fail despite the hook environments being installed; seed those environments into a writable per-session location and point `PRE_COMMIT_HOME` there.

In `logos/logos-agent/workspace/Dockerfile`, the pre-commit cache is placed on the session container's read-only root filesystem even though pre-commit writes to its store while running. Copy the seeded cache to a writable per-session location and set `PRE_COMMIT_HOME` to that location.

</details>

logos/logos-agent/app/sessions.py
```diff
@@ -1493,7 +1505,46 @@ async def _may_try_again(session: dict[str, Any]) -> bool:
...
+        """
+        branch = str(session.get("branch_name") or "")
+        if not branch or not session.get("trigger_ref"):
+            # A session queued by a person is that person's to follow up.
+            return
+        asyncio.create_task(self._take_up_a_red_build(session, branch, pushed_sha))
+
+    async def _take_up_a_red_build(self, session: dict[str, Any], branch: str, pushed_sha: str) -> None:
+        try:
+            status, detail = await github.wait_for_pr_builds(branch, pushed_sha)
```
Claudia-Anthropica: @wasnertobias [high] `wait_for_pr_builds()` watches only `settings.build_workflow`, so it can return success while the separate Logos Test or Logos Lint workflow is red—the exact pre-commit failure this feature is intended to retry. It also turns a polling timeout into a retry although no failed check was observed; inspect all relevant checks for the exact SHA, wait until they are terminal, and retry only on a concrete failure.

In `logos/logos-agent/app/sessions.py`, the failed-check retry path observes only the image-build workflow and treats timeouts as failures. Query all relevant checks for the pushed SHA and queue another attempt only after at least one check has conclusively failed.

</details>

logos/logos-agent/workspace/run_session.py
```diff
@@ -486,6 +486,12 @@ def build_prompt(task: str) -> str:
             "Look at them before you decide anything — on a visual report they "
             "are usually the whole description.\n"
         )
+    # What this container is, as the runner describes it — an operator can
+    # adjust that text, and the page shows exactly what was handed over. The
+    # text below is the fallback for a session started by an older runner.
+    notes = os.environ.get("LOGOS_SESSION_ENVIRONMENT_NOTES", "").strip()
```
Claudia-Anthropica: @wasnertobias [medium] An explicitly empty environment-notes override becomes indistinguishable from an older runner that omitted the variable: `.strip()` yields an empty string and this falls through to the hard-coded default. That violates the documented empty-versus-unset behavior; branch on whether the environment variable is present, allowing a present empty value to append nothing.

In `logos/logos-agent/workspace/run_session.py`, an explicitly empty `LOGOS_SESSION_ENVIRONMENT_NOTES` value incorrectly restores the fallback notes. Distinguish an absent variable from a present empty variable and append no notes for the latter.

</details>

logos/logos-ui/src/app/features/agents/agents.html
```diff
@@ -466,6 +547,18 @@ <h2 class="section-title">Finished</h2>
             <div class="alert alert--error" role="alert">{{ session.error }}</div>
           }
 
+          <details class="prompt">
+            <summary class="prompt__summary">The exact task this session was given</summary>
+            <pre class="prompt__text">{{ session.task }}</pre>
+            @if (instructions(); as text) {
+              <pre class="prompt__text">{{ text.environment_notes }}</pre>
```
Claudia-Anthropica, line 560: @wasnertobias [medium] This renders the current global environment notes for every historical session, not the notes that the selected session received. After an operator edits the setting, old sessions therefore display different text under “The exact task this session was given”; persist the effective notes with each session, expose that snapshot in `SessionSummary`, and render the session field here.

In `logos/logos-ui/src/app/features/agents/agents.html`, historical sessions display the current global environment notes instead of their original prompt. Persist and return each session's effective environment-notes snapshot and render that session-specific value here.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The captured CI checks pass, but instruction delivery and failed-check retry still have blocking correctness gaps. [medium] Sessions created through `POST /sessions` persist `body.task` directly, while only trigger-built tasks call `conventions.for_task()`, so editable house rules never reach sessions launched manually from the Agents page; apply the effective rules in that creation path and add coverage.

logos/logos-agent/app/github.py
```diff
@@ -543,6 +543,19 @@ async def open_pull_request_for(branch: str) -> dict[str, Any] | None:
...
+    """What has been said under an issue, oldest last, and what is missing.
+
+    An issue's description is not always in its body. A title and an empty
+    body with the whole report in the first comment is an ordinary way to
+    file one — and a session handed the title alone can only say that it
+    was handed the title alone.
+    """
+    answer = await _get_all_bounded(f"/repos/{settings.repo_slug}/issues/{number}/comments")
+    entries, missing = _as_conversation({"comments": answer}, number)
+    return entries[-limit:], missing
```
Claudia-Anthropica: @wasnertobias [medium] This slice silently drops every comment before the newest 40 without adding the omission to `missing`, yet `issue_task()` subsequently says this is the whole issue. On a long issue, the first reporter comment—the description this feature is intended to recover—can therefore disappear, and authorization filtering afterward can reduce the retained context further. Return the bounded normalized entries, filter permitted authors first, and then apply the task limit while recording the omitted count.

In `logos/logos-agent/app/github.py`, `issue_conversation()` silently truncates older comments before author filtering and does not report the omission. Return the normalized entries for filtering, then apply the 40-comment task limit after filtering and append the omitted count to `missing`.

</details>

logos/logos-agent/workspace/Dockerfile
```diff
@@ -75,6 +75,22 @@ RUN npm install -g \
...
+COPY logos/.pre-commit-config.yaml /tmp/hooks/.pre-commit-config.yaml
```
Claudia-Anthropica: @wasnertobias [high] Only `logos/.pre-commit-config.yaml` is copied and installed here, but agent sessions operate across the entire monorepo and are instructed to run the root `pre-commit run --files` command. For Iris, Athena, or Memiris changes, the root `ddanier/sub-pre-commit` environment and the corresponding service hooks are absent, so pre-commit must fetch them over the unavailable network and the advertised offline check still fails. Seed the root configuration and every transitively invoked service configuration/environment needed by its hooks.

In `logos/logos-agent/workspace/Dockerfile`, only the Logos pre-commit environments are installed even though the workspace runs repository-root hooks for all services. Include the root and delegated service configurations in the build context and preinstall every hook environment that the root configuration can invoke offline.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The latest commit improves issue context, but the current head still contains the previously identified instruction-delivery, prompt-snapshot, read-only pre-commit-cache, and check-retry defects: `POST /sessions` still stores `body.task` directly, and `wait_for_pr_builds()` remains scoped to the build workflow. The issue-comment path also silently drops older comments at its local limit, while the workspace image seeds only Logos hooks despite serving the whole monorepo. Several captured checks remain in progress, but none is marked required.

Claudia-Anthropica (changes requested):
@wasnertobias The new timeout-reason change is sound, but it does not touch the six unresolved Claudia threads or the prior manual-session instruction-delivery finding. Current source still exhibits those defects: `_take_up_a_red_build()` polls only `settings.build_workflow` and retries `timeout`, while the session image points writable pre-commit state at `/opt/pre-commit` under `ReadonlyRootfs=true`. The snapshot has no failed or required checks, though several non-required checks remain pending.

logos/logos-agent/app/sessions.py
```diff
@@ -1673,6 +1732,31 @@ async def _settle(self, session_id: int, *, exit_code: int | None, error: str |
...
             # too — an answer that never comes is the worst of the three.
             await self._react(session_row, github.REACTION_FAILED)
+            if session_row.get("trigger_ref"):
+                # A request the runner took on and could not finish is a
+                # request nobody is coming back to: the reference counts as
+                # handled forever, so no later pass finds it again, and the
+                # only way back was somebody noticing. Taken up again here,
+                # bounded by the same three attempts as everything else, so
+                # a task that cannot be done stops rather than loops.
+                await self.take_up_again(
```
Claudia-Anthropica: @wasnertobias [medium] `take_up_again()` returns `None` when either the attempt-count query or replacement insert fails, but this call ignores that result. Once the failed session had actually started, `handled_trigger_refs()` keeps its trigger consumed, and `_post_reply()` now abandons any owed reply for a failed row, so a single transient database failure leaves no scheduler path that retries the request. Persist retry intent and have startup or scheduler reconciliation retry it until a replacement row is committed or the attempt limit is reached.

In `logos/logos-agent/app/sessions.py`, the failed-session settlement ignores a transient failure to create the replacement, after which an already-started trigger remains handled and is never retried. Persist the retry intent and make startup or scheduler reconciliation retry it until a replacement row is committed or the three-attempt limit is reached.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The newest commit queues a replacement on the normal path, but the six unresolved Claudia threads and the prior manual-session instruction-delivery finding are unchanged on this head; for example, `POST /sessions` still stores `body.task` directly and `wait_for_pr_builds()` still polls only `settings.build_workflow`. The new settlement path also ignores a failed replacement creation, which permanently loses an already-started trigger after a transient database error; see the inline finding. The snapshot contains no failed or required checks, although several non-required checks were still pending.

Claudia-Anthropica (changes requested):
@wasnertobias The newest commit only changes the optional session timeout; it does not address the outstanding current-head blockers. [high] The check watcher remains in-memory, observes only `settings.build_workflow`, retries timeouts, and the image still provides a read-only `/opt/pre-commit` cache seeded only for Logos. [medium] Manual creation still stores `body.task` without effective rules, empty environment notes still restore defaults, historical sessions render current notes, issue comments are truncated before filtering without omission accounting, and failed settlement still ignores a `None` replacement before abandoning its reply. The snapshot contains no failed or required checks, although several non-required checks remain pending.

Claudia-Anthropica (changes requested):
@wasnertobias The new head only changes agent-created pull requests from draft/generated-body to ready/title-only and leaves the existing unresolved correctness threads untouched. Current source still stores manual session `body.task` verbatim, keeps check follow-up only in an in-memory task that observes the build workflow and retries timeouts, and runs pre-commit with a read-only, Logos-only cache; the prompt snapshot, issue truncation, and failed-replacement paths are likewise unchanged. The captured CI snapshot has no failed or required checks, although several non-required checks remain pending.

logos/logos-agent/app/github.py
```diff
@@ -731,6 +756,37 @@ async def may_push(login: str) -> bool:
...
+    """
+    if not login or not settings.trusted_teams:
+        return None
+    org = settings.repo_slug.split("/", 1)[0]
+    answered = False
+    for team in settings.trusted_teams:
+        try:
+            membership = await _get(f"/orgs/{org}/teams/{team}/memberships/{login}")
+        except Exception as exc:
+            if getattr(exc, "status", None) == 404:
```
Claudia-Anthropica: @wasnertobias [medium] `_get()` raises a plain `GitHubError` without a `status` attribute for every non-200 response. GitHub reports a user who is not in a team with 404, so this branch never marks `answered`; after all configured teams the function returns `None`, and `_writer()` falls back to `may_push()`, allowing write collaborators outside the trusted teams to direct sessions. Probe team visibility separately to distinguish a missing scope, treat a membership 404 from a visible team as `False`, and add coverage using the actual `_get()` error behavior.

In `logos/logos-agent/app/github.py`, non-member 404s lose their HTTP status in `GitHubError`, causing `in_a_trusted_team()` to return `None` and fall back to repository write permission. Preserve and classify response status, establish team visibility separately so missing scope remains unknown, and return `False` for confirmed non-membership.

</details>

logos/logos-agent/README.md
```diff
@@ -341,6 +381,15 @@ that is not queued; and because the row is what the queue is, a restart
 changes nothing about it. GitHub's reaction palette is fixed and has no
 hourglass, so these three are the states it can show.
 
+**An issue is its thread, not its body.** A title, an empty body and the
+whole report in the first comment is an ordinary way to file one, and a
+session handed the title alone can only say so — which is exactly what
+happened on an issue whose description was a maintainer's comment. The
+comments travel with the task now: the people who may write to the
```
Claudia-Anthropica: @wasnertobias [low] This section says the issue reporter’s comments travel with the task even when the reporter lacks write access, but the new policy above and `_issue_conversation()` now exclude the reporter unless they are trusted. Rewrite this paragraph so the operator documentation consistently describes the team-based rule.

In `logos/logos-agent/README.md`, the issue-conversation section contradicts the new team-based trust policy by saying the reporter is included independently. Update the paragraph to state that only comments from trusted accounts are included and that excluded reporter comments are disclosed.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The newest team gate has an authorization fallback bug: normal non-member 404s lose their status and become “unknown,” so write collaborators outside the configured teams can still direct sessions. The seven unresolved Claudia threads and the prior manual-session instruction-delivery finding also remain present on this head—for example, `POST /sessions` still stores `body.task` verbatim, while the check watcher remains an in-memory task that observes only the build workflow. The captured snapshot has no failed or required checks; its pending checks are non-required.

logos/logos-agent/app/conventions.py
```diff
@@ -73,6 +85,102 @@
...
+    now = time.monotonic()
+    if now - _read_at < _CACHE_S:
+        return _cached
+    try:
+        row = await db.get_instructions()
+    except Exception as exc:
+        logger.warning("could not read the agent instructions; using the last known text: %s", exc)
+        return _cached
+    _read_at = now
+    _cached = _from_row(row)
```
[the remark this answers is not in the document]
wasnertobias: Fixed in `e514ec0b3`.

The cache now carries a generation counter. `set_instructions` bumps it, and a
read that finishes afterwards returns its row to its own caller and throws it
away instead of caching it. Without that the page showed the new text while the
next five seconds of sessions were built from the old one — and nothing
anywhere said so.

Covered by `TestAReadThatWasOvertaken`, which holds a read open across a save.

logos/logos-agent/app/sessions.py
```diff
@@ -1493,7 +1505,46 @@ async def _may_try_again(session: dict[str, Any]) -> bool:
...
+
+        Watched in the background because it takes minutes: the session is
+        settled, its workspace is free, and what comes of this is a fresh
+        attempt with the failure in its task.
+        """
+        branch = str(session.get("branch_name") or "")
+        if not branch or not session.get("trigger_ref"):
+            # A session queued by a person is that person's to follow up.
+            return
+        asyncio.create_task(self._take_up_a_red_build(session, branch, pushed_sha))
```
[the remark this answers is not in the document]
wasnertobias: Fixed in `e514ec0b3`.

The intent is on the row now: `agent_sessions.checks_sha` and `checks_watch`
(`pending` / `done`, migration `022-agent-session-check-watch`). It is written
*before* anybody starts watching, so a redeploy in those two minutes cannot lose
it. `resume_check_watches()` picks the pending rows up on startup and on every
scheduler pass; `_start_watching` keeps one poller per session, and a follow-up
older than six hours is closed rather than asked about forever.

`TestTakingUpAWatchAgain` covers the restart, the row with nothing to watch, the
horizon, an unreadable database, and the double-start.

logos/logos-agent/app/sessions.py
```diff
@@ -1493,7 +1505,46 @@ async def _may_try_again(session: dict[str, Any]) -> bool:
...
+        logger.info("the checks of session %s did not pass (%s); taking the work up again", session.get("id"), status)
+        await self.take_up_again(
+            session,
+            note=(
+                f"Your last change to `{branch}` is on the pull request, and its checks did not pass "
+                f"({status}). This is that same piece of work, one round later:\n\n> {detail[:1500]}\n\n"
+                "Fix what the checks are complaining about. `pre-commit run --files <what you changed>` "
+                "runs the same hooks CI does, and it works in here."
+            ),
+        )
```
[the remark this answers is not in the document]
wasnertobias: Fixed in `e514ec0b3`.

`timeout` now returns without queueing anything and without closing the watch:
the checks may still be running, or may never have been queued, and neither is a
failure to fix. The row stays `pending`, so the next pass asks again. Only a
concrete `failure` / `timed_out` / `startup_failure` conclusion takes the work up
again — `cancelled`, `skipped`, `neutral` and `action_required` (a deploy waiting
for a human) do not.

`test_checks_that_have_not_concluded_are_not_a_failure` pins it.

logos/logos-agent/workspace/Dockerfile
```diff
@@ -75,6 +75,22 @@ RUN npm install -g \
...
+RUN uv tool install --python 3.13 pre-commit   && ln -sf /root/.local/bin/pre-commit /usr/local/bin/pre-commit   && cd /tmp/hooks   && git init -q .   && git config user.email agent@logos.invalid   && git config user.name agent   && pre-commit install-hooks   && chmod -R a+rX /opt/pre-commit /root/.local   && chmod a+rx /root   && rm -rf /tmp/hooks
```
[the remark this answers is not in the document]
wasnertobias: Fixed in `e514ec0b3`.

`uv venv --python 3.13 /opt/venv` + `uv pip install --python /opt/venv`, with
`/opt/venv/bin` on the PATH — the same shared layout as the other Logos images,
instead of a tool install under `/root/.local` that needed `chmod a+rx /root` to
be reachable by the unprivileged session user.

Deliberately no `VIRTUAL_ENV`: this venv holds the linters, and a session
building the project makes its own — pointing `uv` at this one would only
confuse it.

logos/logos-agent/workspace/run_session.py
```diff
@@ -486,6 +486,12 @@ def build_prompt(task: str) -> str:
...
+    if notes:
+        return f"{task}{pictures}\n\n{notes}\n"
```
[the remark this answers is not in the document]
wasnertobias: Fixed in `e514ec0b3`.

The branch is on presence now, not truthiness:

```python
if "LOGOS_SESSION_ENVIRONMENT_NOTES" in os.environ:
    notes = os.environ["LOGOS_SESSION_ENVIRONMENT_NOTES"].strip()
    return f"{task}{pictures}\n\n{notes}\n" if notes else f"{task}{pictures}\n"
```

An operator who empties the box has decided nothing should be said there;
answering that with a page of defaults ignores the decision. The fallback is
still there for a session started by a runner too old to send the variable at
all. `TestWhatTheAgentIsTold` covers all three cases, and that the attached
images are named either way.

logos/logos-ui/src/app/features/agents/agents.html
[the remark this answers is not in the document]
wasnertobias: They are already in that block — and that is the part worth being precise
about.

`conventions.for_task()` composes the house rules into the task *before* the row
is written (`triggers.py`), so `session.task` is the task **and** the rules as
they stood for that session. Rendering `text.house_rules` next to it would show
today's rules beside a historical task, which is the same mistake the
environment notes were making.

What was genuinely missing was the other half, and that is fixed in `e514ec0b3`:
the environment notes are stored per session (`agent_sessions.environment_notes`,
written at launch) and the detail renders the session's own copy, falling back to
the current text only for rows from before the column existed — and saying which
of the two it is showing.

logos/logos-ui/src/app/features/agents/agents.ts
```diff
@@ -381,6 +387,69 @@ export class Agents implements OnInit {
...
+      });
+      this.instructions.set(fresh);
+      this.houseRulesDraft.set(fresh.house_rules);
+      this.environmentNotesDraft.set(fresh.environment_notes);
+    } catch (err: unknown) {
+      this.error.set(this.messageOf(err, 'Could not restore the default text.'));
+    } finally {
+      this.savingInstructions.set(false);
+    }
+  }
```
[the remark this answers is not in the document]
wasnertobias: Fixed in `e514ec0b3`, on both sides.

The page now sends `{ reset_house_rules: true }` and nothing else, and refills
only that box. And the API no longer requires both halves: an absent half is left
as stored (`db.UNCHANGED`, a `CASE WHEN` per column in the upsert), so even a
client that sends the other one cannot save it by accident. Null still means
"back to the shipped default" and an empty string still means "say nothing here".

Covered by `resets one half without submitting the other half's draft` in
`agents.spec.ts` and `test_a_half_nobody_mentioned_is_left_alone` in
`test_app.py`.

logos/logos-agent/workspace/Dockerfile
wasnertobias: Correct, and it did fail exactly like that. Fixed in `e514ec0b3`.

The store moved to the session's own home (`PRE_COMMIT_HOME=/workspace/.home/pre-commit`,
on the writable volume). The seeded store stays at `/opt/pre-commit` and stays
read-only; at session start `_seed_hook_store()` copies **only** `db.db` into the
writable one. The rows in it hold the paths of the installed environments, which
pre-commit reads and executes and never writes to — so nothing heavy is copied
and no session can poison what the next one gets.

Verified rather than reasoned about: built the image, ran a container with
`--read-only --network none`, and ran `pre-commit run --config
logos/.pre-commit-config.yaml --files …` in it. Hooks run, black/isort/autoflake
reformat in place, the second run is green.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in `e514ec0b3`.

New `github.wait_for_checks(head_sha)` reads
`/repos/{slug}/commits/{sha}/check-runs` — every check on the commit, which is
where Logos Test and Logos Lint actually are. `wait_for_pr_builds` stays, but
only for the dev-deploy gate, which genuinely needs that one image.

And the timeout is no longer a failure: it returns without queueing and leaves
the follow-up owed, so a poll that ran out of time is asked again rather than
answered with a retry. `TestWatchingEveryCheck` covers a red lint beside a green
build, `action_required`, no checks yet, and an unreadable answer.

logos/logos-agent/workspace/run_session.py
wasnertobias: Fixed in `e514ec0b3` — the test is now on presence of the variable rather than
on the text being non-empty, so an operator who empties the box gets a task with
no notes block instead of the built-in fallback. `TestWhatTheAgentIsTold` pins
the three cases (text, empty, unset).

logos/logos-ui/src/app/features/agents/agents.html
wasnertobias: Fixed in `e514ec0b3`.

`agent_sessions.environment_notes` (migration `022-agent-session-environment-notes`)
holds what the session was handed, written at launch beside the container's
environment. The detail renders the session's own copy and says it is kept as it
was on the day. A row from before the column existed shows the current text with
a hint that says so, rather than passing it off as that session's.

The task half needs no column: the house rules are composed into the task before
it is stored, so `session.task` already is the task of its time.

logos/logos-agent/app/github.py
wasnertobias: Fixed in `e514ec0b3`.

`issue_conversation()` returns everything it read; the cut moved to the caller
and now happens **after** the author filter, in `triggers._issue_conversation()`.
Forty comments from passers-by can no longer push out the one from a maintainer
that the work is actually for. What still does not fit is disclosed as
`"N older comment(s), beyond what fits in a task"` and lands in the same
"incomplete" line as the rest.

`TestWhatFitsInATask` covers both halves.

logos/logos-agent/workspace/Dockerfile
wasnertobias: Fixed in `e514ec0b3`.

All five configs are copied in and installed now — the root one (sub-pre-commit
for iris/athena/memiris plus the local Logos hook) and the four service ones.
They pin different versions of the same tools (black 26.1.0 in Iris where Logos
pins 25.1.0), so seeding one would have had a session reformat Iris with the
wrong black, which is worse than not linting.

Two things fell out of building it: Athena's `nbstripout` pins
`language_version: python3.12`, so 3.12 is installed beside 3.13; and the loop
now `exit 1`s, because the first build silently continued past a hook
environment it had failed to install. `.dockerignore` whitelists the four extra
configs.

Not offline-capable, and the prompt now says so: the `pylint`/`mypy` hooks in
iris/ and memiris/ go through `poetry run`.

logos/logos-agent/app/github.py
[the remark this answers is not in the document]
wasnertobias: Fixed in `e514ec0b3`.

The function no longer cuts at all — it returns what it read, and the caller
limits after filtering by author, adding
`"N older comment(s), beyond what fits in a task"` to the same `missing` list
that already carries the pages it could not read. The task claims to carry the
issue, so a silent cut was a lie — and the oldest comment is often the report
itself.

logos/logos-agent/app/triggers.py
```diff
@@ -491,6 +495,38 @@ async def _mark_comments_scanned(self, moment: datetime) -> None:
...
+        for entry in entries:
+            author = str(entry.get("author") or "")
+            if author and (author.lower() == reporter or await self._writer(author)):
+                allowed.append(entry)
+            else:
+                outsiders += 1
+        if outsiders:
+            missing = [*missing, f"{outsiders} comment(s) from accounts without write access"]
+        return allowed, missing
+
```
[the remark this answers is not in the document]
wasnertobias: Fixed in `e514ec0b3`.

`github.issue_conversation()` lost its `limit`; `triggers._issue_conversation()`
filters by author first and then cuts to `MAX_CONVERSATION`, disclosing what did
not fit. `test_a_maintainer_at_the_end_of_a_long_thread_survives` builds sixty
comments from passers-by with the maintainer's last and asserts their instruction
reaches the task.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in `e514ec0b3`, by the same durable state as the restart case.

`_take_up_a_red_build` only closes the watch when the follow-up exists **or** the
request has used its three attempts:

```python
if taken is not None or not await self._may_try_again(session):
    await self._checks_settled(int(session["id"]))
```

A transient database failure therefore leaves the row `pending`, and the next
scheduler pass tries again. `test_a_follow_up_that_could_not_be_queued_stays_owed`
and `test_a_request_out_of_attempts_stops_being_watched` pin both directions.

logos/logos-agent/app/github.py
wasnertobias: Good catch, and it did defeat the whole point of the change. Fixed in
`7baa5e15d`.

`GitHubError` carries `status` now, and `_get` passes the response code. A 404 on
the membership call is followed by one question about the team itself: if the
team is visible, the 404 is a real "not a member" and the answer is `False`; if
the team cannot be seen either, the question stays unanswered and the coarser
`may_push` rule stands in — which is the case a token reissued without `read:org`
must not turn into a silent repository-wide refusal.

`TestAskingWhoIsInATeam` covers the member, the non-member of a visible team, the
missing scope, a 403, and a pending invitation (not membership).

logos/logos-agent/README.md
wasnertobias: Fixed in `7baa5e15d`. The paragraph now says the comments that travel with a task
are the ones from the trusted teams (`LOGOS_AGENT_TRUSTED_TEAMS`), that everybody
else is left out and counted — the person who opened the issue included, since
anybody can open one — and that where the teams cannot be read at all the coarser
rule stands in.

logos/logos-agent/app/sessions.py
```diff
@@ -1673,6 +1842,31 @@ async def _settle(self, session_id: int, *, exit_code: int | None, error: str |
...
             # too — an answer that never comes is the worst of the three.
             await self._react(session_row, github.REACTION_FAILED)
+            if session_row.get("trigger_ref"):
+                # A request the runner took on and could not finish is a
+                # request nobody is coming back to: the reference counts as
+                # handled forever, so no later pass finds it again, and the
+                # only way back was somebody noticing. Taken up again here,
+                # bounded by the same three attempts as everything else, so
+                # a task that cannot be done stops rather than loops.
+                await self.take_up_again(
```
Claudia-Anthropica: @wasnertobias [medium] This remains separate from the durable check-watch path: for an ordinary started session that fails, `take_up_again()` returns `None` when the attempt-count query or replacement insert encounters a transient database failure, and this call discards that result. The original trigger remains handled and the failed row's reply is abandoned, so no later scheduler pass retries the request; persist generic retry intent and reconcile it until a replacement row exists or the attempt limit is reached.

In `logos/logos-agent/app/sessions.py`, ordinary failed-session settlement still discards a failed `take_up_again()` result, permanently losing the request after a transient database failure. Persist generic retry intent and have startup or scheduler reconciliation retry it until a replacement row is committed or the attempt limit is reached.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias Two medium correctness gaps remain on this head. The generic failed-session settlement still ignores a failed replacement, so the durable check-watch path cited in the reply does not rescue ordinary started sessions after a transient database failure. Separately, `POST /sessions` still persists `body.task` directly, so operator-created sessions never receive the editable house-rules block despite the UI and documentation promising it for every session; pass fresh manual tasks through `conventions.for_task()` before creating the row.

Claudia-Anthropica (changes requested):
@wasnertobias The workspace-ownership backfill is scoped to trigger-created workspaces and the captured head checks succeeded, but two medium correctness gaps remain. In `_settle()`, an ordinary failed triggered session still discards a `None` result from `take_up_again()` while only check-watch failures have durable pending state, so a transient attempt-count or insert failure permanently consumes the trigger. `POST /sessions` also still persists `body.task` directly instead of passing it through `conventions.for_task()`, so operator-created sessions do not receive the editable house-rules block promised for every session.

logos/logos-agent/app/sessions.py
wasnertobias: Right, and it is the case that actually bit: two sessions on production were
killed by the wall clock this morning, their settlements were the only thing
that would have replaced them, and nothing did.

Fixed in `6b958074b` — `resume_retries()`, beside `resume_check_watches()` on
startup and on every scheduler pass.

Derived from the rows rather than written down as intent, which I think is the
better shape here. A row already says everything needed: it failed, it came from
a request, nothing newer has been tried for that request, and the request has
attempts left.

```sql
   AND NOT EXISTS (SELECT 1 FROM agent_sessions newer
                    WHERE newer.trigger_ref = s.trigger_ref AND newer.id > s.id)
   AND (SELECT COUNT(*) FROM agent_sessions tried
         WHERE tried.trigger_ref = s.trigger_ref) < :max_attempts
```

`newer.id > s.id` is what makes "nothing newer" exact: a replacement that is
queued, running, or itself failed is a later row with the same reference, and
any of the three means this row no longer owes anything. The attempt bound is in
the query as well as in `take_up_again`, so a request at its limit does not come
back every fifteen seconds to be refused again.

Three things a flag would not have given: nothing to keep in step with the rows
it describes, no way for the intent and the fact to disagree, and it covers the
sessions that failed before any of this existed. Ran it against production as it
stands and it returns exactly the two sessions the timeout killed at 07:36 and
07:37 — the ones a flag written at settlement could never have found.

`TestARequestThatLostItsReplacement` covers the settlement whose retry failed
being picked up on the next pass, the limit travelling into the query, and a
database that will not answer costing the pass nothing.

PR #896, opened 2026-09-03
`Logos`: Remember rejected sharded checkpoints; opt out per model

LogosOSSAgent:


logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -1380,6 +1380,12 @@ async def _maybe_convert_sharded_checkpoint(
             vc_engine = cfg.engines.vllm if cfg.engines else None
             if vc_engine is None or not getattr(vc_engine, "sharded_checkpoint_enabled", True):
                 return
+            # Per-model opt-out (engines.vllm.model_overrides): the lane
+            # spawner honors the same setting, so converting here would only
+            # build a cache the lane never reads.
+            per_model_ov = (getattr(vc_engine, "model_overrides", None) or {}).get(model_name) or {}
```
Claudia-Anthropica, line 1386: @LogosOSSAgent [medium] The new per-model `true` override cannot opt a model into post-calibration conversion because line 1381 returns on the worker-wide `false` before this override is read. With the global switch and `sharded_checkpoint_convert_on_spawn` both false, an explicitly opted-in model can never receive a sharded checkpoint, even though the spawn path treats it as enabled. Compute the effective per-model/global value first and gate calibration on that value.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, post-calibration conversion returns on the global switch before applying a per-model opt-in. Compute the effective setting from the model override when present, otherwise the global setting, and use it for the early-return gate.

</details>

logos/logos-workernode/logos_worker_node/sharded_checkpoint.py
```diff
@@ -108,9 +122,106 @@ def invalidate_sharded_checkpoint(directory: Path) -> bool:
...
+
+    The rejection verdict is pinned to the engine build that produced the
+    unusable shards: a different build may round-trip a layout the old one
+    could not. A package that is not importable here is simply omitted — the
+    comparison then only sees the fields that could be read.
+    """
+    versions: dict[str, str] = {}
+    for pkg in ("vllm", "torch"):
+        try:
+            versions[pkg] = importlib.metadata.version(pkg)
```
Claudia-Anthropica, line 145: @LogosOSSAgent [medium] This fingerprints packages in the worker's interpreter, while conversion and serving explicitly support vLLM from another environment through `vllm_binary` and `resolve_vllm_python`. In that supported setup, upgrading the actual vLLM/torch environment leaves these recorded worker-package versions unchanged (or absent), so `is_sharded_checkpoint_rejected` keeps suppressing retries indefinitely. Fingerprint the interpreter selected for the lane instead.

In `logos/logos-workernode/logos_worker_node/sharded_checkpoint.py`, rejection records fingerprint the worker interpreter instead of the configured vLLM interpreter. Pass the lane's `vllm_binary` through rejection recording and checking, and obtain the vLLM/torch versions from the interpreter selected by `resolve_vllm_python`.

</details>

Claudia-Anthropica (changes requested):
@LogosOSSAgent The persistent-rejection and per-model override paths have two correctness gaps: calibration cannot honor an opt-in over global disablement, and rejection records fingerprint the wrong Python environment for supported external-vLLM setups. Both can leave a model permanently outside the intended sharded-checkpoint path, so changes are requested. The non-required `pre-commit` check is red, but its captured evidence reports only exit code 1; I could not attribute it to a specific defect and did not treat it as a finding.

wasnertobias:
Closing as superseded by #895.

Both PRs implement the same feature — persisting the “the vLLM loader refused this sharded checkpoint” verdict so the conversion is not re-run on every spawn — with two incompatible designs. They conflict in `sharded_checkpoint.py` and `tests/test_sharded_checkpoint.py`, and merging both would leave two parallel rejection mechanisms with different marker files.

#895 is the more mature implementation: it scopes the record to the interpreter that actually *serves* the configured vLLM binary (resolved via the script's shebang, so separate-venv deployments are handled), keeps the version probe off the event loop, and writes the sidecar atomically. This PR fingerprints the worker venv in-process via `importlib.metadata`, which is exactly the defect #895 already fixed, and its two review findings from 2026-09-04 were never addressed.

The two things this PR has that #895 lacks are being carried over on top of #895:

* the per-model opt-out `engines.vllm.model_overrides.<model>.sharded_checkpoint_enabled` (including the fix for the opt-in-over-global-disable gap raised in review),
* the guard in `logos_bridge.py`, so a rejection skip no longer records a misleading `sharded_conversion_failed` calibration event.

Thanks — the per-model override idea is the better half of this PR and is not being lost.

PR #899, opened 2026-09-04
`Development`: Keep pull request workflow read-only for docs builds

### LINKED ISSUE #480: Invalid workflow file
### Describe the bug

[Invalid workflow file: .github/workflows/pullrequest.yml#L16](https://github.com/ls1intum/edutelligence/actions/runs/23865733541/workflow)
The workflow is not valid. .github/workflows/pullrequest.yml (Line: 16, Col: 3): Error calling workflow 'ls1intum/edutelligence/.github/workflows/docs.yml@8abbe837da9f2ffd58d6de79f6650408b1204008'. The workflow is requesting 'pages: write, id-token: write', but is only allowed 'pages: none, id-token: none'.

### To Reproduce

-

### Expected behavior

-

### Screenshots

_No response_

### Which version of Edutelligence are you seeing the problem on?

-

### What service is this bug related to?

Other/Unsure

### What module is this bug related to?

Other (specify in "Additional context")

### What browsers are you seeing the problem on?

Other (specify in "Additional context")

### Operating System

_No response_

### Relevant log output

```shell

```

wasnertobias:
## What

Fixes the invalid workflow reported in #480:

> .github/workflows/pullrequest.yml (Line: 16): Error calling workflow '.../docs.yml@...'. The workflow is requesting 'pages: write, id-token: write', but is only allowed 'pages: none, id-token: none'.

## Why

`pullrequest.yml` called `docs.yml` via `workflow_call`. GitHub statically validates the called workflow's top-level `permissions` against the caller's — `docs.yml` declares `pages: write` + `id-token: write` (needed by its `deploy` job on `main`), while the PR workflow only granted `contents: read`, so validation failed.

## Change

Keep the PR workflow **read-only** (`contents: read`) and call the three read-only `*_docs-build.yml` workflows directly, instead of granting Pages/OIDC permissions to the caller:

- The locally called reusable workflow is PR-controlled code, so a PR could modify `docs.yml` and execute a job with Pages deployment or OIDC capabilities — the `if: github.ref == 'refs/heads/main'` condition is not a security boundary.
- Pages deployment stays exclusively on the `push-main.yml` path, which already carries `pages: write` + `id-token: write`.

Closes #480

## Summary by CodeRabbit

- **Chores**
  - Documentation checks now build separately for Athena, Atlas, and Iris during pull requests.
  - Pull-request documentation workflows remain read-only, while documentation deployment continues through the main-branch release process.

### DIFF
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -11,6 +11,8 @@ concurrency:
 
 permissions:
   contents: read
+  pages: write
+  id-token: write
 
 jobs:
   docs:
--- a/.github/workflows/pullrequest.yml
+++ b/.github/workflows/pullrequest.yml
@@ -11,10 +11,20 @@ concurrency:
 
 permissions:
   contents: read
-  pages: write
-  id-token: write
 
 jobs:
-  docs:
-    uses: ./.github/workflows/docs.yml
+  # Call the build workflows directly instead of docs.yml: the locally
+  # called reusable workflow is PR-controlled code, so this workflow must
+  # stay read-only and must not carry pages/id-token permissions. Pages
+  # deployment stays on the push-main.yml path.
+  build-athena:
+    uses: ./.github/workflows/athena_docs-build.yml
+    secrets: inherit
+
+  build-atlas:
+    uses: ./.github/workflows/atlas_docs-build.yml
+    secrets: inherit
+
+  build-iris:
+    uses: ./.github/workflows/iris_docs-build.yml
     secrets: inherit


.github/workflows/pullrequest.yml
```diff
@@ -11,6 +11,8 @@ concurrency:
 
 permissions:
   contents: read
+  pages: write
```
Claudia-Anthropica: @wasnertobias [high] Granting `pages: write` and `id-token: write` here lets a future pull request modify the locally called `docs.yml` and execute a job with Pages deployment or OIDC capabilities. The current `deploy` condition is not a security boundary because that condition and the called workflow are themselves evaluated from the pull request; keep this workflow read-only and invoke only the three build workflows from this PR path.

In `.github/workflows/pullrequest.yml`, granting `pages: write` and `id-token: write` allows PR-controlled local reusable-workflow code to obtain deployment/OIDC privileges. Keep this workflow at `contents: read` and call only build-specific reusable workflows, leaving deployment permissions exclusively on the main-push path.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The documentation builds succeeded and deployment was skipped, but this change exposes Pages/OIDC write capabilities to PR-controlled reusable workflow code. Keep pull-request documentation builds read-only and reserve deployment permissions for the main-push path. Separately, the non-required `validate-pr-title` check is red because the title lacks the required backtick-wrapped scope.

.github/workflows/pullrequest.yml
wasnertobias: Done in a81fc2f: reverted to contents: read and call the three read-only *_docs-build.yml workflows directly — the PR path no longer carries pages/id-token, so PR-controlled docs.yml code can no longer reach deployment or OIDC capabilities. Deployment stays on push-main.yml. Title updated to the required scope format as well.

PR #900, opened 2026-09-04
`Logos`: Answer a review request, measure the GPUs, and keep two sessions off one branch

wasnertobias:
## Summary by CodeRabbit

* **New Features**
  * Added support for reviewer-requested agent sessions, including review-only work on protected and forked branches.
  * Pull-request conversations now use the relevant pull-request revision and respect branch permissions.
  * Pull requests can automatically include referenced issues in their closing statement.
  * Added read-only session handling to prevent unintended pushes, including across retries.

* **Bug Fixes**
  * Improved session claiming to prevent conflicting work on the same branch.
  * Improved queue capacity reporting across deployments and providers.
  * Made interruption markers reliably readable without exposing them to session modifications.

* **Documentation**
  * Documented reviewer-triggered sessions and branch access rules.


Claudia-Anthropica (changes requested):
@wasnertobias [high] The existing branch-claim thread remains blocking on this head: `claim_queued_sessions(limit >= 2)` evaluates the new occupancy predicate while same-branch peers are still queued, so one SELECT can return and start multiple rows for the same non-null branch. The capacity diagnostic and prompt changes look sound and captured CI completed successfully, but the batch query must select at most one candidate per non-null branch while preserving null-branch candidates, with regression coverage.

logos/logos-agent/app/capacity.py
```diff
@@ -195,7 +223,18 @@ def parse_scheduler_state(
...
+    # orchestrator's queue, and the queue is keyed by model alone — so a
+    # request on its way to Azure sits in it exactly like one waiting for a
+    # GPU. The runner read that as "a user is waiting", paused its sessions,
+    # gave a GPU back to somebody who was never asking for one, and lost the
+    # turn each session was in the middle of. Pausing frees a GPU; nothing
+    # about that helps a request bound for a cloud provider.
+    #
+    # The cloud never appears below: `logosnode.providers` is the local
+    # fleet, and that is the whole point of reading it here.
+    queue_total = 0
```
Claudia-Anthropica, line 246: @wasnertobias [high] Zeroing the fleet-wide count makes `_live`'s choice between `queue_depth` and `queue_waiting_current` unsafe: these are distinct queues, not competing measurements. A GPU-bound request can remain in the orchestrator queue (`queue_depth > 0`) while vLLM correctly reports zero backend wait, causing this runner to see no waiting users and retain or add GPU work; the newly added test at the bottom currently codifies that failure. Count the model-wide orchestrator backlog once and add the per-deployment backend backlog instead of allowing either value to erase the other.

In `logos/logos-agent/app/capacity.py`, local orchestrator backlog is discarded whenever vLLM reports a numeric backend queue value, even though the two values describe different queue stages. Deduplicate and count `queue_depth` once per model, add the relevant per-deployment `queue_waiting_current` values, and update the regression test so positive scheduler backlog remains visible when backend wait is zero.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The unresolved branch-claim thread remains blocking: the batch query still admits multiple queued rows sharing a non-null branch, while concurrent single-session claims have no shared branch lock. The new capacity calculation also hides local scheduler backlog when vLLM reports no backend wait, as detailed inline. Captured CI completed successfully, but these current-head correctness issues require changes.

Claudia-Anthropica (changes requested):
@wasnertobias [high] Both prior branch-claim blockers remain visible: `claim_queued_sessions` can select multiple queued rows sharing one non-null branch in a single batch, while concurrent `claim_session` calls still lack shared branch serialization. [high] `parse_scheduler_state` still chooses `queue_waiting_current` over `queue_depth`; the orchestrator's own estimator treats these as additive queue stages, so a numeric backend zero can hide positive local scheduler backlog and admit or retain GPU work. The interruption-marker changes have focused tests and captured CI completed successfully, but they do not resolve these two defects.

Claudia-Anthropica (changes requested):
@wasnertobias [high] The capacity blocker remains: `_waiting` and `_live` choose `queue_waiting_current` instead of adding it to the model-wide `queue_depth`, so a backend zero can mask positive scheduler backlog and let GPU work start or remain while users wait. [medium] `claim_queued_sessions(limit >= 2)` still evaluates same-branch candidates while all are queued and can claim multiple rows sharing one non-null branch in a single statement. The interruption and helper-diagnostic changes are covered and the captured checks have no failure, but these two current-head issues still require changes.

logos/logos-agent/app/triggers.py
```diff
@@ -690,9 +750,74 @@ async def _candidates(self, now: datetime) -> list[dict[str, Any]]:
...
+                    # re-adding it, which is a new request and reads as one.
+                    "ref": f"pr-{number}-review-requested-{requester.lower()}",
+                    "kind": "review-request",
+                    "task": await review_request_task(
+                        number, title, str(pull.get("body") or ""), requester, branch=branch
+                    ),
+                    "branch": branch,
+                    # Nothing to push to: read the pull request's own code
+                    # so the review is at least about the right diff.
+                    "read_ref": None if branch else f"refs/pull/{number}/head",
```
Claudia-Anthropica, line 864: @wasnertobias [high] Setting `branch` to `None` does not make this session read-only. `_launch_tracked` replaces every null branch with `branch_for(...)`, records it, and `_finalize` invokes `commit_and_push` with the session token, so any edit made while reviewing a fork creates a branch in this repository despite the prompt saying it cannot push.

In `logos/logos-agent/app/triggers.py`, review and comment candidates marked read-only still receive a derived push branch during launch and are finalized with repository credentials. Persist an explicit no-push capability and have `SessionManager` bypass credentialed commit/push finalization for it, with an end-to-end test proving a modified read-only checkout leaves the remote unchanged.

</details>

logos/logos-agent/app/triggers.py
```diff
@@ -271,6 +271,47 @@ def _inline_block(comments: list[dict[str, Any]]) -> str:
...
+    change" to a request to change it is not much of an answer.
+    """
+    description = (body or "").strip()
+    if len(description) > 4000:
+        description = description[:4000] + "\n\n[description truncated]"
+    return await for_task(
+        f"{requester} asked you to review #{number} ('{title}').\n\n"
+        f"{description or '(the pull request has no description)'}\n\n"
+        f"You are working in a checkout of that pull request's own code, so read the diff "
+        f"against the default branch — `git diff origin/main...HEAD` — and then read the "
```
Claudia-Anthropica, line 298: @wasnertobias [medium] This required diff command is not guaranteed to exist in the constructed checkout. Same-repository reviews clone shallowly with `--branch <feature>`, while reused checkouts rebuild `.git` and fetch only their configured base, so neither path necessarily creates `refs/remotes/origin/main`; the review can therefore fail before inspecting the change it was asked to assess.

In `logos/logos-agent/app/triggers.py`, review tasks require `origin/main` although the checkout preparation path does not guarantee that ref exists. Fetch and materialize the default branch as a remote-tracking ref during preparation, and add regression tests for fresh feature-branch and reused pull-ref checkouts.

</details>

logos/logos-agent/workspace/run_session.py
```diff
@@ -606,10 +648,30 @@ def _agent_command(prompt: str, *, resuming: bool) -> list[str]:
...
+def _pauses_so_far() -> int:
+    """How many times the runner has frozen this session.
+
+    Counted rather than flagged: a run has to know whether it was frozen
+    during *itself*, and a flag set by an earlier pause would make every
+    later failure look like an interruption.
+    """
+    path = Path(os.environ.get("LOGOS_ARTIFACT_DIR", "/artifacts")) / INTERRUPTION_FILE
+    try:
+        return sum(1 for line in path.read_text(encoding="utf-8").splitlines() if line.strip())
```
Claudia-Anthropica, line 761: @wasnertobias [high] `/artifacts` is writable by the untrusted agent, and the runner explicitly hands ownership of this marker to that same user. The agent can append a line during its invocation, making this comparison classify an ordinary failure as a platform pause and grant up to 60 costly continuations; this is not an authoritative runner signal.

In `logos/logos-agent/workspace/run_session.py`, pause classification trusts a marker the untrusted agent can modify. Move the pause signal to runner-owned state exposed read-only to the agent-phase harness, and add a regression test proving agent-written artifact data cannot create paused continuations.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias [high] The prior claim and capacity blockers remain: concurrent claims still only lock their individual rows, the batch claim can select same-branch peers together, and `_live`/`_waiting` still replace a positive orchestrator queue with a numeric backend zero although those are distinct additive stages. The new review path also has authorization and diff-base defects, while the interruption marker trusts agent-writable state, as detailed inline. The captured CI snapshot contains no failed checks, although several non-required checks were still pending.

Claudia-Anthropica (changes requested):
@wasnertobias The existing blockers remain on this head: same-branch claims are neither serialized nor deduplicated, and capacity parsing lets a backend zero hide the distinct orchestrator backlog. The review path still grants credentialed finalization to nominally read-only sessions, trusts an agent-writable pause marker, and does not reliably materialize the refs its prompts require. The incomplete-timeline authorization, mixed-author writable tasks, reusable review-request identity, swallowed pull-request lookup failures, and missing remote-tracking ref also remain visible in the current code. Captured CI has no failed or pending required checks, but these correctness and security defects still require changes.

wasnertobias:
@LogosOSSAgent Resolve conflicts

LogosOSSAgent:
I can't resolve the conflicts from my side, for two independent reasons:

1. **The PR isn't visible to me.** My checkout only contains `main`, currently at `8ea8f70` ("`Logos`: Add dynamic Scheduler"). I checked the local refs, the full commit history, stashes, and worktrees: PR #900 has not been merged (nothing in the history references it), and its branch and diff are not present in my environment. I have no network access or GitHub credentials to fetch it, so I can't even tell you which files are conflicting — I won't guess at that from the diff I don't have.

2. **I can't push to your branch.** My tooling only commits to its own session branch and opens new pull requests from it; it can't push into the branch of an existing pull request. Resolving the conflict means updating your branch, which has to come from you (or someone with access to it).

What clears it: rebase your branch onto current `main` (`8ea8f70`), or merge `main` into it, resolve the hunks, and push. If you paste the conflict output here — the `git merge main` report on the PR branch, or the file list GitHub shows as conflicting — I'll walk through each hunk and say exactly how I'd resolve it, so you can apply it in one pass.

wasnertobias:
All review threads are now resolved. The four still-open findings from the review passes of Sep 4 are addressed in d360285: the two queue stages count additively instead of one erasing the other (capacity.py), a session with no writable branch persists no_push and the finalizer refuses to push for it (triggers.py, sessions.py), the prepared checkout carries the default branch the review diffs against (run_session.py), and the pause signal moved out of the agent-writable artifacts into runner-owned state mounted read-only (run_session.py). Each comes with its regression test, all green.

wasnertobias:
All four findings from the re-review of the merge commit are addressed in 566570310: the workspace's best-candidate comparison now applies the trigger rule to peers as well (db.py), the state directory is made traversable after mkdir so a restrictive runner umask cannot hide the pause mark (sessions.py), a question on a pull request the runner answers for reads that pull request's own code even when the head is not writable (triggers.py), and the claim tests lock their rows the way the database does — the missing await made the fake's row lock a formality (test_db.py). Each comes with a regression test; the suite is at 592 green. The PR body now names the issues the work closes and nothing else (dbff28ce0).

logos/logos-agent/workspace/run_session.py
```diff
@@ -1009,19 +1141,34 @@ def _as_subject(text: str) -> str:
...
+    """The pull request's body, and nothing else.
+
+    The issues the work closes, named the way GitHub reads the reference:
+    merging the pull request closes them. The numbers are the task's — the
+    assigned issue and the ones its body and conversation point at — and
+    the list is the whole body, so nothing beside it can overstate what the
+    change does. What the change does belongs in the commit subject; why it
+    was made belongs to whoever picks it up.
+    """
+    numbers = sorted({int(number) for number in re.findall(r"#(\d+)\b", task)})
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 8ceaf3519. `_closed_issues` now deduplicates with `dict.fromkeys` over the `re.findall` result, so the body lists the references in the order the task names them (`closes #948, #493` for a task that writes `#948` before `#493`) instead of re-sorting them numerically.

Tests: `TestClosedIssues` in `tests/test_run_session.py` — the non-monotonic case (`#948` before `#493`), a reference named twice, and a task without references. Full `tests/test_run_session.py` (76 tests) green, pre-commit clean.

logos/logos-agent/app/triggers.py
```diff
@@ -864,15 +1037,46 @@ def consider(comment: dict[str, Any], number: int, *, root: int | None) -> None:
...
+            # seen the diff. Its title was `#882` for the same reason.
+            other = None if pull else await self._pull_request(number)
+            title = pull["title"] if pull else str((other or {}).get("title") or f"#{number}")
+            # About a pull request's code when it is one this runner
+            # answers for or one it read: the checkout must carry that
+            # pull request's own head even when the head is not a branch
+            # it may push to — its own protected branch, for instance —
+            # or the answer would be written from main, a diff it was
+            # never shown.
+            about_pull = pull is not None or other is not None
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 8ceaf3519. `_writable_head` no longer swallows lookup failures: `None` is now an answer only (the lookup succeeded and the head is confirmed not ours to push — a fork or a protected branch), and a failed lookup raises. A head that cannot be read is carried as its own state (`_HEAD_UNKNOWN`) rather than as `None`:

- `_responsible_pulls` records it when the pre-fetch fails, so the pull request stays responsible (its comments are still considered) without a branch to stand behind.
- A changes-requested review or takeover of a pull request whose head is unknown is left to the next pass — both are read from the repository's state, so nothing is lost.
- A review request whose head lookup fails is not added at all: the reference is never recorded, and the next pass finds the request again.
- A comment conversation about an unknown head becomes a `deferred` candidate: nothing is queued for it, and `poll_once` treats deferred candidates the way it treats work that does not fit — the comment mark does not move past the conversation, so the question is still in the window when the next pass reads the head again.

Regression tests: `TestAHeadThePassCouldNotRead` in `tests/test_triggers.py` — a change request on a responsible pull request whose head lookup fails queues nothing and holds the mark; the next pass (lookup working) lands the session on the branch; a review request is not spent while its head is unreadable and is answered once the lookup succeeds. `FakeRepo` gained `failing_heads` for 5xx lookups. All 96 triggers tests plus the rest of the agent suite (560) green, pre-commit clean.

logos/logos-agent/app/db.py
```diff
@@ -1152,86 +1205,154 @@ async def claim_session(session_id: int, *, trigger_quota: int | None = None) ->
...
+        for branch in sorted({c["branch_name"] for c in candidates if c["branch_name"] is not None}):
+            await db.execute(
+                text("SELECT pg_advisory_xact_lock(hashtext('logos-agent-branch:' || :branch))"),
+                {"branch": branch},
+            )
+        claimed = (
+            (
+                await db.execute(
+                    text(
+                        "SELECT s.id, s.branch_name FROM agent_sessions s WHERE "
```
Claudia-Anthropica, line 1339: @wasnertobias [high] This second query reevaluates `_CLAIMABLE` without restricting results to the branches locked from the first query. Under PostgreSQL's default READ COMMITTED isolation, a newly queued row—or a row that becomes claimable while this transaction waits for another branch lock—can appear here with a non-null branch whose advisory lock this transaction never acquired; a concurrent `claim_session` for a different row on that branch can then admit both sessions. Restrict this query to null branches plus the previously locked branch set, or acquire any newly discovered branch locks before rechecking and claiming.

In `logos/logos-agent/app/db.py`, `claim_queued_sessions` can claim a non-null branch that became eligible after its initial candidate query without holding that branch's advisory lock. Ensure the final claim query returns only null branches or branches already locked by this transaction, or lock newly discovered branches and recheck before claiming; add a barrier-based concurrency regression test covering a branch becoming eligible between discovery and claim.

</details>

logos/logos-agent/workspace/run_session.py
```diff
@@ -1009,19 +1141,36 @@ def _as_subject(text: str) -> str:
...
+    The issues the work closes, named the way GitHub reads the reference:
+    merging the pull request closes them. The numbers are the task's — the
+    assigned issue and the ones its body and conversation point at — and
+    the list is the whole body, so nothing beside it can overstate what the
+    change does. What the change does belongs in the commit subject; why it
+    was made belongs to whoever picks it up.
+    """
+    # First-seen order, deduplicated: the body lists the references the way
+    # the task names them, not in numerical order.
+    numbers = list(dict.fromkeys(re.findall(r"#(\d+)\b", task)))
```
Claudia-Anthropica: @wasnertobias [medium] This scans the entire generated task, which includes the assigned issue's body and retained conversation, and turns every `#N` mention into a GitHub closing keyword. A comment such as “related discussion in #948” therefore makes merging the generated PR close #948 even though only the assigned issue was authorized as work. Derive closures from the assigned issue identity (or explicit closing directives) rather than arbitrary references in task context.

In `logos/logos-agent/workspace/run_session.py`, `_closed_issues` converts every issue reference in the issue body and conversation into a closing directive, so merely related issues can be closed on merge. Generate the closing body from the assigned issue identity or explicitly authorized closure references only, and add a test proving an incidental related-issue mention is not closed.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The prior capacity, no-push, interruption, and unknown-head findings are resolved, and the captured checks completed successfully. However, the batch claim still has a branch-lock race between candidate discovery and the locking query, and generated pull-request bodies can close unrelated issues merely mentioned in task context. These current-head data-integrity issues require changes.

logos/logos-agent/app/sessions.py
```diff
@@ -1012,6 +1078,9 @@ async def _launch_tracked(self, session: dict[str, Any], launch: _Launch) -> Non
                 ),
                 workspace_volume=workspace["volume_name"],
                 artifact_host_path=artifact_host_path,
+                # The agent's container is the one that reads the pause mark;
+                # the helper phases never do, and stay without the mount.
+                state_host_path=str(state_dir(sid)),
```
Claudia-Anthropica: @wasnertobias [high] This passes a path inside the `logos-agent` service container, but Docker interprets bind sources on the daemon host. `logos/docker-compose.yaml` mounts only `agent_artifacts` and declares neither a state volume nor `LOGOS_AGENT_STATE_ROOT`, so the runner writes the marker into its container overlay while the child mounts a different, daemon-host directory; `_pauses_so_far()` therefore never sees the authoritative pause marks in the deployed topology. Back this state with storage shared between the runner and daemon and pass the daemon-visible mountpoint to the child, with a regression test that models the separate filesystems.

In `logos/logos-agent/app/sessions.py`, the child container's state bind uses a runner-container-local path that does not refer to the same directory on the Docker host. Back the interruption state with a named volume mounted into the runner, resolve its daemon-visible mountpoint before creating the child container, and add a deployment-topology regression test proving a runner-written mark is visible read-only inside the child.

</details>

logos/logos-agent/workspace/run_session.py
```diff
@@ -371,20 +381,57 @@ def prepare_checkout(repo_url: str, base_branch: str, branch: str, token: str) -
...
+    # the checkout is the code under discussion and there is nothing here to
+    # push back to.
+    reading = base_branch.startswith("refs/")
+    cloned = not (CHECKOUT / ".git").is_dir()
+    if cloned:
         log(f"cloning {repo_url} at {base_branch}")
         CHECKOUT.parent.mkdir(parents=True, exist_ok=True)
         _clear_checkout()
-        run(["git", "clone", "--depth", "50", "--branch", base_branch, repo_url, str(CHECKOUT)])
+        clone = ["git", "clone", "--depth", "50"]
```
Claudia-Anthropica, line 445: @wasnertobias [medium] Both the pull-request head and default branch are fetched with depth 50. When their merge base is older than those shallow boundaries, both refs exist but `git diff origin/main...HEAD` fails with no merge base; the agent phase has no external egress to deepen the checkout, so the requested review cannot be performed. Have the trusted preparation helper deepen the histories until a merge base exists (or fail preparation) and cover a divergence beyond 50 commits using a transport where `--depth` is honored.

In `logos/logos-agent/workspace/run_session.py`, depth-50 fetches can leave a review checkout without the common ancestor required by `git diff origin/main...HEAD`. Deepen both refs in the trusted preparation phase until `git merge-base` succeeds, fail preparation if it cannot, and add a regression test with a merge base more than 50 commits behind both tips.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The two previous data-integrity findings are fixed on this head: the batch recheck is bound to locked branches, and PR closing text derives only from the assigned issue identity; the snapshot has 49 successful checks. Two current correctness issues remain in deployed state sharing and shallow review checkouts, as detailed inline.

logos/logos-agent/app/db.py
Claudia-Anthropica: @wasnertobias Thanks, this is fixed now: the final claim query is restricted to `locked_branches`, and the new window regression covers the race.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 53f9796e2. The state is now backed by a named volume mounted into the runner (`agent_state` → `/var/lib/logos-agent/state`, `LOGOS_AGENT_STATE_VOLUME` in docker-compose), and the child's bind source is that volume's daemon-visible mountpoint — resolved the same way the artefact volume's is — instead of the in-container state directory. The runner writes the pause mark into the volume through its in-container `state_root`; the child mounts the same storage read-only, so `_pauses_so_far()` reads it.

Tests: `test_the_pause_mark_reaches_the_child_through_the_state_volume` in `tests/test_sessions.py` — the child's state bind source is the state volume's mountpoint, not the in-container state directory, and the mark the runner writes is the bytes the child mounts.

logos/logos-agent/workspace/run_session.py
wasnertobias: Fixed in 93798f18d. The trusted preparation helper now deepens both histories in steps (`git fetch --deepen`, no refspec, so the base's line and the default's line both grow) until `git merge-base origin/<default> HEAD` resolves, and fails the preparation if it still cannot within the ceiling — so a base that cannot be found is a failed session, not a review diff the agent cannot compute. The agent phase is unchanged: it still has no egress to deepen, it is handed a checkout whose diff it can actually run.

Tests: `test_a_merge_base_beyond_the_shallow_history_is_deepened_into_view` and `test_a_merge_base_beyond_the_ceiling_fails_the_preparation` in `tests/test_run_session.py` — over a `file://` transport (where the clone honours `--depth`), a pull request whose merge base is more than fifty commits behind both tips is deepened to its true split point and the diff shows the request's own work; a base beyond the ceiling fails the preparation.

logos/docker-compose.yaml
```diff
@@ -437,12 +437,19 @@ services:
       LOGOS_AGENT_TRIGGERS_ENABLED: "${LOGOS_AGENT_TRIGGERS_ENABLED:-true}"
       LOGOS_AGENT_DEPLOY_ENABLED: "${LOGOS_AGENT_DEPLOY_ENABLED:-false}"
       LOGOS_AGENT_DEV_BASE_URL: "${LOGOS_AGENT_DEV_BASE_URL:-https://logos-dev.aet.cit.tum.de}"
+      # The volume that backs the runner's per-session state (the pause mark).
+      # Named to match the volume below so the child container's bind source
+      # resolves to the same storage the runner writes the mark into.
+      LOGOS_AGENT_STATE_VOLUME: "${LOGOS_AGENT_STATE_VOLUME:-agent_state}"
```
Claudia-Anthropica: @wasnertobias [high] Compose mounts the logical `agent_state` volume under its project-scoped engine name (normally `logos_agent_state`), but this environment value makes `SessionManager` inspect and, if absent, create the separate literal Docker volume `agent_state`. The runner writes pause marks through the Compose-mounted volume while child containers bind from the literal volume's mountpoint, so deployed sessions still cannot observe authoritative pause marks. Give the top-level volume an explicit `name` and pass that exact same expansion here.

In `logos/docker-compose.yaml`, `LOGOS_AGENT_STATE_VOLUME` names a different Docker volume from the project-scoped `agent_state` volume mounted into the runner, so child containers read a separate empty volume. Declare an explicit engine-level name for `agent_state` and pass that identical name through `LOGOS_AGENT_STATE_VOLUME`, then add a Compose-topology regression test.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The shallow-review checkout blocker is fixed, and the snapshot contains 45 successful checks plus one skipped build. The deployed pause-marker sharing defect remains because Compose passes a literal volume name while mounting a project-scoped volume, as detailed inline.

Claudia-Anthropica (changes requested):
@wasnertobias [high] The deployed pause-marker sharing defect remains: Compose passes the literal volume name `agent_state`, while its unnamed top-level volume is project-scoped (normally `logos_agent_state`). `SessionManager` consequently inspects or creates a different literal volume and binds child containers from its empty mountpoint, while the runner writes pause marks into the Compose-mounted volume. The snapshot has 44 successful checks, one skipped check, and one pending non-required documentation check, but this current deployment defect still requires changes.

logos/docker-compose.yaml
wasnertobias: Fixed in 5f99a7612. The compose volume is now given the explicit literal name `logos_agent_state` instead of relying on its project-scoped engine name, and `LOGOS_AGENT_STATE_VOLUME` defaults to that same name — which is also the runner's config default. So `ensure_volume` / `volume_mountpoint` operate on the exact volume Compose mounts at the runner's `state_root`, and the child's bind source is the storage the runner writes the pause mark into.

Tests: `test_the_pause_mark_reaches_the_child_through_the_state_volume` in `tests/test_sessions.py` covers the code half (the child's state bind source is the state volume's daemon-visible mountpoint, not the in-container path, and the mark round-trips through the shared storage); the name agreement itself now lives in docker-compose.

PR #901, opened 2026-09-04
`Logos`: Track last activity to prevent sleeping busy lanes

LogosOSSAgent:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved idle detection to account for request activity that begins and ends between polling cycles.
  * Prevented recently active lanes from being put to sleep prematurely.
  * Preserved sleep behavior for lanes that have genuinely remained idle.

* **Tests**
  * Added coverage for recent activity, idle lanes, activity timing, and provider-specific tracking.


Claudia-Anthropica (dismissed):
@LogosOSSAgent The facade now records request begin/completion activity per provider and model, and the planner advances stale idle timestamps before evaluating sleep actions. The focused tests cover recent and stale activity, provider scoping, facade-to-planner wiring, and the unchanged genuinely-idle sleep path; the captured checks at this exact head report no failures.

wasnertobias:
Conflicts resolved — `e8ba6a04c`. `MERGEABLE`, and the diff dropped from +929/-28 across 7 files to **+311/-1 across 4**.

**Where the conflict came from.** This branch forked off #895's branch and carried five of its sharded-checkpoint commits (`6a1cad215` … `223191172`, 2026-09-04). #895 has since merged as `0c8edcc24`, so `main` holds a *newer* version of that same work — including the review fixes this branch's copy predates. The two rewrote the same regions of `sharded_checkpoint.py` from a common base, which is the conflict.

**How.** All four sharded-checkpoint files are pinned to `main`'s exact content rather than three-way merged. That mattered: only `sharded_checkpoint.py` actually conflicted — `vllm_process.py`, `test_sharded_checkpoint.py` and `test_vllm_process.py` all reported *auto-merged*, which is the quieter hazard. A three-way merge keeps this branch's stale lines wherever `main`'s squash did not touch the same region, so accepting the auto-merge would have silently reverted parts of merged work. `git diff origin/main -- logos/logos-workernode/` is now empty. The PR is back to being only its own change.

No force-push was needed: the repo is squash-merge only (`allow_merge_commit=false`, `allow_rebase_merge=false`), so the extra history collapses on merge anyway.

**One thing I had to add — please look at this.** The squash of #895 **did not include** its last commit, `083b800db`, which fixed six tests in `tests/unit/capacity/test_multi_replica_loads.py`. `0c8edcc24` touches only the eight worker-node files; `test_multi_replica_loads.py` is not among them. So that fix was lost on merge and **`main` is still affected**.

That is not cosmetic. #931 (`11e8ddb34`) added `_load_requires_calibration`, and this test file's helper stubs `_safe_get_profiles` to `{}`, so every `load_lane_manually` there is refused. Six tests fail and three of them **hang forever** — the `TestInFlightLaneIdReservation` scenarios `await` an event only the dispatch callback sets, and the refusal returns before dispatch. The orchestrator CI step runs `coverage run -m pytest` with **no `--timeout`** (only the worker-node step passes one), so `Run Tests` never terminates instead of reporting a failure.

This PR touches the orchestrator, so its own CI would have hung on it. I cherry-picked `083b800db` back in — that is the fourth file in the diff. If it is preferable to land that separately on `main` instead, say so and I will drop it here.

**Local, with the venv CI uses (`logos-orchestrator`):**

| | |
|---|---|
| orchestrator suite | **1581 passed**, 1 xfailed, 6.74s |
| worker-node suite (CI ignores) | **572 passed**, 2 skipped |

`pre-commit` clean on all four files.

Issue #810, opened 2026-08-27
`Logos`: Context routing 404s pinned-model requests when the token estimate overshoots the served window


wasnertobias:
When a client pins a model in the request (e.g. Claude Code calling `/v1/messages` with a specific model) using a large conversation, the orchestrator can answer **404 "No deployment found for model …"** although a deployment of that model exists and the request would actually fit the served context window.

Observed case: a session whose own usage accounting reported ~75 % of the model's 256k (262,144) window was rejected with the 404. Counted exactly with the model's tokenizer, the request is ~200k input tokens plus a 20k `max_tokens` reservation — comfortably below 262,144.

Root cause — two interacting issues in `logos/logos-orchestrator` (both introduced/active since the context-routing feature, #775):

1. **The token estimate is systematically too high for agent traffic.** `required_context_tokens()` in `context_budget.py` counts payload characters at 3.0 chars/token. Tool-heavy assistant payloads (many `tool_use`/`tool_result` blocks, repeated JSON framing, code) tokenize at ≈ 3.7 chars/token with the Qwen tokenizer, so the estimate overshoots the true count by ~20 % (in the observed case the estimate projected ~243k prompt tokens; the tokenizer counts ~202k). The estimate (≈266k) crosses the 262,144 window, while the true requirement (≈225k incl. reservation and margin) fits with headroom.

2. **The context filter can empty out the pinned model.** `_prefer_deployments_with_context_room()` drops every deployment whose known served window is below the estimate. Its fallback ("return the widest if nothing is left") only triggers when *all* deployments are filtered out. If the API key is additionally entitled to other models whose lanes are not loaded (unknown window → always kept), those survive — while *all* deployments of the pinned model are gone. Proxy mode then narrows the already-filtered list to the pinned `model_id`, finds nothing, and raises the 404 — masking that the request was servable.

Impact:
- A servable request fails with a misleading 404 ("no deployment found") that suggests a configuration/permission problem instead of context routing.
- Clients (Claude Code) cannot continue; the session is stuck until the conversation is compacted or the model is switched.
- The orchestrator synthesizes an error the serving engine would have answered more honestly — and in the observed case it would simply have served the request.

To Reproduce:
1. Serve a model on one or more workers with a 262,144-token context window; give the API key additional model entitlements that have no loaded lanes (unknown window).
[...]

PR #903, opened 2026-09-04
`Logos`: Stop offering Claude Code a window it cannot start in

wasnertobias:
## What was wrong

A freshly installed `claude-logos` on `logos-test` answered its **very first message** with a 400 from the worker:

```
This model's maximum context length is 32768 tokens. However, you requested 20000
output tokens and your prompt contains at least 12769 input tokens, for a total of
at least 32769 tokens.
```

That is not a coincidence of one long prompt — it is arithmetic, and it makes a 32,768-token lane unusable for Claude Code no matter what:

| | |
|---|---|
| window | 32,768 |
| headroom (`claude-logos.sh`, 2% floored at 1024) | 1,024 |
| `CLAUDE_CODE_MAX_CONTEXT_TOKENS` handed over | 31,744 |
| `COMPACT_AT` | **−1,256** |
| `HARD_STOP_AT` | 8,744 |
| real vLLM input cap (`window − max_tokens`) | **12,768** |

Claude Code reserves 20,000 tokens for a reply on every request whatever it is told, vLLM charges input and output against the same window, and Claude Code's own system prompt plus tool schemas is ~13,000 tokens before the user has typed anything — the 12,769 in the error message. One token short, with nothing to compact yet, so the session never recovers.

The wrapper's guard could not catch it: it only tested `headroom + reservation >= window`, which is `21024 >= 32768` — passed comfortably while being unusable.

## What this changes

The floor is Claude Code's own opening prompt, and it is enforced where the model gets chosen:

```
floor       = 13000 opening prompt + 20000 reservation + 3000 hard stop     + 1024 headroom = 37024
comfortable = 13000 opening prompt + 20000 reservation + 13000 auto-compact + 1024 headroom = 47024
```

**AI Tools page** (`ai-tools.ts` / `.html`) — per @tobiaswasner's suggestion, the model simply cannot be picked for Claude Code:

- the option is disabled and its label says why (`… — 32,768 tokens, too small for Claude Code`);
- the wizard stops on the model step with the arithmetic spelled out, rather than generating an install command whose first use is an error;
- step 3 is no longer skipped when the single available model is a blocked one — that step is where the explanation lives;
- the pre-selection and a tool switch land on a model Claude Code can use, when there is one;
- between 37,024 and 47,024 the model stays selectable with a "workable but tight" note;
- **OpenCode is untouched.** It is told what to reserve (`min(8192, context/2)`), so a narrow window costs it reply length, not the session.

The figure judged is `max_model_len_current_min` — the one a request meets whichever deployment answers — falling through to the wider figures only when absent. A model no lane is serving is never judged.

**The wrapper** (`.sh` + `.ps1`, revision 1 → 2) — because the page is not the only way in (an existing install, `LOGOS_MODEL=…` per invocation), it refuses to start rather than handing Claude Code a session that 400s:

```
BLOCKED  : this window cannot host a Claude Code session.
           8,744 tokens of input are left after the 20,000 reserved for replies and
           the 1,024 of headroom, and Claude Code needs about 13,000 of that for its own
           system prompt and tool definitions — so the first request is rejected.
           Reserving less fits, at the cost of reply length:
             LOGOS_MAX_OUTPUT_TOKENS=5744 claude-logos
           Otherwise pick a model Logos serves with a wider window (at least 37,024);
           the AI Tools page shows what each one gets.
```

The value it names is measured against the auto-compact point rather than the hard stop — a reservation that only clears the hard stop leaves a session compacting on every turn, which is a working session in name only. `--check` prints all of it and refuses nothing, which is what makes it the thing to run when a session will not start.

**One correction on the side:** the cold-start line claimed the first request brings a lane up "sized against this number". It is not — with no lane up only `max_model_len_overall` is known, which is the widest window the model has *ever* been calibrated for, while the planner sizes the new lane from whatever capacity is free at that moment. That is the second route to the same 400, and it is now stated honestly in the wrapper, on the page and in the troubleshooting table instead of promised.

## Verified

- `ai-tools.spec.ts`, 13 new cases: the thresholds, the reported window that produced the report, the narrow-lane/wide-model case, cold and unknown models, and every gating consequence. `ng test`: **121 passed**. `ng build`: clean.
- The wrapper run end-to-end against a stand-in `/v1/models`:
  - **32,768** → `BLOCKED`, real start exits 1, `LOGOS_MAX_OUTPUT_TOKENS=5744` then gives compact at 13,000 / hard stop at 23,000;
  - **40,000** → runs, "workable but tight";
  - **61,440** → clean, no warning (hard stop 37,212 against a 41,440 cap);
  - **cold** → still sizes from the maximum, now saying what that does and does not promise.
- `bash -n` on the shell wrapper. The PowerShell copy mirrors it line for line; no `pwsh` on this machine to execute it.

## Not in scope

Two things I deliberately left out, both worth their own decision:

1. **Clamping `max_tokens` server-side** to the window of the lane a request is routed to. It would make *any* client's window overestimate a shorter reply instead of a 400 — including the cold-start case above, and OpenCode. It also cuts against the "answer an honest 400" line from #810, so it should be a deliberate call, not a side effect of this fix.
2. **`ng test` does not run in CI** — `logos_test.yml` only builds the UI, so the 121 specs (8 files before this one) are local-only. Adding the step is a one-liner but changes what can go red on every PR.

https://claude.ai/code/session_01JLzCdi7A3JoWJW4qXmnaQD

### DIFF
--- a/logos/docs/context-windows.md
+++ b/logos/docs/context-windows.md
@@ -189,6 +189,54 @@ carries on. If a single turn grows past `window − reserve − 3000` it refuses
 send and asks for a `/compact` instead. Neither is an error the user has to
 recover from — the failure mode this replaces was a 400 from vLLM mid-turn.
 
+#### The floor: 37,024 tokens
+
+The deductions above are fixed, so there is a window below which Claude Code
+cannot run **at all** — and it is much higher than it looks. Its own opening
+prompt (system prompt plus the schemas of every tool it carries) is around
+13,000 tokens before the user has typed anything, and none of it is compactable:
+
+```text
+floor       = 13000 opening prompt + 20000 reservation + 3000 hard stop + 1024 headroom  = 37024
+comfortable = 13000 opening prompt + 20000 reservation + 13000 auto-compact + 1024        = 47024
+```
+
+A 32,768-token lane leaves `32768 − 20000 = 12768` tokens of input — one token
+short of the opening prompt. The session's **first** message comes back as
+
+```text
+This model's maximum context length is 32768 tokens. However, you requested
+20000 output tokens and your prompt contains at least 12769 input tokens
+```
+
+and there is nothing to compact, so it never recovers. Between the floor and
+47,024 the session runs but auto-compaction fires from the first message on.
+
+Two places enforce this, because a model can be chosen in either:
+
+- **The AI Tools page** disables such a model for Claude Code (`claudeCodeFitFor`
+  in `ai-tools.ts`), names the window in the option label and blocks the wizard
+  on the model step with the arithmetic spelled out. OpenCode is unaffected — it
+  is told what to reserve (`min(8192, context/2)`), so a narrow window costs it
+  reply length, not the session. The figure judged is
+  `max_model_len_current_min` — the one a request meets whichever deployment
+  answers — falling through to the wider figures only when it is absent, and a
+  model no lane serves is never judged.
+- **The wrapper** refuses to start and prints what is left, what it costs and
+  the `LOGOS_MAX_OUTPUT_TOKENS` value that would fit — measured against the
+  auto-compact point rather than the hard stop, since a reservation that only
+  clears the hard stop leaves a session compacting on every turn. `--check`
+  prints all of it without refusing anything, which is what makes it the thing
+  to run when a session will not start.
+
+Lowering the reservation is the only lever on the client side: at 32,768 tokens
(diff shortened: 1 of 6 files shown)


Claudia-Anthropica (approved):
@wasnertobias I reviewed the exact-head changes across both wrappers, the AI Tools gating, documentation, and the new Angular coverage. The floor arithmetic is consistent between the UI and default wrapper settings, and blocked selections cannot advance the wizard; I found no blocking defect. The snapshot had no required checks pending or failing, although one non-required worker image build was still in progress.

PR #912, opened 2026-09-06
`Iris`: Rerank global search retrieval and ground answers in searchable entities

Nayer-kotry:
Combines three PRs that shipped separately but form one answer pipeline: #710 (reranked two-stage retrieval), this PR (entity grounding) and the already-merged #913 (inline citations and streaming). #710 and #913 are closed as folded in here; this PR now targets `main` directly. Pairs with the Artemis counterpart ls1intum/Artemis#13711, which pre-fetches entity candidates with the palette's access filtering and renders inline citations; without it the wire additions are a no-op.

## Motivation

Production testing of global search surfaced several gaps at once. Retrieval ranked poorly and the answer bubble sometimes appeared and vanished without an answer, traced to a metadata fetch truncation and no defense against junk results. Questions about due dates, exam schedules, points, channels, FAQs or course offerings went unanswered even when the answer sat in the SearchableEntities collection, because the pipeline retrieved lecture content only. And the answer card was one prose blob with no link to its source chips, shown only after the full generation time because the JSON response format could not stream. This PR reworks retrieval, adds entities to the evidence pool, and switches the answer contract to plain text with inline markers so all three problems share one fix.

### Retrieval (two-stage rerank)

- Retrieval becomes two-stage: fetch a deep candidate set from both lecture collections, rerank it with the configured reranker, and gate the pool. If the reranker is unavailable or times out, search falls back to the fused ordering, so it never degrades below pre-reranker behavior.
- Rerankers are normalized behind one provider-neutral interface (`RerankModel` / `RerankResponse`). The Cohere client conforms to it, and a new `vllm_rerank` client speaks the `/v1/rerank` endpoint vLLM-served cross-encoders expose (Qwen3-Reranker-8B on Logos).
- The rerank cutoff is a junk floor at 0.10, calibrated against what irrelevant candidates score (junk peaks at 0.065), rather than a relevance threshold that sat inside reranker noise.
- Answer-path unit expansion: once a candidate survives the floor, the rest of its lecture unit is fetched by structural join instead of ranked, so siblings do not have to win a ranking slot they should not need.
- Query embeddings use the Qwen3 retrieval instruction prefix. HyDE is removed from the retrieval path (no quality change measured, lower latency, fewer empty-output model errors). The metadata fetch truncation is fixed.

### Entity grounding

- Artemis pre-fetches entity candidates with its full role-based visibility filtering and forwards them in the request DTO (`entityCandidates`, optional): channel membership, exam assignment and role-dependent release rules live in the Artemis database, so Pyris never queries the entity collection itself.
- Pyris renders each candidate into a text card (type, course, verbalized dates, points, duration, language, description). Measured on Qwen3-Reranker-8B, bare titles score 0.05-0.07 (below the junk floor) while rendered cards score 0.49-0.57.
- Entity cards join the shared rerank pool, so one cross-encoder scores entities and lecture content on the same calibrated scale. A representation pass guarantees the best above-floor entity cards (2 slots) and content snippet (1 slot) a place in the answer context.
- A pointer tier admits entity cards from just below the floor (0.08-0.10) only when nothing clears it, answered through a dedicated direct-the-student prompt (the grounded prompt vetoes pointer-only contexts ~80 percent of the time; the navigate prompt was 8/8 in English and German). A null-to-navigate fallback retries once over entity sources when the grounded answer comes back null.
- The grounded prompt receives the Artemis clock ("Today is ...") so semester course copies with conflicting dates resolve toward today.

### Citations and streaming

- The answer contract is plain markdown text with sentence-level inline markers (`worth 10 points.[3]`) instead of a JSON envelope, so a partial snapshot is renderable. A literal `!none!` sentinel carries the no-answer state; a streaming gate buffers output until it can no longer become that sentinel, so a no-answer run never flashes text.
- Markers are sanitized before reaching a student (out-of-range indices stripped, repeated markers collapsed, distinct chains preserved as multi-source citations) and renumbered onto the returned source list after the used-sources filter, so `[n]` always indexes `response.sources` on the wire.
- Wire changes are additive per the freeze rules: `entityCandidates` on the request, `entitySources` and `partialResult`/`partialSeq` on the response and status DTOs, all optional. An old Artemis is a no-op on both sides.

## Testing

- 52-query labeled suite across 18 categories against local Weaviate with the Logos-served Qwen3-Reranker-8B: baseline 19/52, this design 51/52 on retrieval, 36/36 scored end-to-end answers.
- Head-to-head A/B on the test server: false nulls on answerable queries 30 percent to 3 percent, top-1 hit rate 0.33 to 0.78, used-source precision 0.67 to 0.95.
- Semester-copy conflict suite (seeded WS25 clone with the same exercise and exam at different dates): 18/18 correct dates, 18/18 semester-labeled answers.
- Unit tests: card rendering, additive wire shapes, the shared rerank gate, expansion join semantics, floor gating, the rerank response contract for both providers, marker sanitation and renumbering, streaming.
- Manual end-to-end verification of the merged branch against a real course, lecture and exercise, through a real LLM call: a lecture-content question streamed correctly and cited the right slide; an entity-only question ("is there an rnn quiz") was answered and cited from the exercise card with no lecture content available; an off-topic question produced no leaked partial text and a clean null answer.

#### Code Review
- [ ] Code Review 1
- [ ] Code Review 2
#### Manual Tests
- [x] Local end-to-end smoke test (retrieval, entity grounding, citations, streaming, no-answer suppression)
- [ ] Test server pass

## Summary by CodeRabbit

- **New Features**
  - Global Search includes relevant courses, lectures, exercises, and catalog materials.
  - Supports course filtering, semester-aware results, inline citations, and streamed answer drafts.
  - Suggests catalog materials when teaching content is unavailable.
  - Supports additional reranking providers to improve relevance.

- **Bug Fixes**
  - Improved citation validation, duplicate-source handling, and unsupported-answer suppression.
  - Improved result diversity and filtering across related course content.
  - Standardized search dates and times to UTC.
  - Improved fallback behavior when reranking is unavailable.


Nayer-kotry:
@coderabbitai review

iris/src/iris/pipeline/global_search_pipeline.py
```diff
@@ -117,119 +457,239 @@ def __call__(
...
+            (time.perf_counter() - t_answer) * 1000,
+            len(grounded_sources),
+            len(context),
+            len(raw),
+            getattr(answer_usage, "num_input_tokens", None),
+            getattr(answer_usage, "num_output_tokens", None),
         )
-
-        return GlobalSearchResponseDTO(answer=answer, sources=used_sources)
+        return raw
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed, but not with the suggested `sender.on_delta(None)` before `sender.stop()` — that runs after the null-to-navigate fallback has already completed, so it doesn't actually address the window described: the fallback runs a second, non-streamed LLM call while the client keeps showing the first (discarded) draft's raw markers for the whole round-trip. Fixed by clearing the stream via `stream_handler(None)` right when the fallback is decided to run, before it starts, instead of after everything is already done.

Reproduced the stale-draft behavior with a standalone script against the real PartialResultSender/_SentinelGateStreamHandler before fixing, verified the fix closes it, and added a permanent regression test (TestNavigateFallbackClearsTheStream in test_answer_streaming.py). See 0a6042ae2.

iris/tests/test_rerank_model_contract.py
```diff
@@ -0,0 +1,141 @@
...
+        llm_manager=type(
+            "_Manager", (), {"get_llm_by_id": lambda self, _id: _StubReranker()}
+        )(),
+    )
+
+    ranked = handler.rerank(
+        query="q", documents=documents, top_n=2, content_field_name="text"
+    )
+
+    assert ranked == [documents[2], documents[0]]
```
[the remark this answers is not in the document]
Nayer-kotry: This was already fixed in a previous commit (4f571ff07) before this thread was opened — the test at this line now uses a non-empty mixed-validity document list (`[_Doc("valid text"), _Doc(None)]`) instead of `[]`, exercising the filter comprehension as suggested. Resolving.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -289,39 +528,415 @@ def _map_search_objects(
...
+    ) -> list[Any]:
+        """Every object of the given units, fetched by join rather than ranked."""
+        if not unit_ids:
+            return []
+        return collection.query.fetch_objects(
+            filters=Filter.by_property(schema.LECTURE_UNIT_ID.value).contains_any(
+                unit_ids
+            ),
+            limit=settings.global_search_expand_fetch_limit,
+        ).objects
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed and fixed as proposed. `_fetch_unit_objects` now accepts an optional `extra_filter`, and the transcription lane passes `PAGE_NUMBER == -1`, matching the identical filter `_search_video_transcriptions` already applies for the same reason. Added a regression test asserting the filter is actually sent (and that the segment lane keeps every page). See 0a6042ae2.

iris/src/iris/web/routers/pipelines.py
```diff
@@ -352,8 +358,9 @@ def run_global_search_pipeline_worker(dto: GlobalSearchRequestDTO, request_id: s
         return
 
     try:
+        started = time.perf_counter()
         intent = classify_intent(dto.query)
-        logger.debug(
+        logger.info(
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed and fixed as proposed. See 0a6042ae2.

iris/tests/test_lecture_global_search_retrieval_units.py
```diff
@@ -0,0 +1,251 @@
...
+                LectureUnitSchema.BASE_URL.value: "http://localhost:8086",
+                LectureUnitSchema.LECTURE_UNIT_NAME.value: "instance 6's unrelated unit",
+            }
+        )
+        retrieval.lecture_unit_collection = Mock()
+        retrieval.lecture_unit_collection.query.fetch_objects.return_value = Mock(
+            objects=[instance_0_unit, instance_6_unit]
+        )
+
+        result = retrieval._fetch_lecture_units([411])
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed and fixed. Added a dedicated test (test_fetch_lecture_units_sends_the_formula_limit_to_weaviate) asserting `fetch_objects`'s actual `limit` kwarg matches `max(100, len(unit_ids) * 10)`, since neither existing test touched the real call. See 0a6042ae2.

iris/tests/test_rerank_model_contract.py
```diff
@@ -0,0 +1,218 @@
...
+                )()
+            },
+        )(),
+    )
+
+    ranked = handler.rerank("q", documents, top_n=2, content_field_name="text")
+
+    # Only the valid document ever reaches the reranker; the one missing "text" was
+    # filtered out before the call, not passed through as empty content.
+    assert ranked == [documents[0]]
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed and fixed as proposed. The stub reranker now records what it actually receives and the test asserts `received_documents == [["valid text"]]`. See 0a6042ae2.

iris/tests/test_answer_streaming.py
```diff
@@ -0,0 +1,166 @@
...
+        pipeline(
+            query="where is this covered",
+            intent=SearchIntent.TRIGGER_AI,
+            stream_handler=deltas.append,
+        )
+
+        # A reset (None) must reach the client BEFORE the fallback's answer is
+        # generated, not only after __call__ has already returned — otherwise the
+        # discarded first draft stays the last thing shown for the whole fallback.
+        assert None in deltas
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed and fixed as proposed. The test now wraps `_generate_answer` and asserts `deltas[-1] is None` at the point the fallback is invoked (`navigate=True`), not just that `None` appears somewhere in the final list. See 1f6f84e30.

iris/tests/test_lecture_global_search_retrieval_units.py
```diff
@@ -0,0 +1,308 @@
...
+        _VisibilityPolicy.from_context(None),
+    )
+
+    calls = {
+        call.args[0]: call for call in retrieval._fetch_unit_objects.call_args_list
+    }
+    segment_call = calls[retrieval.collection]
+    transcription_call = calls[retrieval.transcription_collection]
+
+    assert len(segment_call.args) == 3, "the segment lane keeps every page"
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed and fixed as proposed. See 1f6f84e30.

Nayer-kotry:
@coderabbit codereview

iris/src/iris/pipeline/global_search_pipeline.py
```diff
@@ -32,20 +38,327 @@
...
+        answer = parsed.get("answer")
+        # Treat null, "" and non-string values as no answer.
+        answer = answer if isinstance(answer, str) and answer else None
+        if answer is None:
+            logger.info("[global-search] outcome=llm_null_json raw=%r", raw[:300])
+        raw_indices = parsed.get("used_sources")
+        used_indices = {
+            i - 1
+            for i in (raw_indices if isinstance(raw_indices, list) else [])
+            if isinstance(i, int) and i >= 1
```
Claudia-Anthropica: @Nayer-kotry [medium] Validate recovered `used_sources` against `num_sources`. Both this JSON branch and the trailing-schema branch accept any positive integer, so `{"answer":"Yes.","used_sources":[99]}` leaves `used_indices` non-empty, bypasses the ungrounded-answer guard, and returns an answer with no corresponding source. Filter to `1 <= i <= num_sources` in both recovery paths.

In `iris/src/iris/pipeline/global_search_pipeline.py`, out-of-range `used_sources` values can make an unsupported answer look grounded while no source is returned. Bound both JSON and trailing-text indices to the available source count before applying the grounding guard.

</details>

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -289,39 +528,430 @@ def _map_search_objects(
...
+        timeout_s: float = _RERANK_TIMEOUT_S,
+    ) -> tuple[float, list[float]] | None:
+        """Rerank candidate snippets; return (duration_ms, per-candidate relevance).
+
+        Returns None when no reranker is configured, on any API failure, or on
+        timeout — the caller then falls back to the fused ordering, so the search
+        can never degrade below pre-reranker behavior. Deliberately does NOT
+        disable itself process-wide on failure (unlike RerankRequestHandler).
+        """
+        if self.reranker_model_id is None or len(candidates) < 2:
```
Claudia-Anthropica: @Nayer-kotry [medium] A one-item candidate pool must still be scored. Returning `None` for `len(candidates) < 2` routes the only hit through the fused fallback, so generation skips `global_search_rerank_floor` and can answer from a lone junk hit. Only skip an empty pool; rerank a single document so the floor remains effective.

In `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, a single candidate bypasses reranking and the configured junk floor. Return early only for an empty candidate list and rerank one-item pools normally.

</details>

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -134,142 +380,135 @@ def search(
                 "Access context yields no accessible courses; skipping search."
             )
             return []
-        query_embedding = self.llm_embedding.embed(query)
+        if entity_sources:
```
Claudia-Anthropica, line 530: @Nayer-kotry [medium] Do not discard pre-authorized entity candidates when the lecture access scope is empty. The preceding early return runs before `entity_sources` are deduplicated or reranked, so a user with no role-based course IDs gets no answer even when Artemis supplied visible public course or catalog entities. Skip the Weaviate content lanes for an empty lecture scope, but continue through the entity-only rerank and pointer path.

In `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, an empty effective lecture-course scope currently drops already-authorized entity candidates. Preserve the empty-scope content restriction while continuing entity-only retrieval and reranking.

</details>

iris/src/iris/pipeline/global_search_pipeline.py
```diff
@@ -117,119 +457,251 @@ def __call__(
...
+        # pointing at the wrong (or out-of-range) position in the final list.
+        ordered_used = sorted(used_indices)
+        answer = renumber_citation_markers(
+            answer,
+            {old + 1: new + 1 for new, old in enumerate(ordered_used)},
         )
+
         self._append_tokens(
-            self.hyde_llm.tokens, PipelineEnum.IRIS_GLOBAL_SEARCH_PIPELINE
+            self.answer_llm.tokens, PipelineEnum.IRIS_GLOBAL_SEARCH_PIPELINE
```
Claudia-Anthropica: @Nayer-kotry [medium] Preserve usage from the first answer call when the navigation fallback runs. `IrisLangchainChatModel.tokens` is overwritten by every invocation, but the pipeline appends it only here after the optional second call, so fallback requests report only the navigation call and omit the initial grounded call from token and cost accounting. Append a usage snapshot after each invocation or aggregate both calls before returning.

In `iris/src/iris/pipeline/global_search_pipeline.py`, navigation fallback overwrites the first answer call's token usage before it is recorded. Record both LLM invocations in `pipeline.tokens` without double-counting the single-call path.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The retrieval and streaming paths are well covered, and the captured CI checks pass. However, the current head still allows invalid source indices to defeat grounding, skips the junk floor for single candidates, drops authorized entity-only searches for an empty lecture scope, and under-reports token usage during navigation fallback.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -40,9 +46,235 @@
...
+    Distinct offerings (series, cross-listings, other courses) have their
+    own keys and are never merged.
+    """
+    if not entity_sources or _DATED_QUERY_RE.search(query):
+        return entity_sources
+    now = now or datetime.now(timezone.utc)
+    best: dict[tuple[str, str, str], EntitySourceDTO] = {}
+    order: list[tuple[str, str, str]] = []
+    for source in entity_sources:
+        key = _twin_key(source)
```
Claudia-Anthropica, line 305: @Nayer-kotry [medium] This key collapses every entity with the same type, normalized course name, and normalized title, even when the entities have different IDs in the same course. For example, two distinct exercises named `Quiz` in one course are reduced to one before reranking, so a relevant exercise and its dates can disappear. Preserve distinct `entity_id` values within a course and apply semester-instance selection only across different course IDs.

In `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, `dedupe_semester_twins` incorrectly merges distinct same-title entities belonging to the same course. Change the grouping so candidates with different `entity_id` values in one `course.id` are preserved, and apply semester preference only between candidate groups from different course IDs.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The four findings from the previous review are fixed on this head, including the entity-only empty-scope path. One correctness issue remains: semester deduplication can discard distinct same-title entities from a single course before reranking. Several non-required CI jobs were still running in the captured snapshot.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Claudia-Anthropica: @Nayer-kotry Thanks, this is fixed now: an empty lecture scope skips only the content lanes while retaining the pre-authorized entity pool, with regression coverage.

iris/src/iris/pipeline/global_search_pipeline.py
Nayer-kotry: Confirmed and fixed as proposed. Bounded both the JSON and trailing-schema recovery paths to `1 <= i <= num_sources`. Reproduced first: fed `{"answer":"Yes.","used_sources":[99]}` through `parse_answer_response` and confirmed it returned a non-null answer with an empty `used_sources` list at the caller, bypassing the ungrounded-answer guard. Added regression tests covering both paths, in-range and out-of-range. See 4acbf3e7d.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Nayer-kotry: Confirmed and fixed as proposed. `_safe_rerank` now only short-circuits on `not candidates` (empty), not `len(candidates) < 2`. Reproduced first: called `_safe_rerank` with a single candidate and a configured reranker and confirmed it returned `None` without ever calling the reranker. Added a regression test with a mocked `LlmManager` confirming a lone candidate now reaches the reranker. See 4acbf3e7d.

iris/src/iris/pipeline/global_search_pipeline.py
Nayer-kotry: Confirmed and fixed as proposed. `self.answer_llm.tokens` is reassigned to a fresh `TokenUsageDTO` per invocation, so it's now appended right after each of the two calls instead of once at the end. Added a regression test simulating both calls and asserting both usage snapshots are recorded, not just the last one. See 4acbf3e7d.

Claudia-Anthropica (changes requested):
@Nayer-kotry The three replied-to findings are fixed on this head: recovered indices are bounded, lone candidates are reranked, and both fallback LLM calls record token usage. The unresolved semester-deduplication defect remains: `_twin_key` still omits `course.id` and `entity_id`, so distinct same-title entities in one course collapse before reranking. Iris Test and Lint succeeded; two non-required container-image builds were still in progress.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Nayer-kotry: Fixed in 6abfa5e2b. The dedup key now discriminates by `course.id`: a key group is only collapsed to one instance when every entity sharing it also shares the same course id (true semester twins are always different course instances), otherwise every entity in the best-represented course is kept. Added a regression test (`test_same_course_same_title_entities_are_both_kept`) and fixed the existing twin tests, which had accidentally been hardcoding `course_id=1` for both sides of every twin pair.

iris/src/iris/pipeline/global_search_pipeline.py
```diff
@@ -117,119 +459,256 @@ def __call__(
...
+        # the used sources, so renumber onto that list. Must run AFTER the
+        # fallback — renumbering before it left a fallback answer's markers
+        # pointing at the wrong (or out-of-range) position in the final list.
+        ordered_used = sorted(used_indices)
+        answer = renumber_citation_markers(
+            answer,
+            {old + 1: new + 1 for new, old in enumerate(ordered_used)},
+        )
+
+        used_lecture = [
```
Claudia-Anthropica, line 759: @Nayer-kotry [medium] Renumber citations against the order actually serialized on the wire. `used_sources` may be ordered `[entity, lecture]`, but these comprehensions serialize it as `sources=[lecture]` and `entitySources=[entity]`, with no field preserving the interleaving; a consumer therefore cannot know which object `[1]` references. Establish the final deterministic cross-type wire order before renumbering, or transmit explicit citation indices/order metadata.

In `iris/src/iris/pipeline/global_search_pipeline.py`, splitting ranked used sources into `sources` and `entitySources` discards the ordering used to number mixed-type citations. Renumber against a deterministic serialized order or add explicit ordering metadata, and cover an entity-before-lecture response with a regression test.

</details>

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -107,169 +366,193 @@ def __init__(self, client: WeaviateClient, local: bool = False):
...
+        auto_cut: bool,
+        telemetry: "_SearchTelemetry",
+    ) -> tuple[list[Any], list[Any]]:
+        """Query both collection lanes in parallel.
+
+        Lanes fetch CANDIDATES at _LANE_DEPTH; the final ordering is decided
+        by the reranker (or the fused scores when no reranker is available).
+        """
+        auto_limit = _AUTOCUT_GROUPS if auto_cut else None
+        lane_depth = max(limit, _LANE_DEPTH)
```
Claudia-Anthropica: @Nayer-kotry [medium] Do not stop candidate acquisition at a fixed 25 rows before visibility mapping. Unit release dates and slide visibility are checked later in `_segment_to_dto`/`_transcription_to_dto`, so 25 higher-ranked unreleased or hidden rows can exhaust this window while a valid row at position 26 is never considered; the previous loop expanded until each lane had enough visible results or was exhausted. Restore adaptive expansion after visibility filtering, or push equivalent visibility constraints into the Weaviate queries before applying this cap.

In `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, the fixed lane depth can hide accessible results because release and visibility filtering happens after retrieval. Expand candidate windows until the requested visible-result count is filled or the lane is exhausted, or enforce all visibility predicates in the database query, and add a regression test with more than 25 filtered hits ahead of a visible hit.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The one-course regression test does not close the prior deduplication issue: mixed-course groups can still discard distinct same-title entities from the non-selected course. Two additional correctness issues remain: mixed lecture/entity citations lose their ranked order when split across response arrays, and fixed-depth retrieval can stop before post-retrieval visibility filtering reaches an accessible result.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Claudia-Anthropica: @Nayer-kotry This remains incomplete on the current head. Once a same-title group spans courses, `result.extend(by_course[best_course_id])` discards every entity from the other course; with two distinct `Quiz` IDs in course 1 and one newer-semester `Quiz` in course 2, both course-1 entities disappear even though only one can be treated as the twin. The new test exercises only the single-course branch—preserve the entire ambiguous group unless actual twins can be paired across courses.

Claudia-Anthropica (changes requested):
@Nayer-kotry The new commit fixes one direction of the deduplication case, but a singleton losing course is still discarded when the selected course has multiple same-title entities, despite no actual twin pairing being established. The two unresolved correctness findings also remain: mixed lecture/entity citations lose their numbered order when split across response arrays, and each retrieval lane is capped at 25 before visibility filtering. The captured Iris Test, Lint, and container-image jobs were still in progress.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Nayer-kotry: Right, my previous fix conflated "multiple courses share this key" with "these are genuine semester twins," so it still discarded a losing course's entities as a group even when they were only coincidentally sharing the key. Fixed in 08d975b54: a losing course's entry is now only dropped when it contributes exactly one entity under the key (the winner's unambiguous twin analog); a losing course contributing more than one entity is evidence it isn't actually a repeated offering, so those entities all survive. Added `test_multiple_distinct_entities_in_the_losing_course_all_survive` covering exactly this.

iris/src/iris/pipeline/global_search_pipeline.py
Nayer-kotry: Fixed in 800a4749a. Renumbering now follows the same lecture-then-entity split order the response actually serializes (all lecture-sourced markers numbered first, then entity-sourced ones), instead of raw ranked position, so a marker always lands on the array the client actually resolves it against. Added a regression test with an entity ranked between two lecture sources.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Nayer-kotry: Fixed in 800a4749a. The search now retries at double the lane depth (capped by the new `global_search_lane_depth_max` setting) whenever the first fetch returns a full window that still leaves fewer than `limit` visible candidates after filtering, stopping once enough are found or a lane returns fewer hits than requested (genuinely exhausted). Added regression tests for both the expansion and the exhaustion stop condition.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -107,169 +378,230 @@ def __init__(self, client: WeaviateClient, local: bool = False):
...
+        self.reranker_model_id = resolve_reranker_model(local)
+
+    def embed_retrieval_query(self, query: str) -> list[float]:
+        """Embed a USER QUERY with the Qwen3 retrieval instruction prefix.
+
+        Query-side only (asymmetric retrieval): documents are ingested raw and
+        must stay raw — both sides carrying the instruction cancels the
+        benefit. Never apply this to document-shaped text.
+        """
+        return self.llm_embedding.embed(QWEN3_RETRIEVAL_INSTRUCTION + query)
```
Claudia-Anthropica: @Nayer-kotry [medium] This unconditionally applies a Qwen3-specific asymmetric-retrieval instruction, but `global_search_pipeline.embedding` remains provider-agnostic and defaults to `oai-embedding-small` in `application.example.yml`. Standard deployments therefore embed the instruction itself as query content with an OpenAI model, rather than invoking the calibrated Qwen3 protocol, which can alter retrieval semantics and invalidate the rerank-floor calibration. Gate this prefix behind an explicit embedding-model capability or configuration so non-Qwen3 providers receive the original query.

In `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, `embed_retrieval_query` applies a Qwen3-specific instruction to every configured embedding provider. Add an explicit capability or configuration switch and apply the prefix only to Qwen3 instruction-aware embeddings, leaving other providers' queries unchanged.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The citation-order fix now matches the serialized source order. However, `dedupe_semester_twins` still drops a singleton losing course when the selected course has multiple same-title entities despite no one-to-one twin pairing, and the query embedding now assumes Qwen3 regardless of the configured provider. The captured snapshot has no required checks or failures, although several non-required jobs remained unfinished.

iris/src/iris/web/routers/search.py
```diff
@@ -19,12 +20,31 @@
...
+        dto.limit,
+        dto.course_ids,
+    )
+    return _traced_lecture_search(dto)
+
+
+@observe(name="Lecture Search", as_type="retriever")
+def _traced_lecture_search(
+    dto: LectureSearchRequestDTO,
+) -> list[LectureSearchResultDTO]:
```
Claudia-Anthropica, line 43: @Nayer-kotry [medium] The new `excludeCourseIds` path is wired only through `GlobalSearchRequestDTO` and the asynchronous answer worker. `LectureSearchRequestDTO` has no equivalent field and this synchronous `/api/v1/search/lectures` call forwards only `course_ids`, so an unrestricted caller with no inclusion ceiling still sees excluded courses in the instant-results list. Add the exclusion field to the lecture-search request and forward it here, with route-level regression coverage.

In `iris/src/iris/web/routers/search.py`, the synchronous lecture-search endpoint does not accept or forward course exclusions, so excluded courses remain visible to unrestricted callers. Add `excludeCourseIds` to `LectureSearchRequestDTO`, pass it as `exclude_course_ids` to the retriever here, and add a regression test for an exclusion-only request.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The new exclusion filter reaches the asynchronous answer pipeline, but the synchronous lecture-results route still cannot receive or forward it, so excluded courses can remain in instant results. Two previously identified correctness issues also remain: `embed_retrieval_query` still applies the Qwen3 prefix to the default `oai-embedding-small`, and `dedupe_semester_twins` still drops a singleton losing course when the selected course has multiple same-title entities. The captured snapshot has no failed or required checks, while several non-required jobs remained unfinished.

iris/src/iris/web/routers/search.py
Nayer-kotry: Fixed in d6b041f7e. LectureSearchRequestDTO now accepts excludeCourseIds and the route forwards it into the same retrieval-layer exclusion support (contains_none) added for the answer path. Added a unit test proving the route forwards it to the retriever, and a wire-shape test for the DTO field; a FastAPI TestClient-based route test would be new infrastructure this codebase doesn't otherwise use, so I matched the existing pattern of testing the route's underlying function directly.

Claudia-Anthropica (changes requested):
@Nayer-kotry The synchronous exclusion path is fixed: `LectureSearchRequestDTO` now accepts `excludeCourseIds` and `_traced_lecture_search` forwards it. Two concrete correctness defects remain on this head: `embed_retrieval_query` still unconditionally prepends the Qwen3 instruction even though the checked-in default is `oai-embedding-small`, and `dedupe_semester_twins` still drops a singleton losing course whenever the selected course has multiple same-title entities, without establishing a one-to-one twin. The captured snapshot has no failed or required checks, while nine non-required checks were still pending.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Nayer-kotry: Fixed in 647430925, mirroring how VllmRerankModel already gates its own Qwen3 instruction fold for the reranker (an optional instruction field, unset by default). OpenAIEmbeddingModel gains an optional retrieval_instruction field; embed_retrieval_query now only folds it in when the resolved model's config actually sets it (looked up via LlmManager, same pattern already used for the reranker two lines below), so the oai-embedding-small default and any other plain OpenAI/Azure entry embed the raw query. Documented with a commented example in llm_config.example.yml. Note: a production Qwen3-Embedding deployment needs retrieval_instruction added to its own model entry to keep the calibrated behavior — that's an operational config change outside this repo, flagged in the commit message.

Claudia-Anthropica (changes requested):
@Nayer-kotry The embedding-instruction fix is correct: the default OpenAI embedding now receives the raw query. The previously identified [medium] semester-deduplication defect remains: `dedupe_semester_twins` still drops a singleton losing course when the selected course contains multiple same-title entities, although no unique twin pairing exists; preserve the ambiguous group unless a one-to-one pairing can be established. The snapshot has no failed or required checks, though several non-required jobs were still running.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -103,173 +392,272 @@ def __init__(self, client: WeaviateClient, local: bool = False):
...
-            ]
-            transcription_scored = [
-                item
-                for item in scored
-                if item[1].lecture_unit.source_type == "lecture_unit_video"
-            ]
-            segment_complete = (
-                len(segment_scored) >= limit or len(seg_objects) < candidate_limit
+            exhausted = (
+                len(seg_objects) < lane_depth and len(trans_objects) < lane_depth
```
Claudia-Anthropica: @Nayer-kotry [medium] The answer path calls this with `auto_cut=True`, so Weaviate can return fewer than `lane_depth` because `auto_limit=2` stopped at a score cliff, not because both lanes are exhausted. If those returned groups are hidden or unreleased, this sets `exhausted=True` and line 592 returns before fetching a visible result below the cutoff. When visibility filtering leaves fewer than `limit` under autocut, retry without autocut before applying the length-based exhaustion check.

In `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, `_search_lanes_until_visible` treats autocut-shortened lane results as exhausted, so hidden or unreleased top score groups can prevent a visible lower-ranked result from being retrieved. When post-retrieval filtering leaves fewer than `limit` while `auto_cut` is enabled, retry the lanes with `auto_limit=None` at increasing depth before using the length-based exhaustion check, and add a regression test where autocut returns fewer than the lane depth but a visible hit exists below the cutoff.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The latest commit fixes the semester-deduplication case by preserving ambiguous multi-entity groups. One [medium] retrieval issue remains: answer-path autocut can masquerade as lane exhaustion after visibility filtering and prevent lower-ranked accessible content from being fetched. The snapshot contains no failed or required checks, while seven non-required checks were still running.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Nayer-kotry: Confirmed and fixed in 65b9507df. auto_cut's score-cliff cutoff is a quality signal, not exhaustion; the length-based exhausted check now only applies once autocut is off. The first time visibility filtering leaves too few results while autocut was on, the search drops autocut and retries the same depth with the fixed limit before trusting exhaustion or expanding further. Added a regression test with autocut returning fewer than lane_depth while a visible hit exists once autocut is disabled at the same depth.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -448,61 +1269,55 @@ def _fetch_slides_by_display_page(
...
+        """Map a segment hit to a DTO; on failure return (None, drop_reason)."""
         if policy is None:
             policy = _VisibilityPolicy.from_context(None)
-        if not is_segment_visible(props, policy.now):
-            return None
-
         snippet = props.get(LectureUnitSegmentSchema.SEGMENT_SUMMARY.value)
-        if not snippet or snippet.startswith(_EMPTY_SEGMENT_PREFIX):
-            return None
+        if not snippet:
```
Claudia-Anthropica, line 1405: @Nayer-kotry [medium] This now accepts known ingestion placeholders such as `There is no content...` as valid snippets. `_safe_rerank` explicitly falls back to fused ordering when the model is missing, times out, or returns an invalid response, and instant search can disable reranking entirely; in those paths the placeholder reaches the UI or answer context even though the base implementation excluded it. Keep a deterministic placeholder exclusion before reranking, or add and filter the proposed structured empty-content flag.

In `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, `_segment_to_dto` no longer rejects the ingestion placeholder prefix, so reranker-disabled and reranker-failure fallbacks can return empty-slide text as real evidence. Restore a provider-independent placeholder filter before candidates enter the fallback path, or persist and filter a structured empty-content flag.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The autocut exhaustion fix is correct: when too few visible candidates survive, the code now retries at the same depth without autocut before trusting lane length. One medium fallback regression remains: known empty-content placeholders can now be returned whenever reranking is disabled or unavailable. The captured Iris Test passed; Lint and container builds were still non-required and in progress.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -40,9 +46,317 @@
...
+            return resolve_model(pipeline_id, "default", role, local=local)
+        except LlmConfigurationError:
+            continue
+    logger.info("[LectureSearch] no reranker configured — using fused ordering")
+    return None
+
+
+def _is_low_information(snippet: str) -> bool:
+    stripped = snippet.strip()
+    if len(stripped) < _LOW_INFO_MIN_CHARS:
```
Claudia-Anthropica: @Nayer-kotry [medium] This unconditional length cutoff classifies valid concise content as junk. For example, `A stack is LIFO.` is a complete, answer-bearing summary, but both DTO mappers discard it here before reranking, so it disappears from both the instant-results UI and answer context. Restrict deterministic filtering to known placeholder signatures or a structured empty-content flag instead of rejecting all text below an arbitrary length.

In `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, `_is_low_information` drops every snippet shorter than 25 characters even when it contains legitimate course information. Remove the generic length-only rejection and retain only explicit placeholder detection or a structured empty-content check, with regression coverage for a concise factual snippet.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The original placeholder-fallback defect is fixed: known empty-content text is filtered before reranking. However, the replacement also rejects every snippet shorter than 25 characters, removing concise legitimate content from both instant results and answer retrieval. The snapshot has no failed or required checks, while several non-required jobs remained in progress.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Claudia-Anthropica: @Nayer-kotry Thanks, the original placeholder-fallback issue is fixed on this head; known empty-content placeholders are again filtered before fallback results are returned.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Nayer-kotry: Fixed in bcc5cca8: dropped `_LOW_INFO_MIN_CHARS` entirely. `_is_low_information` now relies only on the deterministic placeholder-phrase patterns, which target known junk shapes directly instead of guessing from length. Added a regression test asserting `"A stack is LIFO."` is kept.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -40,9 +46,317 @@
...
+# unfiltered straight into the UI or answer context.
+#
+# Deliberately pattern-based rather than a length cutoff: a genuinely short but
+# complete summary ("A stack is LIFO.") is valid content, not junk, and a raw
+# character-count floor cannot tell the two apart.
+_LOW_INFO_PATTERNS = re.compile(
+    r"^there is no content"  # ingestion placeholder (empty slide)
+    r"|no spoken content"  # transcription placeholder (silent/music video)
+    r"|solely of repeated"  # music-only / repeated-symbol transcriptions
+    r"|single (word|phrase|character|letter)",  # one-word junk slides
```
Claudia-Anthropica: @Nayer-kotry [medium] Only the first alternative is anchored, so generic phrases in the remaining alternatives match anywhere in otherwise valid content. For example, `A token can represent a single character.` is classified as low-information and removed from both instant results and answer retrieval before reranking. Restrict these alternatives to the exact ingestion-placeholder shapes instead of matching `single word/phrase/character/letter` or `solely of repeated` in arbitrary explanatory text.

In `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, `_LOW_INFO_PATTERNS` rejects legitimate snippets containing generic phrases such as `single character`. Replace the broad alternatives with anchored, complete placeholder signatures and add regression coverage proving explanatory text containing those phrases is retained.

</details>

iris/src/iris/pipeline/global_search_pipeline.py
```diff
@@ -32,20 +38,329 @@
...
+def _location_label(source: LectureSearchResultDTO) -> str:
+    """Human-readable slide/video position tag for the numbered context."""
+    page = source.lecture_unit.page_number
+    if page == -1:
+        meta = source.lecture_unit.display_meta or "video"
+        return f"Video @ {meta}"
+    return f"Slide {page}"
+
+
+_CITATION_MARKER_RE = re.compile(r"\[(\d+)\]")
```
Claudia-Anthropica: @Nayer-kotry [medium] This pattern treats every numeric bracket in the generated prose as a citation, regardless of its position. A programming answer such as `Use array[0] as the pivot.[1]` loses `[0]`, while an in-range expression such as `element[1]` is incorrectly collected and rendered as a source citation. Since the prompt defines citations as marker chains immediately after claim punctuation, constrain parsing to that position and preserve ordinary bracketed numeric content.

In `iris/src/iris/pipeline/global_search_pipeline.py`, `_CITATION_MARKER_RE` interprets ordinary numeric brackets such as `array[0]` as citations and corrupts answer text or attribution. Recognize only post-claim citation markers and their directly adjacent marker chains, and add a regression test preserving numeric brackets inside programming prose.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The replacement placeholder filter avoids the old length cutoff, but its unanchored generic alternatives still discard legitimate lecture content. Citation sanitation also mistakes ordinary numeric brackets in answer text for source markers; both correctness issues are detailed inline. The captured Iris Test and Lint checks passed, while two non-required image builds remained in progress.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Nayer-kotry: Fixed in 9492653a: anchored every alternative in `_LOW_INFO_PATTERNS` to the start of the snippet, since they describe what the whole summary IS, not a phrase that can appear incidentally inside real content. Added a regression test with your exact example ("A token can represent a single character.") asserting it is kept.

iris/src/iris/pipeline/global_search_pipeline.py
Nayer-kotry: Fixed in 6f591b42: `_CITATION_MARKER_RE` now requires a `[n]` bracket to be immediately preceded by claim punctuation (`.!?`) or another marker (`]`, for chains like `[1][2][3]`), matching exactly how the prompt defines marker placement. `array[0]` and `element[1]` in ordinary prose are left alone; only markers in the two positions the prompt actually produces are recognized. Added regression tests for both your examples: `array[0]` preserved as prose, `element[1]` not collected as a citation, plus a renumbering-path equivalent.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -103,173 +422,292 @@ def __init__(self, client: WeaviateClient, local: bool = False):
...
-            ]
-            transcription_scored = [
-                item
-                for item in scored
-                if item[1].lecture_unit.source_type == "lecture_unit_video"
-            ]
-            segment_complete = (
-                len(segment_scored) >= limit or len(seg_objects) < candidate_limit
+            deduped = _dedupe_by_snippet(scored, telemetry)
+            if len(deduped) >= limit:
```
Claudia-Anthropica: @Nayer-kotry [medium] This return still measures visibility across both lanes together. If five visible slide candidates survive while all 25 transcription candidates are hidden, `len(deduped) >= limit` returns immediately even though the saturated transcription lane may have a visible result at position 26; the healthy segment lane therefore masks the lane that still needs expansion. Track surviving candidates and exhaustion per lane, and continue widening any saturated lane that has fewer than `limit` visible results.

In `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, `_search_lanes_until_visible` stops when the combined visible pool reaches the limit, allowing one healthy lane to prevent expansion of another saturated but fully filtered lane. Track each lane separately, expand any non-exhausted lane lacking enough visible candidates, and add a regression test with visible segment hits masking 25 hidden transcription hits ahead of a visible transcription.

</details>

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -40,9 +46,322 @@
...
+#
+# Every alternative is anchored to the start of the snippet: these describe what
+# the WHOLE summary is, not a phrase that can appear anywhere in one. Unanchored,
+# "single character" or "solely of repeated" match plenty of legitimate content
+# ("A token can represent a single character.") mid-sentence.
+_LOW_INFO_PATTERNS = re.compile(
+    r"^there is no content"  # ingestion placeholder (empty slide)
+    r"|^no spoken content"  # transcription placeholder (silent/music video)
+    r"|^solely of repeated"  # music-only / repeated-symbol transcriptions
+    r"|^single (word|phrase|character|letter)",  # one-word junk slide
```
Claudia-Anthropica: @Nayer-kotry [medium] Anchoring does not make this a complete placeholder signature. A legitimate summary such as `Single character encodings map symbols to integers.` still matches `^single character` and is discarded from both instant and answer retrieval. Require the known placeholder grammar after the noun—such as its expected delimiter or complete suffix—rather than matching every snippet that merely begins with these words.

In `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, the anchored `single word/phrase/character/letter` alternative still rejects legitimate explanatory snippets beginning with those words. Restrict it to the complete ingestion-placeholder shape and add a regression test proving a factual sentence beginning with `Single character` is retained.

</details>

iris/src/iris/pipeline/shared/entity_card_renderer.py
```diff
@@ -0,0 +1,122 @@
...
+        facts.append(f"worth {candidate.max_points:g} points")
+    if candidate.quiz_duration_seconds:
+        minutes = candidate.quiz_duration_seconds // _SECONDS_PER_MINUTE
+        facts.append(f"quiz duration {minutes} minutes")
+    if candidate.programming_language:
+        facts.append(f"programming language {candidate.programming_language}")
+    if candidate.short_name:
+        facts.append(f"short name {candidate.short_name}")
+    if etype == "channel":
+        visibility = "public" if candidate.channel_is_public else "private"
```
Claudia-Anthropica, line 99: @Nayer-kotry [medium] `channel_is_public` is optional and defaults to `None`, but this conditional renders every missing value as `private`. An entity candidate that omits `channelIsPublic` therefore injects a false visibility fact into reranking and the answer context. Emit the public/private detail only when the field is explicitly non-null.

In `iris/src/iris/pipeline/shared/entity_card_renderer.py`, an unknown `channel_is_public` value is rendered as `private`. Only add a channel visibility fact when the value is explicitly `True` or `False`, and add coverage for an omitted value.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The two triggering fixes are present: ordinary numeric brackets are preserved, and the placeholder alternatives are now anchored. However, the current code can still stop adaptive retrieval while one lane remains starved, discard legitimate snippets beginning with a generic placeholder phrase, and describe unknown channel visibility as private; details are inline. The captured Iris Test and Lint checks passed, while two non-required image builds remained in progress.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Nayer-kotry: Fixed in ae4b7cea: exhaustion and "enough visible results" are now tracked per lane instead of on the merged pool. `_Candidate` carries which lane it came from; the loop only stops when both lanes are individually satisfied (exhausted, or already carrying at least `limit` visible results). Added a regression test matching your exact scenario: 5 visible segment hits with all 25 transcription hits hidden now correctly continues expanding until the transcription lane surfaces its visible result.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Nayer-kotry: Fixed in ae4b7cea: the alternative now requires the colon that follows the noun in the actual placeholder shape ("Single word: 'Loading'."), not just the noun itself at the start. Added a regression test with your exact example ("Single character encodings map symbols to integers.") asserting it is kept.

iris/src/iris/pipeline/shared/entity_card_renderer.py
Nayer-kotry: Fixed in 774e7072: only add the visibility fact when `channel_is_public is not None`. Added a regression test asserting a candidate that omits `channelIsPublic` produces no "discussion channel" visibility fact at all.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
```diff
@@ -289,39 +747,438 @@ def _map_search_objects(
...
+        telemetry.reranked = True
+        reranked = sorted(
+            (
+                _Candidate(rel, c.dto, c.unit_key)
+                for rel, c in zip(relevance, candidates)
+            ),
+            key=lambda c: c.score,
+            reverse=True,
+        )
+        floor = settings.global_search_rerank_floor
```
Claudia-Anthropica: @Nayer-kotry [medium] `resolve_reranker_model` falls back to `lecture_retrieval_pipeline.reranker`, which is `cohere` in the checked-in example configuration, but this single 0.10 floor is explicitly calibrated only against Qwen3 scores. Relevance scores are not calibrated across providers, so default deployments can apply a Qwen-specific cutoff to Cohere and incorrectly discard answerable results or admit junk. Make the floor provider/model-specific, or skip absolute gating for rerankers without a calibrated floor, and cover the Cohere fallback.

In `iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py`, `_rerank_and_gate` applies a Qwen3-calibrated global relevance floor to every reranker even though the default fallback resolves Cohere. Move the floor into provider/model-specific configuration and avoid applying an uncalibrated threshold; add regression coverage for the Cohere fallback.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The three triggering fixes are present: lane widening is tracked per lane, the placeholder pattern requires the actual colon shape, and unknown channel visibility is omitted. One [medium] provider-calibration issue remains in the shared reranker gate. The captured Iris Test and Lint checks succeeded; one non-required merge job remained queued.

iris/src/iris/retrieval/lecture/lecture_global_search_retrieval.py
Nayer-kotry: Fixed in f93fbbe5: determined at construction time whether the resolved reranker is actually the calibrated VllmRerankModel (isinstance check on the LlmManager entry), stored as `_reranker_floor_calibrated`. The floor in `_rerank_and_gate` now only applies when that is true; an uncalibrated reranker (Cohere, or any future addition) still orders candidates by score but is not absolutely gated, same as the no-reranker fallback. Added a regression test for the uncalibrated case (a well-below-0.10 score is kept, not dropped) and updated the 4 existing floor tests to explicitly declare a calibrated reranker in their fixture, since they construct the instance via `__new__` and bypass `__init__`.

Claudia-Anthropica (changes requested):
@Nayer-kotry The Cohere fallback no longer receives the Qwen3 floor, but the calibration fix remains model-incomplete. [medium] `_reranker_floor_calibrated` is true for every `VllmRerankModel`, although that wrapper supports arbitrary cross-encoders and the 0.10 floor is calibrated only for Qwen3-Reranker-8B; another vLLM-served model will still be filtered on the wrong scale. Make calibration an explicit model/config capability and cover a non-Qwen vLLM model rather than inferring it from the transport wrapper. The captured CI snapshot is complete with no failed or required checks.

Claudia-Anthropica (changes requested):
@Nayer-kotry The merge from `main` preserves the passthrough-reranker integration, but the previously identified calibration defect remains on this exact head. [medium] `LectureGlobalSearchRetrieval.__init__` still derives `_reranker_floor_calibrated` from `isinstance(..., VllmRerankModel)`, so any arbitrary cross-encoder served through that transport receives the Qwen3-specific 0.10 floor; make calibration an explicit model/config capability and cover a non-Qwen vLLM model. The captured Iris Test passed and no required checks are configured, while Lint and two non-required image builds were still in progress.

Nayer-kotry:
@Claudia-Anthropica Fixed in f3ca8999: added `RerankModel.rerank_floor_calibrated` (default `False`), an explicit per-config-entry capability, replacing the `isinstance(..., VllmRerankModel)` inference — that class is a generic transport wrapper for any vLLM-hosted cross-encoder, not specifically Qwen3-Reranker-8B. Extracted the determination into `_resolve_reranker_floor_calibrated` so it's testable without constructing the retrieval class, and added coverage for a non-Qwen vLLM model (not treated as calibrated), a model that explicitly declares `rerank_floor_calibrated=True`, and an unresolved reranker.

Claudia-Anthropica (approved):
@Nayer-kotry The explicit per-config `rerank_floor_calibrated` capability fixes the remaining calibration defect: generic vLLM rerankers now default to uncalibrated, while only explicitly declared models receive the Qwen3-specific floor. The prior retrieval, citation, streaming, visibility, and entity-grounding findings are resolved on this head with regression coverage. The captured Iris Test and Lint checks passed with no required or failed checks; one non-required amd64 image build remained in progress.

iris/src/iris/pipeline/global_search_pipeline.py
```diff
@@ -32,20 +38,343 @@
...
+# after the claim" per rule 2 — with no sentence punctuation between the
+# equation and the marker, e.g. "$$\hat{y}_i = \theta x$$[1]". Matching every
+# `[n]` regardless of position corrupts ordinary bracketed content in the
+# answer's own prose, e.g. a programming answer's "array[0]" — and, worse,
+# silently mis-attributes a source when that bracketed number happens to fall
+# in 1..num_sources ("element[1]" read as citing source 1). The lookbehind
+# restricts matches to the positions the prompt actually produces: the two
+# punctuation positions, split into their own alternative from the `$$` one
+# since Python's re requires each lookbehind branch to be a fixed width.
+_CITATION_MARKER_RE = re.compile(r"(?:(?<=[.!?\]])|(?<=\$\$))\[(\d+)\]")
```
Claudia-Anthropica: @Nayer-kotry [medium] Treating every `[n]` preceded by `]` as a citation-chain continuation also matches ordinary chained indexing. `Use matrix[0][1] as the pivot.` matches `[1]`; sanitation then records source 1, allowing an uncited answer to pass the grounding guard, while renumbering can rewrite or remove the array index. Match the entire adjacent citation chain only when its first marker follows claim punctuation or closing `$$`, and add a `matrix[0][1]` regression test.

In `iris/src/iris/pipeline/global_search_pipeline.py`, `_CITATION_MARKER_RE` treats the second index in `matrix[0][1]` as a citation because it accepts any marker after `]`, which can falsely ground or rewrite an uncited programming answer. Match an entire marker chain only when its first marker follows valid claim punctuation or closing `$$`, and add chained-index regression coverage.

</details>

iris/tests/test_global_search_intent_classifier.py
```diff
@@ -0,0 +1,39 @@
...
+import iris.pipeline.shared.global_search_intent_classifier as intent_classifier
+from iris.domain.search.search_intent_dto import SearchIntent
+
+# The compiled .onnx weights are gitignored (fetched/built locally, not
+# checked into the repo — see models/global_search_intent/.gitignore), so this
+# test only has a real model to exercise wherever that artifact is present.
+_HAS_ONNX_MODEL = any(intent_classifier._model_dir().glob("*.onnx"))
+
+
+@pytest.mark.skipif(
```
Claudia-Anthropica, line 27: @Nayer-kotry [low] This test never executes in the normal Iris test job: the `.onnx` binary is gitignored, while the workflow runs pytest directly after checkout and dependency installation without fetching the Docker-only artifact. The compatibility branch therefore has no CI coverage; test `_infer` with mocked tokenizer/session inputs so zero-filled `token_type_ids` is exercised without the model file.

In `iris/tests/test_global_search_intent_classifier.py`, the artifact-gated regression test is always skipped in the normal Iris CI checkout because the required ONNX binary is not present. Add an artifact-free unit test with mocked tokenizer and ONNX session inputs that asserts `_infer` supplies zero-filled `token_type_ids` when the graph requires them and the tokenizer omits them.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The `token_type_ids` fallback is correct, but its new regression test is skipped by the normal Iris CI checkout. One [medium] citation-parser defect remains: ordinary chained indexes can be interpreted as source markers and make an uncited answer appear grounded. The captured snapshot has no failed or required checks; non-required Iris test/lint and image-build jobs were still running.

iris/src/iris/pipeline/global_search_pipeline.py
Nayer-kotry: Fixed in 0f055e51b: anchored the whole citation chain on its first marker instead of re-anchoring each marker off a bare preceding `]`. `_CITATION_MARKER_RE` now only matches `(?:\[\d+\])+` runs whose first `[` follows claim punctuation or `$$`; a `_SINGLE_MARKER_RE.finditer()` walk inside the substitution callback still handles per-number validity/dedup within a matched chain. Added `test_chained_indexing_is_not_read_as_a_citation_chain_continuation` covering exactly the `matrix[0][1]` case.

iris/tests/test_global_search_intent_classifier.py
Nayer-kotry: Fixed in 2d0699a66: added `TestInferTokenTypeIdsCompatibility`, which constructs `_IntentClassifier` via `__new__` (bypassing `__init__`, so no real tokenizer/ONNX/joblib files are needed) with mocked `_tokenizer`/`_session`/`_head`, then calls `_infer` directly. One case declares `token_type_ids` in `_input_names` while the mocked tokenizer output omits it, asserting the zero-fill branch fires and feeds a correctly-shaped zero array to `session.run`; the other declares no `token_type_ids` input and asserts it stays out of the fed inputs. This runs unconditionally in CI, unlike the existing skipif-guarded real-model test.

Claudia-Anthropica (dismissed):
@Nayer-kotry The citation-chain fix now anchors the complete run at its first valid claim boundary, preserving chained indexing while retaining citation sanitation and renumbering. The artifact-free intent-classifier tests exercise both `token_type_ids` branches, and the earlier retrieval, streaming, reranker, and entity-grounding findings remain resolved on this exact head. Iris Test, Lint, and the completed builds passed; one non-required merge build remained queued.

iris/src/iris/pipeline/global_search_pipeline.py
```diff
@@ -32,20 +38,368 @@
...
+    # cited_indices is the ground truth of what the rendered text actually cites, scanned from
+    # its own [n] markers; used_indices is only the model's separate JSON self-report, which can
+    # be wrong in both directions. Prefer cited_indices whenever the model wrote any inline
+    # markers at all: it already covers a source cited inline but left off the JSON list (the
+    # prior union's actual intent), while also NOT attaching every source the JSON over-reports
+    # beyond what the text actually references — a union unconditionally did, once observed
+    # live attaching 13 sources to an answer that inline-cited only 2 of them. Only when the
+    # model wrote no inline markers at all — relying solely on the JSON field for attribution —
+    # does used_indices still apply.
+    used_indices = cited_indices if cited_indices else used_indices
```
Claudia-Anthropica: @Nayer-kotry [medium] This conditional cannot distinguish “no markers” from “markers were present but all invalid.” For `{"answer":"Yes.[99]","used_sources":[1]}` with two sources, sanitation removes `[99]` and returns an empty cited set, so this line restores `{0}` from JSON and `_sanitize_and_suppress` returns `Yes.` as grounded to source 1. Track marker presence separately and trust the sanitized citation set—even when empty—whenever a citation chain appeared; add a regression covering an invalid marker plus non-empty `used_sources`.

In `iris/src/iris/pipeline/global_search_pipeline.py`, invalid inline citation markers can be removed while the JSON `used_sources` fallback still makes the answer appear grounded. Track whether a citation chain was present separately from the surviving cited indices, use the sanitized indices whenever markers appeared even if that set is empty, and add regression coverage for an out-of-range marker with a non-empty `used_sources` list.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The latest change correctly drops JSON over-reporting when valid inline markers survive, and the captured CI is fully passing. One grounding edge case remains: when markers are present but all invalid, the empty cited set falls back to the JSON self-report, allowing an uncited answer through. Details are inline.

iris/src/iris/domain/status/global_search_status_update_dto.py
```diff
@@ -16,4 +19,22 @@ class GlobalSearchStatusUpdateDTO(StatusUpdateDTO):
...
     sources: List[LectureSearchResultDTO] = Field(default_factory=list)
+    # Additive (wire freeze rules): old Artemis ignores the unknown field.
+    entity_sources: List[EntitySourceDTO] = Field(
+        default_factory=list, alias="entitySources"
+    )
+    # Short stage name ("searching", "generating") sent alongside a thinking update with no
+    # partial_result yet, so the client can show what is actually happening instead of one
+    # static "thinking" message for the whole wait. Additive (wire freeze rules): old Artemis
+    # ignores the unknown field and falls back to its own generic message.
+    stage: Optional[str] = None
```
Claudia-Anthropica, line 36: @Nayer-kotry [low] `GlobalSearchCallback` reuses one status DTO across `update()`, `finish()`, and `fail()`, but these new stage fields are not included in `StatusCallback._TRANSIENT_RESULT_FIELDS`. After the final stage update, terminal payloads therefore still serialize `stage="generating"` and `stageSources`, contradicting their FINISHED or FAILED run state. Clear both fields after successful running updates and cover terminal serialization with a regression test.

In `iris/src/iris/domain/status/global_search_status_update_dto.py`, stage fields persist on the reused callback DTO and leak the last running stage into FINISHED or FAILED payloads. Register `stage` and `stage_sources` as transient callback fields and add a regression test proving terminal payloads clear them.

</details>

Claudia-Anthropica (dismissed):
@Nayer-kotry The invalid-marker grounding fallback is fixed with direct regression coverage, and the new stage callbacks follow the intended retrieval and generation boundaries. One non-blocking terminal-payload cleanup issue is noted inline. The captured snapshot contains no failed or required checks; several non-required jobs were still running.

iris/src/iris/pipeline/global_search_pipeline.py
Nayer-kotry: Thanks, fixed in 8bd1e2d2b: `markers_present` is now computed on the raw text before `sanitize_citation_markers` strips anything, so `used_indices` only falls back to the JSON self-report when no citation chain was written at all -- an empty `cited_indices` from an all-invalid chain (e.g. `[99]`) is trusted as-is instead of being overridden by `used_sources`. Covered by `test_an_invalid_marker_does_not_fall_back_to_a_populated_used_sources_list`.

iris/src/iris/domain/status/global_search_status_update_dto.py
Nayer-kotry: Thanks, fixed in 440eb57bf: `stage` and `stage_sources` are now registered in `StatusCallback._TRANSIENT_RESULT_FIELDS`, so they're cleared after a successfully-delivered RUNNING update the same way `result`/`display_page_numbers` already are -- a subsequent finish()/fail() terminal payload no longer carries the last stage name. Covered by two new regression tests in test_status_callback_lifecycle.py.

Claudia-Anthropica (dismissed):
@Nayer-kotry The stage lifecycle fix is correct on this head: `stage` and `stage_sources` are now cleared after delivered RUNNING updates, with direct terminal-payload regression coverage. I rechecked the prior retrieval, citation, streaming, reranker, visibility, and entity-grounding fixes against the current code; all retained threads are resolved. The complete exact-head CI snapshot reports Iris Test, Lint, and all builds passing, with no failed or required checks.

iris/src/iris/pipeline/global_search_pipeline.py
```diff
@@ -32,20 +39,456 @@
...
+    FIRST in the sentences it actually writes gets the higher final number anyway — a reader
+    sees "[2]" before "[1]" ever appears, which reads as out of order even though nothing is
+    actually wrong. Any index with no marker in the text at all (defensive; should not happen
+    once sanitize_citation_markers has already run) is appended at the end, ascending, so
+    nothing is silently dropped.
+    """
+    seen: list[int] = []
+    seen_set: set[int] = set()
+    if answer:
+        for match in _SINGLE_MARKER_RE.finditer(answer):
```
Claudia-Anthropica: @Nayer-kotry [low] This scans every numeric bracket, although citation sanitation deliberately preserves ordinary indices. With `Use array[2] as input. First claim.[1] Second claim.[2]`, both sources are cited but `array[2]` puts source 2 first, so renumbering emits the actual citations as `[2]` then `[1]`. Derive first appearances only from chains matched by `_CITATION_MARKER_RE`, and add a regression combining an ordinary numeric index with two real citations.

In `iris/src/iris/pipeline/global_search_pipeline.py`, `_first_appearance_order` treats ordinary numeric brackets as citation appearances and can reverse the numbering of later real citations. Traverse only `_CITATION_MARKER_RE` chains when determining source order and add regression coverage for ordinary bracketed indexing before multiple valid citations.

</details>

Claudia-Anthropica (dismissed):
@Nayer-kotry The new citation-source ordering is carried through the response DTO and terminal callback, and the retained prior findings remain resolved. One non-blocking citation-order edge case is noted inline. Iris Test and Lint passed; two non-required image builds remained in progress, with no required checks configured.

iris/src/iris/pipeline/global_search_pipeline.py
Nayer-kotry: Fixed in bb90e6909: `_first_appearance_order` now scans only markers inside a chain matched by `_CITATION_MARKER_RE`, the same chain-then-marker scoping `sanitize_citation_markers`/`renumber_citation_markers` already use, instead of scanning `_SINGLE_MARKER_RE` against the whole answer. Added a regression reproducing the exact `array[2]` scenario with two real citations after it.

Claudia-Anthropica (approved):
@Nayer-kotry The citation-order fix now considers only genuine citation chains, and the regression covers ordinary numeric indexing before real citations. All retained threads are resolved, and the exact-head Iris Test and Lint checks passed with no failed or required checks; two non-required image builds remained in progress.

PR #920, opened 2026-09-07
`Logos`: Improve Error report handling

julia-rai:
## Summary by CodeRabbit

* **New Features**
  * Added support for creating, processing, tracking, cancelling, and settling batch requests, including local execution and progress updates.
  * Added provider and model eligibility tracking to improve batch routing.
  * Added clearer model calibration results, failure reasons, affected nodes, stages, and calibration timestamps.
  * Added periodic vLLM performance metrics reporting.

* **Bug Fixes**
  * Batch lifecycle actions now revalidate provider permissions before proceeding.
  * Improved fatal error detection and trust-remote-code retry handling.
  * Calibration logs now preserve observed error reasons.


logos/logos-workernode/logos_worker_node/calibration.py
```diff
@@ -306,6 +342,29 @@ class FatalLoadErrorPattern:
             "architecture. Upgrade vLLM (and remove this entry) if support "
             "has been added since this worker was deployed."
         ),
+        domain=_DOMAIN_ENGINE_INIT,
+    ),
+    FatalLoadErrorPattern(
+        needle="trust_remote_code=True",
```
Claudia-Anthropica, line 348: @julia-rai [high] Adding this signal to `_FATAL_LOAD_ERROR_PATTERNS` writes `requires-trust-remote-code` to the persistent unsupported-model list on the first failure. The existing `_retry_with_trust_remote_code_if_needed` then retries with the flag, but `calibrate_model` immediately returns the persisted unsupported entry during Phase 0 without spawning vLLM, so models this flow was designed to retry remain uncalibratable until the file is manually edited.

In `logos/logos-workernode/logos_worker_node/calibration.py`, classifying `trust_remote_code=True` as fatal persists unsupported state before the existing automatic retry and causes that retry to short-circuit. Keep this signal out of persistent fatal classification while preserving the `--trust-remote-code` retry.

</details>

logos/logos-ui/src/app/features/model-error-report/model-error-report.ts
```diff
@@ -1191,12 +1417,123 @@ export class ModelErrorReport implements OnInit, OnDestroy {
...
     node: string,
     log: string,
-    success: boolean
+    success: boolean,
+    unsupportedReason: string | null = null,
+    nodeUnhealthyReason: string | null = null,
+    observedReason: string | null = null,
+    backendStages: readonly BackendStageResult[] | null = null
   ): NodeCalibrationResult {
+    if (backendStages && backendStages.length > 0) {
```
Claudia-Anthropica, line 1520: @julia-rai [medium] Successful calibration events explicitly send `log_text=null`, and successful `CalibrationResult`s leave `stages=null`, so the no-log fallback returns `probes: []`. `getCalibrationChecklistItems()` consequently sees no successful stages; when every node succeeded, the new default Error Report displays “No checklist information available,” and successful nodes never populate the intended checklist.

In `logos/logos-ui/src/app/features/model-error-report/model-error-report.ts`, a successful row with neither backend stages nor log text produces no probes, removing every successful node from the checklist. When `success` is true and both backend stages and probe blocks are absent, build a synthetic probe containing the calibration domains with successful statuses.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The worker-to-UI persistence path is mostly coherent, and the exact-head CI snapshot has no failed or pending checks. Two blocking regressions remain: the new fatal reason defeats the remote-code retry, and successful calibration rows cannot populate the default checklist. I also reviewed the existing inline feedback and did not duplicate it.

Claudia-Anthropica (changes requested):
@julia-rai The two previously reported blocking regressions remain on this exact head. `_FATAL_LOAD_ERROR_PATTERNS` still persists `trust_remote_code=True` as permanently unsupported before the automatic retry, causing that retry to stop at the unsupported-model check. The successful no-log path still returns `probes: []`, so successful-only calibration rows cannot populate the now-default deployment checklist.

logos/logos-ui/src/app/features/model-error-report/model-error-report.ts
```diff
@@ -1191,12 +1432,179 @@ export class ModelErrorReport implements OnInit, OnDestroy {
...
+
+  // Success sends log_text=null and stages=null (see
+  // BackendCalibrationLog) — nothing to parse. Without this,
+  // getCalibrationChecklistItems() sees `probes: []` and the node
+  // silently drops out of every domain's checklist row.
+  private buildSyntheticSuccessProbe(): CalibrationProbeResult {
+    return {
+      probe: 1,
+      status: 'success',
+      stages: CALIBRATION_DOMAINS.map(domain => ({
```
Claudia-Anthropica: @julia-rai [medium] This marks every domain successful even when `summary.tensor_parallel_size` is 1. The worker deliberately omits `Multi-GPU Coordination` for single-GPU attempts, so the default report now claims that coordination succeeded where it never ran, even though the row summary already contains the tensor-parallel size.

In `logos/logos-ui/src/app/features/model-error-report/model-error-report.ts`, `buildSyntheticSuccessProbe` marks Multi-GPU Coordination successful for single-GPU calibrations where that domain is inapplicable. Pass the calibration summary's tensor-parallel size into fallback parsing and omit the multi-GPU domain unless the size exceeds one, with coverage for single- and multi-GPU successes.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The two previously reported blockers are fixed: the trust-remote-code retry no longer persists unsupported state, and successful no-log rows now populate the checklist. One medium correctness issue remains because the synthetic success path reports an inapplicable multi-GPU stage as successful on single-GPU calibrations. The exact-head Logos test, UI static export, and pre-commit checks succeeded; several non-required build and documentation checks were still pending in the snapshot.

logos/logos-workernode/logos_worker_node/calibration.py
Claudia-Anthropica: @julia-rai Thanks, `persist=False` and the non-persistent retry latch now let the trust-remote-code retry re-enter calibration without being blocked by Phase 0.

logos/logos-ui/src/app/features/model-error-report/model-error-report.ts
Claudia-Anthropica: @julia-rai Thanks, the synthetic success probe now keeps successful no-log rows in the deployment checklist.

logos/logos-orchestrator/src/logos/batch_api.py
```diff
@@ -1583,6 +1583,17 @@ async def handle_batch_api_request(request: Request) -> Response:
...
+                # Ownership above only says the team minted this object;
+                # permissions can move after that — re-check them too (see
+                # _assert_still_permitted), against owner itself for a file
+                # resource, else the file it came from, if still on record.
+                input_file = (
+                    owner
+                    if operation.resource == "files"
+                    else db.get_batch_object("file", str(owner.get("input_file_id") or ""))
+                )
+                if input_file is not None:
```
Claudia-Anthropica, line 1595: @julia-rai [high] Revalidation is still bypassable because it runs only when the input-file row exists. Provider-backed file deletion removes that row without protecting batches that reference it, so a caller can delete the input, lose provider/model access, and still poll or cancel the batch through the shared provider credential; mapped result-file rows also carry no model list, so model revocations are not checked there. Persist the required provider/model authorization metadata on batch and result-file rows (or retain their input relationship), then require `_assert_still_permitted` for every provider-backed lifecycle operation instead of skipping it when this lookup returns `None`.

In `logos/logos-orchestrator/src/logos/batch_api.py`, provider-backed lifecycle authorization is conditional on an input-file row that may have been deleted, and result-file mappings lack model metadata. Persist or reliably resolve each object's provider and model authorization metadata, and always revalidate it before forwarding lifecycle operations.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The single-GPU synthetic-success checklist issue is fixed because the summary's tensor-parallel size now filters the multi-GPU stage. The existing retry-cleanup blocker remains: a retryable fatal result leaves `proc` unset, then the unconditional `stop_vllm(proc)` raises before the trust-remote-code retry can inspect that result. A separate lifecycle authorization gap remains below. The snapshot contains no failed checks and no pending required checks; four non-required checks were still pending.

logos/logos-workernode/logos_worker_node/calibration.py
```diff
@@ -1972,15 +2494,24 @@ def _try_start(
             stop_vllm(proc)
             time.sleep(_VRAM_SETTLE_S)
 
+            # Informational only, checked on every failure regardless of
+            # whether the node/fatal classifiers below also match — never
+            # gates blacklisting or node-health state. Reflects the MOST
+            # RECENT failing probe (overwritten, not first-match-wins like
+            # the two latches below), since it's just "what was last seen".
+            observed_pattern = _classify_observed_transient_error(probe_log)
+            if observed_pattern is not None:
```
Claudia-Anthropica, line 2712: @julia-rai [low] This only updates the latch when the current failure matches an observed pattern. If an OOM probe is followed by a different unclassified terminal failure, `_override_error_if_unsupported` persists the earlier `cuda-oom` reason even though `stages` describes the latest probe; clear `_observed_reason_box` when `observed_pattern` is `None` so the stored reason cannot become stale.

In `logos/logos-workernode/logos_worker_node/calibration.py`, the observed-reason latch retains an earlier probe's classification when a later failure has no observed-pattern match. Replace or clear the latch on every probe failure so `observed_reason` always represents the deciding probe.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The exact head still has two previously reported high blockers: provider lifecycle authorization is skipped when the input-file row is absent, and a retryable trust-remote-code failure leaves `proc` unset before unconditional cleanup dereferences it. The legacy log parser also still includes Multi-GPU Coordination for single-GPU attempts and calculates scope against all logs, as existing feedback notes. I found one additional low-severity stale diagnostic issue inline; the snapshot has no failed or pending required checks, while several non-required checks remain pending.

logos/logos-orchestrator/src/logos/batch_api.py
```diff
@@ -1583,6 +1609,15 @@ async def handle_batch_api_request(request: Request) -> Response:
...
                     )
+                # Ownership above only says the team minted this object;
+                # permissions can move after that — re-check them too (see
+                # _assert_still_permitted). owner itself carries the models
+                # list (batches and result files are registered with their
+                # own copy — see _register_upstream_object /
+                # _register_result_files), so this never has to chase the
+                # original input file, which may since have been deleted
+                # independently of the batch/result file that referenced it.
+                _assert_still_permitted(db, auth, provider, owner)
```
Claudia-Anthropica: @julia-rai [high] This now checks only `owner.models`, but every provider batch/result row created before this commit has `models = NULL` because the previous registration path populated models only on uploaded file rows. Such live objects compare permissions against an empty list, so a key retaining provider access after losing model access can still poll, cancel, or download them; this also removes the former fallback to a surviving `input_file_id`. Backfill existing rows and retain an input-row fallback where possible, failing closed when model authorization metadata cannot be recovered.

In `logos/logos-orchestrator/src/logos/batch_api.py`, forwarding uses only `owner.models`, but pre-existing batch and result-file rows contain no model metadata and therefore bypass model revocation checks. Backfill legacy rows, use the surviving input-file relationship as a fallback, and fail closed when the required authorization metadata cannot be recovered.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The provider-level recheck protects newly recorded objects, but existing batch/result rows lack copied model metadata and still bypass model revocation checks, as noted inline. The previous trust-remote-code blocker also remains: `_retryable_fatal_box` makes `_try_start` return `None`, then the outer `finally` calls `stop_vllm(proc)` unconditionally and raises on `proc.pid` before the retry helper can inspect the result. The exact-head snapshot has no failed or pending required checks; several non-required jobs were unfinished.

Claudia-Anthropica (changes requested):
@julia-rai The prior high blockers are fixed: `stop_vllm` safely handles a missing process, and the unrelated batch-lifecycle changes are gone. The existing legacy-parser issue remains: `parseCalibrationProbe` includes Multi-GPU Coordination for single-GPU rows, while the scope helpers divide by all logs, so mixed single-/multi-GPU reports show a false stage and incorrect percentage; pass `tensorParallelSize` through raw-log parsing and use only applicable nodes as the denominator. No required checks are pending or failed in the snapshot.

Claudia-Anthropica (changes requested):
@julia-rai The previously reported legacy-parser issue remains on this exact head: `parseCalibrationProbe` still receives no tensor-parallel size and therefore includes Multi-GPU Coordination for single-GPU raw-log rows, while both scope helpers divide by all logs rather than only nodes where that domain applies. Pass domain applicability through raw-log parsing and calculate each stage percentage using only applicable nodes. The snapshot has no failed or pending required checks.

PR #923, opened 2026-09-07
`Logos`: Serve the Anthropic Messages API on OpenAI-only upstreams and self-scrape cloud models

### LINKED ISSUE #739: `Logos`: Scraping of available models
When an azure deployment is added it does automatically self-scrape for available deployed models.

If I connect a logos instance to another logos instance then it shall do the same for the provided API key. Right now it does not.

Also for any other provider types it shall do that. Maybe we shall even add a dropdown possibility for Logos type?

### LINKED ISSUE #883: `Logos`: Prod connection does not work on Logos Test

wasnertobias:
## Summary

`claude-logos` did not work against a Logos instance that has another Logos instance (or any cloud resource) upstream. Two independent causes, both reproduced on `logos-test` against the `Logos PROD` provider.

**1. The Anthropic Messages API was forwarded, not served.**
`POST /v1/messages` went to the upstream like-for-like (`_cloud_forward_url`). vLLM answers that route natively and so does another Logos instance, which is why a workernode-backed model worked. An Azure or OpenAI resource has no Messages route at all, so the request 404'd before the model saw anything.

**2. Context windows stopped at the upstream.**
`GET /v1/models` derived its window solely from the live logosnode runtime snapshots plus the `model_profiles` high-water mark. A cloud provider contributes to neither, so a model reachable only through one was published with no window — measured on `logos-test`:

| | `GET /v1/models` for `Qwen/Qwen3.8-27B` |
|---|---|
| upstream (`logos.aet.cit.tum.de`) | `max_model_len: 262144`, `…_current_min` / `_current_max` / `_overall`: `262144` |
| downstream (`logos-test`, via the `Logos PROD` cloud provider) | no context field at all |

`claude-logos` then falls back to `LOGOS_CONTEXT_FALLBACK=111200` instead of the 262144 the upstream had measured — "works, but not with full context".

## Changes

**Anthropic Messages translation** — new package `src/logos/anthropic_compat/`

- `common.py` — dialect classification, Anthropic⇄OpenAI content-block mapping, an incremental SSE decoder and an `AnthropicStreamWriter` that emits a protocol-correct Messages event stream (one open content block at a time, gap-free indices, terminal `message_delta` + `message_stop`).
- `chat_completions.py` — Messages ⇄ Chat Completions, request, response and stream. Covers system-block lists, tool calls answered by tool results (reordered so every `role: tool` message follows its assistant call), images as data URIs, `tool_choice`, and `max_tokens` → `max_completion_tokens` for the o-series / gpt-5 families.
- `responses_api.py` — Messages ⇄ the Responses API, for the gpt-5.x deployments `classify_azure_operation` stores against `responses`. Tool calls and results become top-level `input` items; `temperature`/`top_p` are not forwarded (reasoning deployments reject them).
- `pipeline/context_resolver.py` — resolves the upstream's dialect and re-points the forward URL. Azure keeps the operation its stored endpoint names; only the `base_url` branch is re-pointed to `chat/completions`. Request translation happens before model injection and effort normalization, so those act on the body actually sent.
- `main.py` — response translation on the sync, streaming and both error paths. It runs *last*, so the log accumulator, cost enricher and rate limiter all still see the upstream's own OpenAI-shaped frames; only the bytes on the wire change shape.

**Cloud model + context sync** — new `src/logos/sdi/cloud_model_sync.py`

- Mirrors every non-Azure cloud provider's `GET /v1/models` into `models` + `model_provider` (Azure keeps its own control-plane discovery, which yields the deployment id and api-version a URL needs). Runs on startup and every 15 min (`LOGOS_CLOUD_MODEL_SYNC_INTERVAL_S`, `…_ENABLED`).
- Records the context windows the upstream reports in a new `cloud_model_context` table; `_served_context_window_stats` folds them in on the same terms as a lane's — `current_min` keeps the smallest, since a request may be routed to any deployment of the model.
- An upstream that is down, unreachable or lists nothing leaves the catalogue untouched, so its models stay routable.
- A listing whose entries are all `owned_by: "logos"` identifies a Logos upstream. Its `cloud_provider_type` is filled in only while unset; an operator-chosen value is never overwritten, just logged.

**New `logos` cloud provider type** — Liquibase `024`, the SQLAlchemy and Java enums, and the provider dropdown. An upstream Logos instance serves every surface this one does, so its Messages requests are forwarded verbatim rather than translated.

**Shared auth defaulting** — `cloud_auth_header` in `dbutils/types.py`, now used by both the forwarding path and the sync, so the "empty fields mean `Authorization: Bearer {}`" convention cannot drift.

## Testing

`1340 passed, 1 xfailed` (`main`: 1262 passed) — 78 new tests:

- `tests/unit/anthropic_compat/` — 42 tests: dialect classification, both request/response translations, and the full streaming event sequence for each dialect (text, tool calls, split byte boundaries, idempotent close, mid-stream error).
- `tests/unit/pipeline/test_messages_forwarding.py` — 9 tests: where `/v1/messages` is forwarded per provider type. One of these caught a real bug in an earlier revision, where a stored Azure `responses` deployment was wrongly re-targeted to `chat/completions`.
- `tests/unit/main/test_messages_response_translation.py` — 5 tests: the client receives Anthropic while the ledger keeps the upstream's own field names.
- `tests/unit/sdi/test_cloud_model_sync.py` — 19 tests: URL construction, listing parser, and the sync's failure behaviour.
- `tests/unit/main/test_v1_models.py` — 3 tests for the cloud context window on `/v1/models`.

Run: `poetry run pytest tests/unit -v`

## Database Changes

- `logos-webservice/src/main/resources/liquibase/changelog/024_cloud_model_sync.xml`
  - `ALTER TYPE cloud_provider_type_enum ADD VALUE IF NOT EXISTS 'logos'` (`runInTransaction="false"`)
  - `CREATE TABLE cloud_model_context (provider_id, model_name, context_current_min, context_current_max, context_overall, updated_at)`

## Notes

- The translation targets Chat Completions and the Responses API. `thinking` blocks are dropped in both directions: their signature is meaningless to an OpenAI upstream, and synthesising one would break the next turn when the client echoes it back.
- Discovered models are **not** granted to any team automatically — an admin assigns access per team, same as the Azure sync.
- Pruning follows the Azure sync: links to models an upstream no longer lists are removed, but only when the scrape actually succeeded and returned models.

https://claude.ai/code/session_014DQ68LU9qETRsEhgTqcuQP

## Summary by CodeRabbit

* **New Features**
  * Added Anthropic Messages API compatibility for OpenAI Chat Completions and Responses providers, including multimodal content, tools, usage details, errors, and streaming.
  * Added support for Logos instances as cloud providers.
  * Cloud model catalogs and context-window information now synchronize automatically.
  * Added administrative tools for calibration-log retrieval and worker compatibility checks.
* **Bug Fixes**
  * Improved streamed tool-call and refusal handling.
  * Fixed macOS Bash compatibility in the Claude wrapper.
* **UI**
  * Added Logos as a selectable cloud provider type.


logos/logos-orchestrator/src/logos/anthropic_compat/responses_api.py
```diff
@@ -0,0 +1,325 @@
...
+    """One Anthropic message -> the Responses input items it becomes.
+
+    Tool calls and tool results are siblings of the messages here, not fields
+    on them, so an assistant turn that called two tools becomes one message
+    item plus two ``function_call`` items.
+    """
+    role = message.get("role")
+    blocks = normalize_content(message.get("content"))
+    is_assistant = role == "assistant"
+    text_type = "output_text" if is_assistant else "input_text"
```
Claudia-Anthropica: @wasnertobias [high] Assistant history is encoded with `output_text`, but this object is being constructed as an easy input message without the required output-item metadata (`id` and `status`). Responses API request-message content uses `input_text`, including for the assistant role, so a follow-up turn containing assistant text can be rejected with HTTP 400.

In `logos/logos-orchestrator/src/logos/anthropic_compat/responses_api.py`, assistant history is emitted as an incomplete Responses output item and can make multi-turn requests invalid. Encode translated assistant message content as `input_text` like other request messages.

</details>

logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py
```diff
@@ -0,0 +1,373 @@
...
+
+        out: List[bytes] = []
+        delta = choice.get("delta") if isinstance(choice.get("delta"), dict) else {}
+        text = delta.get("content")
+        if isinstance(text, str) and text:
+            out.extend(writer.text(text))
+
+        for call in delta.get("tool_calls") or []:
+            if isinstance(call, dict):
+                out.extend(self._tool_call(writer, call))
```
Claudia-Anthropica: @wasnertobias [medium] Parallel tool-call deltas can be interleaved by index, but the writer accepts arguments only for the currently open block. Once index 1 opens, later fragments for index 0 are silently discarded because index 0 remains in `_open_tools` and is never reopened, corrupting its tool input.

In `logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py`, interleaved streaming deltas for parallel tool calls lose arguments belonging to any call that is not currently open. Accumulate metadata and argument fragments by call index, then emit each complete tool block sequentially before the terminal message events.

</details>

logos/logos-orchestrator/src/logos/sdi/cloud_model_sync.py
```diff
@@ -0,0 +1,336 @@
...
+        pid = provider["id"]
+        name = provider.get("name", f"provider-{pid}")
+        try:
+            url = models_url(provider.get("base_url", ""))
+        except ValueError as exc:
+            logger.warning("Cloud model sync: provider %s (%s) skipped: %s", pid, name, exc)
+            return False, False
+
+        headers = {"Accept": "application/json"}
+        auth = cloud_auth_header(provider.get("auth_name"), provider.get("auth_format"), provider.get("api_key"))
```
Claudia-Anthropica: @wasnertobias [medium] `auth_format` is an unrestricted operator-provided string, and `cloud_auth_header()` can raise `ValueError`, `KeyError`, or `IndexError` while formatting it. This call is outside the provider error handling, so one malformed provider aborts `run_once()` and prevents every later provider from syncing on each cycle.

In `logos/logos-orchestrator/src/logos/sdi/cloud_model_sync.py`, a malformed provider `auth_format` escapes `_sync_provider` and aborts the entire synchronization cycle. Catch formatting failures per provider, log and skip only that provider, and add a test showing a later valid provider is still synchronized.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The previously raised startup-blocking issue is fixed on this head, and captured CI passes. I found three current correctness and resilience issues in Responses translation, parallel tool streaming, and provider synchronization.

logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py
```diff
@@ -0,0 +1,395 @@
...
+    tools = anthropic_tools(payload.get("tools"))
+    if tools:
+        result["tools"] = [{"type": "function", "function": tool} for tool in tools]
+        choice = _tool_choice(payload.get("tool_choice"))
+        if choice is not None:
+            result["tool_choice"] = choice
+
+    effort = _reasoning_effort(payload)
+    if effort:
+        result["reasoning_effort"] = effort
```
Claudia-Anthropica, line 107: @wasnertobias [high] The translation forwards `output_config.effort` as `reasoning_effort` for every Chat Completions model while also forwarding `temperature` and `top_p` unconditionally. Claude Code supplies effort and commonly supplies temperature, but OpenAI/Azure GPT-4.1 rejects `reasoning_effort`, while o-series reasoning models reject the sampling parameters, so common targets can return HTTP 400 before inference.

In `logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py`, request parameters are forwarded without accounting for model capabilities, causing GPT-4.1 and o-series requests to contain unsupported fields. Make parameter translation model-capability-aware by omitting `reasoning_effort` for non-reasoning families and omitting unsupported sampling parameters for reasoning families, with regression tests for both cases.

</details>

logos/logos-orchestrator/src/logos/anthropic_compat/responses_api.py
```diff
@@ -0,0 +1,347 @@
...
+        except (TypeError, ValueError):
+            return []
+        if not isinstance(frame, dict):
+            return []
+
+        # The event name is authoritative; the payload repeats it in "type"
+        # for clients that read only the data lines.
+        event = name or str(frame.get("type") or "")
+
+        if event in ("response.failed", "error"):
```
Claudia-Anthropica: @wasnertobias [medium] `Executor.execute_streaming()` represents a transport failure after partial output as an unnamed `data: {"error": ...}` frame followed by `[DONE]`. Such a frame produces an empty `event` here and is ignored, after which `[DONE]` emits `end_turn` and `message_stop`, making a truncated Responses answer look successful to the Anthropic client.

In `logos/logos-orchestrator/src/logos/anthropic_compat/responses_api.py`, unnamed error frames produced after partial transport failures are ignored and the following `[DONE]` closes the response as successful. Treat any frame containing a dictionary or string `error` as a terminal error before dispatching by event name, and add a regression test using the executor's unnamed recovery-frame shape.

</details>

logos/logos-orchestrator/src/logos/anthropic_compat/common.py
```diff
@@ -0,0 +1,508 @@
...
+        """Emit an Anthropic ``error`` event mid-stream.
+
+        Clients treat this as the terminal event, so nothing follows it — in
+        particular not the ``message_stop`` a successful turn ends with, which
+        would otherwise read as a complete answer.
+        """
+        if self._finished:
+            return []
+        out = self._close()
+        self._finished = True
```
Claudia-Anthropica, line 537: @wasnertobias [medium] Marking the writer finished does not prevent `text()`, `tool_use()`, or `tool_arguments()` from emitting later content. Both translators retain buffered tool calls after `error()` and the HTTP generator subsequently calls `finish()`, so a tool block—and even `message_start` when nothing was previously emitted—can appear after the terminal Anthropic error event.

In `logos/logos-orchestrator/src/logos/anthropic_compat/common.py`, content-writing methods can emit events after `error()` has made the stream terminal. Make every content-writing method a no-op once `_finished` is set, and add error-then-finish tests with buffered tool calls for both stream translators.

</details>

logos/logos-orchestrator/src/logos/anthropic_compat/responses_api.py
```diff
@@ -0,0 +1,347 @@
...
+# ── response ────────────────────────────────────────────────────────────────
+
+
+def _stop_reason(body: Dict[str, Any], *, saw_tool_call: bool) -> str:
+    """Anthropic stop reason for a finished Responses body.
+
+    The Responses API reports completion as a status plus an
+    ``incomplete_details.reason`` rather than a per-choice finish reason.
+    """
+    if saw_tool_call:
```
Claudia-Anthropica: @wasnertobias [medium] A Responses result can be `incomplete` because `max_output_tokens` was reached while a function call was being generated. Checking `saw_tool_call` first reports `tool_use` for that truncated call, causing the client to execute incomplete arguments instead of recognizing the length stop.

In `logos/logos-orchestrator/src/logos/anthropic_compat/responses_api.py`, an incomplete response containing a partial function call is incorrectly assigned the `tool_use` stop reason. Give the incomplete/max-output-token condition precedence over `saw_tool_call`, and test an incomplete response containing a function call.

</details>

logos/logos-orchestrator/src/logos/pipeline/context_resolver.py
```diff
@@ -402,13 +443,20 @@ def _cloud_forward_url(
...
+        an Azure deployment keeps the operation its stored endpoint names, and
+        the dialect follows from that instead.
         """
         if endpoint_fallback and endpoint_fallback.startswith("http"):
             return ContextResolver._align_azure_operation(endpoint_fallback, request_path)
         base = (base_url or "").rstrip("/")
         if not request_path:
             return ContextResolver._merge_url(base_url, endpoint_fallback or "")
-        path = request_path.lstrip("/")
+        path = ContextResolver._upstream_path(request_path, cloud_provider_type).lstrip("/")
```
Claudia-Anthropica, line 487: @wasnertobias [medium] This Messages-path rewrite is bypassed by the earlier absolute-`endpoint_fallback` branch. For a Logos provider with an existing manually configured `.../chat/completions` endpoint—a supported UI configuration that the new sync deliberately preserves—the context is classified `NATIVE`, leaving the Anthropic body unchanged but posting it to Chat Completions, which rejects it.

In `logos/logos-orchestrator/src/logos/pipeline/context_resolver.py`, an absolute per-model endpoint bypasses the Messages-path rewrite even when the provider is a native Logos upstream. For native cloud provider types, route `/v1/messages` to the provider's Messages path instead of an absolute non-Messages endpoint, and add a test for a manually connected Logos model with an absolute Chat Completions endpoint.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The three prior Claudia findings are fixed on this head, but five blocking correctness issues remain in request compatibility, stream termination, and native Logos routing. Common OpenAI model families can receive unsupported parameters, while several Responses/tool-error paths can turn failures or truncated calls into invalid or apparently successful Anthropic streams. Captured CI has no failed checks; one non-required build was still in progress.

logos/logos-orchestrator/src/logos/main.py
```diff
@@ -3284,6 +3289,17 @@ async def _on_capacity_needed(model_name: str, provider_id: int | None = None) -
...
+    # the same pass records the context windows the upstream reports so
+    # /v1/models can republish them.
+    global _cloud_model_sync
+    _cloud_model_sync = CloudModelSyncService(
+        on_models_changed=lambda *, rebuild_classifier: refresh_pipeline_runtime_state(
+            rebuild_model_classifier=rebuild_classifier
+        ),
+    )
+    await _cloud_model_sync.start()
+
```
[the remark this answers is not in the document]
wasnertobias: Fixed in c50c1343c. `start()` no longer awaits the first pass — it only creates the tracked task, and `_loop()` runs the sync immediately and then on the interval. Nothing is lost by not waiting: the catalogue and the context windows both live in the database, so what the previous run stored is already being served while the new pass refreshes it.

Covered by `test_start_returns_before_the_first_sync_completes`, which blocks `run_once` and asserts `start()` returns before it completes.

logos/logos-orchestrator/src/logos/anthropic_compat/responses_api.py
wasnertobias: Fixed in 8075e02ab. Request messages now use `input_text` for every role, the assistant's own history included. You are right that the object was neither shape: `output_text` belongs to an output item, which also carries an `id` and a `status` this translation has nothing to fill in.

Covered by `test_assistant_history_uses_request_content_types`, which asserts every content part is `input_text` and that no item carries `id`/`status`.

logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py
wasnertobias: Fixed in 8075e02ab. Confirmed before changing anything — with deltas ordered index 0, index 1, index 0, the first call ended up with `{"cmd":` and the closing fragment was gone.

Fragments are now accumulated per index in the translator and written out as whole blocks in `finish()`, in the order the indices first appeared. Prose still streams as it arrives; only the tool arguments wait. The Responses translator keyed on `output_index` the same way and got the same treatment.

Covered by `test_interleaved_parallel_tool_arguments_are_not_lost` (both dialects), `test_parallel_tool_blocks_keep_the_order_the_model_produced` and `test_text_still_streams_while_tool_calls_are_collected`.

Follow-up: the first version of that fix read `bool(self._tools)` for the stop reason *after* the flush had cleared it, so a tool call went out as `end_turn` unless the upstream spelled out `finish_reason: tool_calls`. Corrected in b8e4f6bb3.

logos/logos-orchestrator/src/logos/sdi/cloud_model_sync.py
wasnertobias: Fixed in 8075e02ab. The `cloud_auth_header()` call is now guarded per provider, with a message naming the provider whose `auth_format` is not a valid template. Added a general per-provider guard in `run_once()` as well, so no future failure in one provider can cost the others their sync either.

Covered by `test_a_malformed_auth_format_only_skips_its_own_provider` (parametrised over `Bearer {`, `{name}` and `{1}` — ValueError, KeyError, IndexError) and `test_an_unexpected_provider_failure_does_not_stop_the_others`; both assert the *following* provider still syncs.

logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py
wasnertobias: Fixed in dbbb37078. `is_reasoning_model()` — the name-based classification that already chose between `max_tokens` and `max_completion_tokens` — now picks the whole parameter set: a reasoning model gets `reasoning_effort` and neither sampling parameter, everything older gets `temperature`/`top_p` and no `reasoning_effort`.

Covered by `test_the_two_openai_families_get_mutually_exclusive_parameters`, which sends one request carrying all of them and asserts both directions for gpt-4.1-nano, gpt-5.6-luna, o3-mini and openai/gpt-5.1.

logos/logos-orchestrator/src/logos/anthropic_compat/responses_api.py
wasnertobias: Fixed in dbbb37078. An `error` payload is now recognised whatever the frame is called, so the unnamed frame the executor appends is handled instead of falling through to `[DONE]` and the terminal events. The chat/completions translator already did this; the two now agree.

Covered by `test_an_unnamed_error_frame_is_not_mistaken_for_a_completed_turn`, which feeds partial text, then the bare `data: {"error": ...}` frame, then `[DONE]`, and asserts the stream ends on `error` with no `message_stop`.

logos/logos-orchestrator/src/logos/anthropic_compat/common.py
wasnertobias: Fixed in dbbb37078. `AnthropicStreamWriter` now refuses to emit once the stream is closed — `start()`, `text()`, `tool_use()` and `tool_arguments()` all return nothing after `stop()` or `error()`. Both translators additionally drop their buffered tool calls in `error()`, so nothing is left pending.

Covered by `test_no_content_is_emitted_after_a_mid_stream_error` in both dialects, and `test_an_error_before_any_output_emits_only_the_error` for the `message_start`-after-error case you called out.

logos/logos-orchestrator/src/logos/anthropic_compat/responses_api.py
wasnertobias: Fixed in dbbb37078. `_stop_reason()` checks truncation before the tool call, so an `incomplete` response reports `max_tokens` even when a function call was in flight.

The chat/completions side was already correct — `length` maps to `max_tokens` and the `saw_tool_call` override only applies when the mapping produced `end_turn` — so only the Responses side needed it.

Covered by `test_a_truncated_function_call_reports_max_tokens_not_tool_use`.

logos/logos-orchestrator/src/logos/pipeline/context_resolver.py
wasnertobias: Fixed in dbbb37078. Confirmed first: with a Logos provider whose model endpoint is `.../v1/chat/completions`, the context came out `NATIVE` and `prepare_headers_and_payload` left `system` in the body, which was then posted to chat/completions.

`dialect_for()` now lets the resolved URL decide first — it is the surface the request is actually posted to — and falls back to the provider type only when the URL names neither OpenAI surface. That keeps every other case as it was: a Logos provider without a pinned endpoint still resolves to `.../v1/messages` and stays NATIVE, and a workernode's `logosnode://` URL matches nothing either.

Covered by `test_a_pinned_chat_endpoint_wins_over_a_native_provider_type`, `test_a_native_provider_keeps_the_messages_route` and, end to end through the resolver, `test_a_logos_provider_with_a_pinned_chat_endpoint_is_translated`.

wasnertobias:
All review threads are addressed and resolved, each with a regression test that fails without its fix. One item from the first review had no thread of its own, so it is answered here.

**`stream_options.include_usage` on translated streaming requests** — not applicable. `Executor._streaming_payload` already adds it for every non-Responses forward URL, which is exactly where a translated Messages request goes; `tests/unit/pipeline/test_executor_streaming_payload.py` pins that, including the Responses URLs it must *not* be added to. Setting it in `to_chat_completions` as well would only duplicate it. Left a comment at the call site so the question does not come back, plus `test_streaming_switch_is_forwarded_without_stream_options` to pin the division of labour.

Two of the findings were reproduced before being changed, because both failed silently rather than loudly:

- Interleaved parallel tool calls: with deltas ordered index 0, index 1, index 0, the first call reached the client as `{"cmd":` — truncated JSON it would have handed to the tool.
- A Logos provider with a hand-pinned `.../v1/chat/completions` endpoint: the context resolved to `NATIVE` and the Anthropic body kept its `system` field while being posted to chat/completions.

While fixing the first of those I introduced a second defect — reading the tool state after the flush had cleared it, which reported `end_turn` for a turn that ended in a tool call. CodeRabbit caught it; fixed in b8e4f6bb3 with coverage for the `finish_reason: stop` and no-`finish_reason` cases the original test missed.

Unit tests: 1428 passing (1262 on main).

logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py
```diff
@@ -0,0 +1,406 @@
...
+    Only fields with a chat/completions counterpart are carried over — an
+    unknown field is a 400 from OpenAI and Azure, so Anthropic-only fields
+    (``metadata``, ``top_k``, server-side tools) are dropped rather than
+    forwarded.
+    """
+    messages: List[Dict[str, Any]] = []
+
+    system = system_to_text(payload.get("system"))
+    if system:
+        messages.append({"role": "system", "content": system})
```
Claudia-Anthropica: @wasnertobias [high] `is_reasoning_model()` includes `o1`, but the system prompt is emitted before that classification and always uses `role: "system"`. o1 Chat Completions deployments reject system-role messages, and Claude Code supplies a system prompt on every turn, so this model path still returns HTTP 400 before inference.

In `logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py`, o-series requests encode the system prompt using a role those deployments reject. Determine the model family before assembling messages, emit a `developer` instruction message for affected reasoning models, and add an o1 regression test containing a system prompt.

</details>

logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py
```diff
@@ -0,0 +1,406 @@
...
+    if max_tokens is not None:
+        result["max_completion_tokens" if reasoning else "max_tokens"] = max_tokens
+
+    if not reasoning:
+        for name in _PASSTHROUGH_PARAMS:
+            if payload.get(name) is not None:
+                result[name] = payload[name]
+
+    if payload.get("stop_sequences"):
+        result["stop"] = payload["stop_sequences"]
```
Claudia-Anthropica: @wasnertobias [medium] `stop_sequences` is still forwarded as `stop` when `reasoning` is true. Chat Completions rejects `stop` for o3/o4-mini and newer reasoning models, so an otherwise valid Messages request combining one of these models with a stop sequence receives HTTP 400.

In `logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py`, reasoning-model requests still receive the unsupported `stop` parameter. Omit translated stop sequences for reasoning families and add an o3 regression test proving `stop` is absent while non-reasoning models retain it.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The five prior Claudia findings are fixed on this head, and I staged resolution for the three still-current threads. Two Chat Completions compatibility gaps still produce request-time 400s for supported o-series inputs: instruction-role selection and unconditional stop forwarding. Captured CI has no failed checks; several non-required jobs were still in progress.

logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py
wasnertobias: Fixed in 82a1c449e. You are right that the split did not reach this — the system prompt was assembled before the classification existed in the function. The classification is now computed first, and the prompt goes out as `role: "developer"` for a reasoning model and `role: "system"` for everything older.

Covered by the same `test_the_two_openai_families_get_mutually_exclusive_parameters`, now carrying a system prompt and asserting the role for gpt-5.6-luna, o1, o3-mini and openai/gpt-5.1 as well as gpt-4.1-nano.

logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py
wasnertobias: Fixed in 82a1c449e. `stop` moved in with `temperature` and `top_p` under the non-reasoning branch, so it is dropped for o3, o4-mini and the models after them.

Same test, which now sends `stop_sequences` and asserts it survives as `stop` for gpt-4.1-nano and is absent for every reasoning model.

logos/logos-orchestrator/src/logos/anthropic_compat/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+# Where a Messages request is sent when the upstream only speaks
+# chat/completions. Azure upstreams never take this path: their per-model
+# endpoint already names the deployment and its operation, so the URL is used
+# as stored and only the dialect is derived from it.
+CHAT_COMPLETIONS_PATH = "v1/chat/completions"
+
+# Cloud provider types that serve the Anthropic Messages API themselves.
+# "logos" is another Logos instance used as an upstream, which serves every
+# surface this one does.
+_NATIVE_CLOUD_PROVIDERS = frozenset({"anthropic", "logos"})
```
Claudia-Anthropica, line 50: @wasnertobias [high] Classifying an official Anthropic endpoint as native leaves its requests without the mandatory `anthropic-version` header. `ContextResolver.prepare_headers_and_payload()` creates only `Content-Type` and the single stored auth header, while `CloudModelSyncService` similarly sends only `Accept` plus auth to `/v1/models`; both Anthropic endpoints require the version header, and empty auth fields additionally default to Bearer instead of the normal `x-api-key`, so direct Anthropic inference and discovery fail even though this branch declares them supported.

In `logos/logos-orchestrator/src/logos/anthropic_compat/__init__.py`, Anthropic is classified as native although neither inference nor model discovery supplies its mandatory protocol headers. Add provider-aware header preparation for both request forwarding and cloud model synchronization, preserving explicitly configured authentication while supplying the Anthropic version and API-key conventions, with regression tests.

</details>

logos/logos-orchestrator/src/logos/anthropic_compat/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+    path = (forward_url or "").split("?", 1)[0].rstrip("/")
+    if path.endswith("/responses"):
+        return UpstreamDialect.RESPONSES
+    if path.endswith("/chat/completions"):
+        return UpstreamDialect.CHAT_COMPLETIONS
+    if (provider_type or "").lower() == "logosnode":
+        return UpstreamDialect.NATIVE
+    if (cloud_provider_type or "").lower() in _NATIVE_CLOUD_PROVIDERS:
+        return UpstreamDialect.NATIVE
```
Claudia-Anthropica, line 80: @wasnertobias [high] Native streamed Messages requests still pass through `Executor._streaming_payload()`, which injects `stream_options: {"include_usage": true}` into every URL except `/responses`. That field belongs only to Chat Completions; an Anthropic `/v1/messages` endpoint rejects it as unknown, so every streaming request to a native Anthropic provider can fail before inference.

In `logos/logos-orchestrator/src/logos/anthropic_compat/__init__.py`, native Anthropic Messages streams are classified correctly but later receive the Chat-Completions-only `stream_options` field. Make streaming payload preparation omit `stream_options` for `/messages` URLs while retaining it for translated Chat Completions requests, and add a native-Messages regression test.

</details>

logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py
```diff
@@ -0,0 +1,410 @@
...
+
+def from_chat_completion(body: Dict[str, Any], *, model_name: Optional[str] = None) -> Dict[str, Any]:
+    """Translate a chat/completions response body into an Anthropic message."""
+    choices = body.get("choices") if isinstance(body, dict) else None
+    choice = choices[0] if isinstance(choices, list) and choices else {}
+    message = choice.get("message") if isinstance(choice, dict) else {}
+    message = message if isinstance(message, dict) else {}
+
+    content: List[Dict[str, Any]] = []
+    text = message.get("content")
```
Claudia-Anthropica, line 236: @wasnertobias [medium] Chat Completions can return a safety refusal in `message.refusal` with `content: null`, and streaming chunks expose the same text as `delta.refusal`. This translator reads only `content`, so such a response becomes an empty successful Anthropic `end_turn`, hiding the provider's explanation from the client.

In `logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py`, refusal text is discarded in both synchronous responses and streaming deltas. Translate `message.refusal` and `delta.refusal` into Anthropic text content and add regression tests for both response modes.

</details>

logos/logos-orchestrator/src/logos/anthropic_compat/responses_api.py
```diff
@@ -0,0 +1,359 @@
...
+    content: List[Dict[str, Any]] = []
+    saw_tool_call = False
+
+    for item in body.get("output") or []:
+        if not isinstance(item, dict):
+            continue
+        kind = item.get("type")
+        if kind == "message":
+            for part in item.get("content") or []:
+                if isinstance(part, dict) and part.get("type") == "output_text" and part.get("text"):
```
Claudia-Anthropica: @wasnertobias [medium] Responses API message content may contain a `{"type":"refusal","refusal":"..."}` part, with streaming text delivered through `response.refusal.delta`. Accepting only `output_text` and `response.output_text.delta` silently turns these refusals into empty successful Anthropic messages.

In `logos/logos-orchestrator/src/logos/anthropic_compat/responses_api.py`, refusal content and refusal streaming events are ignored. Convert Responses refusal parts and `response.refusal.delta` events into Anthropic text content and cover both synchronous and streaming forms with tests.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The two latest Chat Completions findings are fixed on this head, and all retained threads are resolved. Four current protocol gaps remain: native Anthropic forwarding omits required headers and receives a Chat-Completions-only streaming parameter, while translated refusals are silently dropped in both OpenAI dialects. Captured CI has no failed checks; several non-required jobs are still unfinished.

logos/logos-orchestrator/src/logos/anthropic_compat/__init__.py
wasnertobias: Fixed in 3c0cc174e. You are right that the branch declared support the rest of the code did not provide.

Both conventions now travel with the provider type: `cloud_protocol_headers()` supplies `anthropic-version` (pinned, `LOGOS_ANTHROPIC_VERSION`-overridable) and `cloud_auth_header()` defaults an unset header to `x-api-key` for an Anthropic provider instead of `Authorization: Bearer`. An explicit choice by the operator is still never overridden. Applied on both paths you named — `ContextResolver.prepare_headers_and_payload()` carries them on the `ExecutionContext`, and `CloudModelSyncService` sends them to `/v1/models`.

Covered by `test_an_anthropic_upstream_gets_its_mandatory_version_header` (resolver, asserts no Authorization header), `test_other_cloud_providers_get_no_protocol_headers`, `test_an_anthropic_upstream_is_discovered_with_its_own_conventions` (sync) and the two unit tests on the helpers.

logos/logos-orchestrator/src/logos/anthropic_compat/__init__.py
wasnertobias: Fixed in 3c0cc174e. `_streaming_payload` now adds `stream_options` only for the surface that takes it: `_takes_stream_options()` excludes both `/responses` and `/v1/messages`. Usage on the Messages surface arrives in `message_start` and `message_delta`, so nothing is lost by not asking for it.

Covered by `test_messages_url_skips_stream_options` and `test_is_messages_url_ignores_query_and_trailing_slash`, alongside the existing Responses cases in the same file.

logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py
wasnertobias: Fixed in 3c0cc174e. `from_chat_completion()` falls back to `message.refusal` when `content` is empty, and the stream translator reads `delta.refusal` the same way it reads `delta.content`, so the explanation reaches the client as a text block instead of an empty `end_turn`.

Covered by `test_a_safety_refusal_reaches_the_client` and `test_a_streamed_refusal_reaches_the_client`.

logos/logos-orchestrator/src/logos/anthropic_compat/responses_api.py
wasnertobias: Fixed in 3c0cc174e. A `refusal` content part is now accepted alongside `output_text`, and `response.refusal.delta` is relayed like `response.output_text.delta`.

Covered by `test_a_safety_refusal_reaches_the_client` and `test_a_streamed_refusal_reaches_the_client` in the Responses tests.

logos/logos-orchestrator/src/logos/pipeline/context_resolver.py
```diff
@@ -104,29 +116,28 @@ async def resolve_context(
...
-                )
-                return None
+            # Cloud credentials get the convention the provider form advertises
+            # filled in — see ``cloud_auth_header``, which the model sync uses
+            # against the same providers. A provider that configured a header
+            # but has no key cannot authenticate at all, which is an error
+            # rather than an unauthenticated request.
+            auth_value = auth_format.format(api_key or "")
+            if provider_type != "logosnode":
+                header = cloud_auth_header(auth_name, auth_format, api_key, cloud_type)
```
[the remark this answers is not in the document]
wasnertobias: Fixed in the commit above. You are right that guarding only the once-per-interval sync was a false sense of security — the key went out in the clear on every actual request.

`resolve_context()` now refuses before returning an `ExecutionContext`. Checked on the *resolved* URL rather than `base_url`, since a per-model endpoint can point somewhere else entirely, and only when a credential would actually be attached, so an upstream that serves unauthenticated is unaffected. Same predicate the benchmark path already uses for these credentials: HTTPS, or plain HTTP on loopback only.

Worth flagging as a behaviour change: a cloud provider configured with a plain-HTTP non-loopback URL *and* an API key now stops resolving, with a log line naming the provider. Loopback development setups and unauthenticated upstreams keep working.

Covered by `test_credentials_are_not_sent_over_plain_http`, `test_a_cleartext_per_model_endpoint_is_refused_too`, `test_an_anthropic_provider_on_plain_http_is_refused` (both cases you asked for), plus `test_loopback_http_keeps_working` and `test_an_unauthenticated_cleartext_upstream_keeps_working` to pin what must not break.

logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py
```diff
@@ -0,0 +1,417 @@
...
+        # added by ``Executor._streaming_payload`` for every non-Responses
+        # forward URL, which is where a translated Messages request goes.
+        result["stream"] = payload["stream"]
+
+    tools = anthropic_tools(payload.get("tools"))
+    if tools:
+        result["tools"] = [{"type": "function", "function": tool} for tool in tools]
+        choice = _tool_choice(payload.get("tool_choice"))
+        if choice is not None:
+            result["tool_choice"] = choice
```
Claudia-Anthropica, line 98: @wasnertobias [medium] Anthropic's `tool_choice.disable_parallel_tool_use` is discarded here and by `to_responses`. Both OpenAI surfaces default `parallel_tool_calls` to true, so a request explicitly limited to one tool call can still produce several calls and trigger multiple side effects. Map this flag to `parallel_tool_calls: false` in both translators and cover both dialects with regression tests.

In `logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py`, `to_chat_completions` ignores `tool_choice.disable_parallel_tool_use`, so translated requests can permit parallel tool calls despite the client's explicit restriction. Map the flag to `parallel_tool_calls: false` in both Chat Completions and Responses request translators and add regression tests for both.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The prior findings are fixed on this head, all 22 retained threads are resolved, and captured CI has no failed checks. One Messages compatibility gap remains: both OpenAI request translators discard the client's explicit restriction against parallel tool calls, allowing multiple side-effecting calls instead.

logos/logos-orchestrator/src/logos/anthropic_compat/chat_completions.py
wasnertobias: Fixed in the commit above, in both translators. `disables_parallel_tool_use()` sits in `common.py` so the two dialects cannot drift, and each maps it to `parallel_tool_calls: false`.

Emitted only alongside tools — OpenAI rejects the parameter when no tools are supplied, so setting it unconditionally would have traded one 400 for another.

Covered by `test_disable_parallel_tool_use_is_carried_over` in both dialects (which also pins that an absent or false flag leaves the upstream default alone) and `test_parallel_tool_calls_is_not_sent_without_tools`.

logos/logos-orchestrator/src/logos/anthropic_compat/common.py
```diff
@@ -0,0 +1,537 @@
...
+        out = self.start()
+        out += self._close()
+        self._finished = True
+        out.append(
+            sse(
+                "message_delta",
+                {
+                    "type": "message_delta",
+                    "delta": {"stop_reason": reason or "end_turn", "stop_sequence": None},
+                    "usage": {"output_tokens": max(self._output_tokens, 0)},
```
Claudia-Anthropica: @wasnertobias [medium] In a normal text stream, `text()` emits `message_start` before the Chat Completions usage-only chunk or Responses terminal usage arrives, fixing input/cache usage at zero. `record_usage()` later receives the real counts, but this terminal event publishes only `output_tokens`, so every translated streaming text response reports zero prompt usage to Claude Code; the translators also discard the cost fields added by `_StreamingCostEnricher`. Emit the settled cumulative usage in `message_delta` and add text-before-usage regression coverage for both dialects.

In `logos/logos-orchestrator/src/logos/anthropic_compat/common.py`, translated text streams emit `message_start` before upstream usage arrives and later expose only output tokens, permanently losing settled input/cache usage and enriched cost metadata. Retain the complete normalized terminal usage and emit it cumulatively in `message_delta`, with regression tests for both Chat Completions and Responses streams where text precedes the usage frame.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The previous parallel-tool finding is fixed on this head, all retained threads are resolved, and captured CI completed without failures. One streaming compatibility issue remains: translated text streams lose their settled prompt/cache usage and enriched cost metadata before reaching the Anthropic client.

logos/logos-orchestrator/src/logos/anthropic_compat/common.py
wasnertobias: Fixed in the commit above. Reproduced first — a text-first stream with a terminal usage chunk of 1234 prompt / 500 cached / 0.0042 EUR came out as `message_start.usage.input_tokens = 0` and `message_delta.usage = {"output_tokens": 7}`, so all three were lost.

`message_delta` now carries the whole settled usage — prompt, output, cache read — plus the cost fields, which are collected through `record_usage(extras=...)` from the same upstream usage object the enricher writes into. The Logos-added keys had two definitions, one per dialect; they are now one `usage_extras()` helper in `common.py` so the two cannot drift.

Covered by `test_the_terminal_event_carries_the_settled_usage` in both dialects, each ordering the text before the usage frame exactly as you described.

PR #801, opened 2026-08-26
`Logos`: Add Apple Silicon (MLX) worker node

wasnertobias:
## Summary

Adds a third worker-node variant alongside the CUDA and Ollama builds: Apple
Silicon, serving MLX models through
[vllm-metal](https://github.com/vllm-project/vllm-metal). Target model
`mlx-community/Qwen3.8-27B-8bit` — upstream added support for it in 08/2026
("a 27B hybrid SDPA + GDN linear model on a single Apple Silicon Mac").

No issue number was given — happy to add `Closes #NNN` if one exists.

**Requires vllm-metal ≥ 0.3.0.dev20260826.** 0.2.0 cannot load Qwen3.8 at all.

## Why this image is not a container

**Metal cannot be passed into a container.** Docker on macOS runs a Linux VM
with no GPU passthrough, and Apple's own `container` framework has the same
limitation; MLX inside a container silently falls back to the CPU. Docker hit
this with their own vllm-metal backend (Docker Desktop 4.62) and solved it the
same way this PR does: publish the payload as an OCI artifact, extract it on
the host, run natively.

Running natively is also what preserves orchestrator control — a native process
can fork `vllm serve` subprocesses on command exactly as the CUDA worker does,
so `add_lane` / `delete_lane` / `reconfigure_lane` keep working. A containerised
worker could not reach the GPU to start them.

```
CI (GitHub Actions)                    Mac (native)
┌────────────────────────┐            ┌─────────────────────────────────┐
│ Dockerfile.mlx         │   pull     │ bootstrap-macos.sh              │
│  → source only,        │ ─────────► │  docker create + docker cp      │
│    never run           │            │   → ~/logos-workernode-mlx      │
│ ghcr.io (public)       │            │  launchctl bootstrap            │
└────────────────────────┘            │ logos_worker_node.main          │
                                      │  ├── outbound WS → orchestrator │
                                      │  └── subprocess: vllm serve     │
                                      │        → Metal GPU              │
                                      └─────────────────────────────────┘
```

**Registry:** this image alone goes to **public GHCR**, not Harbor, so a Mac can
bootstrap without Harbor credentials. A GHCR package is private on first push —
it needs one manual flip to public under Package settings.

## Changes

**Worker**
- `logos_worker_node/metal_process.py` (new) — `MetalVllmProcessHandle`
  subclasses `VllmProcessHandle`, overriding only the command line, the
  environment and the CUDA preflight guards. Spawn, readiness polling, log
[...]

PR #927, opened 2026-09-08
`Logos`: Fix add-provider dialog — hide cloud field for local nodes, auto-generate worker key

### LINKED ISSUE #902: `Logos`: Adding a logosnode provider still shows cloud options and does not auto-generate an API key
Two bugs in the "Add Provider" dialog:

**1. Cloud options still shown for `logosnode`**

When adding a provider with Provider Type `logosnode`, the "Cloud Provider" and "Privacy Level" selects are still displayed (only disabled, with `none`/`LOCAL` pre-filled). Since `logosnode` providers are local by definition, these cloud-related fields should not be shown at all.

**2. API key is not auto-generated**

When adding a provider, no API key is generated — the field only accepts manual entry and the webservice stores whatever is submitted as-is. For `logosnode` providers this is especially annoying, because the worker node needs the shared key to authenticate. The orchestrator's bootstrap endpoint `POST /logosdb/providers/logosnode/register` already auto-generates such a key (`secrets.token_urlsafe(48)`), but the UI add-provider flow does not.

**To Reproduce**

1. Logos → Providers → "Add Provider"
2. Set Provider Type to `logosnode`
3. Observe: the "Cloud Provider" and "Privacy Level" fields are still displayed
4. Observe: the API Key field stays empty — nothing is auto-generated

**Expected behavior**

- For `logosnode` (local) providers, no cloud-related options/fields should be shown.
- An API key should be auto-generated for new providers — at least for `logosnode` — and the generated key should be shown to the user so the worker node can be configured with it.

**Relevant code**

- UI dialog: `logos/logos-ui/src/app/features/providers/providers.html` (Add dialog ~L190–300: manual API Key field L224–236, Cloud Provider / Privacy Level selects L277–296) and `providers.ts` (`coercedForType()` L102–112, per-type option lists L73–100)
- Webservice: `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ProviderService.java` — `addProvider()` (L56–73) stores the user-supplied `apiKey` as-is, no key generation
- Orchestrator: `logos/logos-orchestrator/src/logos/main.py` — `logosnode_register` (L4780–4820) is the only place a provider API key is auto-generated (L4791)

wasnertobias:
## Summary

Fixes #902 — two bugs in the Add Provider dialog.

**1. Cloud Provider select shown for local nodes.**
For `logosnode` (local) providers the *Cloud Provider* select was still displayed, only disabled with `none` pre-filled. Cloud vendors don't apply to self-hosted nodes, so the field is now hidden entirely (`@if (addProviderType() === 'cloud')`). The *Privacy Level* select stays **enabled** for logosnode, so a node on hardware outside operator control (e.g. a personal Mac MLX worker) can be marked `THIRD_PARTY_HARDWARE` instead of `LOCAL` — the comment on the issue. Same change applied to the Edit dialog.

**2. No API key auto-generated for new providers.**
The webservice stored whatever key was submitted as-is (i.e. an empty string). For logosnode this is the key the worker node uses to authenticate, so it must not be left blank. `ProviderService.addProvider` now generates a random key when none is supplied for a `logosnode` provider — `SecureRandom` 48 bytes, base64url, no padding, i.e. equivalent to the orchestrator's `secrets.token_urlsafe(48)` used in `logosnode_register`. Cloud providers keep whatever key was sent, unchanged.

The generated key is echoed back in the `add_provider` response and shown to the operator in a copyable follow-up modal (it's only available at creation time, so the UI makes it easy to copy into the worker config).

## Changes

- `logos/logos-ui/src/app/features/providers/providers.html`
  - Add dialog: hide Cloud Provider select for logosnode; keep Privacy Level enabled for both types.
  - Edit dialog: same.
  - New "Provider created" modal displaying the generated shared key with a copy button.
- `logos/logos-ui/src/app/features/providers/providers.ts`
  - Signals + methods for the created-key modal (`createdKey*`, `copyCreatedKey`, `closeCreatedKeyDialog`).
  - `submitAdd` captures the `api_key` from the response and opens the modal when a key was generated.
- `logos/logos-ui/src/app/features/providers/providers.scss`
  - Styles for the created-key modal (`.created-key-intro`, `.created-key-row`).
- `logos/logos-webservice/.../configuration/service/ProviderService.java`
  - Auto-generate + echo API key for logosnode; `generateApiKey()` helper.
- `logos/logos-webservice/.../configuration/ProviderControllerTest.java`
  - 3 new tests: logosnode generates a key when absent, echoes an explicitly supplied key, cloud does not generate one.

## Verification

- `mvn test -Dtest=ProviderControllerTest` → 20 tests, 0 failures, 0 errors (full Liquibase schema incl. changeset 024 applied in Testcontainers Postgres 17).
- `ng build` → success (only pre-existing SCSS budget warnings for unrelated files).

## Out of scope

No new provider type, no `provider_type_enum` migration, no orchestrator changes. `THIRD_PARTY_HARDWARE` remains a **privacy level** (introduced on main in #801), selectable through the now-enabled Privacy Level field.

## Summary by CodeRabbit

* **New Features**
  * Added automatic API key generation when creating Logos worker node providers without a supplied key.
  * Added a one-time dialog displaying the generated API key, with copy-to-clipboard support and confirmation feedback.
  * Cloud Provider selection now appears only for cloud provider types.
* **Bug Fixes**
  * Removed unnecessary disabling of Cloud Provider and Privacy Level fields for Logos worker node providers.

### DIFF
--- a/logos/logos-ui/src/app/features/providers/providers.html
+++ b/logos/logos-ui/src/app/features/providers/providers.html
@@ -276,23 +276,25 @@
         (valueChange)="onAddProviderTypeChange($any($event))"
       />
     </div>
+    @if (addProviderType() === 'cloud') {
     <div class="form-field">
       <label class="field-label" for="add-cloud-type">Cloud Provider</label>
       <app-select
         [id]="'add-cloud-type'"
         [options]="addCloudProviderTypeOptions()"
         [value]="addCloudProviderType()"
-        [disabled]="addLoading() || addProviderType() === 'logosnode'"
+        [disabled]="addLoading()"
         (valueChange)="addCloudProviderType.set($any($event))"
       />
     </div>
+    }
     <div class="form-field">
       <label class="field-label" for="add-privacy">Privacy Level</label>
       <app-select
         [id]="'add-privacy'"
         [options]="addPrivacyLevelOptions()"
         [value]="addPrivacyLevel()"
-        [disabled]="addLoading() || addProviderType() === 'logosnode'"
+        [disabled]="addLoading()"
         (valueChange)="addPrivacyLevel.set($any($event))"
       />
     </div>
@@ -319,6 +321,47 @@
         <i class="pi pi-spin pi-spinner" aria-hidden="true"></i>
       }
       Add Provider
+    </button>
+  </div>
+</app-modal-form>
+
+<!-- ── Created Provider API Key Dialog ─────────────────────────────────────── -->
+<app-modal-form
+  [visible]="createdKeyOpen()"
+  (visibleChange)="$event || closeCreatedKeyDialog()"
+  title="Provider created"
+>
+  <div class="form-body">
+    <p class="created-key-intro">
+      <strong>{{ createdKeyName() }}</strong> is a Logos worker node. Authenticate it
+      with the shared key below — it is shown only now, so copy it to your worker
+      configuration.
+    </p>
+    <div class="form-field">
+      <label class="field-label" for="created-key">Shared API Key</label>
+      <div class="created-key-row">
+        <input
+          id="created-key"
+          class="field-input"
+          type="text"
+          readonly
+          [value]="createdKey()"
+          autocomplete="off"
+          spellcheck="false"
+        />
+        <button class="btn-secondary" type="button" (click)="copyCreatedKey()">
+          @if (createdKeyCopied()) {
+            <i class="pi pi-check" aria-hidden="true"></i> Copied!
+          } @else {
+            <i class="pi pi-copy" aria-hidden="true"></i> Copy
+          }
+        </button>
+      </div>
+    </div>
+  </div>
+  <div class="modal-footer" modal-footer>
+    <button class="btn-primary" type="button" (click)="closeCreatedKeyDialog()">
+      Done
     </button>
   </div>
 </app-modal-form>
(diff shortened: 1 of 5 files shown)


Claudia-Anthropica (dismissed):
@wasnertobias The logosnode-only key generation uses the intended 48-byte URL-safe format, persists and returns the key correctly, and the UI exposes it through the creation flow while applying the requested provider/privacy controls. The captured UI and webservice builds, tests, and lint checks succeeded, including coverage for generated, supplied, and cloud-provider key behavior.

wasnertobias:
Fixed — the modal now opens only when the operator left the key field empty (`payload.api_key === undefined`). An operator-supplied key is echoed back by the backend and no longer triggers the modal, so the "shown only now" wording stays accurate. (c5e6bad4c)

wasnertobias:
Fixed — `generateApiKey()` now reuses a static `SecureRandom` and a static encoder, matching the `ApiKeyFactory` pattern. (c5e6bad4c)

wasnertobias:
Fixed — the assertion now uses `matchesPattern("[A-Za-z0-9_-]{64}")`, so an empty string can no longer pass. (c5e6bad4c)

Claudia-Anthropica (approved):
@wasnertobias The logosnode-only generation path securely creates, persists, returns, and displays the intended 64-character base64url key while preserving explicit and cloud-provider keys. The latest commit resolves the prior modal, SecureRandom reuse, and assertion concerns, and the captured UI/webservice builds, tests, and lint checks succeeded.

PR #908, opened 2026-09-06
`Logos`: Remove the dashboard

LogosOSSAgent:

PR #905, opened 2026-09-04
`Logos`: Hide the Agents menu entry when the runner is absent

LogosOSSAgent:
## Summary by CodeRabbit

* **New Features**
  * The Agents navigation entry now appears only when the agent service is available and responds successfully.
  * The entry remains hidden while availability is checked or when the service is unreachable.

* **Bug Fixes**
  * Prevented the Agents entry from appearing for fallback or unexpected service responses.

* **Tests**
  * Added coverage for availability checks, failures, fallback responses, and role-based menu visibility.

PR #929, opened 2026-09-09
`Logos`: Show the exact cold-start percentage on the stats page

LogosOSSAgent:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved percentage formatting on the Statistics page.
  * Cold-start rates now show precise values, including small percentages, instead of rounding them to whole numbers.
  * Invalid or unavailable percentage values display as `0%`.

* **Tests**
  * Added coverage for standard, fractional, very small, and invalid percentage values.

### DIFF
--- a/logos/logos-ui/src/app/features/statistics/statistics.html
+++ b/logos/logos-ui/src/app/features/statistics/statistics.html
@@ -113,7 +113,7 @@
         <app-stats-kpi-card
           label="Cold starts"
           [accent]="CHART_ROLE.local"
-          [value]="coldPct() + '%'"
+          [value]="formatColdPct()"
         >
           <app-stats-sparkline kpiRight [data]="sparkLocal()" [color]="CHART_ROLE.local" />
           <span kpiHint>
--- a/logos/logos-ui/src/app/features/statistics/statistics.ts
+++ b/logos/logos-ui/src/app/features/statistics/statistics.ts
@@ -23,6 +23,7 @@ import {
   chooseDynamicBucketMs,
   chooseDynamicTargetBuckets,
   extractProviderVramMb,
+  formatPercent,
   formatRangeLabel,
   formatTokenCount as formatTokenCountValue,
   normalizeFeedStatus,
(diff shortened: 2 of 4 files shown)


Claudia-Anthropica (dismissed):
@LogosOSSAgent The focused formatter change preserves fractional cold-start rates for the reported case and adds coverage for invalid, ordinary, and small inputs. The Logos UI checks passed; the non-required vLLM worker prepare failure is unrelated to this UI-only diff.

wasnertobias:
Pushed `4c867aa88`, addressing the open CodeRabbit thread, plus current `main` merged in (clean).

**The finding is valid.** `formatPercent` documents that it never collapses a non-zero share to `"0%"`, but the widening loop stops at `decimals < 6`, so anything below 0.0000005% still went `toFixed(6)` → `"0.000000"` → `"0%"`. `formatPercent(1, 10_000_000_000)` was the counterexample. Past the cap it now returns `<0.000001%`, and the doc comment states the bound. Regression tests cover both sides of the cap: `1/10_000_000_000` and `1/1_000_000_000` bound, while `1/100_000_000` still reads its exact `0.000001%`.

**Verification.** `npm run build` passes on the merged head. The `formatPercent` suite passes — verified on the pre-merge head, where the whole spec bundle still compiles.

**Two pre-existing breakages in `ng test` on `main`, unrelated to this PR** — flagging them because they are invisible to CI (`logos_test.yml` only runs `npm run build`, never `ng test`):

1. `src/app/features/ai-tools/ai-tools.spec.ts:37` — `TS2741: Property 'rate_limit_usage' is missing in type ... but required in type 'MyKey'`. The spec bundle does not compile, so **no** UI unit test can run. Reproduced on a clean `origin/main` checkout (`11e8ddb34`).
2. `src/app/core/layout/shell/shell.spec.ts:152` — the expected menu labels have drifted: the spec still expects `Dashboard`, while the shell now renders `Users` and `Teams`.

Both want their own PR; neither blocks this one. Happy to pick them up if useful.

LogosOSSAgent:
The `formatPercent` fix is good. I verified it against the code: the bound after the widening loop's cap (`pct > 0 && Number(pct.toFixed(decimals)) === 0` → `'<0.000001%'`) is the right place to catch it, the doc comment now states the bound, and all 23 assertions in the spec pass when I run them against the function as it stands on `HEAD`. I also ran the three new cases against the pre-fix code: the two `<0.000001%` cases print `0%` there (the `1/100_000_000` case passes on both sides), so they are genuine regression tests. One caveat on method: `node_modules` cannot be installed in this environment, so I extracted the function verbatim and ran it in node rather than through `ng test`.

On the two pre-existing breakages:

1. **`ai-tools.spec.ts` — confirmed real.** `MyKey` requires `rate_limit_usage` (`my-key.model.ts:41`), and the `KEY` literal at `ai-tools.spec.ts:37` omits it, so the spec bundle cannot type-check and no UI unit test can run. The fix is one line: add `rate_limit_usage: null` to that literal. Yes, pick it up as its own PR — but keep it out of this one: the diff here is confined to the statistics feature, and this PR's CI (the `logos_test.yml` UI job runs only `npm run build`, verified) is not blocked by it.

2. **`shell.spec.ts` — no longer exists on `main`.** The `Dashboard` expectation was added in `094c3a8` (#905), the dashboard itself went away in `357c757` (#908), and the stale `'Dashboard',` line was deleted from the spec in `e3e125e` (#936) — which is an ancestor of `11e8ddb` and already merged into this branch. A case-insensitive grep finds no `Dashboard` anywhere in the UI sources, and the test at `shell.spec.ts:152` (the `arrayContaining` of Statistics/Models/Providers/Policies/Billing) holds against the current nav configuration for its `logos_admin` + SPA-fallback scenario. So that observation predates #936; there is nothing to fix there.

No files changed on this branch: the fix on `HEAD` is correct as is, and neither item belongs in this PR.

PR #939, opened 2026-09-09
`Logos`: Fix calibrate uncalibrated permission handling

julia-rai:
## Summary by CodeRabbit

- **Bug Fixes**
  - Calibration checks now use the worker’s current runtime state, improving identification of models requiring calibration.

### DIFF
--- a/logos/logos-orchestrator/src/logos/main.py
+++ b/logos/logos-orchestrator/src/logos/main.py
@@ -3727,36 +3727,39 @@ def _find_uncalibrated_models_on_provider(provider_id: int) -> list[str]:
     but this lets the API caller see the candidate list up front. Sourced
     from configured_models so models the worker stripped from
     capabilities_models (because they have no profile yet) are visible.
+
+    Reads the live session snapshot directly rather than through
+    _logosnode_facade: the facade's provider entry is populated from DB
+    model_provider links, which sync_logosnode_capabilities prunes to
+    match capabilities_models. A worker with every model uncalibrated
+    reports an empty capabilities_models, so that sync deletes all of its
+    links and the facade drops the provider entirely — leaving nothing
+    here to discover, even though the raw snapshot still has it all.
     """
-    if _logosnode_facade is None:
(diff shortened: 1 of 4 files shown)


logos/logos-orchestrator/src/logos/main.py
```diff
@@ -3727,36 +3727,39 @@ def _find_uncalibrated_models_on_provider(provider_id: int) -> list[str]:
...
             profile is None
-            or profile.base_residency_mb is None
-            or profile.sleeping_residual_mb is None
-            or profile.sleep_l1_transient_host_ram_mb is None
-            or (
-                profile is not None
-                and profile.residency_source == "calibrated"
-                and not profile.kv_cache_to_max_model_len_pairs
-            )
+            or profile.get("base_residency_mb") is None
```
Claudia-Anthropica, line 3759: @julia-rai [medium] Keep this dispatch gate aligned with `LogosBridgeClient._list_uncalibrated_models()`. The worker skips `calibration_unsupported` profiles, treats sleep measurements as optional when `sleep_mode_disabled` or the session sleep level requires it, and explicitly selects `residency_source == "hf"`; this predicate does none of those, so no-sleep/unsupported profiles produce misleading no-op sessions while an otherwise complete HF-derived profile can make the endpoint return “No uncalibrated models” without dispatching the calibration the worker would perform.

In `logos/logos-orchestrator/src/logos/main.py`, `_find_uncalibrated_models_on_provider` uses a selection predicate that disagrees with the worker’s authoritative `_list_uncalibrated_models`, causing false-positive and false-negative calibration decisions. Mirror the worker’s unsupported, sleep-applicability, and HF-source rules here and add focused tests covering those profiles.

</details>

Claudia-Anthropica (changes requested):
@julia-rai Reading the live registry snapshot fixes discovery when capability synchronization removes the facade provider, and the captured checks completed successfully. One medium correctness issue remains: the server’s candidate predicate diverges from the worker’s selector, so the endpoint can suppress required calibration or start sessions the worker treats as no-ops.

PR #941, opened 2026-09-10
`Logos`: Report failed manual lane loads to the UI

wasnertobias:
## Problem

Testing "load a model twice on one worker node" on deioma (Qwen/Qwen3.8-27B ×2), the second load **silently failed**: the VRAM reservation was correctly denied (3 GPUs, first replica already holds GPU 1+2 with a pre-allocated KV pool → no pair fits a second TP=2 replica), but the only trace was one W line in the orchestrator log:

```
planner W VRAM reservation denied for load of Qwen/Qwen3.8-27B: need=94757MB avail=162463MB committed=0MB gpus=0,1
```

The UI's "Loading … this can take minutes" note never resolves, because it only goes away when the lane appears in the status stream — which never happens for a denied load. Three layers of silence:

1. the per-GPU denial details in `vram_ledger.try_reserve_atomic` are DEBUG-only, so the W line reads contradictory (need < avail);
2. `lanes/add` answers 202 and refuses only in the background;
3. nothing ever reports the refusal to the UI.

## What this PR does (fix options 1 + 3 from the analysis)

**Option 1 — the failure reaches the UI** (the important part):

- The planner records the outcome of the most recent manual load per `(provider_id, model)`: `running | succeeded | failed` plus a human-readable `reason`, TTL 1 h. Every exit of `load_lane_manually` records — pre-rejection, capacity gone, executor refusal (reason handed over per lane), confirmation timeout, no free lane id. An in-flight load's `running` entry is never stomped by a second click.
- New `POST /internal/logosnode/lanes/load_status` (internal-secret auth) serves the outcome; `lanes/add` records `running` *before* answering 202 so the UI can't read the previous attempt's `failed`. No outcome → `unknown`, which the UI must not render as a failure.
- Spring: `ProviderController.laneLoadStatus` pass-through with the usual JWT + `LOGOS_ADMIN` gate, plus the new `LaneLoadStatusRequestDTO` and `OrchestratorWorkerAdminClient.getLaneLoadStatus`.
- Traefik: `lanes/load_status` added to the `webservice-logosnode-admin` (priority 202) exact-path rule in both compose files — without this the call would 404 at the orchestrator, as documented in the prod compose comment.
- UI: while the pending note is up, the lane-health panel polls the outcome every 2.5 s (hard stop past the full two-phase attempt, 65 min; stops on provider switch / 404 / terminal answer). A recorded `failed` reopens the picker with `Loading <model> failed: <reason>` and re-offers the model — exactly where a synchronous refusal lands.

**Option 3 — the log line tells the whole story**:

- The denial reason is recorded per lane (`not enough free VRAM for this model: it needs ~48.4 GB in total (~24.2 GB on each of 2 GPUs), but the worker reports only 63.4 GB free (GPU 0: 96595MB, GPU 1: 31705MB, GPU 2: 31705MB)`).
- The W line is enriched with `need-per-GPU=…MB per-GPU-free=[…]`, so it no longer reads as a bug.

Option 2 (synchronous 409 pre-check) is deliberately not included — the pre-check can't see per-GPU feasibility anyway and the background refusal is now reported, keeping the PR focused.

## Follow-up fixes (from the review rounds)

- Stale action feedback on provider switch: the lane-health and GPU panels reset unload/sleep/wake/calibration state — including in-flight answers — when the worker changes.
- Outcome-poll hardening: per-session generation token (a delayed answer from a superseded retry of the same model is dropped), the poll survives a plain picker close, and the cap covers the full command + confirmation attempt.
- Denial diagnostics report the reservation gate's effective free VRAM (raw minus committed) instead of the raw snapshot.
- `StatsV2WebSocketHandler` pushes VRAM deltas on every 1 s tick instead of every fifth: a loaded lane reaches the UI ~1 s after the worker reports it, not up to 5 s later.
- `ai-tools.spec.ts` fixture carries the `rate_limit_usage` field `MyKey` gained in #847, so the UI test build on main is healed.

## Deploy note

The **live** `/opt/logos/docker-compose.yaml` on the host needs the same one-line addition to the `webservice-logosnode-admin` rule (`|| Path('/api/logosdb/providers/logosnode/lanes/load_status')`) when the new image goes out, or the UI's poll 404s (it degrades gracefully — the note just behaves as before until the route is there).

## Tests

- Orchestrator: 20 new planner-outcome tests (`test_manual_load_outcome.py`) + 7 endpoint tests (`test_internal_lane_load_status_endpoint.py`); full `tests/unit/capacity/` green (395 passed, 1 xfailed), related internal-endpoint tests green.
- UI: 7 new poll-behavior tests in `lane-health-panel.spec.ts` (zoneless timer-spy style, no fakeAsync); full suite green — **except a pre-existing break on main**: `ai-tools.spec.ts` is missing the `rate_limit_usage` field that `MyKey` gained in #847, so `ng test` fails at the build step on a clean checkout. That break is fixed in this PR: the spec's `KEY` literal now carries `rate_limit_usage`, so `ng test` builds green on the branch.

## Summary by CodeRabbit

- **New Features**
  - Added load-status tracking for background model loads, including running, succeeded, failed, and unknown states with operator-facing reasons.
  - The lane health panel now monitors load progress automatically and reports failures more clearly.
  - Added routing support for load-status information in development and production environments.
  - VRAM availability updates are now checked every second.

- **Bug Fixes**
  - Prevented stale feedback after switching providers or workers.
  - Improved handling of delayed, rejected, failed, and completed lane loads.
  - Added safeguards for unavailable status services and outdated responses.
  - Improved calibration-state handling when changing workers.
  - Capacity errors now provide more accurate provider and GPU availability details.

### DIFF
--- a/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
+++ b/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
@@ -6737,13 +6737,22 @@ class CapacityPlanner:
                                 provider_id,
                                 lane_id,
                             )
-                            # The model is loaded — the operator's goal is met.
-                            # Report it as success so the UI note resolves; but
-                            # not over a "running" entry: a concurrent attempt
-                            # for this same lane may still be executing and
-                            # owns the final state.
+                            # The model is loaded — the operator's goal is met,
+                            # so the UI note resolves as success. The hold-back
+                            # is a "running" entry that *names a lane*: it
+                            # belongs to a dispatch still executing, and that
+                            # dispatch's executor settles it. The exception is
+                            # the endpoint's lane-less "running" placeholder:
+                            # it is this attempt's own, and this attempt is
+                            # about to return without dispatching, so no
+                            # executor will ever settle it — leaving it would
+                            # pin the note at "Loading" for the outcome TTL
+                            # although the lane is already serving.
                             current = self.get_manual_load_outcome(provider_id, model_name)
-                            if not (current and current.get("status") == "running"):
+                            held_by_dispatch = (
+                                bool(current) and current.get("status") == "running" and "lane_id" in current
+                            )
+                            if not held_by_dispatch:
                                 self.record_manual_load_outcome(provider_id, model_name, "succeeded", lane_id=lane_id)
                             return False
                         logger.info(
--- a/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
+++ b/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
@@ -169,9 +169,13 @@ class CapacityPlanner:
     LANE_LOAD_COMMAND_TIMEOUT_S = 1800
 
     # How long the outcome of a manual load stays queryable by the UI. Must
-    # outlive the whole attempt: the worker-command budget above plus the
-    # confirmation polling that follows it.
-    MANUAL_LOAD_OUTCOME_TTL_SECONDS = 3600.0
+    # outlive the whole attempt with margin: the _lane_lock wait, the
+    # LANE_LOAD_COMMAND_TIMEOUT_S command budget, the confirmation polling of
+    # the same length that follows it, plus polling/processing overhead. A
+    # TTL that only covered 1800 s + 1800 s could expire the "running" entry
+    # before the terminal outcome is written, and the UI would read an
+    # unknown (and keep polling) on a slow-but-healthy load.
+    MANUAL_LOAD_OUTCOME_TTL_SECONDS = 7200.0
 
     # Minimum tenure: after a model wakes/loads, give it at least this
     # long to serve its queue before it can be drained for another model.
@@ -1839,6 +1843,17 @@ class CapacityPlanner:
 
     def get_lane_action_failure(self, provider_id: int, lane_id: str) -> str | None:
         return self.__dict__.get("_lane_action_failure", {}).get(self._lane_key(provider_id, lane_id))
+
+    def clear_lane_action_failure(self, provider_id: int, lane_id: str) -> None:
+        """Drop a recorded reason before a new attempt starts.
+
+        The map is per lane id, and the planner reuses lane ids across
+        attempts of the same model — without this, a new attempt that fails
+        without recording a fresh reason would report the previous attempt's
+        failure to the operator. Every executor failure path records a new
+        reason, so clearing here cannot hide information.
+        """
+        self.__dict__.get("_lane_action_failure", {}).pop(self._lane_key(provider_id, lane_id), None)
 
     def _reconcile_load_failures(self, provider_id: int, lanes: List[LaneSchedulerSignals]) -> None:
         """Keep the load-failure state in step with the lanes a worker reports.
@@ -6689,6 +6704,7 @@ class CapacityPlanner:
                         params=self._build_load_params(model_name, lane_id, profile, capacity, provider_id),
                         reason="Manual load requested by an operator",
                     )
+                    self.clear_lane_action_failure(provider_id, lane_id)
                     self.record_manual_load_outcome(provider_id, model_name, "running", lane_id=lane_id)
                     confirmed = await self._execute_action_with_confirmation(
                         action, timeout_seconds=self.LANE_LOAD_COMMAND_TIMEOUT_S
@@ -8322,9 +8338,19 @@ class CapacityPlanner:
                         # did not (or no per-GPU data exists) — say which GPUs
                         # fell short, or this line reads as a bug: need <
                         # avail, yet the reservation was denied.
+                        # Report the gate's effective numbers (raw free minus
+                        # what in-flight operations already committed), not
+                        # the raw snapshot — raw free can look sufficient
+                        # while the reservation is correctly denied.
+                        _vram = self._vram_ledger
+                        effective_avail = _vram.get_effective_available_mb(action.provider_id, raw_avail)
                         gpu_free_parts: list[str] = []
                         if _per_gpu_free:
-                            gpu_free_parts = [f"GPU {dev}: {free:.0f}MB" for dev, free in sorted(_per_gpu_free.items())]
+                            gpu_free_parts = [
+                                f"GPU {dev}: "
+                                f"{_vram.get_gpu_effective_available_mb(action.provider_id, dev, free):.0f}MB"
+                                for dev, free in sorted(_per_gpu_free.items())
+                            ]
                         tp_size = len(self._parse_gpu_device_ids(_lane_gpus)) if _lane_gpus else 0
                         per_gpu_need = (
                             _estimated_load_vram / tp_size * self.VRAM_SAFETY_MARGIN
@@ -8337,22 +8363,24 @@ class CapacityPlanner:
                         )
                         if tp_size:
                             reason += f" (~{per_gpu_need / 1024.0:.1f} GB on each of {tp_size} GPUs)"
-                        reason += f", but the worker reports only {raw_avail / 1024.0:.1f} GB free"
+                        reason += f", but only {effective_avail / 1024.0:.1f} GB are effectively free on the worker"
                         if gpu_free_parts:
                             reason += " (" + ", ".join(gpu_free_parts) + ")"
                         reason += "."
                         self.record_lane_action_failure(action.provider_id, action.lane_id, reason)
                         _per_gpu_free_str = ", ".join(gpu_free_parts) or "unknown"
                         _per_gpu_detail = (
-                            f" need-per-GPU={per_gpu_need:.0f}MB per-GPU-free=[{_per_gpu_free_str}]"
+                            f" need-per-GPU={per_gpu_need:.0f}MB per-GPU-effective=[{_per_gpu_free_str}]"
                             if _per_gpu_free
                             else ""
                         )
                         logger.warning(
                             "VRAM reservation denied for load of %s: "
-                            "need=%.0fMB avail=%.0fMB committed=%.0fMB gpus=%s%s",
+                            "need=%.0fMB effective_avail=%.0fMB "
+                            "(raw=%.0fMB committed=%.0fMB) gpus=%s%s",
                             action.model_name,
                             _estimated_load_vram,
+                            effective_avail,
                             raw_avail,
                             self.get_pending_vram_mb(action.provider_id),
                             _lane_gpus or "unknown",
--- a/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
+++ b/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
@@ -168,6 +168,11 @@ class CapacityPlanner:
     # workers and spurious load-failure cooldowns.
     LANE_LOAD_COMMAND_TIMEOUT_S = 1800
 
+    # How long the outcome of a manual load stays queryable by the UI. Must
+    # outlive the whole attempt: the worker-command budget above plus the
+    # confirmation polling that follows it.
+    MANUAL_LOAD_OUTCOME_TTL_SECONDS = 3600.0
+
     # Minimum tenure: after a model wakes/loads, give it at least this
     # long to serve its queue before it can be drained for another model.
     # Without this, a freshly-woken model has 0 active requests, making
@@ -436,6 +441,14 @@ class CapacityPlanner:
         # must keep blocking allocation until the lane serves or leaves the
         # worker. Reconciled every cycle in _reconcile_load_failures.
         self._load_failed_lane_ids: set[tuple[int, str]] = set()
+
+        # Outcome of the most recent manual load per (provider_id, model_name),
+        # for the statistics UI: the "Load lane" endpoint answers 202 and the
+        # load runs in the background, so without this the UI could never tell
+        # a slow load from a denied one. Entries expire after
+        # MANUAL_LOAD_OUTCOME_TTL_SECONDS (a load gets LANE_LOAD_COMMAND_TIMEOUT_S
+        # plus confirmation polling, so the window covers the whole attempt).
+        self._manual_load_outcomes: dict[tuple[int, str], Dict[str, Any]] = {}
 
         # Workers the cycle is currently skipping, and when the skipping began.
         # Backs the stuck-worker warning in _note_unplannable: a worker no
@@ -1746,6 +1759,84 @@ class CapacityPlanner:
             self._lane_load_failure_until.pop(key, None)
             return False
         return True
+
+    # ------------------------------------------------------------------
+    # Manual-load outcomes — the "Load lane" button answers 202 and the
+    # load runs in the background, so the UI can only learn the outcome by
+    # asking. Without this, a denied load (no room on the worker) looks
+    # identical to a slow one: the lane simply never appears, and the
+    # operator's only trace is a W line in the orchestrator log.
+    # ------------------------------------------------------------------
+
+    def record_manual_load_outcome(
+        self,
+        provider_id: int,
+        model_name: str,
+        status: str,
+        *,
+        lane_id: str | None = None,
+        reason: str | None = None,
+    ) -> None:
+        """Record the state of the most recent manual load of *model_name*.
+
+        *status* is one of ``"running"``, ``"succeeded"``, ``"failed"``;
+        *reason* is a human-readable explanation, meant for the operator who
+        clicked "Load lane" (e.g. which GPUs lacked how much VRAM). The entry
+        replaces any previous one for the (provider, model) pair — a retry
+        supersedes the attempt it retries.
+        """
+        entry: Dict[str, Any] = {"status": status, "updated_at": time.time()}
+        if lane_id is not None:
+            entry["lane_id"] = lane_id
+        if reason is not None:
+            entry["reason"] = reason
+        self._manual_load_outcome_store()[(provider_id, model_name)] = entry
+        self._purge_manual_load_outcomes()
+
+    def get_manual_load_outcome(self, provider_id: int, model_name: str) -> Dict[str, Any] | None:
+        """The most recent manual-load outcome for (provider, model), or None.
+
+        None means "no manual load of this model is known" — the caller must
+        not render it as a failure. Expired entries are dropped on read so
+        the store cannot grow unbounded.
+        """
+        store = self._manual_load_outcome_store()
+        entry = store.get((provider_id, model_name))
+        if entry is None:
+            return None
+        if time.time() - float(entry.get("updated_at") or 0.0) > self.MANUAL_LOAD_OUTCOME_TTL_SECONDS:
+            store.pop((provider_id, model_name), None)
+            return None
+        return dict(entry)
+
+    def _manual_load_outcome_store(self) -> Dict[tuple[int, str], Dict[str, Any]]:
+        """The outcome map, created on demand so harness planners that skip
+        __init__ work unchanged."""
+        store = self.__dict__.get("_manual_load_outcomes")
+        if store is None:
+            store = self._manual_load_outcomes = {}
+        return store
+
+    def _purge_manual_load_outcomes(self) -> None:
+        """Drop expired entries; bounded work, called on every record."""
+        now = time.time()
+        store = self._manual_load_outcome_store()
+        expired = [
+            key for key, entry in store.items() if now - float(entry.get("updated_at") or 0.0) > self.MANUAL_LOAD_OUTCOME_TTL_SECONDS
+        ]
+        for key in expired:
+            store.pop(key, None)
+
+    # The executor (_execute_action_with_confirmation) fails for reasons the
+    # manual-load path cannot know on its own — a denied VRAM reservation, a
+    # worker rejection, a confirmation timeout. It records a per-lane
+    # human-readable reason here; load_lane_manually picks it up for the
+    # outcome it records.
+    def record_lane_action_failure(self, provider_id: int, lane_id: str, reason: str) -> None:
+        self.__dict__.setdefault("_lane_action_failure", {})[self._lane_key(provider_id, lane_id)] = reason
+
+    def get_lane_action_failure(self, provider_id: int, lane_id: str) -> str | None:
+        return self.__dict__.get("_lane_action_failure", {}).get(self._lane_key(provider_id, lane_id))
 
     def _reconcile_load_failures(self, provider_id: int, lanes: List[LaneSchedulerSignals]) -> None:
         """Keep the load-failure state in step with the lanes a worker reports.
@@ -6488,6 +6579,10 @@ class CapacityPlanner:
         rejection = self.manual_load_rejection_reason(provider_id, model_name)
         if rejection is not None:
             logger.warning("Refusing manual load of %s on worker=%s: %s", model_name, provider_id, rejection)
+            # Normally the endpoint already answered 409 with this text before
+            # creating the task; record it anyway so a race between the two
+            # checks leaves the UI with a failure, not a hanging "Loading".
+            self.record_manual_load_outcome(provider_id, model_name, "failed", reason=rejection)
             return False
 
         # A load of this model already in flight on this worker: a second
@@ -6496,6 +6591,10 @@ class CapacityPlanner:
         # so the pick below would hand out the next free suffix and place a
         # second copy of the very model the operator is already waiting for.
         if any(model.lower() == model_name.lower() for model in self._inflight_load_models(provider_id).values()):
+            # Deliberately no outcome record: the in-flight load (this click's
+            # first copy, or a planner one) owns the entry and will write the
+            # final state. Overwriting "running" with a no-op here would make
+            # the UI drop its waiting note while the load is still going.
             logger.info(
                 "Manual load of %s on worker=%s is a no-op: a load of the model is already in flight",
                 model_name,
@@ -6533,6 +6632,16 @@ class CapacityPlanner:
                                 provider_id,
                                 lane_id,
                             )
+                            # The model is loaded — the operator's goal is met.
+                            # Report it as success so the UI note resolves; but
+                            # not over a "running" entry: a concurrent attempt
+                            # for this same lane may still be executing and
+                            # owns the final state.
+                            current = self.get_manual_load_outcome(provider_id, model_name)
+                            if not (current and current.get("status") == "running"):
+                                self.record_manual_load_outcome(
+                                    provider_id, model_name, "succeeded", lane_id=lane_id
+                                )
                             return False
                         logger.info(
                             "Manual load of %s on worker=%s: lane %s is held by %s; taking the next suffix",
@@ -6552,6 +6661,7 @@ class CapacityPlanner:
                             provider_id,
                             rejection,
                         )
+                        self.record_manual_load_outcome(provider_id, model_name, "failed", reason=rejection)
                         return False
 
                     capacity = self._safe_get_capacity(provider_id)
@@ -6560,6 +6670,13 @@ class CapacityPlanner:
                             "Refusing manual load of %s on worker=%s: capacity snapshot went away before dispatch",
                             model_name,
                             provider_id,
+                        )
+                        self.record_manual_load_outcome(
+                            provider_id,
+                            model_name,
+                            "failed",
+                            lane_id=lane_id,
+                            reason="the worker stopped reporting its capacity before the load could start",
                         )
                         return False
 
@@ -6572,13 +6689,33 @@ class CapacityPlanner:
                         params=self._build_load_params(model_name, lane_id, profile, capacity, provider_id),
                         reason="Manual load requested by an operator",
                     )
-                    return await self._execute_action_with_confirmation(
+                    self.record_manual_load_outcome(provider_id, model_name, "running", lane_id=lane_id)
+                    confirmed = await self._execute_action_with_confirmation(
                         action, timeout_seconds=self.LANE_LOAD_COMMAND_TIMEOUT_S
                     )
+                    if confirmed:
+                        self.record_manual_load_outcome(provider_id, model_name, "succeeded", lane_id=lane_id)
+                    else:
+                        # The executor recorded a human-readable reason for
+                        # every failure mode it knows; fall back to a generic
+                        # one if this one did not (a new branch that forgot it).
+                        reason = self.get_lane_action_failure(provider_id, lane_id)
+                        if reason is None:
+                            reason = "the load was not confirmed by the worker"
+                        self.record_manual_load_outcome(
+                            provider_id, model_name, "failed", lane_id=lane_id, reason=reason
+                        )
+                    return confirmed
             logger.warning(
                 "Manual load of %s on worker=%s: no free lane id after re-reading the runtime",
                 model_name,
                 provider_id,
+            )
+            self.record_manual_load_outcome(
+                provider_id,
+                model_name,
+                "failed",
+                reason="no free lane id was available on the worker",
             )
             return False
         finally:
@@ -8154,6 +8291,9 @@ class CapacityPlanner:
                         self._facade.get_provider_name(action.provider_id) or action.provider_id,
                         action.lane_id,
                     )
+                    self.record_lane_action_failure(
+                        action.provider_id, action.lane_id, "another load is already using this lane"
+                    )
                     return False
                 if self._lane_exists_in_runtime(action.provider_id, action.lane_id):
                     logger.warning(
@@ -8178,14 +8318,42 @@ class CapacityPlanner:
                         per_gpu_free=_per_gpu_free,
                     )
                     if _reservation_id is None:
+                        # The aggregate check above passed but the per-GPU one
+                        # did not (or no per-GPU data exists) — say which GPUs
+                        # fell short, or this line reads as a bug: need <
+                        # avail, yet the reservation was denied.
+                        gpu_free_parts: list[str] = []
+                        if _per_gpu_free:
+                            gpu_free_parts = [
+                                f"GPU {dev}: {free:.0f}MB"
+                                for dev, free in sorted(_per_gpu_free.items())
+                            ]
+                        tp_size = len(self._parse_gpu_device_ids(_lane_gpus)) if _lane_gpus else 0
+                        per_gpu_need = (
+                            _estimated_load_vram / tp_size * self.VRAM_SAFETY_MARGIN if tp_size else _estimated_load_vram * self.VRAM_SAFETY_MARGIN
+                        )
+                        reason = (
+                            f"not enough free VRAM for this model: it needs "
+                            f"~{_estimated_load_vram / 1024.0:.1f} GB in total"
+                        )
+                        if tp_size:
+                            reason += f" (~{per_gpu_need / 1024.0:.1f} GB on each of {tp_size} GPUs)"
+                        reason += f", but the worker reports only {raw_avail / 1024.0:.1f} GB free"
+                        if gpu_free_parts:
+                            reason += " (" + ", ".join(gpu_free_parts) + ")"
+                        reason += "."
+                        self.record_lane_action_failure(action.provider_id, action.lane_id, reason)
                         logger.warning(
                             "VRAM reservation denied for load of %s: "
-                            "need=%.0fMB avail=%.0fMB committed=%.0fMB gpus=%s",
+                            "need=%.0fMB avail=%.0fMB committed=%.0fMB gpus=%s%s",
                             action.model_name,
                             _estimated_load_vram,
                             raw_avail,
                             self.get_pending_vram_mb(action.provider_id),
                             _lane_gpus or "unknown",
+                            f" need-per-GPU={per_gpu_need:.0f}MB per-GPU-free=[{', '.join(gpu_free_parts) or 'unknown'}]"
+                            if _per_gpu_free
+                            else "",
                         )
                         return False
                 elif _reservation_id is None and _estimated_load_vram > 0:
@@ -8228,6 +8396,9 @@ class CapacityPlanner:
                             action.lane_id,
                             details=str(exc),
                         )
+                        self.record_lane_action_failure(
+                            action.provider_id, action.lane_id, f"the worker rejected the load: {exc}"
+                        )
                         return False
                     except Exception as exc:
                         logger.error(
@@ -8236,6 +8407,9 @@ class CapacityPlanner:
                             action.model_name,
                             action.lane_id,
                             exc,
+                        )
+                        self.record_lane_action_failure(
+                            action.provider_id, action.lane_id, f"the load command could not be sent: {exc}"
                         )
                         return False
                 else:
@@ -8266,6 +8440,11 @@ class CapacityPlanner:
                                 self._facade.get_provider_name(action.provider_id) or action.provider_id,
                             )
                             self._clear_inflight_add(action.provider_id, action.lane_id)
+                            self.record_lane_action_failure(
+                                action.provider_id,
+                                action.lane_id,
+                                "the worker accepted the load but rolled the lane change back",
+                            )
                             return False
                         self._registry.update_desired_lanes(action.provider_id, desired)
                         # Inflight entry now committed to registry — clear it
@@ -8279,6 +8458,9 @@ class CapacityPlanner:
                             exc,
                         )
                         self._clear_inflight_add(action.provider_id, action.lane_id)
+                        self.record_lane_action_failure(
+                            action.provider_id, action.lane_id, f"the load command could not be sent: {exc}"
+                        )
                         return False
 
             elif action.action == "stop":
@@ -8421,6 +8603,11 @@ class CapacityPlanner:
                     action.provider_id,
                     action.lane_id,
                     details="confirmation timeout",
+                )
+                self.record_lane_action_failure(
+                    action.provider_id,
+                    action.lane_id,
+                    f"the worker accepted the load but the lane did not reach the loaded state within {int(timeout_seconds)} s",
                 )
             # Sleep confirmation timeout: the command was sent, so the lane
             # is likely sleeping even though we couldn't verify.  Clear the
--- a/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
+++ b/logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
@@ -8338,19 +8338,20 @@ class CapacityPlanner:
                         # did not (or no per-GPU data exists) — say which GPUs
                         # fell short, or this line reads as a bug: need <
                         # avail, yet the reservation was denied.
-                        # Report the gate's effective numbers (raw free minus
-                        # what in-flight operations already committed), not
-                        # the raw snapshot — raw free can look sufficient
-                        # while the reservation is correctly denied.
+                        # Report the gate's effective numbers, not the raw
+                        # snapshot: at provider level the gate compares
+                        # needed against raw free minus what in-flight
+                        # operations already committed (raw free can look
+                        # sufficient while the reservation is correctly
+                        # denied). Per-GPU, _get_per_gpu_free already returns
+                        # the ledger-adjusted free the gate reads — use it
+                        # directly, subtracting commitments again would
+                        # under-report.
                         _vram = self._vram_ledger
                         effective_avail = _vram.get_effective_available_mb(action.provider_id, raw_avail)
                         gpu_free_parts: list[str] = []
                         if _per_gpu_free:
-                            gpu_free_parts = [
-                                f"GPU {dev}: "
-                                f"{_vram.get_gpu_effective_available_mb(action.provider_id, dev, free):.0f}MB"
-                                for dev, free in sorted(_per_gpu_free.items())
-                            ]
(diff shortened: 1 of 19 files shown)


logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -168,6 +168,11 @@ class CapacityPlanner:
     # workers and spurious load-failure cooldowns.
     LANE_LOAD_COMMAND_TIMEOUT_S = 1800
 
+    # How long the outcome of a manual load stays queryable by the UI. Must
+    # outlive the whole attempt: the worker-command budget above plus the
+    # confirmation polling that follows it.
+    MANUAL_LOAD_OUTCOME_TTL_SECONDS = 3600.0
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 23caef87f: `MANUAL_LOAD_OUTCOME_TTL_SECONDS` is now 7200 s — twice the combined command + confirmation budget (1800 s + 1800 s) — with the comment spelling out what it must outlive: the `_lane_lock` wait, `send_command`, `_poll_confirmation`, and polling/processing overhead. The TTL tests key off the constant, so they track the new value.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -1747,6 +1760,84 @@ def _lane_is_in_load_failure_cooldown(
...
+    # The executor (_execute_action_with_confirmation) fails for reasons the
+    # manual-load path cannot know on its own — a denied VRAM reservation, a
+    # worker rejection, a confirmation timeout. It records a per-lane
+    # human-readable reason here; load_lane_manually picks it up for the
+    # outcome it records.
+    def record_lane_action_failure(self, provider_id: int, lane_id: str, reason: str) -> None:
+        self.__dict__.setdefault("_lane_action_failure", {})[self._lane_key(provider_id, lane_id)] = reason
+
+    def get_lane_action_failure(self, provider_id: int, lane_id: str) -> str | None:
+        return self.__dict__.get("_lane_action_failure", {}).get(self._lane_key(provider_id, lane_id))
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 23caef87f: new `clear_lane_action_failure(provider_id, lane_id)`, called in `load_lane_manually` right where the attempt records "running" (before the executor runs). Every executor failure path records a fresh reason, so clearing at attempt start cannot hide information; an attempt that fails without a new reason now gets the generic fallback instead of the previous attempt's. Added `test_clear_lane_action_failure_drops_only_that_lane` and `test_a_failed_attempt_cannot_inherit_the_previous_reason`.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -8178,14 +8318,42 @@ async def _execute_action_with_confirmation(
...
+                            f"not enough free VRAM for this model: it needs "
+                            f"~{_estimated_load_vram / 1024.0:.1f} GB in total"
+                        )
+                        if tp_size:
+                            reason += f" (~{per_gpu_need / 1024.0:.1f} GB on each of {tp_size} GPUs)"
+                        reason += f", but the worker reports only {raw_avail / 1024.0:.1f} GB free"
+                        if gpu_free_parts:
+                            reason += " (" + ", ".join(gpu_free_parts) + ")"
+                        reason += "."
+                        self.record_lane_action_failure(action.provider_id, action.lane_id, reason)
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 23caef87f: the denial reason and the log line now report the gate's effective numbers instead of the raw snapshot — `get_effective_available_mb` (raw − committed) at provider level and `get_gpu_effective_available_mb` per GPU, the same public accessors the rest of the planner uses; the log line mirrors the ledger's own denial log (`need / effective_avail / (raw, committed)`). The per-GPU requirement was already the gate's formula — `vram_mb / len(gpus) * VRAM_SAFETY_MARGIN`, with the planner passing its fixed margin (1.0) into `try_reserve_atomic` — so that part is unchanged.

logos/logos-orchestrator/tests/unit/capacity/test_manual_load_outcome.py
```diff
@@ -0,0 +1,268 @@
+"""The recorded outcome of a manual load — what the statistics UI can ask for.
```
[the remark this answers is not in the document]
wasnertobias: Already addressed before this review round: the formatting was committed in f2bafe7e5 and CI's pre-commit has been green since. Re-ran `pre-commit run --config logos/.pre-commit-config.yaml --all-files` locally on the current head — all hooks pass.

logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.ts
```diff
@@ -497,6 +523,102 @@ export class LaneHealthPanel implements OnChanges {
...
+    }
+    if (Date.now() > this.loadStatusPollDeadline) {
+      // The orchestrator has long since resolved this load; keep the note,
+      // stop re-asking for an outcome that no longer exists.
+      this.stopLoadStatusPoll();
+      return;
+    }
+    this.statisticsService
+      .getLaneLoadStatus(providerId, model)
+      .then((outcome) => this.applyLoadStatus(outcome, model))
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 23caef87f: each `startLoadStatusPoll` bumps `loadStatusPollGeneration`, and every request captures its generation at dispatch time; a response — or a 404/501 stop — from a superseded session is dropped in both `then` and `catch`, without touching the active session's timer. Regression test `drops a delayed answer from a superseded session of the same model` covers the race: first session's request in flight, a retry starts session 2, the delayed "failed" lands and is discarded, and session 2 keeps polling.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/controller/ProviderController.java
```diff
@@ -150,6 +151,20 @@ public ResponseEntity<?> addLane(@RequestBody AddLaneRequestDTO req) {
...
+    @PreAuthorize("hasAuthority('" + Role.Names.LOGOS_ADMIN + "')")
+    public ResponseEntity<?> laneLoadStatus(@RequestBody LaneLoadStatusRequestDTO req) {
+        if (req.providerId() == null || req.model() == null || req.model().isBlank())
+            return ResponseEntity.badRequest().body(Map.of("error", "provider_id and model are required"));
+        try {
+            return workerAdminClient.getLaneLoadStatus(req.providerId(), req.model());
+        } catch (RestClientResponseException e) {
+            return ResponseEntity.status(e.getStatusCode()).body(parseOrWrap(e.getResponseBodyAsString()));
+        } catch (Exception e) {
+            return ResponseEntity.status(503).body(Map.of("error", e.getMessage()));
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 23caef87f — the 503 path now uses the existing null-safe `errorBody(e)` helper, like sleep/wake already do.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/dto/LaneLoadStatusRequestDTO.java
```diff
@@ -0,0 +1,6 @@
+package de.tum.cit.aet.logos.logoswebservice.configuration.dto;
+
+public record LaneLoadStatusRequestDTO(
+        Integer providerId,
+        String model
+) {}
```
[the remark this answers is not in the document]
wasnertobias: Hardened in 23caef87f: `LaneLoadStatusRequestDTO.providerId` now carries an explicit `@JsonProperty("provider_id")`, which binds correctly whether or not the runtime mapper applies the snake_case naming strategy — safe in both worlds. One data point against the literal 400: the other lane DTOs (`AddLaneRequestDTO`, `DeleteLaneRequestDTO`, `SleepLaneRequestDTO`, `WakeLaneRequestDTO`) use the same annotation-free camelCase pattern, and their snake_case endpoints work in production (load/unload/sleep/wake/calibrate all exercised on the live worker), so the deployed runtime does bind `provider_id` today. Pinning the name anyway removes the dependency on which mapper wins.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/orchestrator/OrchestratorWorkerAdminClient.java
```diff
@@ -56,6 +56,16 @@ public ResponseEntity<Map> addLane(int providerId, Map<String, Object> lane) {
...
     }
 
+    /**
+     * Asks the orchestrator for the outcome of the most recent manual load of a
+     * model. The load itself runs in the background after addLane's 202, and a
+     * refusal there (e.g. not enough VRAM) is otherwise only a log line — the
+     * statistics UI polls this while its "Loading" note is up.
+     */
+    public ResponseEntity<Map> getLaneLoadStatus(int providerId, String model) {
+        return post("/internal/logosnode/lanes/load_status", Map.of("provider_id", providerId, "model", model));
```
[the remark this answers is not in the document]
wasnertobias: Leaving this one out of scope: the Spring → orchestrator hop is plain internal HTTP by design — all eight worker-admin calls in `OrchestratorWorkerAdminClient` share the same `post()` path against `logos.orchestrator.url`, and the bearer credential only ever travels on the docker network. Moving that hop to TLS/mTLS is a webservice-wide integration change (CodeRabbit itself rates it "Heavy lift") that predates this PR — worth a separate issue if the team wants it.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -8178,14 +8334,57 @@ async def _execute_action_with_confirmation(
...
+                        # the raw snapshot — raw free can look sufficient
+                        # while the reservation is correctly denied.
+                        _vram = self._vram_ledger
+                        effective_avail = _vram.get_effective_available_mb(action.provider_id, raw_avail)
+                        gpu_free_parts: list[str] = []
+                        if _per_gpu_free:
+                            gpu_free_parts = [
+                                f"GPU {dev}: "
+                                f"{_vram.get_gpu_effective_available_mb(action.provider_id, dev, free):.0f}MB"
+                                for dev, free in sorted(_per_gpu_free.items())
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 461f37ffd: the per-GPU part of the diagnostic now uses the `_get_per_gpu_free()` values directly — they are already ledger-adjusted, and they are exactly what the reservation gate reads, so passing them through `get_gpu_effective_available_mb` again subtracted committed twice and under-reported. The provider level keeps the single subtraction: `available_vram_mb` comes from the worker snapshot's physical free memory (`logosnode_provider.get_capacity_info`), so raw − committed there matches the gate's provider check. One related observation, out of scope here: the gate itself reads the already-adjusted per-GPU input and subtracts commitments once more (`try_reserve_atomic`: `gpu_effective = gpu_avail − gpu_committed`), so with in-flight reservations on the same GPU the per-GPU check is more conservative than the ledger bookkeeping alone would suggest. That is pre-existing behavior shared by every reservation call site that passes `_get_per_gpu_free()` output — happy to look at it separately if the team wants.

logos/logos-ui/src/app/features/statistics/components/worker-gpu-panel/worker-gpu-panel.ts
```diff
@@ -45,6 +54,18 @@ export class WorkerGpuPanel {
   private statisticsService = inject(StatisticsService);
 
   calibrateState = signal<CalibrateState>({ kind: 'idle' });
+  /** Worker the in-flight calibrate call was started on — its answer must not
+   *  land under a worker the operator has moved on to. */
+  private calibrateProvider: string | null = null;
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (!changes['activeProvider']) return;
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 461f37ffd: `ngOnChanges` now compares the *resolved* worker (the explicit selection, or the online-first fallback when it is null) against the previously resolved one, so the feedback resets when the fallback flips — worker leaves the sample list or goes offline — even though `activeProvider` never changed. The in-flight guard additionally rejects the answer when `resolvedActiveProvider` no longer matches the worker the call went to. Added the regression test: null selection, calibrate started on the fallback worker, that worker leaves the list mid-request, fallback flips to the next worker — the state resets on the flip and the delayed answer is dropped.

logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.ts
```diff
@@ -461,6 +502,7 @@ export class LaneHealthPanel implements OnChanges {
     this.selectedModel.set(null);
     this.addError.set(null);
     this.pickerProviderId = null;
+    this.stopLoadStatusPoll();
```
Claudia-Anthropica: @wasnertobias [medium] `closePicker()` is also invoked by the picker’s normal Close button while an accepted load may still be pending. Opening and closing the picker during that wait clears the only status interval but leaves `acceptedModel` and `acceptedLaneIds` intact, so a later background refusal once again leaves “Loading …” indefinitely. Keep the active outcome poll running on an ordinary picker close; stop it only when the pending attempt resolves, is superseded, the provider changes, or the component is destroyed.

In `logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.ts`, `closePicker()` cancels an active load-status poll without clearing the pending load state, allowing a background failure to remain undisplayed. Preserve the poll on ordinary picker closes and stop it explicitly only when the pending attempt is resolved, superseded, abandoned by a provider change, or destroyed.

</details>

logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.ts
```diff
@@ -227,6 +228,37 @@ export class LaneHealthPanel implements OnChanges {
...
+  private static readonly LOAD_STATUS_POLL_INTERVAL_MS = 2500;
+  /**
+   * Hard stop for the outcome poll. The orchestrator resolves every manual
+   * load within its load command timeout (30 min) — refusal, confirmed lane,
+   * or a recorded timeout — so well past that the live outcome is gone
+   * (planner restart, entry aged out) and the poll can only re-ask for the
+   * same "unknown". The lane-appearance check in ngOnChanges keeps owning the
+   * note from there on.
+   */
+  private static readonly LOAD_STATUS_POLL_CAP_MS = 31 * 60 * 1000;
```
Claudia-Anthropica: @wasnertobias [medium] The 31-minute cap is shorter than the backend’s complete failure window. A manual load gives `send_command` up to 1,800 seconds and then calls `_poll_confirmation` with another 1,800-second timeout, so the terminal failure can be recorded nearly 60 minutes after dispatch, plus any lane-lock wait; this poll can stop before that outcome exists and leave the pending note indefinitely. Extend the cap to cover the full two-phase attempt with margin, or continue polling until a terminal outcome/provider change.

In `logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.ts`, `LOAD_STATUS_POLL_CAP_MS` expires before the orchestrator’s possible 30-minute command plus 30-minute confirmation sequence completes. Make the polling lifetime cover the complete backend attempt with margin, or keep polling until a terminal outcome or provider change, and update the cap regression test accordingly.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The status route and outcome propagation are well covered, and all captured checks succeeded. Two UI lifecycle gaps can still recreate the indefinite pending state this change is intended to eliminate.

logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.ts
wasnertobias: Fixed in fd00665f4: `closePicker()` no longer stops the outcome poll — a plain Close (the picker button) is not the end of an accepted load, the pending note stays up, and the poll is what surfaces a background refusal minutes later. The provider switch now stops the poll explicitly (it previously relied on the closePicker side effect); the other stop paths are unchanged (terminal answer, supersession by a new attempt, 404/501, cap, destroy). Added a regression test: accepted load, picker closed via `closePicker()`, poll still running, a late "failed" still lands as the error and re-offers the model.

logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.ts
wasnertobias: Fixed in fd00665f4: the cap is now 65 min, covering the full two-phase attempt — the 30 min `send_command` budget plus the 30 min `_poll_confirmation` that follows it, plus lock wait and overhead — with the comment documenting the reasoning. The cap test is keyed to the constant, so it tracks the new value; its rationale comment now says "full two-phase attempt" instead of "load command timeout".

wasnertobias:
@Claudia-Anthropica review again

Claudia-Anthropica:
@wasnertobias got it — starting the review right away.

wasnertobias:
@Claudia-Anthropica where is your review?

logos/logos-ui/src/app/features/statistics/components/worker-gpu-panel/worker-gpu-panel.ts
```diff
@@ -163,11 +192,18 @@ export class WorkerGpuPanel {
...
+    this.calibrateProvider = active;
     this.calibrateState.set({ kind: 'loading' });
 
     try {
       const body = await this.statisticsService.calibrateUncalibrated(pid);
+      // The operator can switch workers while the call is in flight — or the
+      // fallback worker can change under a null selection — and the answer
+      // belongs to the worker it was asked for, so a stale one is dropped
+      // rather than shown under the panel the operator is looking at now.
+      if (this.calibrateProvider !== active || this.resolvedActiveProvider !== active) return;
```
Claudia-Anthropica: @wasnertobias [medium] `calibrateProvider` distinguishes workers but not separate attempts on the same worker. If a calibration on A remains pending while the operator switches A → B → A and starts another calibration, the first response passes this guard because both calls captured A and the shared field contains A; it then replaces the newer request's loading state and re-enables submission. Track a unique token for each calibration request and apply a completion only while that exact token and provider remain current; add an A → B → A regression test.

In `logos/logos-ui/src/app/features/statistics/components/worker-gpu-panel/worker-gpu-panel.ts`, the stale-response guard uses only provider identity, allowing an earlier calibration to overwrite a newer request after switching away and back to the same provider. Track a unique token per calibration attempt, accept completion only when both the token and resolved provider still match, and add regression coverage for the A → B → A sequence.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The two previous pending-load blockers are resolved on this head, and all captured build, test, and lint checks succeeded. One calibration lifecycle race remains: returning to the same worker allows an older in-flight response to overwrite a newer calibration attempt, so changes are still required.

logos/logos-ui/src/app/features/statistics/components/worker-gpu-panel/worker-gpu-panel.ts
wasnertobias: Fixed in 3f12fa0a9: a `calibrateGeneration` token advances on every calibration start *and* on every worker-change reset, and a completion only applies while its exact token, the captured provider, and the current resolved worker all still match — so the first answer of an A → B → A sequence is dropped instead of replacing the newer request's loading state. Added the regression test for exactly that sequence (the shared mock now keeps a settler per call so both answers can be delivered in order).

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -6496,6 +6608,10 @@ async def load_lane_manually(self, provider_id: int, model_name: str) -> bool:
         # so the pick below would hand out the next free suffix and place a
         # second copy of the very model the operator is already waiting for.
         if any(model.lower() == model_name.lower() for model in self._inflight_load_models(provider_id).values()):
+            # Deliberately no outcome record: the in-flight load (this click's
```
Claudia-Anthropica: @wasnertobias [medium] This assumes every in-flight same-model load will publish the manual outcome, but planner-initiated loads call `_execute_action_with_confirmation` directly and never call `record_manual_load_outcome`. If such a load is already running when `/lanes/add` records `running`, this task exits here; should the planner load fail, nothing replaces that entry, so the UI eventually stops polling while retaining its pending note. Make same-model admission atomic before recording `running` and return a synchronous conflict when another owner already holds the load.

In `logos/logos-orchestrator/src/logos/capacity/capacity_planner.py`, an accepted manual load can exit as a no-op behind a planner-initiated same-model load that never publishes a terminal manual outcome. Atomically reject an existing same-model in-flight owner before the endpoint records `running` and returns 202, and add regression coverage for a failing planner-owned load.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias All captured build, test, and lint checks succeeded, and the previous calibration blocker is fixed. The existing lane-action completion and VRAM day-switch races remain reproducible, and an uncovered planner/manual-load overlap can also leave the new status permanently `running`.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java
```diff
@@ -327,9 +327,14 @@ private void tick() {
...
-                    pushVramDelta(session, state);
-                }
+                // VRAM deltas ride every tick: a lane the worker just loaded
+                // reaches the UI within one second of its status report
+                // instead of waiting up to five for the next vram cadence.
+                // The fetch is cursor-scoped (new snapshots only) and the
+                // provider-status hop is cached for 3 s, so the per-tick cost
+                // stays small; the push itself is still skipped when nothing
+                // moved (no new samples, cursor, or connection state).
+                pushVramDelta(session, state);
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 488e1ffb6.

A `vramDayGeneration` token on the session state is bumped in `handleInit` and `handleSetVramDay`. `pushVramDelta` captures the generation before the `getVramStats` call and discards the result when it returns against a superseded window, so a day change landing mid-query can no longer send previous-day samples or overwrite the fresh cursor and connection-state baseline.

Regression test added (`a_stale_vram_delta_is_dropped_after_a_day_change`): the day switch is made to complete inside the in-flight delta query, and the test asserts the stale delta is neither pushed nor written back to the new day's state.

Claudia-Anthropica (changes requested):
@wasnertobias The captured CI has no failed or pending required check, and the calibration-attempt fix is present. Three medium races remain: old lane-action completions can clear a newer provider’s action state; manual admission can record `running` before a planner-owned same-model load makes the task exit without a terminal outcome; and `pushVramDelta` can read the old day before reading the new generation, allowing stale data to pass its check. Guard actions by attempt token, atomically claim manual-load admission before returning 202, and capture the VRAM day/generation/cursor coherently.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java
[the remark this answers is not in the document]
Claudia-Anthropica: @wasnertobias The generation check still has a capture race: `pushVramDelta` reads `vramDay` before `vramDayGeneration`, so `set_vram_day` between those reads pairs the old day with the new generation and the stale payload passes. Capture a coherent day/generation/cursor snapshot or serialize the transition, and cover this interleaving.

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
Claudia-Anthropica: @wasnertobias This remains on the current head: `internal_logosnode_add_lane` records `running` and schedules the task without atomically checking or claiming same-model ownership; `load_lane_manually` can then return at `_inflight_load_models`, while the planner owner never publishes a terminal manual outcome. Make admission atomic before returning 202 and add the planner-owned failure regression.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java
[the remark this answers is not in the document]
wasnertobias: Fixed in 621fe2cca.

The capture order is now generation -> day -> cursor, and the window changes (init, set_vram_day) bump the generation before they write the new day and cursor — the bump is the commit point of the change. So a stale day can never be paired with a fresh generation: a delta that captures before the change holds the old generation and is discarded when the bump lands, and one that reads the new generation already reads the new day. The guard is now the pure predicate `isCurrentVramWindow(generation, day, state)` (generation match plus, for explicitly set days, a day match), and the new test pins exactly the interleaving you described: a snapshot with the old day and the fresh generation is not a current window, while a coherent snapshot of the state's window still passes. (The interleaving itself cannot be driven end-to-end in a unit test — the switch would have to land between two consecutive volatile reads — so the predicate is the coverage; the existing end-to-end test keeps covering a switch that completes inside the in-flight query.)

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
wasnertobias: Fixed in 621fe2cca.

Admission is now atomic before the 202: `manual_load_admission_rejection` is a check-and-claim on the event loop (a per-(provider, model) marker plus the same-model in-flight check), called in the endpoint before it answers — a second click, or a load the planner is already bringing up, gets a 409 the operator reads instead of a 202 whose background task no-ops. The refused click no longer records "running" either, so it cannot strand a terminal outcome.

For the planner-owned case that slips past admission (the planner starts the model between admission and the task's check), the no-op branch now rides the in-flight load's lane: it records "running" against that lane, and the executor wrapper (`_execute_action_with_confirmation`) settles the riding "running" outcome for every load it runs — planner's included — with the per-lane failure reason on failure. Settle only touches the lane the entry names, so a same-model load on another lane cannot pre-empt the operator's own attempt, and the endpoint's lane-less placeholder is never settled by a foreign load.

Regression added: `test_a_planner_owned_inflight_failure_reaches_the_manual_outcome` (admit -> planner claims the model -> no-op rides -> planner failure settles the outcome as failed with the reason), plus the executor settle wiring (success/failure, foreign lane, lane-less placeholder) and the endpoint's 409 admission test.

Claudia-Anthropica (changes requested):
@wasnertobias The planner/manual-load overlap fix is present and the captured CI has no failed or pending required check, but two current-head races still block approval. In `LaneHealthPanel`, every unload/sleep/wake completion clears the shared action signal before validating its attempt, so an older request on provider A can clear a newer provider B request after an A → B switch and permit duplicate dispatch. In `StatsV2WebSocketHandler`, `handleSetVramDay` publishes the new generation before the new day, allowing the tick thread to capture a new-generation/old-day window that passes `isCurrentVramWindow` if it completes while the writer is paused; synchronize the window transition and snapshot and cover that in-progress interleaving.

wasnertobias (commented):
Both races are fixed in 8039c658e:

**StatsV2WebSocketHandler (window transition vs. snapshot).** The day/cursor/generation trio is now one immutable `VramWindow` record behind an `AtomicReference`. `handleInit`/`handleSetVramDay` transition with a single `.set()`, `pushVramDelta` snapshots with a single `.get()`, and the write-back goes through `compareAndSet(captured, …)` — a delta spanning a window change loses the CAS whatever it fetched. The reference identity *is* the generation, so the fresh-generation/stale-day pairing is no longer expressible even if the writer pauses mid-transition; `isCurrentVramWindow` reduces to a reference-equality check. The stale-delta interleaving test is kept (now asserting on the window reference), and the unit test on the predicate pins that a swapped-out snapshot is stale whatever its in-flight query fetched, and that a superseded snapshot of the moved-in window still does not pass.

**LaneHealthPanel (action-signal ownership).** Unload/sleep/wake completions no longer clear the shared in-flight signal before validating the attempt. A new `actionStillOwned` helper releases the signal and posts the error only when the provider is unchanged *and* the signal still names the attempt's lane — both halves matter, because lane ids are per-worker and can collide across workers. A regression test pins the A → B interleaving: a late completion of provider A's call leaves provider B's still-running signal and its in-flight guard untouched, and the signal still settles normally for the attempt that owns it.

Verification: webservice 429/429 green (incl. the updated live-push tests), UI 184/184 green (incl. the new ownership test), pre-commit clean.

logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.ts
```diff
@@ -345,22 +380,47 @@ export class LaneHealthPanel implements OnChanges {
...
+   * the provider is unchanged AND the signal still names the attempt's lane.
+   * A finishing attempt that failed both must not clear the signal — after an
+   * A → B switch, B's own in-flight attempt holds it, and clearing it here
+   * would release the in-flight guard while B's call is still running, so the
+   * next click dispatches a duplicate action to the same lane. The provider
+   * half of the check matters even when the lane ids collide, because a lane
+   * id is per-worker and the same id can be in flight on two workers at once.
+   */
+  private actionStillOwned(pid: number | null, signal: () => string | null, laneId: string): boolean {
+    return this.providerId === pid && signal() === laneId;
```
Claudia-Anthropica: @wasnertobias [medium] This still identifies an action only by provider and lane. If an unload on A remains pending while the operator switches A → B → A and starts another unload of the same lane, the old A completion passes both checks and clears or errors the newer request, reopening its duplicate-dispatch guard; sleep and wake have the same race. Track a unique token for every action attempt and require that token in the completion guard, with A → B → A regression coverage.

In `logos/logos-ui/src/app/features/statistics/components/lane-health-panel/lane-health-panel.ts`, provider and lane identity cannot distinguish separate unload/sleep/wake attempts after an A → B → A switch. Add a unique token per action attempt, require it when settling each completion, and add an A → B → A regression test.

</details>

logos/logos-orchestrator/src/logos/capacity/capacity_planner.py
```diff
@@ -6533,6 +6737,14 @@ async def load_lane_manually(self, provider_id: int, model_name: str) -> bool:
                                 provider_id,
                                 lane_id,
                             )
+                            # The model is loaded — the operator's goal is met.
+                            # Report it as success so the UI note resolves; but
+                            # not over a "running" entry: a concurrent attempt
+                            # for this same lane may still be executing and
+                            # owns the final state.
+                            current = self.get_manual_load_outcome(provider_id, model_name)
+                            if not (current and current.get("status") == "running"):
```
Claudia-Anthropica: @wasnertobias [medium] The endpoint always records a lane-less `running` outcome before this task starts. When the facade lane list lags the runtime and this branch discovers that the selected lane already contains the requested model, this condition preserves that placeholder even though no executor remains to settle it; `load_status` therefore stays `running` until expiry and the UI note hangs. Record success for the current lane-less placeholder while preserving only a lane-named outcome owned by another in-flight executor, and cover the endpoint-style sequence in a regression test.

In `logos/logos-orchestrator/src/logos/capacity/capacity_planner.py`, the lane-exists no-op preserves the endpoint's lane-less `running` placeholder, leaving it without an executor that can publish a terminal outcome. Set that placeholder to `succeeded` when the requested model is already present, preserve only lane-owned in-flight outcomes, and add regression coverage.

</details>

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java
```diff
@@ -351,41 +369,65 @@ private void tick() {
...
             Object sid = payload.get("last_snapshot_id");
-            state.vramCursor = sid instanceof Number n ? n.intValue() : 0;
-            state.prevVramMetaSig = vramMetaSig(payload);
-            send(session, Map.of("type", "vram_init", "payload", payload));
+            int cursor = sid instanceof Number n ? n.intValue() : 0;
+            // The payload belongs to the window it was captured for; if a
+            // window change swapped the reference meanwhile, that window's
+            // init is responsible for the push — the stale baseline is
+            // dropped together with the stale payload.
+            if (state.vramWindow.compareAndSet(window, new VramWindow(window.day(), cursor, vramMetaSig(payload)))) {
```
Claudia-Anthropica: @wasnertobias [medium] After `handleSetVramDay` installs the empty new-day window, the tick thread can snapshot that same reference while `pushVramInit` is fetching. If the delta returns first, it wins the CAS and sends `vram_delta`; this initialization then loses its CAS and sends no `vram_init`, so the client appends the new day's rows to the old series instead of replacing it. Keep deltas disabled while a window is initializing (or serialize initialization and deltas), and test the delta-finishes-before-init interleaving.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java`, a new-window delta can win the atomic reference before `pushVramInit`, suppressing the required initialization and mixing day series in the client. Prevent deltas from updating an initializing window until its init is installed, and add a regression test where the delta completes first.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The previously reported fresh-generation/stale-day interleaving is fixed, and the captured build, test, and lint checks succeeded. Three current-head edge cases remain: an A → B → A lane action can settle a newer request, a lane-exists no-op can leave an accepted outcome permanently `running`, and a day-change delta can pre-empt its required initialization. These require changes before approval.

wasnertobias (commented):
All three edge cases are fixed in 20bf56122.

1. **A → B → A lane action settling a newer request** — the settle gate in `LaneHealthPanel` no longer matches on (provider, lane id), which is not a unique attempt identity: lane ids are per-worker and can collide, and an A → B → A switch re-uses both the provider and the lane name for a newer attempt while the older one is still in flight. Each action kind now has a monotonic attempt counter, bumped by every new attempt and by the provider switch (which abandons the in-flight calls); a settling attempt may touch the shared signal and error only while its counter is still the current one. Regression test pins the exact A → B → A interleaving with a colliding lane id: the older attempt settling last leaves the newer attempt's signal and error untouched.

2. **Lane-exists no-op leaving an accepted outcome permanently `running`** — in `_load_lane_manually`, the no-op now settles the endpoint's *lane-less* "running" placeholder (this attempt's own, recorded when the 202 went out): the attempt returns without dispatching, so no executor would ever settle it, and the UI note sat at "Loading" for the 2 h outcome TTL although the lane was already serving. The result is recorded as `succeeded` on the existing lane. A *lane-named* "running" entry still belongs to an in-flight dispatch whose executor settles it, and is left alone — the existing test for that hold-back is unchanged and green. New regression test covers the placeholder case.

3. **Day-change delta pre-empting its required initialization** — `VramWindow` now carries whether the window's init has gone out (`baselineSent`). After a day change, the init and the first tick delta query the same (day, cursor 0) and race for the write-back: whichever lost the `compareAndSet` dropped its push, and a delta winning that race would owe the viewer a full-day `vram_init` that never came. Deltas now skip windows whose baseline has not gone out — the init runs on the websocket thread the moment the window moves, so the next tick already sees the baseline. New test drives the interleaving (delta runs inside the init's in-flight query) and asserts the payload arrives as `vram_init`, never as a shadowing `vram_delta`.

Verification: webservice 430/430, UI 185/185, orchestrator unit 1628 passed + 1 xfailed (pre-existing documented gap), pre-commit clean.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java
```diff
@@ -351,41 +371,73 @@ private void tick() {
...
+            // cursor and connection-state baseline.
+            VramWindow window = state.vramWindow.get();
+            // A window whose init has not gone out yet is not consumable:
+            // init and the first delta after a day change query the same
+            // (day, cursor 0) and race for the write-back — whichever loses
+            // the compareAndSet drops its push, and a delta winning that race
+            // would owe the viewer a full-day "init" that never comes. The
+            // init runs on the websocket thread the moment the window moves,
+            // so the next tick already sees the baseline.
+            if (!window.baselineSent()) return;
```
Claudia-Anthropica: @wasnertobias [medium] When `getVramStats` throws in `pushVramInit`, the catch sends an error but leaves the current window with `baselineSent == false`. Every subsequent one-second tick returns here, so a transient database or orchestrator failure permanently freezes VRAM updates until another client-driven initialization or reconnect. Retry initialization for the unchanged window from the tick path and add a transient-failure-then-success regression test.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java`, a failed `pushVramInit` leaves `baselineSent` false and makes every future delta tick return without recovery. Retry initialization for the same still-current window until it succeeds, preserve the atomic window-change guard, and add a transient init failure/recovery test.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The three previously reported current-head races are fixed, and the captured source, test, and lint checks succeeded. The non-required Agent Workspace build is red only because Harbor's storage quota is exhausted and is unrelated to this diff; however, the new VRAM baseline guard has a transient-failure recovery hole, so changes remain required.

wasnertobias (commented):
Fixed in dcccfc323.

The hole: when the websocket thread's init query failed transiently, the error init went out, but the window kept `baselineSent=false` — and the delta's guard then skipped that window on every subsequent tick, so the session's vram feed sat behind one failed query until a manual day change or a reconnect re-initialised the window.

The tick now retries the owed init instead of skipping: `pushVramDelta` delegates to `pushVramInit` while the captured window has no baseline, and `pushVramInit` bails out when the window's baseline already went out (a concurrent init or an earlier retry established it), so the retry cannot restate the full day. The two inits race the write-back like any two pushes — the compareAndSet loser drops. The init's catch also logs a warn like the delta's did. The preemption test keeps pinning the in-flight interleaving (the delta's answer now runs it once; the retry's own query must not re-trigger it), and a new test drives the failure/recovery sequence: error init out, window still owed, next tick's delta recovers, and the full-day payload arrives as a `vram_init` — never as a `vram_delta`.

Webservice 431/431, pre-commit clean.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java
```diff
@@ -351,41 +371,83 @@ private void tick() {
...
+            int cursor = sid instanceof Number n ? n.intValue() : 0;
+            // The payload belongs to the window it was captured for; if a
+            // window change swapped the reference meanwhile, that window's
+            // init is responsible for the push — the stale baseline is
+            // dropped together with the stale payload.
+            if (state.vramWindow.compareAndSet(window, new VramWindow(window.day(), cursor, vramMetaSig(payload), true))) {
+                send(session, Map.of("type", "vram_init", "payload", payload));
+            }
         } catch (Exception e) {
+            log.warn("[ws/stats/v2] vram_init error: {}", e.getMessage());
```
Claudia-Anthropica, line 392: @wasnertobias [medium] The retry now lets `pushVramInit` for day A run on the tick thread while `handleSetVramDay` initializes day B. If A's `getVramStats` throws after B has successfully installed and sent its baseline, this catch unconditionally sends a `vram_init` error; the UI then shows “Failed to load VRAM data” for the current day until a later delta changes, even though B succeeded. Capture the attempted `VramWindow`, emit the error only if that exact window is still current, and add a regression where the old retry fails after the new-day init succeeds.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java`, a failed retry for a superseded VRAM window unconditionally emits an error after a newer day may already have initialized. Capture the attempted window, suppress the error unless that exact window is still current, and add the stale-retry failure regression.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The previous transient-init recovery blocker is fixed on this head, and the snapshot contains no failed or pending required checks. However, the retry introduces an unguarded stale-error path: an old window's failed retry can publish a `vram_init` error after a newer day has initialized, so changes are still required.

Claudia-Anthropica (changes requested):
@wasnertobias The transient-init recovery path is present, and no required check is failed or pending; the non-required Agent Workspace build failed only because Harbor's storage quota is exhausted, unrelated to this diff. The unresolved current-head blocker remains: `pushVramInit` unconditionally sends its error after a superseded window's retry fails, allowing an old day-A failure to replace a successfully initialized day-B view. This exact issue is already captured in the retained non-outdated thread, so no duplicate inline comment is added.

wasnertobias (commented):
Fixed in 659849061.

The catch in `pushVramInit` published the error init unconditionally: a slow failed query captured on an older window would send its error after a day change had already initialised the newer day, overwriting the viewer's good baseline with a failure for a day they no longer look at. The success path already dropped the stale payload via the compareAndSet; the error path now drops the stale failure the same way — `window` is hoisted out of the try and the error goes out only while `isCurrentVramWindow(state, window)` holds. A still-current window keeps reporting its failure on every retry, as before.

New test drives the interleaving: the day1 retry's query is in flight, `set_vram_day(day2)` runs inside it and establishes day2's baseline, then the stale query fails — the assertions pin that day2's `vram_init` is what the viewer is left with and that no "Failed to load VRAM data" error went out on top of it.

Webservice 432/432, pre-commit clean.

Claudia-Anthropica (changes requested):
@wasnertobias The new current-window check narrows the stale-init path, and the snapshot has no failed or pending required checks. It does not close the retained race: after `isCurrentVramWindow` returns true, the websocket thread can initialize day B before the tick thread calls `send`, letting the stale day-A error still arrive last. The same CAS-then-send gap exists for successful init and delta publication, so window transitions and publication need one ordering mechanism before approval.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java
Claudia-Anthropica: @wasnertobias [medium] This remains a check-then-send race: after `isCurrentVramWindow(state, window)` succeeds, `handleSetVramDay` can install and send day B before this thread reaches `send`, after which day A's stale error is still published last. Serialize the window transition/current check and publication under one per-session ordering mechanism—including successful init and delta sends—and add a regression that switches days after the stale path passes its guard but before it publishes.

wasnertobias (commented):
Fixed in ff37537d1 — window transitions and publication now share one ordering mechanism.

The gap was real: the compareAndSet (and the isCurrentVramWindow check) guard the *state*, but the send is a separate step — between the check/CAS and the send, the websocket thread could run set_vram_day and publish day B's init, so the superseded window's push (its stale error, but also a successful init or a delta) still arrived last and overwrote the good view.

SessionState now carries a `vramLock`, and every vram critical section runs under it:

* the tick's delta — the whole capture → query → CAS → send section;
* every transition — the window swap *and* its init push, together, in both `handleInit` (the swap moved next to its init) and `handleSetVramDay`.

Because both sides serialize on the same monitor, a push for window W always reaches the viewer before the transition to W' and its init can run: the day-B init can no longer land between a stale check/CAS and its send, for the error path, the success path, or deltas alike. The compareAndSet stays in place as the state-level guard, and the stale-error check from the previous round is kept as defense in depth (under the lock it now only matters for the direct-call paths).

Lock order is always `vramLock -> session` (`send` is the only acquirer of the session monitor and never takes the vram lock), so there is no inversion. The delta's retry path calls `pushVramInit` without re-acquiring, so nothing nests.

Webservice 432/432, pre-commit clean.

Claudia-Anthropica (approved):
@wasnertobias The shared per-session VRAM lock now orders each capture/query/CAS/send sequence before any subsequent window transition and initialization, closing the retained stale-publication race. The manual-load outcome flow, UI polling and action-attempt guards are consistent with their regression coverage. The snapshot has no failed or pending required checks; several non-required build jobs were still in progress.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/websocket/StatsV2WebSocketHandler.java
Claudia-Anthropica: @wasnertobias Thanks, this is fixed now: VRAM window transitions and capture-to-send publication share the same per-session lock, so a superseded window cannot publish after the newer baseline.

PR #704, opened 2026-07-23
`Logos`: Add first-class Whisper support

bassner:
## Motivation

Logos' inference pipeline was built around JSON requests and responses. Whisper uses multipart file uploads, can return non-JSON formats, and is billed by audio duration. As a result, supporting it properly requires more than forwarding one additional path: the request must survive authentication, scheduling, cloud/worker transport, async jobs, privacy logging, response relay, and billing without losing OpenAI compatibility.

## Summary

This PR makes OpenAI-compatible Whisper transcription and translation a first-class Logos API surface across direct cloud execution, Azure deployment routing, Logos worker nodes, async jobs, usage logging, and billing.

It follows the OpenAI Audio API contract and Azure OpenAI deployment-scoped audio routes.

```text
OpenAI SDK / curl
  -> authenticate
  -> bounded multipart parser
  -> JSON-safe internal envelope
  -> existing classification + scheduling pipeline
  -> OpenAI / Azure / Logos worker node
  -> JSON, text, SRT, VTT, or SSE response
  -> sanitized logging + duration billing
```

## Changes

- Adds explicit `POST /v1/audio/transcriptions` and `POST /v1/audio/translations` routes before the `/v1/{path:path}` catch-all.
- Accepts OpenAI SDK-compatible `multipart/form-data`, preserving duplicate form fields and file metadata through scheduling and worker-node JSON transport.
- Supports OpenAI and Azure cloud providers plus Logos worker-node backends.
- Preserves JSON, verbose JSON, text, SRT, VTT, and streaming response content types.
- Rewrites Whisper standard JSON/text/SRT/VTT calls to one metered `verbose_json` upstream request and renders the requested OpenAI-compatible representation locally, retaining duration usage without a second provider call.
- Tracks fractional duration as integer `audio_milliseconds` and imports per-second catalogue prices with the corresponding billing scale.
- Adds Azure transcription/translation route alignment and the documented `2025-04-01-preview` audio API version.
- Authenticates before parsing uploads, enforces one file, bounded text fields, a 25 MiB file limit, and a 30 MiB total request limit for declared and chunked bodies.
- Excludes audio contents and credentials from usage logs and durable async-job records.
- Adds the direct `python-multipart` dependency and user-facing documentation.
- Requires no Whisper-specific UI fork: the existing generic provider, deployment, model, API-key, and permission forms already expose every required setting; Logos has no inference playground.

## API Compatibility

| Dimension | Behavior |
|---|---|
[...]

PR #942, opened 2026-09-10
`Logos`: Install vLLM with the audio extra so transcription can decode uploads

### LINKED ISSUE #925: `Logos`: Audio transcription always fails — worker image installs vLLM without the audio extra (no soundfile/av)
## Symptom

Every request to `/v1/audio/transcriptions` (and `/v1/audio/translations`) against a local Whisper deployment fails with the identical 400, regardless of the uploaded file:

```json
{"error": {"message": "Invalid or unsupported audio file."}}
```

Verified with a valid 16 kHz and a valid 22.05 kHz WAV, MP3, FLAC — and even a plain text file renamed to `.wav`: all return the same error. A working decoder would distinguish those, so the failure is not format validation — it happens before any decoding is attempted.

The string "Invalid or unsupported audio file." appears nowhere in the Logos source (Logos' own validation messages are distinct, e.g. omitting `model` yields "Audio transcription requests require a 'model' form field"), so the request passes Logos' checks and the error is produced by vLLM on the worker.

## Root cause

The worker image installs **plain vLLM without the `audio` extra**:

- `logos/logos-workernode/Dockerfile:68` — `ARG VLLM_PIP_SPEC="vllm==v0.28.0"`
- `logos/logos-workernode/Dockerfile:194` — `uv pip install --no-cache "$VLLM_PIP_SPEC"` (no `[audio]`)
- `logos/logos-workernode/requirements.txt` — 7 packages, none audio-related; no `ffmpeg` apt package either

vLLM 0.28.0's audio decoding path (`vllm/multimodal/media/audio.py`, `load_audio`) has exactly two backends, both optional:

1. `soundfile` (primary) — missing in the image → `ImportError` → fall back
2. `av` / PyAV (fallback) — missing in the image → `ImportError: Please install vllm[audio] for audio support`

`vllm/entrypoints/speech_to_text/base/serving.py` (`_decode_and_chunk_speech`) then wraps **any** non-`ValueError` exception into `ValueError("Invalid or unsupported audio file.")` → the 400 above. With both backends absent, every upload fails at the same point — exactly matching the observed behaviour.

Whisper support landed in #704 (commit `fb23b983`) with a complete orchestrator side (multipart round-trip through base64 preserves filename, bytes and content type), but the commit never touched the worker Dockerfile/requirements.

### Correction of a common misdiagnosis

**librosa is NOT part of vLLM 0.28.0's audio stack.** The official `vllm[audio]` extra is `av` + `soundfile` + `mistral_common[audio]` (the latter = soundfile + soxr + opencv for Voxtral models). vLLM does not import librosa anywhere in the transcription path (only a comment references it). So the check

```
python -c "import librosa, soundfile"
```

fails on `librosa` even after a correct fix. The correct check is:

```
python -c "import soundfile, av"
```

## Verified live (prod worker, deipapa, container `logos-workernode`, vLLM 0.28.0)

Running the exact function the endpoint calls (`load_audio` on a 4 s, 22.05 kHz mono WAV with real speech):

```
# as shipped (no audio deps):
ERROR ... [audio.py:237] Failed to load audio via soundfile: ImportError('Please install vllm[audio] for audio support')
DECODE FAILED: ImportError: Please install vllm[audio] for audio support
cause: ModuleNotFoundError: No module named 'av'

# after `pip install soundfile av` in the same container (uninstalled again afterwards, container left unchanged):
DECODE OK: shape=(66256,) sr=16000 samples=66256 (~4.1s)
```

So the missing dependencies are both **necessary** (no other decode path exists for this endpoint) and **sufficient** (same file decodes and resamples correctly once installed). Note that `soundfile` alone would not be enough: any input not already at the model's sample rate (16 kHz for Whisper — i.e. most real-world files) is resampled via `resample_audio_pyav`, which requires `av`.

There is no Logos-side workaround (the OpenAI-compatible endpoint accepts arbitrary audio formats); the fix has to land in the worker image.

## Proposed fix

Preferred (canonical, tracks vLLM's own definition of the audio dependency):

```dockerfile
ARG VLLM_PIP_SPEC="vllm==v0.28.0[audio]"
```

Alternative (slimmer, skips opencv/soxr that only Voxtral needs): add `soundfile` and `av` to `logos/logos-workernode/requirements.txt`.

Either way, the worker image must be rebuilt and rolled out; until then, local Whisper transcriptions are unusable.

## Side note (separate from this bug)

At the time of verification (2026-09-08), `openai/whisper-large-v3` (registered, bound to providers deipapa/deimama) was **not loaded on any worker** — the orchestrator returned `No deployment found for model 'openai/whisper-large-v3'` (404) before a request could reach a lane. So the bug above is only observable while a Whisper lane is actually loaded.

wasnertobias:
## Motivation

Closes #925. Every request to `/v1/audio/transcriptions` (and `/v1/audio/translations`) against a local Whisper deployment fails with an identical 400, whatever the uploaded file:

```json
{"error": {"message": "Invalid or unsupported audio file."}}
```

I hit this while wiring a small tool against Logos and reproduced it against prod with six independently generated, valid files (WAV 16 kHz, WAV 44.1 kHz, MP3, M4A, OGG, WebM) plus a 16 KB pure-silence WAV — all rejected identically. A working decoder would distinguish those, so nothing is being decoded at all.

## Description

**Root cause.** The string `Invalid or unsupported audio file.` does not exist anywhere in Logos; it is raised by vLLM in `vllm/entrypoints/speech_to_text/base/serving.py`, which wraps *any* failure of `load_audio` into that one message. `vllm/multimodal/media/audio.py::load_audio` has exactly two backends and both are optional dependencies:

1. `soundfile` — primary
2. `av` / PyAV — fallback, and the resampler (`resample_audio_pyav`) for any input not already at Whisper's 16 kHz, i.e. most real files

Per PyPI, vLLM 0.29.0's `audio` extra is `av`, `scipy`, `soundfile`, `soxr`, `mistral_common[audio]`. The worker image installed plain `vllm==v0.29.0`, so neither backend was present and every upload failed before decoding began. PyAV bundles FFmpeg, so no `ffmpeg` apt package is needed.

**This was never a regression.** `git log --all -S` over the workernode tree finds `soundfile`, `av` and `[audio]` in no commit on any ref: #704 added the full orchestrator/worker multipart round-trip for Whisper but never touched the Dockerfile or `requirements.txt`, and no later commit did either. So the endpoint has been unusable since the day it shipped.

**Changes:**

1. **`logos/logos-workernode/Dockerfile`** — pin `vllm[audio]==v0.29.0`, with a comment recording why the extra is load-bearing. Note the spelling: extras precede the version specifier (PEP 508). The `vllm==v0.28.0[audio]` form suggested in the issue is **not a valid requirement** — `packaging.requirements.Requirement` and pip both reject it (`Expected comma (within version specifier)…`), so it would have broken the build rather than fixed the bug. I chose the extra over hand-listing `soundfile`+`av` in `requirements.txt` so the dependency tracks vLLM's own definition; the cost is `scipy`/`soxr`/`opencv` (via `mistral_common[audio]`, only Voxtral needs them).

2. **`.github/workflows/logos_update-vllm.yml`** — the reason this needs more than a one-line pin change. The workflow runs hourly and rewrites the pin on every vLLM release; as written it would revert this fix within the hour:
   - the *check* step's `sed` captured `vllm==v?([^"]+)`, which does not match a pin carrying extras at all → `CURRENT` empty → never equal to the PyPI version → a spurious "Update vLLM" PR **every run**;
   - the *update* step replaced the whole line with a hardcoded `vllm==v$VERSION`, silently dropping the extra.

   The check step now reads the version past any extras, and the update step captures the existing extras and re-attaches them to the bumped pin. Bare pins keep working unchanged. The PR-body text the workflow generates was adjusted to match.

3. **`logos/logos-workernode/docker-compose.dev.yml`** — the dev default (`${VLLM_PIP_SPEC:-vllm}`) now mirrors the image, so a dev worker can decode audio too and the bug can't hide until prod.

## Tests

New `logos/logos-workernode/tests/test_vllm_audio_extra.py` (stdlib only, in the style of `test_installer_pins.py`) guards both halves of the failure — the pin and the automation that would strip it:

- the Dockerfile pin carries `[audio]`, asserted on the parsed *shape* so the unbuildable `vllm==vX.Y.Z[audio]` spelling can't slip in;
- the dev-compose default carries it too;
- the workflow's two `sed` expressions are **compiled out of the YAML verbatim** and exercised against pins with one extra, several extras, an `rc` version, and no extra — so the tests check the expression the workflow really runs, plus the live Dockerfile line;
- the rewrite must re-attach `${EXTRA}` and must not contain the hardcoded bare form.

Verified as genuine regressions: on the unfixed tree all **4 fail**, with the fix all **4 pass**.

I also ran both `sed` expressions against the real files: the check step reads `0.29.0` from the new pin (so no hourly PR spam), the update step reads `[audio]`, and a simulated bump to `v9.9.9` yields `ARG VLLM_PIP_SPEC="vllm[audio]==v9.9.9"` — the extra survives. `pre-commit run --files` passes on all four files; the workflow YAML additionally `yaml.safe_load`s.

## What I could not verify here

The image build and a live transcription against a rebuilt worker — no Docker/GPU in this environment. The dependency argument is settled independently, though: the issue author verified on the prod worker that `pip install soundfile av` in the same container makes the identical file decode correctly (`DECODE OK: shape=(66256,) sr=16000`), and uninstalling reverts to the error. Remaining step is the CI image build + Whisper-lane rollout; `check-workernode-build` diffs the Dockerfile, so this triggers a rebuild automatically.

One deployment note for review: the transcription path is only exercisable once a Whisper lane is actually loaded on a worker — that is scheduling state, not part of this change.

---

https://claude.ai/code/session_012ULn1EiP4Rn2GQLYVsfvee

### DIFF
--- a/.github/workflows/logos_update-vllm.yml
+++ b/.github/workflows/logos_update-vllm.yml
@@ -62,7 +62,11 @@ jobs:
           VERSION: ${{ steps.version.outputs.version }}
           BRANCH: ${{ steps.version.outputs.branch }}
         run: |
-          CURRENT=$(sed -nE 's/^ARG VLLM_PIP_SPEC="vllm==v?([^"]+)"$/\1/p' "$DOCKERFILE")
+          # The pin may carry extras (vllm[audio]==vX.Y.Z). Capture only the
+          # version: matching the extras as part of it would make the pin never
+          # compare equal to the PyPI version, and this workflow would open the
+          # same "update" PR every hour.
+          CURRENT=$(sed -nE 's/^ARG VLLM_PIP_SPEC="vllm(\[[^]]*\])?==v?([0-9][0-9a-z.]*)"$/\2/p' "$DOCKERFILE")
           echo "Current pin: v$CURRENT"
           if [ "$CURRENT" = "$VERSION" ]; then
             echo "Dockerfile already pins vllm==v$VERSION — nothing to do."
@@ -81,8 +85,13 @@ jobs:
         env:
           VERSION: ${{ steps.version.outputs.version }}
         run: |
-          sed -i -E "s|^ARG VLLM_PIP_SPEC=.*$|ARG VLLM_PIP_SPEC=\"vllm==v$VERSION\"|" "$DOCKERFILE"
-          grep -F "vllm==v$VERSION" "$DOCKERFILE" >/dev/null || {
+          # Carry the current extras (e.g. [audio]) over to the new pin.
+          # Rewriting the line as a bare "vllm==v<version>" would silently
+          # uninstall the features that depend on them — the [audio] extra is
+          # what makes /v1/audio/transcriptions able to decode uploads at all.
+          EXTRA=$(sed -nE 's/^ARG VLLM_PIP_SPEC="vllm(\[[^]]*\])?==.*"$/\1/p' "$DOCKERFILE")
+          sed -i -E "s|^ARG VLLM_PIP_SPEC=.*$|ARG VLLM_PIP_SPEC=\"vllm${EXTRA}==v$VERSION\"|" "$DOCKERFILE"
+          grep -F "vllm${EXTRA}==v$VERSION" "$DOCKERFILE" >/dev/null || {
             echo "::error::Failed to update VLLM_PIP_SPEC in $DOCKERFILE"
             exit 1
           }
@@ -133,7 +142,7 @@ jobs:
 
           ## Description
 
-          Updates the \`VLLM_PIP_SPEC\` build arg default in \`$DOCKERFILE\` to \`vllm==v$VERSION\`, and the \`VLLM_REASONING_EFFORT_VALUES\` snapshot label in \`$EFFORT_MODULE\` alongside it. Torch and FlashInfer are resolved from vLLM's own dependency pins at build time, so no other changes are needed.
+          Bumps the version in the \`VLLM_PIP_SPEC\` build arg default in \`$DOCKERFILE\` to v$VERSION — extras carried by the pin (e.g. \`[audio]\`, which the speech-to-text endpoints need) are preserved — and the \`VLLM_REASONING_EFFORT_VALUES\` snapshot label in \`$EFFORT_MODULE\` alongside it. Torch and FlashInfer are resolved from vLLM's own dependency pins at build time, so no other changes are needed.
 
           Please check the release notes for new \`ChatCompletionRequest.reasoning_effort\` values: only the label is bumped automatically, the tuple is not. Runtime normalization is unaffected either way — unknown values fall back to the template family's default.
 
--- a/logos/logos-workernode/Dockerfile
+++ b/logos/logos-workernode/Dockerfile
@@ -61,7 +61,17 @@
 # =============================================================================
 
 ARG BASE_IMAGE=ubuntu:24.04
-ARG VLLM_PIP_SPEC="vllm==v0.29.0"
+# The [audio] extra is required, not cosmetic: the speech-to-text endpoints
+# (/v1/audio/transcriptions, /v1/audio/translations) decode uploads through
+# vllm.multimodal.media.audio.load_audio, whose only two backends are
+# soundfile (primary) and av/PyAV (fallback, and the resampler to Whisper's
+# 16 kHz for any input that isn't already at that rate). Both ship solely in
+# this extra; without them every upload fails with the same opaque 400,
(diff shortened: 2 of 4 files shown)


Claudia-Anthropica (approved):
@wasnertobias The Dockerfile and development Compose configuration now use a valid PEP 508 `vllm[audio]` requirement, while the production build retains the Dockerfile default. The updater preserves extras across version bumps, and the collected regression tests cover both the dependency pins and workflow expressions. The captured snapshot contained no failed or required checks, though several optional builds were still running.

PR #950, opened 2026-09-12
`Iris`: Lecture unit ingestion reliability

Nayer-kotry:
> ⚠️ **Companion PR** — this pairs with **[Artemis#13798](https://github.com/ls1intum/Artemis/pull/13798)** (`feature/iris/ingestion-reliability` → `develop`). Both should be reviewed and merged together; Artemis's reconciliation walk and lease/heartbeat protocol depend on the census and worker endpoints added here.

### Motivation and Context
The previous lecture-ingestion write path was delete-then-insert with no verification: a crash between the delete and the insert could leave a unit's index rows empty or half-written, and nothing detected or corrected it afterward — only a manual re-trigger noticed. This PR reworks ingestion into a convergent, self-verifying pipeline, adds the census/worker surface Artemis's reconciler depends on, and closes several gaps a live review pass found in that surface (ghost rows inflating counts, silent read truncation, an over-generous worker claim response).

### Description
**Write-then-purge + convergence** — each run writes its generation under client-assigned ids, verifies it, then purges other rows, so a crash can never leave a unit empty.

**Generations + ghost-robust counting** — every row carries an `ingestion_run_id`; counts are confirmed against the object store (`confirmed_generations`/`confirmed_rows`) so scan-only "ghost" rows (index/scan-visible but object-store-missing) never inflate them or block convergence.

**Fingerprint + end-of-run audit** — the content fingerprint is stamped, and the index is read back and checked (ghost-aware) before a unit is certified.

**Census** — per-course endpoint reporting what the index actually holds (generations, page coverage, display numbers, language, truncation), backing Artemis's reconciliation.

**Pull-based worker** — health-check upstream discovery with a lease/heartbeat, wedged-unit diagnostics, and a claim response clamped to the requested slot count.

**Display-page reconciliation**, stale same-slide segment pruning on the new deterministic uuid, and typed, bounded write retries.

**Configurable tunables** — vision retry attempts, read-completeness/convergence limits, and language-detection thresholds now live in `LectureIngestionSettings` (with validated bounds) instead of hardcoded module constants, matching the rest of Iris's settings.

### Steps for Testing
Prerequisites:
- A running Artemis instance (from the companion PR's branch) with a course containing a lecture unit
- A small lecture attachment (e.g. a short PDF)

1. Upload a lecture unit with an attachment from Artemis and confirm ingestion completes (`DONE`).
2. Hit the census endpoint (`GET` the course's ingestion census) and confirm it reports the unit with correct page coverage, generation count, and no truncation flag.
3. Re-upload the same content unchanged — confirm the write path skips re-ingestion (fingerprint match).
4. Simulate a crash mid-run (kill the process between the write and the purge) — confirm the next run converges the unit to a single generation instead of leaving duplicate/empty rows.
5. Kill the worker process mid-claim — confirm its lease lapses and Artemis's reconciler recovers the run within the stall window.

## Summary by CodeRabbit

* **New Features**
  * Added automatic ingestion worker discovery, capacity management, leasing, and heartbeats.
  * Added ingestion census reporting with completeness, quality, language, and generation metrics.
  * Added stage-level progress updates and stable process identifiers in health responses.
  * Added deterministic language detection with configurable fallback behavior.
  * Added safe reuse of unchanged lecture content and quality-based re-ingestion.

* **Reliability**
  * Improved ingestion verification, retry handling, error classification, and stale-data cleanup.
  * Added structured failure reporting for ingestion stages.


iris/src/iris/pipeline/ingestion_audit.py
```diff
@@ -0,0 +1,407 @@
...
+        run_ids: set = set()
+        stale_versions = 0
+        for chunk in chunks:
+            version = chunk.properties.get(
+                LectureUnitPageChunkSchema.PAGE_VERSION.value
+            )
+            if version != manifest.attachment_version:
+                stale_versions += 1
+            run_ids.add(
+                chunk.properties.get(LectureUnitPageChunkSchema.INGESTION_RUN_ID.value)
```
Claudia-Anthropica, line 198: @Nayer-kotry [high] Make the final audit ghost-aware before evaluating structural state. Unlike convergence and the census, these direct scans treat object-store-missing ghosts as real, so an undeletable ghost can produce mixed generations, count mismatches, or duplicate unit rows forever and prevent every otherwise healthy run from reaching `FINISHED`. Confirm scanned UUIDs with `fetch_object_by_id` and perform the page, transcript, segment, and unit-row checks only on confirmed objects.

In `iris/src/iris/pipeline/ingestion_audit.py`, the audit treats scan-visible but object-store-missing ghost rows as real and can therefore reject a healed unit forever. Filter every audited collection to object-store-confirmed rows before calculating generations, counts, pages, or unit-row cardinality.

</details>

iris/src/iris/pipeline/lecture_unit_segment_summary_pipeline.py
```diff
@@ -232,98 +320,60 @@ def _create_summary(self, transcriptions, slides) -> str:
...
+            LectureUnitSegmentSchema.CONTENT_FINGERPRINT.value: self.lecture_unit_dto.content_fingerprint,
+        }
+        embedding = self.llm_embedding.embed(summary)
+
+        def upsert():
+            # Insert under the deterministic id; if it already exists (a prior
+            # attempt landed, or a previous run wrote this slide) replace it in
+            # place. Both paths are idempotent, so retrying is duplicate-safe.
+            try:
+                self.lecture_unit_segment_collection.data.insert(
```
Claudia-Anthropica, line 368: @Nayer-kotry [medium] This deterministic insert does not migrate segment rows written by earlier versions, which used random UUIDs. On the next recomputation, each old row remains while a new deterministic row is inserted for the same slide; the stale-range purge keeps both, and the audit collapses pages into a set so it does not detect the duplicate. Remove or replace legacy same-slide rows when adopting the deterministic UUID and add a migration regression test.

In `iris/src/iris/pipeline/lecture_unit_segment_summary_pipeline.py`, deterministic segment inserts leave pre-existing random-UUID rows for the same slide in place, creating permanent duplicates that the current audit misses. During upsert, purge every same-slide row except the deterministic UUID and add coverage starting with a legacy row.

</details>

iris/src/iris/vector_database/batch_verify.py
```diff
@@ -0,0 +1,290 @@
...
+            limit=limit,
+            return_properties=properties,
+        ),
+        retry=retry,
+    ).objects
+    sample_ids_by_generation: dict = {}
+    for stored_object in objects:
+        generation = stored_object.properties.get(run_id_property)
+        sample = sample_ids_by_generation.setdefault(generation, [])
+        if len(sample) < sample_per_generation:
```
Claudia-Anthropica, line 204: @Nayer-kotry [medium] Limiting confirmation to the first three scanned IDs can discard a real generation. Scan order is unrelated to object-store liveness, so when those three IDs are ghosts but a later row from the same run is real, the census excludes every valid chunk in that generation and reports missing content. Continue checking candidates until one live object is found or all scanned IDs for that generation are exhausted.

In `iris/src/iris/vector_database/batch_verify.py`, `confirmed_generations` samples only three IDs per generation, allowing three leading ghosts to hide later real rows and causing valid chunks to be discarded. Confirm IDs until a live row is found or every scanned ID in that generation has been checked.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The exact head still substantiates the unresolved security and convergence findings already present in the review: upstream announcements can redirect credentials, concurrent purges can delete newer generations, and reuse, metadata-only, and census paths remain inconsistent. I also found three distinct data-integrity defects in the ghost-aware audit and migration paths, documented inline. The captured CI snapshot is complete and successful, but these high and medium issues require changes.

Claudia-Anthropica (changes requested):
@Nayer-kotry The exact head still substantiates the unresolved findings already present: worker upstream URLs remain unvalidated while credentials are forwarded, token prefixes are logged, and claimed-slot accounting counts deduplicated jobs. The ingestion paths also retain cross-generation purge races, incomplete summary-reuse and metadata-only handling, capped census scans, three-ID generation sampling, ghost-sensitive auditing, and legacy segment duplicates. Iris test, lint, and image-build checks succeeded in the captured snapshot, but these current high/medium defects still require changes.

iris/src/iris/pipeline/ingestion_audit.py
```diff
@@ -0,0 +1,407 @@
...
+        if self.expects_pdf_content:
+            return self.expected_pages
+        if self.expects_transcript:
+            return set(
+                range(
+                    min(self.transcript_slide_numbers),
+                    max(self.transcript_slide_numbers) + 1,
+                )
+            )
+        return set()
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed and fixed.

`_get_slide_range()` in `lecture_unit_segment_summary_pipeline.py` used to `return 0, 0` when neither the PDF-chunk aggregate nor the transcript aggregate found any rows — exactly the page-0 placeholder path you traced, silently mismatching the audit's `expected_segment_pages` (empty for a unit with no PDF and no transcript).

Dug into when this branch can actually be reached: a unit dispatched by Artemis always carries either a PDF or a transcript in the normal case, and a video-with-zero-transcript-segments unit can't get this far either (`isEnriched` stays False on an empty segment list on the Artemis side, so TRANSCRIBING never advances and the unit times out via existing stuck-detection instead of reaching this code). The real trigger is a corrupt or genuinely empty PDF: Artemis's `hasPdf` check is purely `link.endsWith(".pdf")`, with no validation that the file actually has readable pages.

Fixed by raising `IngestionStageError(NO_INGESTIBLE_CONTENT, ...)` at that point instead of returning `(0, 0)`, failing the run explicitly at the real cause instead of writing a stray page-0 segment for the audit to reject later with no indication why.

New regression coverage: `tests/test_lecture_unit_segment_summary_pipeline_slide_range.py` — the raise, plus two tests pinning that the normal PDF-only and transcript-only spans are unaffected.

iris/src/iris/pipeline/lecture_unit_pipeline.py
```diff
@@ -65,6 +70,38 @@ def _filter(lecture_unit: LectureUnitDTO):
...
+        """Reuse the stored summary and vector when it provably fits the content.
+
+        Only allowed when every content sub-pipeline structurally skipped this
+        run AND the stored row's fingerprint stamp equals the current content
+        fingerprint: same inputs, already summarized, already embedded. This is
+        what makes a metadata-only or reconcile re-dispatch of an unchanged
+        unit cost zero LLM calls. Any doubt falls through to a full recompute.
+        """
+        if not lecture_unit.content_unchanged or not lecture_unit.content_fingerprint:
+            return None
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed and fixed.

Traced the write ordering in `lecture_ingestion_update_pipeline.py::_run_ingestion`: `LectureUnitPipeline(...)()` writes the segment rows and then the unit row's fingerprint stamp, in that order, under `batch_update_lock` — before `IngestionAudit.verify()` runs, with no try/except between the stamp and the audit. Before this fix, `content_unchanged` was derived purely from whether every content sub-pipeline structurally skipped. A run whose earlier attempt stamped the unit row but whose segments the audit then rejected (e.g. the page-0-placeholder case from the sibling thread) would be treated as safely reusable on retry: `LectureUnitPipeline._try_reuse_stored_summary` would reuse the stale/incomplete summary and vector, skip regenerating segments entirely, and fail the identical audit check forever (bounded only by Artemis's `MAX_REVIVALS`, since `INGESTION_AUDIT_FAILED` isn't in `PERMANENT_ERROR_KEYS`).

Fixed by adding `segments_are_complete(client, dto)` as an additional, short-circuited condition on `content_unchanged`: it re-reads the unit's stored segments against the same manifest and ghost-aware confirmation the audit itself uses, and only lets the run treat the unit as unchanged when the segments genuinely already cover it. The `and` short-circuit means this extra Weaviate read only happens when every sub-pipeline already skipped, not on every ordinary run.

Verified live against the real running Weaviate instance, not a mock: captured a real segment row from a DONE unit, deleted it to simulate "unit row stamped, segments incomplete," confirmed `segments_are_complete()` correctly flips to `False` and the audit raises citing the exact missing page, then restored the row and confirmed both return to green.

New regression tests in `test_ingestion_audit.py`: `test_segments_are_complete_is_true_when_nothing_is_expected`, `test_segments_are_complete_true_when_stored_segments_cover_the_manifest`, `test_segments_are_complete_false_when_a_segment_is_missing`.

iris/src/iris/pipeline/ingestion_audit.py
Nayer-kotry: Confirmed and fixed as described.

Page chunks and transcriptions now route through `confirmed_generations` (already used elsewhere for convergence) instead of a raw `fetch_objects` scan, so a coexisting scan-visible-but-object-store-missing generation no longer reads as "multiple generations coexist." Segments and the unit row have no run-id property to group generations by, so I added `IngestionAudit._confirmed_rows()`, which checks each scanned row directly against `fetch_object_by_id` and keeps only confirmed ones — used in both `_verify_segments` and `_verify_unit_row` before computing page coverage / row cardinality.

One thing worth noting from digging in before applying this as a blanket fix: page chunks already had a different, independent protection (`_converge_to_single_generation` in the ingestion pipeline, which escalates to a full delete-and-rewrite and raises `STALE_CONTENT_DELETE_FAILED` if convergence can't be reached), so for chunks the gap this closes is a narrow eventual-consistency window right after convergence. Segments, transcriptions, and the unit row had no equivalent upstream protection at all, so the audit-level fix is the only protection they have.

New regression tests in `test_ingestion_audit.py`: `test_page_chunk_audit_ignores_a_ghost_generation`, `test_segment_audit_ignores_a_ghost_row_but_still_catches_a_real_gap` (proves the exclusion doesn't also hide a genuinely missing page), `test_unit_row_audit_ignores_a_ghost_duplicate`.

iris/src/iris/vector_database/batch_verify.py
Nayer-kotry: Confirmed and fixed as described.

Removed the `sample_per_generation` cap entirely rather than just raising it — once removed, no caller passed it explicitly (confirmed via grep), so I deleted the parameter instead of leaving an inert one, per pylint's unused-argument flag on the alternative. `confirmed_generations` now iterates every scanned id of a generation and stops at the first live confirmation, so it only checks further than the first id when the earliest scanned rows for that generation happen to be ghosts — exactly the case the fixed sample size could misjudge as fake. Cost is unchanged for the common case, where a real generation confirms on the first id checked.

Full Iris suite (515/515), black, and pylint (10.00/10) all pass with this change.

iris/src/iris/ingestion/worker.py
```diff
@@ -0,0 +1,380 @@
...
+            # lapses and Artemis reclaims it (via lease expiry — findStuckStates skips leased rows).
+            # A thread that keeps draining is reclaimed and re-skipped each cycle, so escalate: the
+            # unit stays blocked until that thread exits (a wedged one, only on a worker restart).
+            if skips >= _WEDGED_UNIT_SKIP_THRESHOLD:
+                logger.warning(
+                    "Unit %d skipped %d times: a prior run's thread is still alive; the unit stays "
+                    "blocked until that thread exits (token %s...)",
+                    unit_id,
+                    skips,
+                    token[:8],
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed already fixed in this branch. The three log statements that previously logged `authentication_token[:8]` (around claim/skip/dedup handling) now log `unit_id` and `upstream.url` instead — no token or token fragment appears in any log line in `worker.py`. This was covered by the earlier commit on this branch that also fixed the SSRF/redirect and slot-accounting issues in the same file. Verified live: the running instance's logs show entries like `Claimed ingestion job for unit 485 from http://localhost:8083`, with no token anywhere.

iris/src/iris/web/routers/ingestion_census.py
```diff
@@ -0,0 +1,291 @@
...
+from ...vector_database.lecture_unit_schema import LectureUnitSchema
+from ...vector_database.lecture_unit_segment_schema import (
+    LectureUnitSegmentSchema,
+)
+
+logger = get_logger(__name__)
+
+router = APIRouter(prefix="/api/v1", tags=["ingestion_census"])
+
+_UNIT_ROW_LIMIT = 10_000
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed. Live-reproduced by temporarily lowering the fetch cap and confirming both truncation sources (the course-wide unit-row scan, and the per-unit confirmed_generations chunk fetch) silently under-reported with no signal -- one scenario made 8 real DONE units look identical to "no unit row at all" (unitRowCount=0), which is the same signal the reconciler uses to decide what's orphaned. Added a per-unit `truncated` flag (chunk fetch capped) and a course-level `truncated` flag (unit-row scan capped), confirmed both fire correctly at the lowered cap and both read false with the real 10,000 limit, byte-identical to the pre-fix baseline otherwise. Commit 6d16cfe5e.

iris/src/iris/pipeline/lecture_unit_segment_summary_pipeline.py
Nayer-kotry: Fixed. Confirmed via git archaeology that the deterministic-uuid scheme is genuinely new in this branch (the code it replaced always inserted under a fresh random uuid), so any upgraded, non-fresh Weaviate instance has legacy rows that would collide with their deterministic replacement on the very next re-ingest -- a routine, not rare, event. Live-reproduced twice against the real running instance: an isolated call to the prune method with a manually-inserted legacy row (duplicate survived on the old 2-arg call, removed once the fix's uuid-exclusion condition was added), and a full ingestion run through Artemis+Iris with a genuinely modified PDF to force real segment regeneration.

Fixed by threading the set of uuids the run just wrote through to the prune step, which now also deletes any in-range row whose id isn't one of them, alongside the existing out-of-range cleanup. Callers that don't pass the new parameter (the two pre-existing unit tests) get exactly the prior range-only behavior, confirmed unchanged. Commit dcd47812a.

iris/src/iris/config.py
```diff
@@ -151,6 +151,31 @@ class TranscriptionSettings(BaseModel):
...
+    enabled: bool = Field(default=True)
+    capacity: int = Field(default=2)
+    poll_interval_seconds: float = Field(default=2.0)
+    heartbeat_interval_seconds: float = Field(default=5.0)
+    # Hostnames (case-insensitive, port ignored) an announced upstream must match to be
+    # registered. Empty (the default) accepts any http(s) URL, matching today's zero-config
+    # discovery for local and single-tenant deployments. Set this where Iris is reachable by
+    # parties other than its own Artemis installations, so a valid API key cannot point the
+    # worker's authenticated outbound requests at an arbitrary internal or external address.
+    allowed_upstream_hosts: list[str] = Field(default_factory=list)
```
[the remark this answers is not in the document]
Nayer-kotry: Not a bug — this is a documented, intentional tradeoff, not a silent gap.

`allowed_upstream_hosts`'s own docstring/comment at the point of declaration (config.py) states explicitly that an empty default accepts any http(s) URL, matching the zero-config discovery local and single-tenant deployments rely on, and that operators should set it wherever Iris is reachable by parties other than its own Artemis installation. `_is_allowed_upstream` implements exactly that documented fail-open-when-unconfigured behavior.

The announce path (`register_upstream`) is only reachable through the health endpoint, which is gated by `TokenValidator()` — so the actual threat model here is "a party who already holds a valid Iris API key can redirect the worker's own outbound claim/heartbeat requests," which is precisely the class of risk this config option (added earlier in this same PR for this same reason) exists to close for externally-reachable/multi-tenant deployments. It requires a trusted credential already, and the mitigation is already in place and documented for anyone who needs it.

`tests/test_ingestion_worker.py` already covers the allowlist-configured rejection path. No code change here — closing as working as documented.

iris/src/iris/pipeline/ingestion_audit.py
```diff
@@ -0,0 +1,468 @@
...
+                ]
+            return []
+
+        stored_pages = {
+            int(row.properties[LectureTranscriptionSchema.PAGE_NUMBER.value])
+            for row in rows
+            if row.properties.get(LectureTranscriptionSchema.PAGE_NUMBER.value)
+            is not None
+        }
+        expected_pages = set(manifest.transcript_slide_numbers)
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed real, but deliberately not fixing this one tonight -- flagging for an explicit decision instead.

Traced it precisely: `chunk_transcription()` merges same-slide segments, then splits any slide whose merged text exceeds 1200 characters (~200 words, ~1-2 minutes of speech -- a genuinely common length for a lecture slide's transcript, not a rare edge case) into multiple chunks via `split_text_semantically`, each carrying the same page number but a distinct id. `_verify_transcriptions()` does reduce stored rows to a pure page-number coverage set with no per-slide count check, exactly as described.

One thing worth being precise about on reachability: `write_batch_with_retry` already verifies every object in a batch either lands or the whole run fails, so a chunk lost during THIS run's own write already fails loudly before reaching the audit. The actual gap is narrower -- a post-write loss (one of a multi-chunk slide's rows becoming an object-store ghost after a successful write) -- which is the same ghost-row class as the confirmed_generations finding on batch_verify.py:216 in this same review pass; they compound.

The correct fix mirrors what page chunks already have (`chunk_counts_by_page`/`expected_chunk_counts`, computed right after chunking, persisted on the unit row, compared in `_verify_page_chunks`) -- but for transcriptions that means a new persisted field threaded through the chunking pipeline, the DTO, and the audit: a genuine schema/design change across 3+ files, not a same-function guard addition like the other fixes in this batch. Given the scope, I'd rather this go through an explicit decision than be implemented unilaterally overnight. Real finding, moderate severity given it needs a specific post-write ghost event on a multi-chunk slide -- happy to implement once the approach is confirmed.

iris/src/iris/vector_database/batch_verify.py
```diff
@@ -0,0 +1,291 @@
...
+    real_generations: set = set()
+    for generation, candidate_ids in ids_by_generation.items():
+        for object_uuid in candidate_ids:
+            found = fetch_with_retry(
+                lambda uid=object_uuid: collection.query.fetch_object_by_id(uid),
+                retry=retry,
+            )
+            if found is not None:
+                real_generations.add(generation)
+                break
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed real, but this is a deliberate cost/correctness tradeoff rather than a bounded bug fix -- flagging for an explicit decision rather than implementing unilaterally.

Traced precisely: every one of the three counting callers (`_verify_page_chunks`, `_verify_transcriptions`, and the census's chunk-count loop) filters the raw scanned `objects` down to rows whose run-id is in `real_generations`, then trusts ALL of those rows for exact counting -- but a generation is only confirmed by finding ONE live row and stopping there. A ghost sibling within an otherwise-real generation (plausible: `write_batch_with_retry`'s own verification only catches synchronous batch failures, not a later async ghosting of an already-acknowledged write, which this session directly observed live multiple times via Weaviate's "store is read-only due to resource pressure" error) would inflate a per-page count check while carrying its original, correct-looking page number -- exactly the kind of gap the audit exists to catch, silently passing. Only `_distinct_page_run_ids` in the ingestion pipeline is unaffected, since it only checks generation convergence and already discards `objects`.

The reason I'm not just fixing this: closing it means confirming every candidate row of a generation instead of stopping at the first live one, and the function's own docstring is explicit that the early-stop is what keeps this the same cost as a capped sample in the common (fully-real) case. Removing it multiplies `fetch_object_by_id` calls by generation size across 3 call sites that run on every audit and every periodic census call, not a one-off cost. That's a real tradeoff between steady-state query load and correctness under a rare corruption scenario, which feels like your call to make rather than mine tonight. A cheaper middle ground worth considering: only pay the extra confirmation cost when a generation actually HAS more than one candidate row to distinguish (the common single-row-per-generation case stays exactly as cheap as today). Happy to implement whichever direction you'd like.

iris/src/iris/ingestion/worker.py
```diff
@@ -0,0 +1,414 @@
...
+            response = self._post(
+                upstream, "claim", {"bootId": BOOT_ID, "maxJobs": slots}
+            )
+            response.raise_for_status()
+        except http_requests.exceptions.RequestException as e:
+            self._log_failure(upstream, "claim", e)
+            return 0
+        upstream.claim_failures = 0
+        jobs = (response.json() or {}).get("jobs") or []
+        return sum(1 for job in jobs if self._start_job(job, upstream))
```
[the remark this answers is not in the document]
Nayer-kotry: Fixed exactly as proposed. Added a client-side clamp (`jobs[:slots]`) so the worker never starts more than the requested capacity, regardless of what the upstream returns. Verified with a new test that fails without the fix (5 jobs started against a capacity of 2) and passes with it; full worker suite and Iris suite green.

iris/src/iris/web/routers/ingestion_census.py
```diff
@@ -0,0 +1,302 @@
...
+                )
+            ],
+        )
+        entry.segment_count = result.total_count or 0
+        entry.segment_page_min = _metric(
+            result, LectureUnitSegmentSchema.PAGE_NUMBER.value, "minimum"
+        )
+        entry.segment_page_max = _metric(
+            result, LectureUnitSegmentSchema.PAGE_NUMBER.value, "maximum"
+        )
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed and fixed for both metrics, exactly as described, with one implementation nuance: transcription rows do carry an ingestion-generation property, so that loop now uses `confirmed_generations` (mirroring the page-chunk loop right above it). Segments have no such property, so I extracted a shared `confirmed_rows()` function (from `IngestionAudit`'s existing per-row confirmation, which is now a thin wrapper over it) into `batch_verify.py` and use it directly for segments.

One thing worth being explicit about: a genuine object-store ghost (index-visible, object-store-missing) can't actually be fabricated through Weaviate's normal client API -- it's an internal corruption/interrupted-write state, not something insert/delete can reproduce on demand (I confirmed this the hard way: a "ghost injection" attempt against the real running instance just created a normal, fully-confirmable row). So the ghost-exclusion behavior itself is covered by a unit test that mocks the confirmation boundary directly, the same approach `IngestionAudit`'s own existing ghost-row tests already use for this exact reason. What I could and did verify live: the census response for real, healthy data (no ghosts) is byte-identical before and after this change.

Full suite (521 tests), black, and pylint (10.00/10) all pass.

Nayer-kotry:
@coderabbitai review

Nayer-kotry:
@coderabbitai review

iris/src/iris/config.py
```diff
@@ -151,6 +151,75 @@ class TranscriptionSettings(BaseModel):
...
+        ),
+    )
+    language_detection_max_chars: int = Field(
+        default=10_000,
+        description="Max characters of deck text fed to the language detector per run.",
+    )
+    default_language: str = Field(
+        default="en",
+        description="Fallback language (ISO 639-1) when detection is skipped or fails.",
+    )
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed and fixed — verified each claim before applying: `vision_max_attempts<=0` does empty `range(1, value+1)`, `skip_check_fetch_limit<=0` does make `len(chunks) >= limit` true unconditionally, `convergence_max_escalations<0` does empty `range(value)` (0 correctly stays valid, per your suggestion), and the min/max language bounds had no relative ordering guarantee.

Applied your proposed diff essentially as-is (`ge=1`/`ge=0` per field, plus the `@model_validator(mode="after")` cross-field check), matching the existing `MemirisSettings`/`LangfuseSettings` validator pattern already in this file. Added `tests/test_lecture_ingestion_settings.py` covering each bound (including that `convergence_max_escalations=0` is accepted and `language_detection_max_chars == language_detection_min_chars` is accepted). Defaults unaffected — full suite (530 tests) + black + pylint (10/10) all pass.

Commit `8fa5a98f6`.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
```diff
@@ -486,30 +854,47 @@ def interpret_image(
...
+                )
+            except Exception as e:
+                last_error = e
+                logger.warning(
+                    "Slide vision attempt %d/%d failed: %s",
+                    attempt,
+                    settings.lecture_ingestion.vision_max_attempts,
+                    e,
+                )
+
```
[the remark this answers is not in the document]
Nayer-kotry: Confirmed real — verified directly rather than taking this at face value: `OpenAIChatModel.chat` does run its own retry loop with exponential backoff (`_retry_after_openai_error`, `backoff_factor = 2`, `max_retries=0` on the SDK client precisely because this loop is the only retry layer), while `OllamaModel`'s custom transport explicitly says it exists "to speed up request performance and avoid default retry/backoff behavior" and calls `self._client.chat` with no retry wrapper at all. So for an Ollama-backed vision model, `interpret_image`'s loop is exactly as tight as described.

Deliberately not fixing this one tonight, for the same reason as the transcription chunk-merge finding earlier in this review: it's a genuine design decision, not a same-function guard addition. This loop's `except Exception` is a catch-all today — JSON-parse failures, empty-description failures, and actual transient provider errors all land in the same branch. Doing this properly (bounded backoff for transient provider errors only, immediate retry for parse/empty-description failures, no duplication of OpenAI's own backoff) needs either typed exceptions surfaced from each provider's client or a heuristic classifier here, and touches provider-specific behavior across at least the Ollama and OpenAI chat paths.

Also worth noting explicitly: this is pre-existing behavior my change tonight didn't introduce or worsen — my diff here only swapped `VISION_MAX_ATTEMPTS` for the now-configurable `settings.lecture_ingestion.vision_max_attempts`, the retry loop body is untouched. Flagging for an explicit decision on the right classification/backoff approach rather than picking one unilaterally.

Nayer-kotry:
@coderabbitai review

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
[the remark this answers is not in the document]
Nayer-kotry: Fixed rather than deferred — no follow-up issue needed.

Checked the rest of the codebase first for other instances of this pattern: every other manual retry loop here (`write_retry.py`, `whisper_client.py`, `status_update.py`, `ingestion_status_callback.py`) already backs off between attempts. `interpret_image` was the one outlier with zero delay, so this brings it in line with the existing house pattern rather than introducing a new one.

Implementation: split the `except Exception` into `except ValueError` (covers `json.JSONDecodeError`, a `ValueError` subclass, plus the explicit empty-`academic_description` raise) which retries immediately, and a remaining `except Exception` for genuine provider errors, which sleeps `min(2**(attempt-1), 8)` seconds before the next attempt — capped, and skipped on the final attempt since there's nothing left to retry into. No provider-type detection needed: this doesn't duplicate OpenAI's backoff because by the time an exception reaches this loop, `OpenAIChatModel`'s own internal retry/backoff has already run and exhausted; one short delay before the next outer attempt is negligible on top of that and is exactly what was missing for Ollama, which has none of its own.

Added `test_interpret_image_json_decode_error_retries_immediately` and asserted `mock_sleep` call counts in the two existing retry tests (immediate-retry path: `assert_not_called()`; backoff path: `call_count == vision_max_attempts - 1`) to lock in both branches. 531/531 tests pass, black + pylint (10/10) clean.

Commit `88657b865`.

iris/src/iris/pipeline/ingestion_audit.py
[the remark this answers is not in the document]
Nayer-kotry: No follow-up issue — closing this out as a deliberate deferral. The reasoning above stands as the record if this needs picking up later.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
```diff
@@ -363,49 +566,139 @@ def restore_display_page_numbers_from_existing_chunks(self) -> None:
...
-                except Exception as e:
-                    logger.error("Error updating lecture unit", exc_info=e)
-                    raise
+            # One retry budget for the whole swap: a transient store condition
+            # re-submits only the dropped chunks, never the vision/embedding work.
+            retry = WeaviateWriteRetry.for_request()
+            written_ids = write_batch_with_retry(
+                self.collection, prepared_chunks, "lecture page chunks", retry=retry
+            )
+            purge_other_rows(
```
Claudia-Anthropica: @Nayer-kotry [high] This purge can still delete a newer generation. Both `batch_update_lock` and `lecture_update_lock` are process-local, while lease expiry or revocation explicitly permits an old thread to drain as another process claims the unit; if that old writer reaches this unit-wide purge last, it deletes the newer writer's UUIDs. Add an authoritative per-unit storage-level ownership fence to every page-chunk, transcription, and unit-row purge, with a two-writer interleaving test proving the newer generation survives.

In `iris/src/iris/pipeline/lecture_ingestion_pipeline.py`, an older revoked or lease-expired writer can purge rows inserted by a newer writer in another process. Add an authoritative storage-level generation-ownership fence to every destructive ingestion purge and add an interleaving regression test that proves the newer generation survives.

</details>

iris/src/iris/vector_database/batch_verify.py
```diff
@@ -0,0 +1,312 @@
...
```
Claudia-Anthropica, line 216: @Nayer-kotry [medium] The helper still stops after the first live UUID and returns every scanned row. Audit and census callers then retain all rows whose generation is in `real_generations`, so an object-store-missing sibling in an otherwise-live generation is counted as real and can inflate exact counts, coverage, and version checks. Confirm every candidate UUID and return confirmed UUIDs or rows for exact-count callers; keep a generation-only fast path only for convergence checks.

In `iris/src/iris/vector_database/batch_verify.py`, confirming one UUID marks every scanned sibling in that generation as real, allowing ghost siblings into audit and census counts. Return row-level confirmation results and make every exact-count caller filter by confirmed UUID rather than only by generation ID.

</details>

iris/src/iris/pipeline/ingestion_audit.py
```diff
@@ -0,0 +1,458 @@
...
+                    max(self.transcript_slide_numbers) + 1,
+                )
+            )
+        return set()
+
+
+def build_manifest(dto: IngestionPipelineExecutionDto) -> IngestionManifest:
+    """Derive the manifest from the request inputs alone."""
+    page_count = 0
+    pdf_base64 = dto.lecture_unit.pdf_file_base64
```
Claudia-Anthropica, line 96: @Nayer-kotry [medium] An omitted `pdfFile` or transcript is interpreted here as proof that the unit should have no stored content, although the ingestion path treats the same omission as structurally unchanged and the summary-reuse path explicitly supports metadata/reconcile redispatches. Such a redispatch reuses the stored unit row and then fails the audit because its healthy chunks or transcript still exist. Distinguish omitted payloads from explicit content removal and audit reused content against its persisted certified manifest.

In `iris/src/iris/pipeline/ingestion_audit.py`, omitted content payloads are treated as empty content even though the ingestion path treats omission as a reusable metadata or reconciliation dispatch. Add an explicit presence/removal distinction, use the persisted manifest when content is omitted, and cover an unchanged redispatch without binary content.

</details>

iris/src/iris/pipeline/lecture_unit_segment_summary_pipeline.py
```diff
@@ -140,33 +233,59 @@ def _get_slides(self, slide_number: int):
...
+
+        Uses a server-side aggregate min/max rather than fetching objects and
+        reducing in Python: an unbounded ``fetch_objects`` returns only the
+        client's default page, so a unit larger than that page (or one whose
+        chunk set is inflated by coexisting generations) yielded a truncated
+        range and summaries were produced for only the first slides. The
+        aggregate scans all matching rows and is unaffected by page size or
+        generation count.
+        """
+        slide_span = self._aggregate_page_number_span(
```
Claudia-Anthropica: @Nayer-kotry [medium] This aggregate derives the segment span from raw index-visible rows across every generation, without object-store confirmation. A stale ghost row on an outlying page therefore expands the span, causes spurious segment summaries to be written, and makes the final manifest-based audit fail on every retry. Derive the range from the request manifest or object-store-confirmed rows of the active generation and add a ghost-page regression test.

In `iris/src/iris/pipeline/lecture_unit_segment_summary_pipeline.py`, raw aggregate min/max allows an object-store-missing stale page row to expand the segment range and permanently fail the audit. Derive the range from the current manifest or confirmed active-generation rows and add regression coverage for an outlying ghost page.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The current head still has four blocking integrity defects in the convergence and audit paths. Storage-wide purges remain protected only by process-local locks, while row-level ghost confirmation, omitted-payload audit semantics, and segment-range derivation can still produce wrong counts or make healthy units fail indefinitely. The captured Iris tests, lint, builds, and deployment check succeeded, but they do not exercise these interleavings and corruption states.

Claudia-Anthropica (changes requested):
@Nayer-kotry This exact head still has the four previously raised integrity defects. The merged `current_job_guard` only consults process-local `_running_jobs` and `lecture_update_lock`, so a revoked or expired writer in another Iris process can still purge a newer generation; the stale-segment prune also remains outside the guard. `confirmed_generations` still stops after one live UUID while exact-count callers trust every sibling, `build_manifest` still converts omitted payloads to empty expectations, and `_get_slide_range` still aggregates raw cross-generation rows, leaving the ghost-count, metadata-redispatch, and ghost-range failures reachable. No required check was pending or failed in the snapshot, although several non-required checks were still in progress.

Claudia-Anthropica (changes requested):
@Nayer-kotry The follow-up commit only changes a pylint annotation in a test, so the four current-head integrity defects remain. Destructive purges still rely on process-local ownership while segment pruning remains outside that guard; `confirmed_generations` still promotes ghost siblings after one live UUID; `build_manifest` still treats omitted payloads as empty; and `_get_slide_range` still aggregates raw cross-generation rows. No required check was pending or failed in the snapshot, although several non-required checks were still running.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
Nayer-kotry: Fixed in 0c90289fe: `_prune_stale_segments` now runs inside the same `batch_update_lock` / `current_job_guard` used by every other destructive write in this pipeline, so a revoked/expired writer's stale prune is fenced the same way the rest of the purge/insert paths already are.

The deeper ask — an authoritative storage-level ownership fence for a genuine cross-process race — I'm deferring, with rationale: I checked this repo's actual production deployment (`docker/pyris-production.yml`, `docker/pyris-production-internal.yml`) and Iris runs as a single `pyris-app` docker-compose service with `restart: unless-stopped`, no `deploy.replicas`, and there's no k8s manifest anywhere in the repo. So the two-process interleaving this finding describes isn't reachable in the current topology — the in-process `_running_jobs`/`lecture_update_lock` guard is the correct-sized fix for that. If Iris is ever horizontally scaled, this becomes a real gap and would need a DB/object-store-level generation-ownership check at purge time (e.g. a conditional delete keyed on the current live generation), which is a meaningfully bigger change than anything else in this PR — flagging it for that future rather than building it against a topology that doesn't exist today.

iris/src/iris/vector_database/batch_verify.py
Nayer-kotry: Fixed in 0c90289fe. Kept `confirmed_generations` unchanged as a generation-only fast path (its one caller, the convergence check in `lecture_ingestion_pipeline.py`, only needs to know a generation is live, not confirm every row), and instead moved every exact-count caller onto the existing `confirmed_rows` helper, which checks each candidate row individually:

- `ingestion_audit.py`: `_verify_page_chunks` and `_verify_transcriptions` now compute the generation-matched rows first, then filter them through `confirmed_rows` before counting/comparing against the manifest.
- `ingestion_census.py`: the chunk and transcription census loops do the same — `confirmed_rows(db.lectures, generation_matches)` / `confirmed_rows(db.transcriptions, generation_matches)`.

A ghost sibling in an otherwise-live generation is now excluded from all four counts rather than riding along on the generation's one confirmed UUID.

iris/src/iris/pipeline/ingestion_audit.py
Nayer-kotry: I traced this on the Artemis side rather than changing code, and I don't think the premise holds against the current dispatch path — pushing back rather than fixing.

Both dispatch entry points (`addLectureUnitToPyrisDB`, the push webhook, and `prepareLectureUnitIngestion`, the pull/worker-claim path) go through the same `processAttachmentVideoUnitForUpdate` in `PyrisWebhookService.java`. There, `base64EncodedPdf` is only ever `""` when `attachmentVideoUnit.getAttachment() == null` — a genuinely deleted attachment. Whenever an attachment exists, the full PDF is always re-encoded and sent, regardless of `contentFingerprint` or `forceReingest`. The transcript DTO follows the same shape: it's only omitted when `lectureTranscriptionsRepositoryApi.findByLectureUnit_Id(...)` is genuinely empty.

So there isn't actually a 'metadata/reconcile redispatch that reuses the stored unit row while omitting content' path on the Artemis side — an omitted payload here always means the attachment/transcript was actually removed, in which case treating it as 'no stored content' is correct, not a false failure. If there's a specific redispatch call site you have in mind that skips `processAttachmentVideoUnitForUpdate`, happy to take another look at that one — but across everything reachable from `addLectureUnitToPyrisDB`/`prepareLectureUnitIngestion` today, this scenario isn't reachable.

iris/src/iris/pipeline/lecture_unit_segment_summary_pipeline.py
Nayer-kotry: Fixed in 0c90289fe. Replaced the raw `aggregate.over_all` min/max with `_confirmed_page_number_span`: it fetches the candidate rows (bounded by `settings.lecture_ingestion.skip_check_fetch_limit`, the same bound `confirmed_generations` uses elsewhere in this pipeline), filters them through `confirmed_rows` so an object-store-missing ghost is dropped before the reduce, and only then takes min/max in Python over what's left. A ghost row on an outlying page can no longer widen the range.

Rewrote `test_lecture_unit_segment_summary_pipeline_slide_range.py` around the new fetch-and-confirm shape, including two regression cases for exactly this: `test_a_ghost_row_does_not_expand_the_range` (a confirmed 1–6 span stays 1–6 even with an unconfirmed row at page 99) and `test_all_ghost_rows_falls_back_like_no_rows_at_all` (all-ghost input raises `NO_INGESTIBLE_CONTENT`, same as no rows at all).

iris/src/iris/pipeline/lecture_unit_segment_summary_pipeline.py
```diff
@@ -152,33 +260,67 @@ def _get_slides(self, slide_number: int):
...
+        Fetches rather than aggregates, unlike the raw min/max this replaced, because
+        confirming against the object store needs the actual candidate ids; the fetch
+        is bounded the same way confirmed_generations is elsewhere in this pipeline --
+        a unit with more rows than that would mean an unrealistic page count.
+        """
+        retry = getattr(self, "_retry", None)
+        rows = fetch_with_retry(
+            lambda: collection.query.fetch_objects(
+                filters=unit_filter,
+                limit=settings.lecture_ingestion.skip_check_fetch_limit,
```
Claudia-Anthropica: @Nayer-kotry [medium] This fetch silently treats a capped scan as complete. Because the limit counts rows rather than pages, a unit with at least `skip_check_fetch_limit` chunks can omit its true minimum or maximum from the returned subset; the pipeline then writes and prunes the wrong segment span, and the final audit fails on every retry. Paginate the full matching set or reject a cap hit before deriving the span.

In `iris/src/iris/pipeline/lecture_unit_segment_summary_pipeline.py`, `_confirmed_page_number_span` treats a fetch that reaches `skip_check_fetch_limit` as complete, so min/max may come from a truncated subset and drive an incorrect segment span. Paginate and confirm the entire matching set, or reject a cap hit with a typed failure before writing or pruning segments.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The latest commit fixes exact-count ghost filtering and the ghost-expanded segment range, while the repository's single-instance Iris deployment and new prune guard resolve the prior purge concern. Two medium blockers remain: omitted PDF/transcript fields are still interpreted as empty content in `build_manifest`, causing unchanged reconciliation dispatches to fail audit, and the new span scan trusts a capped row set and can repeatedly fail large units. Several non-required checks were still in progress in the snapshot; no required check was pending or failed.

iris/src/iris/pipeline/lecture_unit_segment_summary_pipeline.py
Nayer-kotry: Fixed in 828a15ad0. Confirmed this against an existing precedent already in the codebase: `check_if_attachment_needs_update` uses the same `skip_check_fetch_limit` bound and already treats a capped scan as untrustworthy (`if len(chunks) >= limit: return True`, i.e. force re-ingest rather than trust a partial sample). `_confirmed_page_number_span` had no equivalent guard.

Added the same rule: if the fetch returns at least `skip_check_fetch_limit` rows, it now raises `IngestionStageError(PAGE_RANGE_FETCH_CAPPED)` instead of deriving a span from the possibly-truncated subset — pagination isn't needed here since there's no safe way to reuse a partial min/max across pages without re-scanning anyway, and this stage has no fallback like "just re-ingest"; failing the run explicitly is the correct outcome for a stage that's actually mid-run.

Added `test_a_capped_fetch_raises_instead_of_trusting_a_truncated_span` (patches the limit down to 2 rows) to `test_lecture_unit_segment_summary_pipeline_slide_range.py`.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
```diff
@@ -263,27 +333,170 @@ def __call__(self) -> (str, []):
...
+        instead of "less than" also re-ingests after a version rollback), the
+        chunks cover exactly pages 1..page_count, all rows belong to a single
+        ingestion generation (mixed run ids mean a crashed write left old and
+        new rows side by side), and the stored per-page chunk counts match
+        exactly when the unit row recorded them (a crash inside a batch flush
+        can drop chunks below page granularity). A partially ingested unit
+        therefore self-heals through re-ingestion on its next dispatch.
+        """
+        chunks = fetch_with_retry(
+            lambda: self.collection.query.fetch_objects(
```
Claudia-Anthropica, line 462: @Nayer-kotry [medium] This completeness check evaluates raw scan-visible rows, whereas the final audit confirms every UUID. An all-ghost but structurally complete result therefore returns `False`, every retry skips replacement, and the audit then fails because none of those rows exists in the object store; `check_if_transcription_needs_update` has the same permanent loop. Confirm candidates before permitting either skip—or force re-ingestion when any candidate is unconfirmed—and add all-ghost regression coverage.

In `iris/src/iris/pipeline/lecture_ingestion_pipeline.py`, the attachment skip check can accept an all-ghost row set and permanently prevent repair. Filter candidates through object-store confirmation before allowing a skip, apply the same rule to the analogous transcription skip check, and add all-ghost regression tests.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The capped page-range scan is now rejected before any write, resolving that finding. Two blocking correctness paths remain: the PDF/transcript skip checks trust scan-visible ghosts and can permanently prevent repair, while omitted content is still indistinguishable from removal and neither interpretation correctly handles persisted rows. The snapshot has no required checks, though several non-required checks were still running.

iris/src/iris/pipeline/ingestion_audit.py
Claudia-Anthropica: @Nayer-kotry The current Iris code still contradicts the proposed producer-only interpretation: `LectureUnitPageDTO` defaults an omitted `pdfFile` to `""`, so Iris cannot distinguish omission from deletion, and `_run_ingestion` skips the page/transcript pipeline without deleting existing rows. Thus an omitted redispatch makes `build_manifest` reject reused content, while a genuine removal leaves stale rows that the same audit rejects. Preserve presence versus removal explicitly, then reuse the persisted manifest on omission and purge stored content on explicit removal.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
Nayer-kotry: Fixed in 82104b913, for both `check_if_attachment_needs_update` and `check_if_transcription_needs_update` (same bug, same fix, in `transcription_ingestion_pipeline.py`).

Both now confirm the scanned rows via `confirmed_rows` before permitting a skip: if the confirmed count doesn't match the scanned count, that's forced to `True` (needs update) instead of falling through to `False`. This closes the permanent-loop case exactly as described — an all-ghost-but-structurally-complete result can no longer make the skip check say "fine" forever while the manifest audit fails every retry.

Added `test_attachment_needs_update_when_all_chunks_are_ghosts` and `test_transcription_needs_update_when_all_rows_are_ghosts` as the requested all-ghost regression coverage.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
```diff
@@ -212,46 +262,67 @@ def __call__(self) -> (str, []):
...
-                self.restore_display_page_numbers_from_existing_chunks()
+            pdf_path = save_pdf(self.dto.lecture_unit.pdf_file_base64)
+            doc = None
+            try:
+                doc = fitz.open(pdf_path)
+                force_reingest = self.dto.lecture_unit.force_reingest
+                if not force_reingest and not self.check_if_attachment_needs_update(
+                    doc.page_count
+                ):
+                    self.course_language = self._resolve_course_language(doc)
```
Claudia-Anthropica: @Nayer-kotry [medium] This skip path ignores the newly authoritative course language. If the requested language changes while the attachment version and page structure remain unchanged, the pipeline skips regeneration, returns the new language here, and rewrites the unit row/census with that language even though the persisted chunk text, embeddings, and `course_language` still reflect the old one; reconciliation then loses the mismatch signal. Include the resolved requested language in skip eligibility and force full page re-ingestion when it differs from the stored chunk language.

In `iris/src/iris/pipeline/lecture_ingestion_pipeline.py`, the attachment skip path can stamp a newly requested course language on the unit row while retaining page chunks generated under the previous language. Compare the resolved requested language with stored chunk language before skipping and force full page regeneration when they differ.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The latest commit correctly makes both structural skip paths reject unconfirmed rows. Blocking correctness issues remain: the existing empty/omitted-content path neither purges nor safely reuses persisted content, and the PDF skip path can relabel a unit’s language without regenerating its stored chunks. The non-required amd64 image build failed only because Docker Hub reset the BuildKit manifest connection, so that failure is unrelated to this change.

iris/src/iris/pipeline/ingestion_audit.py
Nayer-kotry: You're right, and I traced it further before fixing — this is real, independent of my earlier (correct, but narrower) point about Artemis never partially resending. Fixed in b637eb972.

Confirmed on the Iris side: `_run_ingestion`'s `has_pdf`/`has_transcript` `False` branches did nothing but send a heartbeat — no purge. Since Artemis rebuilds the full payload from its current DB state on every dispatch (verified: `processAttachmentVideoUnitForUpdate` is the sole DTO builder for every dispatch type), an absent PDF/transcript in the request always means Artemis currently has neither for this unit — removed, video source changed, or never present. Left alone, the manifest audit correctly expects zero content, finds the stale rows, and fails forever with nothing to clean them up — exactly the loop you described.

Added `_purge_stale_page_chunks` and `_purge_stale_transcription`, called from those same `else` branches, guarded by the same `batch_update_lock`/`current_job_guard` pattern as every other destructive write here. This mirrors `TranscriptionIngestionPipeline._replace_prepared_chunks`'s existing "transcript present but empty → clear what's stored" handling, just one level up for "absent from the request entirely." `delete_many` is idempotent, so this is safe to call unconditionally (including the common case where nothing was ever stored, e.g. a video-only or PDF-only unit).

Added `test_run_ingestion_purges_stale_content_when_pdf_and_transcript_are_absent" verifying both purges fire.

Claudia-Anthropica (changes requested):
@Nayer-kotry The new purge helpers resolve the absent PDF/transcript cleanup path, but the existing language-consistency blocker remains on this exact head. `check_if_attachment_needs_update()` still neither requests nor compares the stored chunk language; after a skip, `_resolve_course_language()` returns the newly requested language and the unit row is rewritten with it while the retained chunks remain in the prior language. No required check is pending or failed in the snapshot, although four non-required Iris checks remain in progress.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
Nayer-kotry: Fixed in 41ea42ad1. Resolved the requested language once per run (the same `_resolve_course_language` the skip branch already called) and compare it against each stored chunk's `course_language` alongside the existing version check — a null (legacy row, written before this field existed) or mismatched stored language now forces full re-ingestion instead of skipping, exactly like a version or page-count mismatch already does.

Added coverage in `test_attachment_needs_update_is_structural`: a language-changed case (chunks stamped "en", request now "de") and a null-stored-language legacy case, both asserting `True`.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
Nayer-kotry: Closing this as deferred rather than fixed: the in-process part of this finding is fixed (see the reply above, 0c90289fe), but the cross-process storage-level ownership fence is out of scope for this PR — it isn't reachable in the current single-instance deployment (no replicas, confirmed via docker/pyris-production*.yml), and building it now would be sizing the fix for a topology that doesn't exist. Tracking it as follow-up work for if/when Iris is horizontally scaled, at which point it would need a DB/object-store-level generation-ownership check at purge time.

iris/src/iris/pipeline/lecture_ingestion_update_pipeline.py
```diff
@@ -479,12 +554,35 @@ def _run_ingestion(
...
+        )
+        # `and` short-circuits: the extra read only happens when every content
+        # sub-pipeline already skipped, not on every ordinary run.
+        content_unchanged = structurally_unchanged and segments_are_complete(
+            client, self.dto
+        )
         callback.update()
-        lecture_unit_dto = self._build_lecture_unit_dto(language)
+        stage_started_at = time.monotonic()
+        lecture_unit_dto = self._build_lecture_unit_dto(language, content_unchanged)
```
Claudia-Anthropica: @Nayer-kotry [medium] `language` is first populated by page ingestion but overwritten with `transcription.language` when a transcript exists; transcript-only units also never use `dto.lecture_unit.course_language`. Passing that value here stamps the unit row and census with the transcript language even though the declared course language is authoritative, so reconciliation can repeatedly detect a false mismatch on an otherwise healthy unit. Resolve the ledger language from the declared course language first and keep the page and transcript language results separate.

In `iris/src/iris/pipeline/lecture_ingestion_update_pipeline.py`, the language passed into the unit-row DTO can be transcript metadata instead of the authoritative declared course language. Resolve the unit-row language from `dto.lecture_unit.course_language` first, falling back only when it is absent, and do not let transcript processing overwrite the page/course result.

</details>

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
```diff
@@ -212,46 +262,68 @@ def __call__(self) -> (str, []):
...
+                        lecture_unit_slide_dto=self.dto.lecture_unit,
+                        base_url=self.dto.settings.artemis_base_url,
+                    )
+                )
+            finally:
+                if doc is not None:
+                    doc.close()
+                cleanup_temporary_file(pdf_path)
+            self._record_chunk_manifest_and_quality(chunks)
+            if force_reingest and self._previous_generation_scores_better():
```
Claudia-Anthropica: @Nayer-kotry [medium] This retention decision bypasses the new language eligibility check. If a quality-forced run accompanies a declared-language change and its score is lower, this branch keeps chunks whose text, embeddings, and `course_language` use the old language, then returns the newly resolved language for the replacement unit row; the audit does not compare chunk language, so it certifies that inconsistency. Permit quality-based retention only when the stored chunks already match `requested_language`; otherwise replace them regardless of score.

In `iris/src/iris/pipeline/lecture_ingestion_pipeline.py`, forced quality re-ingestion can retain a better-scoring generation written in an obsolete course language while stamping the unit row with the new language. Apply the same stored-chunk language check before retaining the previous generation and force replacement whenever it differs from `requested_language`.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The latest comparison fixes the original unchanged-PDF skip path, and the resolved storage-fence concern remains non-blocking for the current single-instance deployment. Two language-consistency paths remain: transcript processing can override the authoritative course language in the unit ledger, and forced quality re-ingestion can retain old-language chunks while certifying the newly requested language. The captured tests, lint, and builds succeeded but do not cover these combinations.

iris/src/iris/pipeline/lecture_ingestion_update_pipeline.py
Nayer-kotry: Fixed in 612f1d7e5. `_run_ingestion` now resolves the declared `dto.lecture_unit.course_language` first and lets it win over whatever a content sub-pipeline reports; the fallback (whatever `language` ended up being from the page or transcript pipeline) only applies when no course language is declared. The transcript pipeline's own return value (the transcript's own language metadata, or Whisper's detection) can no longer overwrite it.

`language` itself is otherwise unused elsewhere in `_run_ingestion` (only fed into `_build_lecture_unit_dto`), so this is a minimal, localized fix — no restructuring of the page/transcript sub-pipelines needed.

Didn't add a dedicated unit test for this one: exercising it properly needs both the real PDF and transcript sub-pipelines running (Weaviate collections, embeddings, etc. all mocked), which is a lot of scaffolding for what's a 3-line priority fix. Verified instead via the full suite (732 passed) plus the sibling fix below, which shares the same test infrastructure and does have dedicated coverage.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
Nayer-kotry: Fixed in 612f1d7e5. `_previous_generation_scores_better` now takes `requested_language` and checks it against the stored unit row's own `course_language` property before even looking at the score — a mismatch forces replacement regardless of which one scored better, same as the structural skip check already does.

Added `test_quality_reingest_replaces_a_better_stored_generation_in_the_wrong_language`: a stored generation that scores higher (0.9) but is stamped "de" while the request wants "en" — asserts replacement proceeds (`insert`/`embed` happen, `kept_previous_generation` is `False`) despite the worse score.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
```diff
@@ -263,27 +337,200 @@ def __call__(self) -> (str, []):
...
+        """True when the stored generation's quality beats this run's result.
+
+        Only permitted when the stored generation was itself written in the
+        currently requested language: retaining an old-language generation while
+        stamping the newly resolved language onto the unit row would create the
+        same mismatch the structural skip check guards against, and the audit
+        does not compare chunk language to catch it.
+        """
+        stored = self._fetch_stored_unit_row_properties()
+        stored_language = stored.get(LectureUnitSchema.COURSE_LANGUAGE.value)
```
Claudia-Anthropica: @Nayer-kotry [medium] This reads the unit ledger, not the page chunks this branch retains. Before this PR, `_run_ingestion` could stamp the transcript language onto the unit row after page chunks were written using the independently detected PDF language; if that ledger value matches `requested_language`, this returns true and keeps mismatched chunks, then the audit certifies them because it never compares chunk language. Confirm that every retained, object-store-backed chunk has `course_language == requested_language` before comparing quality, rejecting null/legacy values, and add a regression where the unit row matches the request but its chunks do not.

In `iris/src/iris/pipeline/lecture_ingestion_pipeline.py`, quality-based retention checks the unit-row language instead of the retained page chunks, allowing previously inconsistent chunks to survive. Require every confirmed stored page chunk to match `requested_language` before retaining the generation, and add regression coverage where the ledger matches but chunk languages differ.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The new unit-row priority fixes the transcript overwrite, but forced-quality retention still checks the unit-row language rather than the stored chunks it retains. Existing rows can have a transcript-derived unit language matching the declared language while their page chunks carry a different independently detected language, so this head can retain and certify mismatched chunks. No required check is pending or failed in the snapshot; several non-required checks are still running.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
Nayer-kotry: Good catch — you're right that the ledger's own course_language wasn't a trustworthy signal here, for exactly the reason you name: it could have been stamped by the same transcript-overwrite bug this session already fixed elsewhere, so a stale value could coincidentally match `requested_language` while the actual retained chunks don't.

Fixed in e63dbcb35: `_previous_generation_scores_better` now calls a new `_stored_chunks_match_language`, which scans the unit's stored page chunks directly (not the unit row) and requires every object-store-confirmed chunk to carry `requested_language`. A capped scan, zero rows, or any unconfirmed (ghost) chunk fails closed to "replace" — the same safe default this pipeline's other structural checks already use.

Added two tests: one matching your exact scenario (a stored unit row that would coincidentally look fine, but stored chunks in "de" while "en" is requested — replacement proceeds), and one for a ghost-only stored chunk scan (also forces replacement).

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
```diff
@@ -263,27 +337,229 @@ def __call__(self) -> (str, []):
...
+
+        Only permitted when the stored *chunks* -- the content retention would
+        actually keep -- are themselves in the currently requested language.
+        The unit row's own course_language stamp is not a trustworthy proxy for
+        this: a run before this check existed could have stamped the unit row
+        with the transcript's language while the page chunks kept their own,
+        independently-resolved language, so a ledger-only comparison can find a
+        coincidental match and retain chunks that don't actually agree with it.
+        """
+        if not self._stored_chunks_match_language(requested_language):
```
Claudia-Anthropica: @Nayer-kotry [medium] This gate proves only that every scanned row is live and uses the requested language; it does not require one ingestion run, the current attachment version, complete page coverage, or the stored manifest counts. After a crash between writing and purging, two confirmed same-language generations satisfy this check; when the stored score is higher, the forced retry keeps them, and the audit then rejects the multiple generations on every subsequent retry. Require retained chunks to pass the same structural completeness checks as the normal skip path before comparing quality scores; otherwise proceed with the prepared replacement.

In `iris/src/iris/pipeline/lecture_ingestion_pipeline.py`, forced quality retention accepts live same-language chunks without verifying that they form one complete, current generation, which can preserve a crash-damaged set that the audit repeatedly rejects. Require the stored chunks to pass the normal structural completeness criteria before retaining them, and add a mixed-generation forced-retry regression test.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The direct chunk-language check resolves the latest mismatch, and the captured Iris test, lint, and build checks succeeded. One medium correctness issue remains: forced quality retention can preserve a structurally incomplete or mixed-generation chunk set, leaving the final audit to reject the same state on every retry.

iris/src/iris/pipeline/lecture_ingestion_pipeline.py
Nayer-kotry: Confirmed and fixed in ae033ff4d.

Verified the gap exactly as described: `_previous_generation_scores_better` only checked that every object-store-confirmed stored chunk was live and in the requested language, which is a strict subset of what the normal skip path (`check_if_attachment_needs_update`) requires before it trusts stored chunks — current attachment version, single ingestion generation (one `ingestion_run_id`), complete page coverage, and a chunk-count match to the recorded manifest, in addition to language and confirmation. A crash between writing a new generation and purging the old one can leave two confirmed, same-language generations coexisting; the language-only check would see all rows match and, if the stored ledger score were higher, keep that mix — which the post-write audit would then reject on every subsequent retry with no way to self-heal.

Fix: `_previous_generation_scores_better` now calls `check_if_attachment_needs_update(page_count, requested_language)` first (threading `page_count` through from the already-open PDF doc) and only proceeds to the score comparison when that says the stored chunks are structurally complete and current. This reuses the exact same tested check the normal skip path relies on, rather than duplicating a narrower version of it — so `_stored_chunks_match_language` is removed entirely as now-redundant.

Added `test_quality_reingest_replaces_a_crash_mixed_generation_even_with_a_higher_score`, which reproduces the described scenario directly: three confirmed same-language chunks split across two `ingestion_run_id`s, with a higher stored quality score. Confirmed it fails before the fix (kept the mix) and passes after (forces replacement). Also rebuilt the three existing quality-reingest tests' stored-chunk fixtures to be structurally complete aside from the one signal each test isolates (language mismatch, ghost/unconfirmed row), since they now exercise the full structural check rather than a language-only one.

Full suite: 734 passed. black/isort/pylint (10.00/10) clean.

iris/src/iris/config.py
```diff
@@ -151,6 +151,89 @@ class TranscriptionSettings(BaseModel):
...
+    upstream configuration here and Iris keeps no standing knowledge of its
+    callers. It claims lecture ingestion jobs from every discovered upstream
+    when it has free capacity and renews a lease for every run it executes on
+    the fixed heartbeat interval. capacity is shared across all upstreams and
+    replaces Artemis's global max-concurrent-jobs as the effective
+    parallelism: in pull mode this process only ever takes what it can run.
+    """
+
+    enabled: bool = Field(default=True)
+    capacity: int = Field(default=2)
```
Claudia-Anthropica: @Nayer-kotry [medium] These worker controls accept zero and negative values. `capacity=0` continues idle heartbeats—keeping Artemis in pull mode—while never claiming work, and non-positive polling or heartbeat intervals make `Event.wait()` return immediately, creating a tight request loop. Require `capacity >= 1` and both intervals to be strictly positive, with validation coverage.

In `iris/src/iris/config.py`, `IngestionWorkerSettings` accepts non-positive capacity and intervals that can stall ingestion or create busy request loops. Add Pydantic bounds requiring positive values and add regression tests for rejected zero and negative settings.

</details>

iris/src/iris/ingestion/worker.py
```diff
@@ -0,0 +1,386 @@
...
+        if upstream.heartbeat_failures > 0:
+            logger.info(
+                "Worker heartbeat to %s recovered after %d failures",
+                upstream.url,
+                upstream.heartbeat_failures,
+            )
+        upstream.heartbeat_failures = 0
+        revoked = (response.json() or {}).get("revokedJobTokens") or []
+        for token in revoked:
+            # The pipeline thread cannot be killed safely; it is left to finish,
```
Claudia-Anthropica: @Nayer-kotry [medium] A revoked token is only logged; the run's cancellation event is neither retained nor signaled. Until another claim for the same unit arrives, `current_job_guard` still considers this thread current, so it can continue destructive vector writes and occupy capacity after Artemis says the lease is no longer owned. Store the cancellation event in `_ActiveRun` and set it on revocation, while retaining the run until its thread exits.

In `iris/src/iris/ingestion/worker.py`, revoked jobs continue running because `_ActiveRun` does not retain or signal their cancellation event. Store each run's cancellation event and set it for every revoked token so existing cancellation checkpoints stop the thread safely, while tracking it until termination.

</details>

iris/src/iris/pipeline/lecture_unit_segment_summary_pipeline.py
```diff
@@ -152,33 +261,79 @@ def _get_slides(self, slide_number: int):
         ).objects
 
     def _get_slide_range(self) -> Tuple[int, int]:
-        slides = self.lecture_unit_page_chunk_collection.query.fetch_objects(
-            filters=self._get_lecture_slide_filter()
-        ).objects
+        """Full page-number span of the unit, over every ghost-free chunk of every generation.
```
Claudia-Anthropica, line 281: @Nayer-kotry [medium] Only the page-range scan is ghost-filtered. `_get_slides()` and `_get_transcriptions()` still return raw scan-visible rows, so an undeletable ghost on an otherwise valid page is concatenated into the regenerated summary; the final audit then confirms the newly written segment row and certifies content derived from stale, non-retrievable data. Confirm the per-slide rows before building each prompt and cover a real-plus-ghost sibling regression.

In `iris/src/iris/pipeline/lecture_unit_segment_summary_pipeline.py`, the confirmed span is followed by raw per-slide reads that can feed object-store ghosts into persisted summaries. Filter both slide and transcription rows through `confirmed_rows` before summary generation and add regression coverage proving ghost content is excluded.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The latest structural-retention change resolves the prior mixed-generation retry loop by reusing the full attachment completeness check. Three medium blockers remain around worker configuration, revoked-run cancellation, and ghost rows entering regenerated segment summaries. No required check was pending or failed in the snapshot, although several non-required checks were still running.

iris/src/iris/config.py
Nayer-kotry: Confirmed and fixed in a0c0498fb.

Verified both mechanisms: `capacity=0` makes `_free_slots()` return `max(0, 0 - len(self._active))` which is always 0, so the claim loop never claims work, while the poll and heartbeat loops (which don't check capacity) keep running unconditionally — matching "idle heartbeats... keeping Artemis in pull mode." And `poll_interval_seconds`/`heartbeat_interval_seconds` feed directly into `self._stop.wait(seconds)`; `threading.Event.wait()` with a non-positive timeout returns immediately, confirmed via the stdlib docs.

Fix: `capacity` now requires `ge=1`, and both intervals require `gt=0`, matching the `Field(ge=1, ...)` convention `LectureIngestionSettings` already uses for its own bounds. Added `test_ingestion_worker_settings.py` with regression coverage for each rejected value plus the valid defaults.

iris/src/iris/ingestion/worker.py
Nayer-kotry: Confirmed and fixed in a0c0498fb.

Verified the gap: `_ActiveRun` had no field for the cancel_event `_start_job` created, and the revoked-token handler only logged. The pipeline already has cooperative-cancellation checkpoints (`current_job_guard` / `raise_if_cancelled`) that `add_job` signals for same-worker supersession, but nothing was wired to signal them on an Artemis-side revocation, so a revoked run really would keep running to natural completion — and if it finished after whichever run now owns the unit, its own write-then-purge could clobber that newer run's content.

Fix: `_ActiveRun` now retains `cancel_event`, set from `_start_job`; the heartbeat loop's revoked-token handling calls `run.cancel_event.set()`. This reuses the exact same signal the pipeline already checks at every stage, so a revoked run stops at its next checkpoint instead of running to completion. The run stays in `_active` (still counted against capacity) until its thread actually exits, per your note — no change needed there since `_prune_finished` already only drops runs whose thread is no longer alive.

Added two tests: one confirming `cancel_event.set()` is called on a revoked token with a matching active run, one confirming a revocation for an unknown/already-finished token doesn't crash.

iris/src/iris/pipeline/lecture_unit_segment_summary_pipeline.py
Nayer-kotry: Confirmed and fixed in a0c0498fb.

Verified exactly as described: `_get_slide_range` already confirms its scan against the object store via `_confirmed_page_number_span`, but `_get_slides`/`_get_transcriptions` — the per-slide reads that actually build each summary's prompt (`_create_summary` concatenates every row in `slides`/`transcriptions`) — returned the raw `fetch_objects(...).objects` scan, unfiltered. A ghost sibling for a real slide's page number would be concatenated straight into the summary text, and since the write of the summary row itself succeeds and is confirmed, the manifest-based audit has nothing to catch.

Fix: both methods now filter their scan through `confirmed_rows` before returning, the same call `_confirmed_page_number_span` already uses in this file. Added `test_lecture_unit_segment_summary_pipeline_slides.py` covering a real-plus-ghost sibling for both `_get_slides` and `_get_transcriptions`, plus the ghost-only case.

Claudia-Anthropica (approved):
@Nayer-kotry The current head resolves the remaining worker-setting, revoked-run cancellation, and ghost-row summary findings with targeted regression coverage. The captured Iris tests, lint, builds, and other checks all succeeded, and no unresolved blocking issue remains.

PR #952, opened 2026-09-13
`Iris`: Document the instructional support level for instructors

thynguyentumde:
### Summary

The instructor documentation covers enabling Iris, custom instructions, variants and rate limits, but not the **instructional support level** — the per-course setting that decides how much Iris gives away when it helps a student. It is arguably the most consequential pedagogical choice an instructor makes about Iris, and it was undocumented.

This adds a page for it. Documentation only: no code changes.

### Motivation

Two gaps made this worth writing:

- The support level is not mentioned anywhere in `docs/instructor/`. An instructor configuring Iris has no reference for what Low, Moderate and High actually do, or which to pick.
- It is easy to confuse with custom instructions. They solve different problems — the level controls *how much* is given away, custom instructions control *what* is emphasised and in what style — so the page states the distinction explicitly and warns against using custom instructions to restate the level.

### Description

Adds `docs/instructor/support-level.md`, covering:

- **The three levels**, each with the example conversation the settings page shows and its "best suited for" guidance. All three use the same student question, so the levels can be read against each other.
- **Choosing a level**, framed around what the exercise is meant to teach rather than how capable the students are.
- **What the level does not change** — the global restriction on complete solutions and full code blocks that applies regardless of the setting, and the fact that the level is per course and cannot be carved out per exercise.
- **Support level vs. custom instructions**, as a short table plus the rule of thumb.

The level descriptions, example conversations, "best suited for" text and the global-restriction wording are taken from the current Artemis UI strings rather than paraphrased, so the page matches what an instructor actually sees. `MODERATE` as the default reflects `IrisSupportLevel` on the Artemis side, including its fallback for absent or unrecognised values.

Also adds the page to `instructorSidebar` (directly after Enabling Iris, since it is part of initial setup) and links it from the **Next Steps** of `enabling-iris.md` and `custom-instructions.md`.

### Testing

1. `cd iris/docs && npm install && npm run build` — succeeds. Docusaurus fails the build on broken internal links, so this also verifies the cross-links resolve.
2. `npm run serve`, then open `/docs/instructor/support-level`. The page renders, with **Instructional Support Level** in the instructor sidebar under Enabling Iris.
3. From **Enabling Iris** and **Custom Instructions**, the Next Steps links reach the new page.

### Note for reviewers

While writing this I noticed that `lecture-ingestion.md` documents **"Send All Lectures to Iris"** and **"Send Unit to Iris"** buttons. Those no longer exist in Artemis — neither the UI code nor the translation strings are present, and lecture content is now processed automatically on publish. FAQ ingestion is still manual and still accurate. That is out of scope here, but worth a separate issue.

https://claude.ai/code/session_017uwtuGHLdzh1WSnjGASmp7

## Summary by CodeRabbit

- **Documentation**
  - Added guidance on the Instructional Support Level, including Low, Moderate, and High settings.
  - Documented how to configure each level, choose an appropriate setting, and understand its effect on student assistance.
  - Clarified limits that apply at every level and how support levels differ from custom instructions.
  - Added navigation links and sidebar access to the new documentation page.


iris/docs/docs/instructor/support-level.md
```diff
@@ -0,0 +1,88 @@
...
+
+**Best suited for** courses where students have some experience but benefit from concrete nudges and contextual pointers.
+
+### High
+
+Iris provides step-by-step explanations with analogies and small code examples to illustrate concepts.
+
+> **Student:** My sorting algorithm is not working. Can you help?
+>
+> **Iris:** The issue is in your inner loop condition `j <= arr.length`. Arrays are zero-indexed, so the last element is at `arr.length - 1`. Try changing it to `j < arr.length` and see if that resolves the IndexOutOfBoundsException.
```
Claudia-Anthropica: @thynguyentumde [medium] This example promises that High support identifies the exact faulty condition and supplies its replacement. The current programming prompt explicitly says High may point to the specific area or concept but must “NOT reveal the fix,” while the global rules forbid exercise-specific fixes at every level, so this documents behavior Iris is instructed not to produce.

In `iris/docs/docs/instructor/support-level.md`, the High example claims Iris supplies an exact exercise-specific fix even though the implementation forbids revealing fixes at every support level. Replace this example and the matching summary on line 51 with a High response that gives detailed conceptual guidance and identifies the relevant area without supplying the corrected condition.

</details>

Claudia-Anthropica (changes requested):
@thynguyentumde The sidebar and cross-links are wired correctly, and the Moderate default matches `PipelineExecutionSettingsDTO`. However, the High-level example contradicts the current Iris prompt contract and would mislead instructors about the assistance students receive.

Claudia-Anthropica (approved):
@thynguyentumde The revised High-support example now matches the current prompt contract: it localizes the relevant area, explains the concept with an unrelated example, and withholds the corrected exercise condition. The Moderate default, sidebar entry, and cross-links also match the implementation and documentation structure. The captured snapshot designated no required checks, although several non-required checks were still running.

PR #953, opened 2026-09-14
`Logos`: Improve calibration

julia-rai:
## Summary by CodeRabbit

* **New Features**
  * Added the ability to stop active calibration sessions from the worker GPU panel.
  * Added Metal (Apple Silicon) calibration support.
  * Calibration now supports generative, embedding, and transcription models.

* **Improvements**
  * Calibration schedules now default to 03:00–08:00.
  * Calibration prioritizes idle GPUs to reduce disruption to running workloads.
  * Improved handling of memory limits, retries, and calibration failures.
  * Improved vLLM executable discovery across supported environments.

* **Bug Fixes**
  * Improved model detection, probe selection, calibration feedback, and metadata cache handling.


logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -1803,21 +1844,36 @@ def _establish_host_ram_floor_for_probe() -> bool:
...
-                        ),
-                    )
+                    if is_metal_backend():
+                        # No TP escalation, no KV sweep, no sleep/wake — see
+                        # calibration_metal's module docstring for why.
+                        result = await loop.run_in_executor(
+                            None,
+                            lambda p=plan: calibrate_model_metal(
+                                p,
+                                vllm_binary=_DEFAULT_VLLM,
```
Claudia-Anthropica, line 1914: @julia-rai [high] This bypasses the supported Metal binary and launch configuration. In the documented macOS installation the worker runs from its own `.venv`, vLLM lives under `LOGOS_METAL_VENV/bin`, and the launchd environment does not add that directory to `PATH`; passing `_DEFAULT_VLLM` to a direct `Popen` therefore makes normal Metal calibration fail with `FileNotFoundError`. Even where `vllm` happens to be on `PATH`, the probe omits `cfg.engines.metal` environment settings and several production lane flags, so the resulting profile measures a different process configuration.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, Metal calibration bypasses the production Metal binary resolver, command builder, and environment configuration, causing supported deployments to fail or produce mismatched profiles. Launch calibration through the same Metal binary, command, and environment resolution used by production lanes.

</details>

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -963,6 +963,7 @@ async def _run_hf_compatibility_precheck(
                 "fit_tp_idle": None,
                 "fit_tp_current": None,
                 "unsupported_reason": None,
+                "model_kind": "generative",
```
Claudia-Anthropica: @julia-rai [medium] Returning `generative` before fetching HF metadata means every Metal model loses the new model-kind classification. For example, the documented `--runner pooling` configuration is then passed to `calibrate_model_metal`, which calls the default `/v1/completions` warmup, accepts its failure as load-only success, and may persist a footprint before the real embedding request's lazy allocations. Keep model classification active on Metal while skipping only the CUDA-specific VRAM checks, and apply that kind to the Metal functional probe.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, the Metal precheck hard-codes every model as generative, causing pooling and transcription calibration to probe the wrong endpoint and accept an incomplete measurement. Classify Metal models before bypassing CUDA-only checks and pass the detected kind through the Metal warmup with the corresponding failure handling.

</details>

logos/logos-workernode/logos_worker_node/calibration.py
```diff
@@ -3388,6 +3540,13 @@ def _is_fatal(result: CalibrationResult) -> bool:
         # vllm_compat._FATAL_LOAD_ERROR_PATTERNS, not just two of them).
         return result.unsupported_reason is not None
 
+    def _is_capacity_exhausted(result: CalibrationResult) -> bool:
+        """True when weights/kv don't fit anywhere in the kv search range
+        at this tp — a real VRAM shortfall, not a config/arch quirk. A
+        lower tp needs *more* VRAM per GPU, not less, so it can't recover
+        from this (see ``min_feasible_tp``'s same math in the HF precheck)."""
+        return "exceed available GPU VRAM" in (result.error or "")
```
Claudia-Anthropica: @julia-rai [medium] This text is not evidence of a genuine capacity OOM. `_calibrate_model_probe` emits the same “likely exceed available GPU VRAM” message whenever no KV probe starts, including TP-specific configuration failures such as an attention-head count not divisible by the hardware-max TP; `_skip_capacity_fallback` then suppresses the existing lower-TP recovery that was intended for exactly that case. Carry an explicit capacity-OOM reason from `_capacity_oom_box` instead of inferring it from the generic terminal message.

In `logos/logos-workernode/logos_worker_node/calibration.py`, `_is_capacity_exhausted` treats a generic no-working-KV message as proof of OOM and incorrectly skips recovery at a compatible lower tensor-parallel size. Record an explicit CUDA-capacity-OOM signal on `CalibrationResult` and gate the fallback suppression only on that signal.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The exact head still has three blocking correctness gaps. Metal sessions bypass the supported vLLM resolver/runtime configuration and discard model-kind detection, while CUDA TP fallback now mistakes generic startup failures for proven VRAM exhaustion. Please address the inline findings before merging.

Claudia-Anthropica (changes requested):
@julia-rai The new head fixes the prior model-kind propagation and explicit CUDA capacity-OOM handling. Two [medium] Metal runtime-parity defects remain in the current unresolved threads: calibration drops each model's `env_overrides`, and its binary resolver fails where production still supports PATH, sibling, and module resolution, so supported configurations either measure the wrong environment or cannot calibrate. Captured test and lint checks passed, while the non-required vLLM worker image build was still in progress.

Claudia-Anthropica (changes requested):
@julia-rai The current head still has the two blocking Metal runtime-parity defects captured in unresolved threads: `_build_metal_calibration_env` omits the plan's per-model `env_overrides`, and calibration returns before using production's PATH, sibling, and module binary fallbacks. Calibration can therefore measure a different environment from the serving lane or reject deployments that production supports. Captured test and lint checks passed; several non-required image builds remained in progress.

Claudia-Anthropica (changes requested):
@julia-rai The new commit correctly routes stop-calibration through the JWT-protected Spring controller and secret-gated internal endpoint. Two existing [medium] Metal defects remain: calibration still drops each model's `env_overrides`, and its binary resolver still returns early instead of preserving production's generic fallbacks, so supported configurations can be measured incorrectly or rejected. The snapshot contains no required checks or captured failures, although several non-required checks remain unfinished.

logos/logos-workernode/logos_worker_node/calibration_metal.py
```diff
@@ -0,0 +1,327 @@
...
+        *vllm_prefix,
+        "serve",
+        model,
+        "--host",
+        host,
+        "--port",
+        str(port),
+        "--dtype",
+        str(plan.get("dtype") or "auto"),
+        "--max-model-len",
```
Claudia-Anthropica: @julia-rai [medium] This command still does not mirror a production Metal lane: it hard-codes `--max-model-len auto` and never emits `mm_processor_cache_gb`, although `plans_from_config` carries both and `MetalVllmProcessHandle._build_cmd` applies them. The shipped Metal configuration depends on `max_model_len: 32768` and `mm_processor_cache_gb: 0` because the full context does not fit and the default processor cache reserves 4 GB, so calibration can fail for a lane that starts or persist a residency measured under a different footprint.

In `logos/logos-workernode/logos_worker_node/calibration_metal.py`, `_build_metal_calibration_cmd` ignores the per-model `max_model_len` and `mm_processor_cache_gb` settings, so calibration does not match production memory configuration. Emit these flags with the same precedence and defaults as `MetalVllmProcessHandle._build_cmd` and add command-parity tests.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The prior Metal binary/environment and endpoint-response findings are fixed on this head. The retained CodeRabbit request-shape finding remains: the UI sends `provider_id` to `StopCalibrationRequestDTO.providerId` while MVC uses an unconfigured `ObjectMapper`, so the stop action is rejected before reaching the orchestrator. Metal calibration also still discards memory-shaping settings honored by production lanes; see the inline finding.

logos/logos-workernode/logos_worker_node/calibration_metal.py
```diff
@@ -0,0 +1,351 @@
...
+    # vllm_process.py): a model configured with mm_processor_cache_gb: 0
+    # (config.example.mlx.yml) frees the 4 GB the default reserves, so
+    # measuring under the default instead would record a residency the
+    # served lane never actually has room for.
+    mm_processor_cache_gb = plan.get("mm_processor_cache_gb")
+    if mm_processor_cache_gb is None:
+        mm_processor_cache_gb = _DEFAULT_MM_PROCESSOR_CACHE_GB
+    cmd.extend(["--mm-processor-cache-gb", str(mm_processor_cache_gb)])
+    extra_args = plan.get("extra_args") or []
+    cmd.extend(str(a) for a in extra_args)
```
Claudia-Anthropica, line 171: @julia-rai [high] `extra_args` are appended after the trusted loopback bind, so a configured `--host 0.0.0.0` wins during argument parsing and exposes the unauthenticated calibration vLLM server to the LAN. The production Metal command builder explicitly restores `--host` and `--port` after extras to prevent this escape.

In `logos/logos-workernode/logos_worker_node/calibration_metal.py`, per-model extra arguments can override the calibration server's loopback host and port. Re-append the trusted loopback `--host` and calibration `--port` after every extra argument and add an override regression test.

</details>

logos/logos-workernode/logos_worker_node/calibration.py
```diff
@@ -3211,6 +3356,27 @@ def calibration_gpu_slice(available_gpus: int) -> list[int]:
...
+    """
+    n = int(available_gpus) if available_gpus else 0
+    if n < 1:
+        return []
+    slice_size = 1 << (n.bit_length() - 1)
+    busy = {int(i) for i in busy_gpus}
+    idle = [i for i in range(n) if i not in busy]
+    if len(idle) >= slice_size:
+        return idle[:slice_size]
+    return list(range(slice_size))
```
Claudia-Anthropica: @julia-rai [medium] When there are some idle GPUs but fewer than `slice_size`, this fallback discards every idle GPU and selects the naive prefix. On a three-GPU node with GPUs 0 and 1 busy, it chooses `[0, 1]` and tears down two serving lanes even though `[2, 0]` would preserve one of them.

In `logos/logos-workernode/logos_worker_node/calibration.py`, the partial-idle fallback ignores available idle GPUs and can stop more serving lanes than necessary. Construct the calibration slice from all idle GPUs first and fill only the remaining slots from busy GPUs, with a partial-idle regression test.

</details>

logos/logos-workernode/logos_worker_node/calibration.py
```diff
@@ -2562,7 +2695,7 @@ def _cancelled() -> bool:
                 "  [2.5/6] Warming up engine (1-token completion, capturing CUDA graphs — may take a couple minutes)..."
             )
         warmup_t0 = time.perf_counter()
-        warmup_ok = warmup_inference(base_url, model, timeout_s=600.0)
+        warmup_ok = warmup_inference(base_url, model, timeout_s=600.0, model_kind=model_kind)
```
Claudia-Anthropica, line 2796: @julia-rai [medium] Once this functional probe enters the blocking HTTP request, `cancel_event` is not observed for up to 600 seconds; the post-wake probe has the same problem, and Metal can block for 120 seconds. The stop handler returns after waiting only 15 seconds, so the UI can report success while the calibration subprocess continues consuming the GPU for minutes.

In `logos/logos-workernode/logos_worker_node/calibration.py`, functional-probe HTTP waits do not observe calibration cancellation, allowing the subprocess to continue after a successful stop response. Propagate `cancel_event` through all warmup probes and make cancellation interrupt the HTTP wait and tear down the spawned vLLM process promptly, with a regression test.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The latest commit fixes the prior request-binding and Metal command-parity findings. The exact head still permits Metal calibration to escape its loopback bind, can evict more serving lanes than necessary when selecting GPUs, and cannot promptly interrupt long functional probes. All 49 captured CI checks passed, but these current code-path defects remain blocking.

logos/logos-workernode/logos_worker_node/calibration_metal.py
```diff
@@ -0,0 +1,394 @@
...
+
+
+def _is_metal_capacity_failure(returncode: int | None, log_tail: str) -> bool:
+    """True when a failed Metal probe looks like a memory-capacity failure.
+
+    Combines two independent signals (neither alone is reliable): the
+    process was killed by a signal consistent with an OS OOM kill, or the
+    log mentions a known memory-allocation failure marker.
+    """
+    if returncode is not None and returncode <= _METAL_OOM_SIGNAL_EXIT_CODE:
```
Claudia-Anthropica: @julia-rai [medium] Python reports signal termination as `-signal_number`, so `returncode <= -9` classifies SIGSEGV (`-11`), SIGTERM (`-15`), and other non-OOM crashes as capacity failures. That value is persisted and then used to skip this model on every provider with no greater capacity, allowing an unrelated crash to suppress future calibration cluster-wide.

In `logos/logos-workernode/logos_worker_node/calibration_metal.py`, `_is_metal_capacity_failure` treats every signal number at least as large as SIGKILL as an OOM and can persist false capacity floors. Accept only an explicit SIGKILL as signal-based capacity evidence, or require a verified memory marker for other signals, and add regression tests for SIGSEGV and SIGTERM.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The latest commits fix the prior Metal command-parity finding, but all three non-outdated Claudia threads remain visible on this exact head: `extra_args` still follow the loopback binding, partial-idle selection still returns the naive prefix, and warmup HTTP requests still do not observe `cancel_event`. The new capacity-floor code also treats several unrelated process signals as OOM evidence, which can persist a false cluster-wide skip; see the inline finding. The snapshot contains no required checks or captured failures, though several non-required builds remain unfinished.

logos/logos-orchestrator/src/logos/capacity/calibration_orchestrator.py
```diff
@@ -367,6 +371,38 @@ def _provider_has_active_requests(self, provider_id: int) -> bool:
...
+        compared against this provider's own ``devices.total_memory_mb``.
+        """
+        try:
+            snap = self._registry.peek_runtime_snapshot(provider_id)
+        except Exception:
+            return frozenset()
+        if not isinstance(snap, dict):
+            return frozenset()
+        devices = (snap.get("runtime") or {}).get("devices") or {}
+        provider_capacity_mb = float(devices.get("total_memory_mb") or 0.0)
```
Claudia-Anthropica, line 392: @julia-rai [medium] This method never verifies that the target provider's device mode is Metal. In a mixed cluster, a floor recorded by a Metal node is compared against a CUDA provider's `total_memory_mb`, causing that CUDA model to be omitted from both `_provider_has_uncalibrated_models` and the worker's `skip_models` even though a Metal working-set floor says nothing about CUDA VRAM requirements.

In `logos/logos-orchestrator/src/logos/capacity/calibration_orchestrator.py`, `_capacity_skip_models` applies Metal-specific capacity floors to non-Metal providers and can suppress valid CUDA calibration. Return an empty set unless the target provider reports `devices.mode == "metal"`, and add mixed Metal/CUDA regression coverage.

</details>

logos/logos-workernode/logos_worker_node/calibration_metal.py
```diff
@@ -0,0 +1,400 @@
...
+    proc: subprocess.Popen[str] | None = None
+    try:
+        proc = _spawn_vllm_metal(cmd, log_path, worker_metal_config, plan.get("env_overrides"))
+        wait_ready(base_url, ready_timeout_s, proc, cancel_event=cancel_event)
+
+        if cancel_event is not None and cancel_event.is_set():
+            result.error = "cancelled"
+            return result
+
+        served = warmup_inference(base_url, model, model_kind=model_kind)
```
Claudia-Anthropica, line 390: @julia-rai [medium] A failed generative warmup is accepted as a load-only success without checking `proc.poll()`. If lazy allocation OOMs or otherwise crashes vLLM after readiness, the HTTP helper converts the disconnect to `False`, after which this path samples post-exit memory and sets `result.success = True`; it also bypasses the new capacity-floor classifier.

In `logos/logos-workernode/logos_worker_node/calibration_metal.py`, a vLLM process that exits during a generative warmup can be recorded as a successful load-only calibration. Before accepting a nonfatal warmup failure, detect an exited subprocess, return a failed result through the existing return-code/log capacity classification, and add a crash-during-warmup regression test.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The SIGKILL and loopback fixes are correct, but two unresolved Claudia threads remain current: partial-idle selection still chooses the naive prefix, and warmup HTTP calls still ignore cancellation. The new Metal capacity floor is also applied to CUDA providers in mixed clusters, while a vLLM process that dies during a generative warmup can still be persisted as a successful load-only profile; see the inline findings. Captured CI has no failures or required checks, although one non-required build was still in progress.

logos/logos-workernode/logos_worker_node/calibration.py
Claudia-Anthropica: @julia-rai This remains unresolved on the current head: line 3381 still returns the naive prefix whenever there are too few idle GPUs, so with busy GPUs `{0, 1}` it discards idle GPU `2` and selects two busy GPUs. Build the slice from idle GPUs first and fill only the remaining slots from busy GPUs.

logos/logos-workernode/logos_worker_node/calibration.py
Claudia-Anthropica: @julia-rai This remains unresolved on the current head: the 600-second warmup calls still use blocking `urllib` helpers without receiving `cancel_event`, while the stop handler waits only 15 seconds. Propagate cancellation through each functional probe so stopping promptly interrupts the request and tears down vLLM.

logos/logos-workernode/logos_worker_node/hf_model_info.py
```diff
@@ -43,6 +52,64 @@
...
+# second-guessed by an incidental architecture-name substring (e.g. a
+# CausalLM wrapper class
+# that happens to contain "Embedding" in its name).
+_PIPELINE_TAG_KIND: dict[str, str] = {
+    "text-generation": "generative",
+    "text2text-generation": "generative",
+    "image-text-to-text": "generative",
+    "feature-extraction": "pooling",
+    "sentence-similarity": "pooling",
+    "text-classification": "pooling",
```
Claudia-Anthropica: @julia-rai [medium] `text-classification`, `text-ranking`, sequence-classification, and reranker models are all collapsed into `pooling`, but `warmup_inference` sends every pooling model to `/v1/embeddings` and treats a non-200 response as fatal. These vLLM models normally serve `/classify`, `/score`, or `/rerank`—paths Logos explicitly proxies—so a healthy classifier or Qwen3 reranker can be marked uncalibratable.

In `logos/logos-workernode/logos_worker_node/hf_model_info.py`, classification and ranking models are classified as generic pooling models and consequently probed through the incompatible `/v1/embeddings` endpoint. Preserve the task distinction and route each detected kind through an endpoint-appropriate functional probe with the matching request body.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The latest commit fixes the CUDA/Metal capacity-floor and idle-GPU findings, and all 49 captured checks passed. One prior [medium] defect remains on this head: the CUDA and Metal warmup requests still block for up to 600 and 120 seconds without observing `cancel_event`, although the stop RPC returns after 15 seconds. The model-kind classifier also routes classifier and reranker models through `/v1/embeddings` and can fatally reject healthy lanes; see the inline finding.

Claudia-Anthropica (changes requested):
@julia-rai The exact head still contains both blocking defects from the previous review. `warmup_inference` uses blocking `urllib` calls without `cancel_event`, so CUDA and Metal probes can continue for up to 600 and 120 seconds after the stop RPC returns; classification and ranking models are also still collapsed into the embeddings-only `pooling` probe, rejecting healthy classifier and reranker lanes. The snapshot reports no required checks or failures, although several non-required checks were unfinished.

logos/logos-workernode/logos_worker_node/hf_model_info.py
Claudia-Anthropica: @julia-rai This remains unresolved on the exact head: `_PIPELINE_TAG_KIND` and `_POOLING_ARCH_MARKERS` still collapse classification and ranking models into `pooling`, while the `pooling` probe only calls `/v1/embeddings`. Preserve separate model kinds and probe their actual serving endpoints.

logos/logos-workernode/logos_worker_node/calibration.py
```diff
@@ -1022,6 +1049,92 @@ def warmup_inference(base_url: str, model: str, timeout_s: float = 120.0) -> boo
...
+    body = {"model": model, "input": "hi"}
+    status, _ = _post(f"{base_url}/v1/embeddings", body=body, timeout_s=timeout_s)
+    return status == 200
+
+
+def probe_classification(base_url: str, model: str, timeout_s: float) -> bool:
+    """One ``/classify`` request — sequence-classification models' real
+    serving path (vLLM never serves these from ``/v1/embeddings``)."""
+    body = {"model": model, "input": "hi"}
+    status, _ = _post(f"{base_url}/classify", body=body, timeout_s=timeout_s)
```
Claudia-Anthropica, line 1063: @julia-rai [medium] This validates classifiers against vLLM's unprefixed `/classify` endpoint, but the Logos user-facing router only registers `/pooling`, `/score`, `/rerank`, `/tokenize`, and `/detokenize`. A classifier can therefore pass calibration and be advertised while clients cannot send the same request through Logos: bare `/classify` returns 404, and `/v1/classify` forwards the wrong upstream path.

In `logos/logos-workernode/logos_worker_node/calibration.py`, the new classification probe validates an unprefixed `/classify` endpoint that Logos does not expose to clients. Add `/classify` to the native vLLM route list in `logos/logos-orchestrator/src/logos/routers/user_facing.py` and add a forwarding route test.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The new head correctly separates classifier and reranker probes, but two blocking gaps remain. CUDA and Metal warmups still enter blocking `urllib` calls for up to 600/120 seconds without observing `cancel_event`, while the stop RPC returns after 15 seconds. Classification is also calibrated through `/classify` even though Logos exposes no corresponding user-facing route; the snapshot captured no failed or required checks, although several non-required checks were unfinished.

logos/logos-workernode/logos_worker_node/calibration_metal.py
```diff
@@ -0,0 +1,433 @@
...
```
Claudia-Anthropica, line 390: @julia-rai [medium] The new `_resolve_probed_model_kind` fallback is only called by the CUDA probe. Metal uses the same HF-derived kinds, but this path sends `reranking` directly to `/rerank` and treats a missing route as fatal, so a Qwen3-style checkpoint that vLLM actually serves generatively remains uncalibratable on Metal even though CUDA now handles it. Resolve the live endpoint before this warmup and add a Metal regression for a missing task route.

In `logos/logos-workernode/logos_worker_node/calibration_metal.py`, Metal calibration does not apply `_resolve_probed_model_kind`, so HF task/served-route mismatches still fail instead of using the confirmed generative route. Resolve the model kind against live OpenAPI before the Metal warmup and add a regression test for a missing rerank or classify route.

</details>

Claudia-Anthropica (changes requested):
@julia-rai The new head fixes the open `/classify` routing thread, which can now be resolved. Two blocking gaps remain: CUDA and Metal warmups still block for up to 600/120 seconds without observing `cancel_event`, so the stop RPC can report success while the probe keeps consuming resources; additionally, the new live-endpoint fallback is applied only to CUDA, leaving Metal vulnerable to the same HF-task/served-route mismatch. Captured CI has no failed or required checks; the non-required Read the Docs check was pending.

logos/logos-workernode/logos_worker_node/calibration.py
Claudia-Anthropica: @julia-rai Thanks, `/classify` is now registered as a native vLLM route and covered by the conformance test, so this is fixed.

Claudia-Anthropica (changes requested):
@julia-rai The merge-only head fixes the `/classify` route, so that thread can be resolved, but two blocking calibration defects remain. Functional probes still use uncancellable `urllib` calls, allowing CUDA and Metal processes to remain active for up to 600 and 120 seconds after the stop RPC's 15-second wait; Metal also passes the HF-derived model kind directly to the fatal warmup without CUDA's live-endpoint fallback, so task/route mismatches remain uncalibratable. The snapshot contains no required checks or captured failures, although several non-required checks are unfinished.

Claudia-Anthropica (changes requested):
@julia-rai The Metal live-endpoint fallback is fixed on this head. [medium] The CUDA warmups at `calibration.py:2787`/`:3018` and Metal warmup at `calibration_metal.py:383` still enter blocking `urllib` calls without `cancel_event`, while `_handle_stop_calibration_session` returns after 15 seconds; propagate cancellation through every functional probe and promptly terminate the spawned vLLM process. The non-required `Test` check is red, but the snapshot contains only a generic exit code and no test report, so I could not attribute that failure to this change.

Claudia-Anthropica (changes requested):
@julia-rai The exact head still contains the prior [medium] cancellation defect despite the thread being marked resolved: CUDA warmups at `calibration.py:2787` and `:3018` and the Metal warmup at `calibration_metal.py:383` still call synchronous `urllib` probes without `cancel_event`. `_handle_stop_calibration_session` returns after 15 seconds, but these calls can hold the calibration subprocess for their 600/120-second timeouts before the `finally` teardown runs; propagate cancellation through every HTTP probe and terminate vLLM promptly. All 49 captured CI checks passed.

PR #964, opened 2026-09-14
`Logos`: Replay assistant history as output_text and report the registry context window

### LINKED ISSUE #962: `Logos`: Luna Integration error report
2 errors emerge for the Luna integration with Claude Code:
 
1. context  : 111,200 tokens (an estimate — Logos reports no size for this model). Image

2. First request usually gets a quick response, second request receives:
API Error: 400 Invalid value: 'input_text'. Supported values are: 'output_text' and 'refusal'.
Image

wasnertobias:
Fixes #962.

The Luna integration (Claude Code → Logos → Azure, Responses dialect) failed in
two ways; both are addressed here.

## Second request 400s: `Invalid value: 'input_text'`

The Messages→Responses translation emitted `input_text` for the content parts
of *every* message, replayed assistant history included. The Responses API
validates assistant content against `output_text` and `refusal` only, so the
first request of a conversation succeeded and the second — the one carrying
the previous answer — was rejected with

> 400 Invalid value: 'input_text'. Supported values are: 'output_text' and 'refusal'.

The `id`/`status` that `output_text` is associated with belong to the wrapping
message item, not to the part, so the part needs nothing the translation cannot
supply. `responses_api.py` now picks the part type by role: `input_text` for
user text, `output_text` for replayed assistant text.

## "Context: 111,200 tokens (an estimate — Logos reports no size for this model)"

The startup line comes from the claude-logos wrapper: it sizes the session from
`GET /v1/models` and falls back to a blind constant when the model carries no
`max_model_len*` field. A model served by a cloud upstream that publishes no
window of its own (Azure included) had no window from any source, so it always
took the fallback.

The upstream registry the webservice already refreshes the model capabilities
from carries `max_input_tokens` for exactly those models, so this hands that
number through to the model endpoints:

- **webservice** — `ModelCapabilitiesUpdaterService` reduces the window across
  all matching registry entries to the smallest (a request may land on any of
  them) and stores it in `model_capabilities.max_input_tokens` (new nullable
  column, Liquibase 034); a refresh that drops the window clears it again.
- **orchestrator** — `DBManager.get_catalog_context_by_model()` +
  `_catalog_context_by_model()` (TTL-cached like the other context lookups,
  fails open) fold the value into `_served_context_window_stats()` **only for
  models no measured source knows**: a provider serves a catalog model at its
  full published size, so the single number is `current_min`, `current_max`
  and `overall` alike, and a local lane running the model narrower is never
  advertised wider.

After this, the wrapper's startup line for an Azure-served reasoning model
shows the published window (1,050,000 tokens for the gpt-5.6 family) instead of
the estimate. No wrapper revision is needed — it already reads the fields this
now fills.

## Tests

- `tests/unit/anthropic_compat/test_responses_api.py` — assistant history is
  now asserted as `output_text` (including the tool-call interleave case)
- `tests/unit/main/test_v1_models.py` — catalog window published for a
  cloud-only model; measured figures (live lane, cloud self-report) win over
  the catalog; a catalog entry never surfaces for a model outside the listing
- `ModelCapabilitiesUpdaterServiceTest` (new) — smallest window across
  matching registry entries, models unknown to the catalog untouched, a
  missing window stores NULL
- `LiquibaseBaselineTest` / `LiquibaseChangelogOrderTest` — the full changelog
  applies to a fresh Postgres and the includes stay strictly increasing

All orchestrator unit tests pass (1837 passed; one pre-existing broken test
module fails identically on main), all webservice tests pass (450).

## Summary by CodeRabbit

* **New Features**
  * Model context-window information is now populated from the model catalog when live or historical measurements are unavailable.
  * The smallest valid catalog-reported context window is used, with measured and provider-reported values taking precedence.
  * Catalog context-window data refreshes daily.

* **Bug Fixes**
  * Assistant conversation history is now translated with the correct content type, preventing request failures during multi-turn interactions.

* **Documentation**
  * Updated guidance explains context-window sources and troubleshooting for missing model limits.


logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
```diff
@@ -1380,6 +1380,40 @@ def get_cloud_context_by_model(self) -> Dict[str, Dict[str, int]]:
...
+        know (or that it lists without a window) is absent and callers treat
+        it as unknown rather than zero.
+        """
+        rows = self.session.execute(
+            text(
+                """
+                SELECT m.name, c.max_input_tokens
+                FROM model_capabilities c
+                JOIN models m ON m.id = c.model_id
+                WHERE c.max_input_tokens IS NOT NULL AND c.max_input_tokens > 0
```
Claudia-Anthropica: @wasnertobias [medium] This query returns catalog windows for every model with a capabilities row, including models connected only to `logosnode` providers. `_served_context_window_stats()` turns each result into `current_min`, `current_max`, and `overall`, so an offline or never-calibrated local model is reported by `/v1/models` and `/internal/model_context_windows` as currently serving the registry limit, and warmup returns `status: serving` despite there being no live lane. Restrict this fallback to models with a cloud `model_provider` association, or otherwise retain provider/source information and avoid synthesizing current fields for local-only models.

In `logos/logos-orchestrator/src/logos/dbutils/dbmanager.py`, `get_catalog_context_by_model()` exposes catalog windows for local-only models, causing theoretical registry limits to become current served windows. Restrict the query or fallback to cloud-backed model-provider associations and keep local-only catalog values out of `current_min` and `current_max`.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The Responses history change and catalog persistence are internally consistent, and the captured CI snapshot has no failures, though several non-required jobs remain pending. One blocking correctness issue remains: the catalog fallback is not scoped to cloud-backed models, so offline local-only models can be advertised as currently serving a theoretical registry window.

logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
wasnertobias: Fixed in 7c827f255 — the query is now scoped exactly as suggested: it joins through `model_provider` to `providers` and only returns rows for models associated with a `provider_type = 'cloud'` provider, so a local-only model (logosnode-only association) no longer gets a synthesized `current_min`/`current_max`/`overall` from the registry.

The reasoning in the docstring: the catalog value stands in for what a provider *serves* — a cloud upstream serves a catalog model at its published size (and is always reachable), while a local model's window is a property of its calibrated lane. With no lane up, there is nothing being served, so no `current_*` field may be claimed for it. A hybrid model (cloud + local associations) still gets the figure when nothing measured is known, because its cloud deployment does serve at that size — and the fold in `_served_context_window_stats()` is untouched: measured sources still take precedence over the catalog.

Added unit tests: the per-model reduction to the smallest published window, the positive-value filter, and the cloud scoping of the query itself (source-level assertion, following the pattern this file already uses for the `upsert_model_profiles` conflict clause). `context-windows.md` documents the scoping as well.

PR #965, opened 2026-09-14
`Logos`: Forward each lane's native vLLM metrics through one export endpoint

### LINKED ISSUE #934: `Logos`: Support vLLM Exporters
vLLM does support Prometheus /metrics endpoint which would exist per Lane of a workernode in Logos
https://docs.vllm.ai/en/v0.7.2/getting_started/examples/prometheus_grafana.html

Right now scraping /metrics is not supported as the metrics endpoint is not statically reachable. We could add a funcitonality in the orchestrator which does forward those information for every lane in one export endpoint.

bensofficial:
## Summary

Closes #934.

vLLM exposes a native Prometheus `/metrics` endpoint per lane, but it binds to a dynamically allocated `127.0.0.1` port on the worker host — never reachable by Prometheus, or even by the orchestrator, directly. This PR forwards that data through the infrastructure that already exists, without opening any new network path:

- **Worker node**: merges every running lane's raw `/metrics` text (relabeled with `lane_id`/`model` so replicas of the same model never collide into one ambiguous series), and pushes the merged blob to the orchestrator over the existing outbound bridge websocket, on its own cadence (`vllm_metrics_interval_seconds`, default 15s) — deliberately separate from the runtime-status loop, since vLLM's counters change every tick and would otherwise defeat that loop's dedupe signature.
- **Orchestrator**: stores each worker's latest pushed snapshot on its session (naturally forgotten on disconnect), relabels it with `worker_id`, and merges all connected workers' snapshots into its own existing `/metrics` registry via a custom Prometheus collector. Prometheus keeps scraping the one endpoint it already does — no new target, no service discovery.

## Design notes

- Metric merging (HELP/TYPE dedup per name + per-source relabeling) lives in a small pure function (`vllm_metrics_merge.py`), duplicated once on each side rather than introduced as a new shared dependency, matching this repo's existing convention of independent `prometheus_metrics.py` modules per service.
- The collector reads only cached, already-pushed data at scrape time — no live fan-out HTTP calls, so one slow/offline worker just means a stale or missing snapshot for it, never a slow or failed scrape.
- Along the way, fixed a real bug the new collector's tests surfaced: `import logos.main as _main` is unreliable when called repeatedly at arbitrary times (as a `collect()` method is), because `logos/__init__.py` intentionally replaces `sys.modules["logos"]` with `logos.main` itself once fully initialized — the collector now does a direct `sys.modules.get("logos.main")` lookup instead.

## Test plan

- [x] `logos-workernode`: full existing suite (377 tests incl. 5 new) — `PYTHONPATH=../src:. pytest tests/test_logos_bridge.py tests/test_lane_models.py tests/test_lane_manager.py tests/test_vllm_process.py tests/test_model_profiles.py tests/test_vllm_metrics_export.py`
- [x] `logos` orchestrator: full unit suite (1842 passed incl. 5 new; 3 pre-existing failures unrelated to this change, confirmed present on `main` too) — `pytest tests/unit/`

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus monitoring for vLLM metrics across active worker lanes.
  * Worker metrics are periodically collected, labeled by lane and worker, merged, and forwarded to the orchestrator.
  * Added configuration for controlling the metrics reporting interval.
* **Bug Fixes**
  * Failed or unavailable lane metrics no longer prevent other metrics from being exported.
* **Tests**
  * Added coverage for metrics collection, merging, labeling, forwarding, and filtering inactive lanes.


logos/logos-workernode/logos_worker_node/vllm_metrics_export.py
```diff
@@ -0,0 +1,64 @@
...
+    Lanes are queried concurrently. A lane whose vLLM process doesn't answer
+    in time is skipped — one stuck lane must not blank the export for every
+    other lane on this worker.
+    """
+    if not endpoints:
+        return ""
+    async with httpx.AsyncClient() as client:
+        texts = await asyncio.gather(*(_fetch_lane_metrics_text(client, port) for _, _, port in endpoints))
+
+    sources = [({"lane_id": lane_id, "model": model}, text) for (lane_id, model, _port), text in zip(endpoints, texts) if text]
```
Claudia-Anthropica: @bensofficial [high] This new line is 127 characters long, exceeding the Logos flake8 limit of 120; therefore the captured failing pre-commit check is attributable to this diff.

In `logos/logos-workernode/logos_worker_node/vllm_metrics_export.py`, the new `sources` comprehension violates the configured 120-character line limit. Reformat the comprehension across multiple lines so pre-commit passes.

</details>

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -346,6 +351,39 @@ async def _status_refresh_loop(self, ws) -> None:
...
+            except asyncio.CancelledError:
+                raise
+            except Exception:  # noqa: BLE001
+                logger.debug("Failed to collect/send vLLM metrics", exc_info=True)
+
+    async def _send_vllm_metrics(self, ws) -> None:
+        lane_manager = self._app.state.lane_manager
+        endpoints = lane_manager.running_vllm_endpoints()
+        metrics_text = await collect_vllm_metrics_text(endpoints)
+        if not metrics_text:
```
Claudia-Anthropica: @bensofficial [medium] When the final lane stops, collection returns an empty string and this branch sends nothing, so the orchestrator retains and exports the session's previous nonempty snapshot until the worker disconnects. Stopped-lane metrics can therefore remain visible indefinitely.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, an empty endpoint set does not clear the orchestrator's cached metrics snapshot. Send an explicit empty `vllm_metrics` update when no endpoints remain, while distinguishing that case from transient fetch failures.

</details>

logos/logos-workernode/logos_worker_node/vllm_metrics_export.py
```diff
@@ -0,0 +1,64 @@
...
+    def __init__(self, families):
+        self._families = families
+
+    def collect(self):
+        return iter(self._families)
+
+
+async def _fetch_lane_metrics_text(client: httpx.AsyncClient, port: int) -> str | None:
+    try:
+        resp = await client.get(f"http://127.0.0.1:{port}/metrics", timeout=_FETCH_TIMEOUT_S)
```
Claudia-Anthropica: @bensofficial [medium] The exporter hard-codes both `/metrics` and a 3-second timeout even though `VllmEngineConfig` exposes `metrics_path` and `metrics_timeout_seconds` (defaulting to 5). A worker using a supported nondefault path will never export metrics, and even the configured default timeout is ignored.

In `logos/logos-workernode/logos_worker_node/vllm_metrics_export.py`, metric requests bypass the configured metrics path and timeout. Thread `VllmEngineConfig.metrics_path` and `metrics_timeout_seconds` into the collection call and use them for each request.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The forwarding implementation leaves stale series after the last lane stops and ignores the existing configurable vLLM metrics endpoint settings. The exact head also contains a new line that violates Logos' 120-character flake8 limit, concretely accounting for the captured red pre-commit check, so these issues need changes before approval.

bensofficial:
@Claudia-Anthropica thanks for the review — addressed in a929e443c:

- **Stale series after the last lane stops**: the worker now always pushes its merged vLLM metrics, even when empty. `peek_vllm_metrics` already treated an empty string as "nothing to report," so the missing piece was the worker going quiet instead of explicitly signaling "zero lanes now" — fixed by removing that early return in `_send_vllm_metrics`.
- **Ignored configurable metrics endpoint settings**: `collect_vllm_metrics_text` now takes `metrics_path`/`timeout_s`, sourced from the worker's `AppConfig.engines.vllm` (the existing `VllmEngineConfig.metrics_path` / `metrics_timeout_seconds` fields), instead of a hardcoded `"/metrics"` and 3s timeout.
- **flake8 120-char violation**: rewrote the offending comprehension as an explicit loop; confirmed no line over 120 chars in any touched file.

Also added, since it came up while fixing the above: incoming `vllm_metrics` payloads are now validated (must be a string, capped at 4MB) before being cached, rather than coerced with `str()` unconditionally.

Full test suites pass on both sides (377 workernode, 1846 orchestrator) with 9 new tests covering these fixes specifically. Re-requesting your review.

logos/logos-orchestrator/src/logos/routers/logosnode.py
```diff
@@ -42,6 +42,36 @@
...
+    every Prometheus scrape of this orchestrator reads from.
+    """
+    if not isinstance(value, str):
+        logger.warning(
+            "Dropping vllm_metrics from provider %s: metrics_text was %s, not a string",
+            provider_id,
+            type(value).__name__,
+        )
+        return None
+    if len(value.encode("utf-8", errors="ignore")) > _MAX_VLLM_METRICS_BYTES:
```
Claudia-Anthropica: @bensofficial [medium] Using `errors="ignore"` makes escaped lone surrogates contribute zero bytes to this limit, so an authenticated worker can bypass the 4 MiB application cap and send a payload up to the transport's 256 MiB allowance. The resulting large string is cached and reparsed on every Prometheus scrape.

In `logos/logos-orchestrator/src/logos/routers/logosnode.py`, `_validated_vllm_metrics_text` ignores invalid Unicode while measuring the payload, allowing escaped lone surrogates to bypass the size cap. Encode with strict UTF-8, reject `UnicodeEncodeError`, and compare the resulting byte length before caching.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The prior stale-series, configurable-endpoint, and line-length findings are fixed on this head. However, the new UTF-8 size validation can still be bypassed with escaped lone surrogates, leaving the orchestrator vulnerable to oversized cached metrics payloads.

bensofficial:
@Claudia-Anthropica good catch — fixed in 0c21dcf3. Switched from `errors="ignore"` to `errors="surrogatepass"` for the size check, so a string of lone surrogates now counts its real encoded byte size instead of collapsing to 0 bytes and slipping under the cap. Added a regression test for exactly that case. Re-requesting review.

logos/logos-orchestrator/src/logos/routers/logosnode.py
```diff
@@ -42,6 +42,41 @@
...
+            provider_id,
+            type(value).__name__,
+        )
+        return None
+    # "surrogatepass", not "ignore": a lone surrogate (valid inside a JSON
+    # string escape, e.g. an unpaired \uD800) encodes to zero bytes under
+    # "ignore", so a string built mostly out of them would sail under the
+    # cap while still costing real memory — surrogatepass counts it instead
+    # of dropping it, closing that bypass.
+    if len(value.encode("utf-8", errors="surrogatepass")) > _MAX_VLLM_METRICS_BYTES:
```
Claudia-Anthropica: @bensofficial [medium] `surrogatepass` counts lone surrogates but still accepts them, so a small payload containing one in a valid HELP string or label can enter a parsed metric family. `generate_latest` then strictly UTF-8 encodes the merged exposition and raises `UnicodeEncodeError`, allowing one authenticated worker to make every `/metrics` scrape fail until the snapshot is cleared. Use strict UTF-8 encoding during validation and reject `UnicodeEncodeError` before caching.

In `logos/logos-orchestrator/src/logos/routers/logosnode.py`, `_validated_vllm_metrics_text` accepts unpaired surrogates via `errors="surrogatepass"`, but prometheus-client later strictly encodes the merged exposition, so a valid metric containing such a character can crash every `/metrics` scrape. Encode with strict UTF-8, catch `UnicodeEncodeError`, and reject the payload before calling `on_vllm_metrics`.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The earlier stale-series, configurable-endpoint, formatting, and size-cap-bypass findings are fixed on this head. One Unicode validation gap remains: `surrogatepass` admits invalid Unicode values that can make Prometheus serialization fail, so the current head still needs a change.

bensofficial:
@Claudia-Anthropica right — surrogatepass counted but didn't reject. Fixed in 3c902745: validation now encodes strictly and rejects on UnicodeEncodeError, so an unpaired surrogate never reaches the cache regardless of payload size. Added a regression test with a small, otherwise-valid metric line carrying one poisoned label value. Re-requesting review.

Claudia-Anthropica (approved):
@bensofficial The exact head resolves all prior findings: it clears stale snapshots, honors the configured vLLM endpoint settings, and rejects invalid or oversized UTF-8 before caching. Focused tests cover collection, relabeling, forwarding, cache clearing, and validation, while the captured snapshot has no failed checks and shows pre-commit passing. Several non-required checks were still pending at capture, so this approval does not imply that the entire CI matrix had completed.

PR #966, opened 2026-09-14
`Logos`: Fix missing Batches menu icon and add batch creation timestamps

bensofficial:
## Summary
- The Batches nav item used `piIcon: 'layer-group'`, which is not a real PrimeIcons 7 icon name — it silently rendered as an empty box instead of an icon. Swapped it for `'clone'` (a stacked-rectangles glyph PrimeIcons actually ships), which reads well for "batches" of requests.
- Batch ids are opaque random tokens by design (`batch_<hex>`, matching the OpenAI Batch API shape scripts rely on), so the "Your batches" table gave no way to tell entries apart at a glance. Added a **Created** column using the `created_at` timestamp the backend already returns, formatted with `toLocaleString()`.

## Test plan
- [x] `ng build` (development config) compiles cleanly with no new type errors.
- [x] Verified the icon bug and fix visually against the deployed page (https://logos.aet.cit.tum.de/batches): injecting `pi-layer-group` renders nothing, `pi-clone` renders the intended glyph.
- [x] Previewed the new "Created" column layout against the live table styling with a real completed batch — renders as `9/14/2026, 9:12:32 PM` and fits without layout issues.
- [ ] Could not exercise the full page end-to-end locally (`ng serve`) since it depends on the orchestrator/Keycloak backend, which isn't reachable in this environment.


Claudia-Anthropica (approved):
@bensofficial The Created column correctly converts the API's epoch-second `created_at` value to a localized timestamp and handles missing values. The replacement icon exists in the declared PrimeIcons version, and the captured UI static export passed; I found no blocking issues.

PR #972, opened 2026-09-14
`Logos`: Count cold and warm starts over local requests only

### LINKED ISSUE #928: Logos: Stats page
Right now it states "0 %", but it shall show the exact percentage instead!

COLD STARTS
0%
694 of 317.265 local starts

wasnertobias:
## What

The stats page KPI card shows the cold-start share as "N of M **local** starts" — but `M` (cold + warm) also counted cloud requests: `warmStarts` aggregated every request whose row was not a cold start, cloud included. The denominator was larger than the set its label describes, and the share was understated (issue #928, comment; the rounding half of the issue was fixed in #929).

## Changes

- `findTotals`: restrict the cold/warm `COUNT(*) FILTER` aggregates to local requests — the exact same `privacy_level` predicate `localRequests` already uses — so cold + warm now adds up to exactly the local request count.
- `findModelBreakdown`: the same local-only rule for the per-model cold/warm counts (same API response, same field semantics).

No UI change: the card's `coldStarts + warmStarts` denominator picks up the corrected numbers automatically.

## Test

New `requestLogStats_countsColdAndWarmStartsOnLocalRequestsOnly`: seeds a cloud provider plus one warm and one cold request on it (via `JdbcTemplate`, since a method-level `@Sql` would replace the class-level seeds) and asserts both land in neither count — 4 requests, 2 cloud / 2 local, 1 cold / 1 warm in totals and in the model breakdown.

`RequestLogStatsControllerTest` (18 tests) and all other suites sharing the operations seed (StatsV2, Stats, ProviderPerformance, TeamActivity, RequestLog, Billing, Vram, TeamActivityExport) pass.

Fixes #928


Claudia-Anthropica (approved):
@wasnertobias The totals and per-model aggregates now use the same provider-privacy predicate as `localRequests`, excluding cloud rows from both sides of the cold-start denominator while preserving the cold/warm partition for local rows. The focused integration test covers cloud rows marked both warm and cold at both aggregation levels, and all captured test, build, and lint checks succeeded.

Claudia-Anthropica (approved):
@wasnertobias The totals and per-model aggregates now apply the same provider-privacy predicate as `localRequests`, so cloud requests no longer affect the local cold-start denominator. The focused integration test covers both cold and warm cloud rows, and all captured test, build, and lint checks succeeded.

PR #974, opened 2026-09-15
`Logos`: Answer review comments in their own threads

### LINKED ISSUE #945: `Logos`: Agent shall re-request review and resolve comments
A review is currently only answered in one big comment, but it shall be re-requested and individually answered and or resolved.

wasnertobias:
Fixes #945.

A changes-requested review was answered with one big comment on the pull
request: every inline thread stayed open, and the reviewer's "changes
requested" stood until they happened to look at the branch again. A review
is several questions, so the answer now goes back question by question —
and the pull request is put back in front of the reviewer instead of
waiting for them to notice.

## What changed

**The review task** (`triggers.py`) names each inline comment by its id and
asks the agent for one reply file per comment — `replies/<comment id>.md` —
saying what it changed and how it verified it, or why no change was needed.
A point that lives in the review body alone still goes into `reply.md`,
where it is posted as a single comment as before.

**The runner** (`sessions.py`) delivers a review answer thread by thread:

- each `replies/<id>.md` is posted into that comment's own thread (only
  comment ids the review actually carries are answered),
- the answered threads are resolved afterwards — a thread a person already
  resolved is left as they left it,
- the summary file is posted when the agent wrote one,
- once the session's work has landed, the review is re-requested from the
  reviewer whose name is on the review the session answered (a failed
  session pushed nothing, so it does not ask for a re-review of the same
  code).

**New GitHub calls** (`github.py`): a single review by id, the
`requested_reviewers` endpoint, and the two GraphQL pieces the REST API
does not cover — the pull request's review threads (so a comment's id finds
the thread that resolution acts on) and `resolveReviewThread`.

**Retries without duplicates:** delivery is what it always was (posted at
settlement, retried by the scheduler sweep until it lands), but a review
answer is several posts instead of one. What already reached GitHub is
recorded in the session's state directory as it goes — the runner's own
directory, out of the agent's reach — so an attempt that dies halfway
resumes where it stopped instead of answering a thread twice.

## Fallbacks

A review session whose row does not name the review it answered, or whose
review has been deleted, posts the summary where a single answer always
went. A session that wrote nothing — no per-comment files and no summary —
is handled exactly as before: taken up again, or left alone when it failed.

## Testing

- the review task: comment ids in the task text, the per-file protocol, the
  summary file's remaining role, and comments without ids;
- the GitHub helpers: the review lookup, the re-request's exact call, the
  thread map (including pagination and comment-less threads), one mutation
  per resolved thread, a declined resolution not stopping the rest, and
  GraphQL errors raising;
- the delivery: each comment answered in its own thread, the summary
  alongside, an id the review does not carry not answered, no re-request
  for work that did not land, a retry that resumes without duplicates, the
  summary-only fallbacks, somebody else's resolution left alone, and a
  corrupt delivery state.

All 646 logos-agent tests pass; the touched files are clean under
`pre-commit`.

## Summary by CodeRabbit

* **New Features**
  * Review feedback can receive individual replies directly in each inline comment thread.
  * All inline review comments are included and clearly identified.
  * Answered review threads are automatically resolved, and reviewers are asked to review changes again.
  * Review responses include a summary for feedback provided in the overall review.
  * Large responses are split into complete, ordered messages without losing content.

* **Bug Fixes**
  * Interrupted or retried deliveries resume safely without duplicating replies.
  * Improved handling of failed, missing, deleted, or incomplete review responses.
  * Delivery markers are validated against the configured account to prevent false confirmations.
  * Large reviews and paginated comment listings are handled more reliably.


logos/logos-agent/app/github.py
```diff
@@ -602,6 +644,80 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+    when it could not resolve what it was given — the thread moved, or the
+    token may not act on it; that is logged and the rest still goes on. A
+    call that reports an error raises instead, and the caller retries the
+    whole sweep rather than claiming a resolution that did not happen.
+    """
+    for thread_id in thread_ids:
+        if not thread_id:
+            continue
+        data = await _graphql(
+            "mutation ResolveReviewThread($threadId: ID!) { resolveReviewThread(threadId: $threadId) { threadId } }",
```
Claudia-Anthropica: @wasnertobias [high] GitHub's `resolveReviewThread` mutation accepts an `input: ResolveReviewThreadInput!` and returns a payload containing `thread`; this query instead passes a top-level `threadId` argument and selects a nonexistent payload field. Every attempted resolution therefore returns GraphQL errors, preventing delivery from reaching the reviewer re-request.

In `logos/logos-agent/app/github.py`, the `resolveReviewThread` mutation uses an invalid argument and response selection. Invoke `resolveReviewThread(input: {threadId: $threadId})`, select `thread { id }`, and make the test assert the exact request and real response shape.

</details>

logos/logos-agent/app/sessions.py
```diff
@@ -2282,6 +2270,226 @@ async def _post_reply_once(self, session_id: int) -> None:
...
+                        exc,
+                    )
+                    return
+                if body:
+                    per_comment[comment_id] = body
+
+        summary = await self._read_answer(session_id)
+        if summary is None:
+            return
+        if not per_comment and not summary and not state["summary_posted"]:
```
Claudia-Anthropica: @wasnertobias [medium] After a first pass posts all per-comment files, `per_comment` is empty on retry because their IDs are already recorded in state. With no summary file, this condition calls `_no_answer` before retrying thread resolution or the reviewer re-request, so a transient failure after the last reply abandons/restarts an already-answered session instead of resuming it.

In `logos/logos-agent/app/sessions.py`, the no-answer guard ignores already-posted per-comment replies recorded in `state["replied"]`. Treat the session as answerless only when no new or persisted reply or summary exists, and add retry coverage for per-comment replies without `reply.md`.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The per-thread design is well covered at the unit boundary, but two current-head defects prevent reliable end-to-end delivery: GitHub rejects the resolution mutation, and a no-summary retry is misclassified as having no answer. The snapshot has no required pending checks; three non-required build jobs were still in progress when captured.

logos/logos-agent/app/github.py
wasnertobias: Confirmed — the live schema rejects the first draft with three validation errors (`missingRequiredArguments: input`, `argumentNotAccepted: threadId`, and `threadId` not existing on `ResolveReviewThreadPayload`), so every resolution raised and the delivery never reached the re-request. The mutation now sends `resolveReviewThread(input: {threadId: $threadId})` and selects `thread { id }`; the test pins the exact query and variables plus the real response shape, and I verified the corrected mutation against the API before pushing.

Along the same line: a thread GitHub refuses outright (a status-less GraphQL error — gone, or not actionable by the token) is now recorded and left open instead of stalling the sweep, so the re-request no longer waits on a thread that will never resolve.

logos/logos-agent/app/sessions.py
wasnertobias: Right — after a first pass posts all per-comment files, a retry finds `per_comment` empty, and without a `reply.md` it took the no-answer path: restart or abandonment instead of resuming an already-answered session. The guard now also requires that no per-comment reply was persisted in the delivery state, and there is retry coverage for exactly this shape: per-comment replies without a summary, a first pass that dies after the last reply, and a second pass that resumes without re-answering a thread or restarting the session.

logos/logos-agent/app/sessions.py
```diff
@@ -2282,6 +2270,241 @@ async def _post_reply_once(self, session_id: int) -> None:
...
+            await self._resolve_answered_threads(session_id, number, state, state_path)
+
+        if summary and not state["summary_posted"]:
+            url = await github.post_issue_comment(number, self._truncate_reply(summary))
+            state["summary_posted"] = True
+            self._write_review_reply_state(state_path, state)
+            await db.add_event(session_id, EventKind.PULL_REQUEST, {"url": url, "reply": True})
+            logger.info("session %s posted its summary on pull request %s", session_id, number)
+
+        if (
```
Claudia-Anthropica, line 2563: @wasnertobias [medium] This re-request is reached after any nonempty artifact, even when `valid` contains review comments absent from `state["replied"]`. For example, given comments 101 and 102 but only `replies/101.md`, the runner resolves 101, re-requests the reviewer, and stamps delivery complete, so 102 is never answered or retried.

In `logos/logos-agent/app/sessions.py`, review delivery succeeds after only a partial set of per-comment reply files, permanently leaving omitted review comments unanswered. Compare the actionable review-comment IDs with persisted and new replies before re-requesting, and route an incomplete response through the replacement path.

</details>

logos/logos-agent/app/sessions.py
```diff
@@ -2282,6 +2270,241 @@ async def _post_reply_once(self, session_id: int) -> None:
...
+            if not thread_id or thread_id in handled:
+                continue
+            if info.get("resolved"):
+                state["resolved_threads"].append(thread_id)
+                self._write_review_reply_state(state_path, state)
+                continue
+            try:
+                await github.resolve_review_threads([thread_id])
+            except github.GitHubError as exc:
+                if exc.status is not None:
```
Claudia-Anthropica: @wasnertobias [medium] `_graphql` gives every HTTP-200 GraphQL error `status=None`, so this branch also swallows rate-limit, server, and query-validation errors, records the still-open thread in `resolved_threads`, and proceeds to re-request the reviewer. Those errors are not evidence that the thread is permanently unactionable.

In `logos/logos-agent/app/sessions.py`, all status-less GraphQL errors are treated as permanent resolution refusals even though many are retryable or indicate a broken request. Preserve and classify the GraphQL error type, mark only explicitly permanent failures handled, and re-raise all other GraphQL errors for retry.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The two previously reported defects are fixed: the resolution mutation now matches GitHub's schema, and summary-less retries recognize persisted replies. Two delivery paths can still silently leave threads unanswered or unresolved while marking the review response complete. No required checks were configured in the snapshot; several non-required build, test, and lint jobs were still pending.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 41dd00aeb. Delivery now compares the actionable comments — those with a non-empty body; an empty-body comment is never shown to the agent in the task, so expecting its reply would make the delivery incomplete forever — against what is already persisted plus the new reply files. An incomplete answer is routed through the no-answer path: the session is taken up again and nothing is posted or re-requested. Covered by a new test: two questions asked, one answered, and the delivery holds with the session taken up again.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 41dd00aeb. GitHubError now carries the GraphQL error type (_graphql extracts it from the error entries), and the resolution sweep swallows only the explicitly permanent refusals — the thread is gone, or this token may never act on it (NOT_FOUND, FORBIDDEN, UNAUTHENTICATED). Every other error — rate limits, server errors, unknown types, and an unclassified one — is re-raised, so the whole delivery fails and is tried again. Covered by tests: the permanent-refusal test now passes graphql_type="NOT_FOUND", a new test verifies a RATE_LIMITED refusal leaves the thread unrecorded and stamps the attempt undelivered, and the GraphQL error test asserts the type reaches the caller.

logos/logos-agent/app/sessions.py
```diff
@@ -2282,6 +2275,258 @@ async def _post_reply_once(self, session_id: int) -> None:
...
+
+        try:
+            review = await github.review(number, review_id)
+        except github.GitHubError as exc:
+            if exc.status == 404:
+                # The review is gone — deleted by its author or a
+                # moderator. Its threads are gone with it, and only the
+                # summary can still be said where it always went.
+                logger.info("the review session %s answered no longer exists; posting its summary only", session_id)
+                await self._post_single_reply(session_id, session, target)
```
Claudia-Anthropica, line 2462: @wasnertobias [medium] An inline-only review session is explicitly told not to write `reply.md`, so if this lookup later returns 404, `_post_single_reply` sees no summary and cycles through replacement sessions until the attempt cap without posting any of the already-written answers. Preserve the per-comment content in this fallback—such as by combining the reply files into an issue comment—and add a deleted-review test without `reply.md`.

In `logos/logos-agent/app/sessions.py`, the deleted-review fallback discards per-comment artifacts when no `reply.md` exists, causing repeated replacement sessions and no delivered answer. Post the available per-comment replies as a fallback summary and cover the no-`reply.md` case.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The current head fixes the partial-artifact and GraphQL-classification defects, but two concrete delivery dead ends remain. Reviews with more than 30 actionable inline comments cannot satisfy the new completeness check because the task exposes only the first 30, while a deleted review discards the normal per-thread-only artifacts instead of falling back to a summary. No required checks were configured in the snapshot; the remaining in-progress jobs are non-required.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 841a92b5a. When the summary file is absent, the single-reply fallback now combines the per-comment answers into the one comment an answer always becomes there — each under a heading naming the comment it answers. An unreadable reply file still counts as an attempt, so a delivery that could only read part of them is retried instead of posting an incomplete set as the whole. The requested test is in: a deleted review, no reply.md, two per-comment answers — they are combined and posted as the single comment, with no re-request and the attempt stamped delivered.

logos/logos-agent/app/sessions.py
```diff
@@ -2224,51 +2237,40 @@ async def _post_reply_once(self, session_id: int) -> None:
...
-                # blinked, not a decision. Left owing an answer, so the next
-                # sweep tries again: abandoning it first is how a request
-                # disappears between two failures.
-                logger.info("session %s could not be taken up again yet; leaving it owing an answer", session_id)
+            # A review answered thread by thread has no summary file. Its
+            # per-comment answers may still exist even where their threads
+            # do not (the review was deleted) — combined, they are still an
+            # answer, and an unanswered review is what they were written
+            # against.
+            body = await self._combined_review_answers(session_id)
```
Claudia-Anthropica: @wasnertobias [medium] When a deleted review contains both a body-level point and inline comments, `reply.md` is nonempty, so this branch never calls `_combined_review_answers`; `_post_single_reply` posts only the summary, silently discards every per-comment answer, and marks the attempt delivered.

In `logos/logos-agent/app/sessions.py`, the deleted-review fallback ignores per-comment artifacts whenever `reply.md` exists. Combine the summary with every per-comment answer before sending, and add coverage for a deleted review containing both body-level and inline answers.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The no-summary deleted-review case is fixed, but a deleted review that has both `reply.md` and inline reply files still posts only the summary and marks delivery complete. The previously reported >30-comment dead end also remains: `_inline_block` exposes only 30 comments while `_post_review_reply` requires answers for every actionable comment, so replacement sessions cannot complete such a review. The captured CI snapshot has no failed or required checks, though several non-required jobs were still pending.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 39f686a9d. The single-reply fallback now reads both parts and sends whatever mix was written: the summary on top, every per-comment answer under it (each under a heading naming its comment), the per-comment answers alone when there is no summary, or the summary alone when there are no per-comment files. An unreadable reply file still counts as an attempt rather than as an absence. The requested coverage is in: a deleted review with both a body-level point and inline answers, posted as the one combined comment and stamped delivered.

logos/logos-agent/app/sessions.py
```diff
@@ -2282,6 +2288,290 @@ async def _post_reply_once(self, session_id: int) -> None:
...
+            c["id"]
+            for c in comments
+            if isinstance(c, dict) and isinstance(c.get("id"), int) and str(c.get("body") or "").strip()
+        }
+        if actionable - (set(state["replied"]) | set(per_comment)):
+            await self._no_answer(session_id, session)
+            return
+
+        for comment_id in sorted(per_comment):
+            url = await github.reply_to_review_comment(
```
Claudia-Anthropica: @wasnertobias [medium] The durable `replied` marker is written only after `reply_to_review_comment` returns. If GitHub creates the reply but its response is lost, or the process exits before the state write, the next sweep posts the same answer again; the retry test only fails after all reply state has already been saved and does not cover this ambiguous-POST window.

In `logos/logos-agent/app/sessions.py`, a remotely accepted review reply can be duplicated when its POST response is lost before local state is saved. Add a stable hidden per-session/per-comment marker to every reply, reconcile the thread for that marker before retrying the POST, and add coverage for a remotely accepted POST that raises locally.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The deleted-review fallback now correctly preserves both summary and per-comment replies. [medium] The >30-comment dead end remains: `_inline_block` exposes only 30 IDs while delivery requires every actionable comment, so replacement sessions can never complete. [medium] Retry state is still written only after the non-idempotent reply POST returns, allowing response-loss retries to duplicate thread answers; no required checks were configured, while several non-required jobs were pending in the captured snapshot.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in cd25ab956. Every posted answer now carries a hidden HTML-comment marker naming the session and the comment it answers. Before a reply goes out, the thread that owns the comment is asked for that marker (paged GraphQL over the thread's comment bodies): a found answer is recorded as delivered in the state without posting it again, so a remotely accepted POST whose confirmation was lost — or a crash between POST and state write — is reconciled instead of duplicated. A look-up that cannot cover every thread raises and the sweep is retried, rather than risking the duplicate. Coverage: the requested test (a remotely accepted POST that raises locally — second pass finds the marker, no second POST, attempt stamped delivered), plus tests for the marker look-up itself, including its pagination.

logos/logos-agent/app/github.py
```diff
@@ -602,6 +652,134 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+# A thread's whole conversation, paged. A reply that a lost POST
+# confirmation left behind is one of a thread's comments, and only the
+# bodies can say whether it is there.
+_REPLY_THREAD_QUERY = """
+query ReviewThreadReply($owner: String!, $name: String!, $number: Int!, $after: String) {
+  repository(owner: $owner, name: $name) {
+    pullRequest(number: $number) {
+      reviewThreads(first: 100, after: $after) {
+        nodes {
+          comments(first: 100) {
```
Claudia-Anthropica: @wasnertobias [medium] Only the outer `reviewThreads` connection is paginated; each thread's `comments` connection is permanently limited to its first 100 entries. If GitHub accepts a reply after 100 existing comments and its response is lost, the marker is outside this slice, `review_reply_is_in_thread` returns false, and the retry posts the answer again—the failure mode this reconciliation is intended to eliminate.

In `logos/logos-agent/app/github.py`, `review_reply_is_in_thread` scans only the first 100 comments of each thread, so it can miss a previously accepted marked reply and duplicate it. Resolve the owning thread and paginate its `comments` connection to completion before returning false, with a regression test covering a marker after the first 100 comments.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The marker reconciliation works for ordinary threads, but two current-head delivery bounds remain. The existing >30-comment dead end persists: `_inline_block` exposes only 30 actionable IDs while `_post_review_reply` requires answers for all of them. The new marker lookup also scans only the first 100 comments in each thread, so a lost POST on a longer thread is duplicated; no required checks were configured, while captured non-required jobs were pending.

logos/logos-agent/app/github.py
wasnertobias: Fixed in 9b8c8c5c5. The look-up is now two-stage: the thread map resolves the thread the comment starts, and that thread's comments connection is then paged to completion (node query on the thread id) before the reply may be ruled out — a marker past the first hundred comments of a busy thread is still found. Both ceilings now raise rather than shrink their results: a thread map that cannot be completed hands the sweep back to the retry instead of letting a delivery claim success with a thread left unresolved, and a comment that starts no thread of its own (a reply inside somebody else's thread) is logged and let through, as the POST for it goes into a thread the map does not name. Regression test included: a thread with a full first page of comments and the marker on the second.

logos/logos-agent/app/github.py
```diff
@@ -602,6 +652,134 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+            if isinstance(comment_id, int) and comment_id not in mapped:
+                mapped[comment_id] = {"thread": node.get("id"), "resolved": bool(node.get("isResolved"))}
+        page = connection.get("pageInfo") or {}
+        if not page.get("hasNextPage"):
+            return mapped
+        after = str(page.get("endCursor") or "")
+    logger.warning(
+        "the review threads of #%s hit the %s-page ceiling; newer threads were not mapped", number, _MAX_PAGES
+    )
+    return mapped
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 9b8c8c5c5: the page-ceiling branch of the thread map now raises GitHubError instead of returning an incomplete map, so the delivery records the attempt as undelivered and retries it.

logos/logos-agent/app/github.py
```diff
@@ -602,6 +652,134 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+# confirmation left behind is one of a thread's comments, and only the
+# bodies can say whether it is there.
+_REPLY_THREAD_QUERY = """
+query ReviewThreadReply($owner: String!, $name: String!, $number: Int!, $after: String) {
+  repository(owner: $owner, name: $name) {
+    pullRequest(number: $number) {
+      reviewThreads(first: 100, after: $after) {
+        nodes {
+          comments(first: 100) {
+            nodes { databaseId body }
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 9b8c8c5c5: after locating the owning thread through the thread map, the look-up pages that thread's comments connection to completion before returning false.

logos/logos-agent/app/triggers.py
```diff
@@ -339,10 +348,16 @@ async def review_task(
...
         f"move, and some of it may already be addressed. Fix what is still valid, add "
         f"regression coverage for it, and run the tests and linters of the part you "
-        f"touched. Write your reply to the review into `$LOGOS_ARTIFACT_DIR/{REPLY_FILE}` "
-        f"— in English, saying for each point what you changed and how you verified it, "
-        f"or why it needed no change; the runner posts it for you. Do not merge the pull "
-        f"request and do not force-push."
+        f"touched.\n\n"
+        f"Answer each inline comment where it was made: one file per comment under "
+        f"`$LOGOS_ARTIFACT_DIR/{REPLY_DIR}/`, named after the id the comment is listed "
+        f"with — `{REPLY_DIR}/<comment id>.md`. Each one in English, saying what you "
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 9b8c8c5c5: the task no longer truncates the inline comments to the first thirty — every comment that carries a body reaches the agent, which is the same set the delivery requires answers for. Each comment's body still carries its per-comment length cap.

logos/logos-agent/app/sessions.py
```diff
@@ -2282,6 +2300,302 @@ async def _post_reply_once(self, session_id: int) -> None:
...
+            state["replied"].append(comment_id)
+            self._write_review_reply_state(state_path, state)
+            await db.add_event(session_id, EventKind.PULL_REQUEST, {"url": url, "reply": True, "comment": comment_id})
+            logger.info("session %s answered comment %s on pull request %s", session_id, comment_id, number)
+
+        if state["replied"]:
+            await self._resolve_answered_threads(session_id, number, state, state_path)
+
+        if summary and not state["summary_posted"]:
+            url = await github.post_issue_comment(number, self._truncate_reply(summary))
```
Claudia-Anthropica: @wasnertobias [medium] The summary still uses a non-idempotent comment POST and records `summary_posted` only after the response returns. If GitHub accepts this POST but its response is lost, the next sweep sees the flag unset and posts the same summary again—the exact ambiguous-POST case the inline marker logic handles.

In `logos/logos-agent/app/sessions.py`, a remotely accepted review-summary POST can be duplicated when its response is lost before `summary_posted` is persisted. Add a stable per-session marker to review summaries, reconcile existing pull-request comments for that marker before retrying the POST, and add coverage for an accepted summary POST that raises locally.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The current head fixes the previously reported pagination and comment-completeness defects, and the captured CI has no failed checks. One duplicate-delivery window remains for review summaries because they lack the reconciliation mechanism now used for inline replies.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 83db656f5. The single answer now carries a hidden mark naming its session (the per-thread replies already had theirs), and before the POST goes out the pull request's comments are looked up for that mark: a summary GitHub accepted but whose confirmation never arrived is found on the pull request and recorded as posted, instead of being said twice. The same treatment covers the combined answer the deleted-review fallback posts as one comment — the same non-idempotent POST one line away. Coverage: the requested test (an accepted summary POST that raises locally — second pass finds the mark, no second POST, summary_posted recorded), a matching test for the single-reply path, and a test for the marker look-up itself.

logos/logos-agent/app/sessions.py
```diff
@@ -2224,51 +2261,56 @@ async def _post_reply_once(self, session_id: int) -> None:
...
-            # thread where somebody is waiting.
-            body = body[:_MAX_REPLY_CHARS].rstrip() + "\n\n_[answer truncated]_"
-            logger.info("the answer of session %s was truncated to fit a GitHub comment", session_id)
+        body = self._truncate_reply(body)
+        if target.startswith("issue:"):
+            # The answer's POST is not idempotent: a confirmation lost on
+            # the way back leaves it posted without the state knowing, and
+            # the mark on the pull request is the answer to whether
+            # posting again would say it twice.
+            marker = _answer_marker(session_id)
```
Claudia-Anthropica: @wasnertobias [medium] `_post_review_reply` uses `_answer_marker(session_id)` for the summary, while this fallback uses the identical marker for a different payload containing all per-comment answers. If the summary posts, the reviewer re-request fails, and the review is deleted before retry, this lookup finds the old summary and stamps delivery complete without posting the now-lost thread answers. Give the combined fallback a distinct or content-bound marker and cover that sequence with a regression test.

In `logos/logos-agent/app/sessions.py`, summary-only delivery and the deleted-review combined fallback share one marker, allowing an existing summary to suppress delivery of the combined inline answers. Use a distinct or content-bound marker for the fallback and add a test where the review disappears after its summary was already posted.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The current head fixes the previously reported ambiguous summary-POST retry for normal delivery, and the captured CI has no failed or required checks, though non-required jobs remained pending. One fallback race remains: summary-only and deleted-review combined answers share a marker, so deleting the review after its summary was posted can make the retry mark the combined answer delivered while dropping every inline answer.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 8b445f82a. The summary now carries its own mark (the single answer's mark stays with the answer), so a posted summary can no longer stand for the combined answer. The regression test covers exactly that sequence: pass one answers the thread and posts the summary, dies on the re-request; the review is deleted; pass two finds it gone and posts the combined answer under its own mark — the summary's mark does not suppress it.

logos/logos-agent/app/github.py
```diff
@@ -602,6 +652,140 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+        connection = (data.get("repository") or {}).get("pullRequest", {}).get("reviewThreads") or {}
+        for node in connection.get("nodes") or []:
+            first = ((node.get("comments") or {}).get("nodes") or [None])[0]
+            comment_id = (first or {}).get("databaseId")
+            if isinstance(comment_id, int) and comment_id not in mapped:
+                mapped[comment_id] = {"thread": node.get("id"), "resolved": bool(node.get("isResolved"))}
+        page = connection.get("pageInfo") or {}
+        if not page.get("hasNextPage"):
+            return mapped
+        after = str(page.get("endCursor") or "")
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 8b445f82a: both pagination loops now validate the next cursor whenever hasNextPage is true — an empty or repeated endCursor raises GitHubError instead of re-reading the same page until the ceiling. Covered by tests for both loops.

logos/logos-agent/app/github.py
```diff
@@ -602,6 +652,140 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+    # be looked for, and the POST goes ahead as before.
+    threads = await review_thread_map(number)
+    thread_id = str((threads.get(comment_id) or {}).get("thread") or "")
+    if not thread_id:
+        logger.warning(
+            "comment %s on #%s starts no thread of its own; its posted answer cannot be looked for",
+            comment_id,
+            number,
+        )
+        return False
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 8b445f82a: the thread map now fetches each thread's comments (not just its starter) and keys the map by every comment id in it, so a comment that is a reply inside another thread resolves to its owning thread. The same map feeds both the marker reconciliation and the thread resolution. A comment a thread holds beyond its first hundred stays out of the map, where resolution then simply leaves the thread open — the safe direction — and that limit is documented on the map.

logos/logos-agent/app/github.py
```diff
@@ -987,6 +1171,18 @@ async def post_issue_comment(number: int, body: str) -> str:
...
 
+async def issue_comment_contains(number: int, marker: str) -> bool:
+    """Whether the pull request's comments already carry the marked answer.
+
+    A comment POST is not idempotent, and a confirmation that never
+    arrived leaves the posted comment behind without the delivery state
+    ever learning of it — this is how the next pass finds it instead of
+    posting the same answer twice.
+    """
+    comments = await _get_all(f"/repos/{settings.repo_slug}/issues/{number}/comments")
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 8b445f82a: issue_comment_contains now reads the listing through the bounded helper and raises GitHubError when the listing is incomplete, instead of answering no from a truncated read. Covered by a test.

logos/logos-agent/app/sessions.py
```diff
@@ -2224,51 +2261,56 @@ async def _post_reply_once(self, session_id: int) -> None:
...
+            if await github.issue_comment_contains(int(target.partition(":")[2]), marker):
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 8b445f82a: the marker look-up and the POST now sit under the same error handler in _post_single_reply — a failed look-up records the attempt as undelivered and returns, so nothing escapes to the settlement. Covered by a test.

logos/logos-agent/app/github.py
```diff
@@ -602,6 +652,154 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+# is the thread, not the comment, that resolution acts on.
+_THREADS_QUERY = """
+query ReviewThreads($owner: String!, $name: String!, $number: Int!, $after: String) {
+  repository(owner: $owner, name: $name) {
+    pullRequest(number: $number) {
+      reviewThreads(first: 100, after: $after) {
+        nodes {
+          id
+          isResolved
+          comments(first: 100) {
```
Claudia-Anthropica, line 731: @wasnertobias [medium] The thread map still reads only `comments(first: 100)` from each thread and does not paginate that nested connection. If an actionable review comment is a reply after position 100, its ID is absent from the map; `review_reply_is_in_thread` consequently returns `False` without examining the owning thread, so a lost-success retry duplicates the answer and resolution skips the thread.

In `logos/logos-agent/app/github.py`, `review_thread_map` omits comment IDs beyond the first 100 comments in each thread, preventing reconciliation and resolution for those replies. Paginate each thread's comments when building the map, or otherwise resolve every target comment's owning thread without relying on the capped nested connection, and add regression coverage for a target after comment 100.

</details>

logos/logos-agent/app/github.py
```diff
@@ -602,6 +652,154 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+            "comment %s on #%s is in no review thread; its posted answer cannot be looked for",
+            comment_id,
+            number,
+        )
+        return False
+    after: str | None = None
+    for _ in range(_MAX_PAGES):
+        data = await _graphql(_THREAD_COMMENTS_QUERY, {"threadId": thread_id, "after": after})
+        connection = (data.get("node") or {}).get("comments") or {}
+        if any(isinstance(c, dict) and marker in str(c.get("body") or "") for c in connection.get("nodes") or []):
```
Claudia-Anthropica: @wasnertobias [medium] Marker reconciliation accepts a match in any comment body: this GraphQL query retrieves no author, and `issue_comment_contains` likewise ignores `user.login`. Once a session ID is visible in one posted reply, another participant can place the expected marker in an unanswered thread or pull-request comment; the runner then records the artifact as delivered, resolves/re-requests, and never posts the agent's actual answer.

In `logos/logos-agent/app/github.py`, reply and issue-comment marker lookups trust marker text authored by any account, allowing repository participants to suppress pending agent answers. Retrieve each matching comment's author and recognize delivery markers only when `login` equals the configured `settings.github_login`, with tests proving participant-authored markers are ignored.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The distinct summary marker fixes the previously reported deleted-review fallback race. Two reconciliation gaps remain: replies beyond a thread's first 100 comments cannot be mapped, and marker matches are trusted regardless of their author. The captured snapshot has no failed or required checks; several non-required jobs were still in progress.

logos/logos-agent/app/github.py
wasnertobias: Fixed in c6a97f7c0: the map now paginates each thread's nested `comments` connection to completion (raising at the page ceiling, like the outer list, so a partial map can never silently drop a target). A new regression test covers a target after comment 100 being read out of the thread's second page and still mapping to its thread, and a second one asserts the ceiling raises.

logos/logos-agent/app/github.py
wasnertobias: Fixed in c6a97f7c0: both marker look-ups now fetch the author (`author { login }` in the thread query, `user.login` on REST issue comments) and recognize a marker only when that login equals the configured `settings.github_login`. New tests prove a participant-authored marker is ignored on both the thread path and the issue-comment path.

logos/logos-agent/app/github.py
```diff
@@ -602,6 +652,202 @@ async def review_comments(number: int, review_id: int) -> list[dict[str, Any]]:
...
+    participant can write the expected marker into an unanswered thread
+    and suppress the answer that is still owed. A marker counts only when
+    the account it was posted by is the account the runner posts with.
+    """
+    if not isinstance(comment, dict):
+        return False
+    if marker not in str(comment.get("body") or ""):
+        return False
+    author = comment.get("author") or comment.get("user") or {}
+    return isinstance(author, dict) and author.get("login") == settings.github_login
```
Claudia-Anthropica: @wasnertobias [medium] GitHub account names are case-insensitive, and this module's `verify_identities` explicitly accepts `LOGOS_AGENT_GITHUB_LOGIN=logosossagent` when the API returns `LogosOSSAgent`; this exact comparison then rejects markers written by that same account. In that supported configuration, a lost POST confirmation is never reconciled and the retry duplicates inline replies or summaries. Normalize both logins (for example with `strip().casefold()`) and cover both REST and GraphQL marker lookups with differently cased configured/API logins.

In `logos/logos-agent/app/github.py`, `_is_our_marker` compares GitHub logins case-sensitively even though the configured identity is accepted case-insensitively, so valid configurations can fail reconciliation and duplicate posts after lost confirmations. Normalize both configured and returned logins before comparing and add REST and GraphQL regression tests with case-only differences.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The nested thread pagination and author-bound marker checks resolve both previous findings. The captured build, lint, and test checks succeeded; the non-required CodeRabbit check remained pending. One supported identity configuration still defeats marker reconciliation because the new author comparison is case-sensitive.

logos/logos-agent/app/github.py
wasnertobias: Fixed in 0b5496149: `_is_our_marker` now compares both logins with `strip().lower()`, exactly the normalization `verify_identities` uses to accept the configured identity, so a supported casing of the same account reconciles the lost confirmation instead of duplicating the answer. Two new regression tests cover the case difference on both look-ups: a GraphQL thread marker authored as "Logos" against a configured "logos", and a REST issue comment authored as "logosossagent" against the default "LogosOSSAgent".

wasnertobias: Follow-up, refined in 3213ee997: rather than case-folding both sides of the comparison, the marker is now recognized by the exact login spelling the API verified the token as at startup (`verify_identities` remembers it), falling back to the configured name only when the identity could not be verified. The API spells the account's name identically on every comment it posts, so a supported any-casing configuration still recognizes its own markers — and a differently cased login that merely matches the configured name apart from case is not the account, so it can no longer make the retry skip an answer the session still owes. Tests cover both look-ups in both directions (verified spelling recognized, differently cased login rejected).

Claudia-Anthropica (changes requested):
@wasnertobias [medium] Remembering the verified API spelling fixes normal startup, but the supported degraded-startup path still leaves marker reconciliation case-sensitive. `verify_identities()` explicitly continues when every `/user` lookup fails, so `_verified_login` remains unset; with an accepted differently cased `github_login`, line 708 compares that configuration exactly with GitHub's canonical comment author and misses a remotely accepted marker, allowing the retry to duplicate an inline reply or summary. Normalize the configured fallback when no verified spelling is available and cover an identity-lookup failure followed by marker reconciliation; the captured build, lint, and test checks otherwise succeeded.

logos/logos-agent/app/sessions.py
```diff
@@ -2224,52 +2273,59 @@ async def _post_reply_once(self, session_id: int) -> None:
...
+            await self._no_answer(session_id, session)
             return
-        if len(body) > _MAX_REPLY_CHARS:
-            # GitHub refuses a comment above its length limit outright, and
-            # an answer nobody receives is worse than a shortened one on a
-            # thread where somebody is waiting.
-            body = body[:_MAX_REPLY_CHARS].rstrip() + "\n\n_[answer truncated]_"
-            logger.info("the answer of session %s was truncated to fit a GitHub comment", session_id)
         try:
+            body = self._truncate_reply(body)
```
Claudia-Anthropica: @wasnertobias [medium] A deleted review can have many per-comment answer files—the task now deliberately includes every actionable comment—but this path combines them all, truncates the result to 60,000 characters, and then records delivery as complete. Answers after the cutoff are silently lost and never retried; split fallback delivery into bounded chunks with durable per-chunk markers/state, and add a regression test whose combined answers exceed the cap.

In `logos/logos-agent/app/sessions.py`, deleted-review fallback truncates a multi-answer payload and marks omitted answers delivered. Split the fallback into bounded, independently reconciled chunks and test a combined response over 60,000 characters.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The previous pagination and identity findings are fixed on this head, and the captured build, lint, and test checks succeeded. One fallback still loses valid output: when a deleted review's combined summary and per-comment answers exceed the 60,000-character cap, the comment is truncated but the session is marked fully delivered. The current tests exercise only short combined fallbacks.

Claudia-Anthropica (changes requested):
@wasnertobias The new head correctly reuses the thread map per delivery, and the captured checks show no failures, though one non-required workspace build remained in progress. The previous fallback defect remains visible: `_post_single_reply` still combines every deleted-review answer, truncates the result at line 2314, appends one session marker, and then records the entire reply as delivered, permanently omitting answers beyond 60,000 characters. Split this fallback into bounded chunks with durable per-chunk markers and state, and cover an over-limit combined response.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in 333b1465d: the fallback no longer truncates. The combined answer is split into chunks bounded by the comment cap (line-aware, hard cut only for a run longer than the cap) and posted in order, each chunk under its own durable mark — the first chunk keeps the answer's original mark, so a delivery that started under the old single-comment scheme still reconciles its posted part on retry. Each chunk is reconciled on its own mark before its POST, so a lost confirmation resumes after the chunks already on the pull request and says none of them twice; delivery is recorded complete only when the last chunk is in. Regression coverage: a combined answer over the cap is posted in two chunks with every character of it present and nothing truncated, a retry that loses the first chunk's confirmation resumes after it and posts only what was still owed, and an overlong single answer is split rather than shortened.

logos/logos-agent/app/sessions.py
```diff
@@ -2224,56 +2285,71 @@ async def _post_reply_once(self, session_id: int) -> None:
...
-            # GitHub refuses a comment above its length limit outright, and
-            # an answer nobody receives is worse than a shortened one on a
-            # thread where somebody is waiting.
-            body = body[:_MAX_REPLY_CHARS].rstrip() + "\n\n_[answer truncated]_"
-            logger.info("the answer of session %s was truncated to fit a GitHub comment", session_id)
         try:
-            url = await self._send_reply(target, body)
+            url = ""
+            for index, chunk in enumerate(self._reply_chunks(body)):
+                if target.startswith("issue:"):
```
Claudia-Anthropica: @wasnertobias [medium] Chunk reconciliation runs only for `issue:` targets, although `_post_single_reply` also serves `review_comment:` targets. If an oversized inline answer posts its first chunk and a later chunk fails, the retry starts from chunk zero and duplicates every previously accepted chunk before completing.

In `logos/logos-agent/app/sessions.py`, `_post_single_reply` splits `review_comment:` answers into multiple posts but protects only `issue:` chunks from replay, so a later-chunk failure duplicates prior chunks on retry. Add durable per-chunk reconciliation for review-comment targets, or retain single-post truncation for that target, and cover a failure after the first successful chunk.

</details>

logos/logos-agent/app/sessions.py
```diff
@@ -2282,6 +2358,348 @@ async def _post_reply_once(self, session_id: int) -> None:
...
+        breaks at a line end where it can, so the split does not tear a
+        line in half; a run longer than the limit is cut anyway.
+        """
+        chunks: list[str] = []
+        while len(body) > _MAX_REPLY_CHARS:
+            cut = body.rfind("\n", 0, _MAX_REPLY_CHARS)
+            if cut <= 0:
+                cut = _MAX_REPLY_CHARS
+            chunks.append(body[:cut].rstrip())
+            body = body[cut:].lstrip("\n")
```
Claudia-Anthropica: @wasnertobias [medium] The splitter is not lossless: `rstrip()` removes trailing spaces and preceding blank lines from one chunk, while `lstrip("\n")` removes the selected newline and any following blank lines from the next. This can alter Markdown formatting, and the new test misses it because it counts only the repeated `a` and `b` characters.

In `logos/logos-agent/app/sessions.py`, `_reply_chunks` deletes whitespace around every line-based split even though fallback delivery promises to preserve the complete answer. Split without consuming boundary characters and add a regression test asserting that concatenating the unmarked chunks reproduces the original body exactly.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The previous 60,000-character truncation is removed, but current chunk delivery can still duplicate earlier chunks for inline review-comment targets after a later POST fails. The splitter also removes boundary whitespace, so over-limit output is not preserved exactly. The snapshot has no failed or required checks; several non-required jobs were still in progress.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in e5cd8b73d: the split consumes no characters at the boundary. A chunk breaks at a line end, the line end stays with the chunk that ends there, and there is no rstrip/lstrip anywhere — the chunks concatenated are the answer, exactly (a run with no line end in the window is hard-cut, which is the only case where a split lands mid-line). Regression coverage: a body over the cap whose boundaries carry trailing spaces and blank lines splits into chunks whose concatenation reproduces the body exactly.

logos/logos-agent/app/sessions.py
wasnertobias: Fixed in e5cd8b73d: the review_comment target got the same durable per-chunk reconciliation as the issue target. Every chunk carries its mark (the first chunk the answer's original mark, the continuations their own), and before each POST the thread the chunk goes into is asked for that mark — the thread map is built once for the delivery and passed in. A retry therefore resumes after the chunks already in the thread. Regression coverage: an oversized inline answer whose second chunk fails posts the first chunk once, and the retry posts only the second — no chunk is duplicated.

Claudia-Anthropica (dismissed):
@wasnertobias The two previously blocking defects are fixed on this head: inline chunk retries reconcile every chunk in its owning thread, and `_reply_chunks` preserves boundary whitespace exactly. The captured build, lint, and test checks succeeded; the duplicate-resolution observation already raised by CodeRabbit is non-blocking.

Claudia-Anthropica (approved):
@wasnertobias The current head correctly deduplicates thread resolution within a delivery sweep, including already-resolved and permanently refused threads, with regression coverage for two answered comments sharing one thread. The earlier pagination, marker reconciliation, and lossless chunk-delivery findings remain fixed, and the captured exact-head build, lint, and test checks all succeeded.

PR #976, opened 2026-09-15
`Logos`: Enable free-VRAM lane replication by default

wasnertobias:
## What

`LOGOS_REPLICATE_ON_FREE_VRAM` is now **on by default**. A hot model auto-scales onto genuinely free VRAM — an extra intra-node lane and/or a cross-provider replica — without an operator pre-loading each copy. The flag remains as an **opt-out**: set `LOGOS_REPLICATE_ON_FREE_VRAM=false` to restore the old single-lane behaviour.

The parse is inverted to the `not in ("0", "false", "no")` idiom already used by `LOGOS_REPLICA_FIRST_EVICTION` directly above it.

## Why

After a workernode redeploy the capacity planner re-spawned **only the first lane** per model; every additional lane on free VRAM was silently skipped with:

```
Skipping additional lane of … extra copies need LOGOS_REPLICATE_ON_FREE_VRAM
and eff=… ≥ replication floor=2.0
```

Observed on deioma (Qwen3.8-27B): single lane while its queue ran deep and free GPUs idled for hours — the surplus lanes had to be loaded **manually** from the UI each time. The feature this flag gates was built precisely for that case (see the production note in the `DEMAND_REPLICATION_FLOOR` comment), so keeping it off by default defeated its purpose.

## Behaviour that does *not* change

The scale-out stays bounded by the same gates as before — flipping the default only removes the manual opt-in:
- sustained demand: `eff ≥ DEMAND_REPLICATION_FLOOR` (2.0)
- free VRAM **without eviction** — a replica never pushes out another model's lane
- at most one additional lane per worker + one cross-provider replica per cycle; idle replicas are reaped by the idle path

## Verification

- `logos-orchestrator` unit suite (Python 3.13): **1594 passed, 1 xfailed** (pre-existing known-gap)
- `pre-commit run --config logos/.pre-commit-config.yaml --all-files` (the CI lint gate): all hooks pass

Tests that exercised the default were updated to reflect it: the `_planner` harness now *pins* the flag off explicitly (it is on by default), and `test_disabled_by_default_returns_empty` → `test_disabled_returns_empty`.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Dynamic intra-node scale-out is now enabled by default, allowing replication onto available VRAM.

* **Configuration**
  * Set `LOGOS_REPLICATE_ON_FREE_VRAM` to `false`, `0`, or `no` to disable the behavior.


logos/logos-orchestrator/tests/unit/capacity/test_best_first_scenarios.py
```diff
@@ -198,7 +198,7 @@ def _planner(providers: List[_MockProvider]) -> CapacityPlanner:
     planner._lane_load_failure_until = {}
     planner._cross_provider_best_first = True
     planner._replica_first_eviction = True
-    planner._replicate_on_free_vram = False  # opt-in; tests turn it on
+    planner._replicate_on_free_vram = False  # on by default; pin off unless a test enables it
```
Claudia-Anthropica, line 201: @wasnertobias [medium] This fixture creates the planner with `__new__` and forces `_replicate_on_free_vram` to `False`; every enabled-path test subsequently forces it to `True`, so none exercises the new constructor default. The production default could therefore be reverted to `false` while these tests remain green. Add a constructor-level test that unsets `LOGOS_REPLICATE_ON_FREE_VRAM` and asserts replication is enabled, along with an explicit `false` opt-out assertion.

In `logos/logos-orchestrator/tests/unit/capacity/test_best_first_scenarios.py`, the modified harness bypasses initialization and does not test the newly enabled default. Add tests that instantiate `CapacityPlanner` through `__init__`, verify an unset `LOGOS_REPLICATE_ON_FREE_VRAM` enables replication, and verify `false` disables it.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The implementation and documentation consistently make free-VRAM replication opt-out, and the captured Logos test and lint checks pass. However, the only test edit bypasses `CapacityPlanner.__init__`, so the new default—the sole behavioral change—has no regression coverage and reverting the default string would leave the suite green.

Claudia-Anthropica (changes requested):
@wasnertobias The constructor now enables free-VRAM replication when the variable is unset, but the modified scenario fixture still bypasses `CapacityPlanner.__init__`, pins the field off, and manually enables it in positive-path tests. No test sets or unsets `LOGOS_REPLICATE_ON_FREE_VRAM` or asserts its parsed constructor value, so reverting the sole behavioral change would leave the suite green. The captured Logos checks pass, but this unresolved regression-coverage gap remains.

Claudia-Anthropica (approved):
@wasnertobias The new constructor-level tests now cover the enabled-by-default behavior, accepted truthy values, explicit opt-out spellings, and whitespace normalization through the real initialization path. The implementation and documentation are consistent, and the captured Logos test and lint checks succeeded; several non-required build jobs were still running in the snapshot.

logos/logos-orchestrator/tests/unit/capacity/test_best_first_scenarios.py
Claudia-Anthropica: @wasnertobias Thanks, the new constructor-level tests now cover the default-on behavior and explicit opt-out values, so this is fixed.

Claudia-Anthropica (approved):
@wasnertobias The implementation and documentation consistently make free-VRAM replication opt-out while retaining the existing demand and no-eviction gates. Constructor-level tests now cover the unset default, enabled and disabled spellings, case normalization, and whitespace handling, and the captured test, lint, and build checks succeeded at this exact head.

PR #977, opened 2026-09-15
`Logos`: Add admin model access page with hosting providers and grant matrix

### LINKED ISSUE #975: `Logos`: Add admin model detail page with provider hosting and team access matrix
## Motivation

On 2026-09-15 we found 7 prod teams (`tep`, `hephaestus`, `skill`, `patrick`, `TeamLisi`, `Hyperion`, `ITG`) holding a model grant for `Qwen/Qwen3-Embedding-8B` (model 38) while lacking the provider grant for `Logos Mac1` (provider 24) — the **only** provider hosting that model. The grant looked valid in every list (all read paths join `team_model_permissions` × `team_provider_permissions` via `model_provider`), so the model simply never routed, and nothing in the admin UI makes this state discoverable: there is no view that shows, per model, which providers host it and which teams actually have both the model grant and a grant for at least one host provider.

Permission grants are two independent dimensions (model, provider) managed via different endpoints — and `PUT /admin/teams/{id}/model-permissions` accepts arbitrary model IDs with no cross-check against the team's provider grants — so orphaned model grants (model without any reachable host provider) are easy to create and currently only findable via raw SQL:

```sql
SELECT t.id, t.name, m.name AS model
FROM teams t
JOIN team_model_permissions tmp ON tmp.team_id = t.id
JOIN models m ON m.id = tmp.model_id
WHERE NOT EXISTS (
    SELECT 1
    FROM model_provider mp
    JOIN team_provider_permissions tpp
      ON tpp.provider_id = mp.provider_id AND tpp.team_id = t.id
    WHERE mp.model_id = m.id
)
ORDER BY t.id, m.name;
```

## Proposal

Add a **model detail page** to the admin UI (extending the existing admin *Models* page, `logos-ui/src/app/features/models/`), reachable by clicking a model row:

1. **Model facts** — name, aliases, description, tags, catalog price, context windows (most of this is already fetched by `POST /api/logosdb/get_models`).
2. **Hosting providers** — all providers from `model_provider` for this model (name, type, privacy level). This is the anchor of the page: it answers "where can this model actually run?".
3. **Team access matrix** — one row per team, showing:
   - model grant (`team_model_permissions`) ✔/✘
   - provider grant ✔/✘ per hosting provider (`team_provider_permissions`)
   - effective access = model grant ∧ ≥1 host-provider grant — **highlight orphaned grants** (model grant, no host provider) and, as a secondary signal, teams that have a host provider but no model grant.
4. **Custom-permission API keys** (section or tab) — keys with `use_custom_permissions = true` that have a model or provider grant touching this model (`api_key_model_permissions` / `api_key_provider_permissions`), with the same effective-access highlight, since team-level grants are ignored for these keys.
5. *Nice to have* — recent usage per hosting provider (count / last request from `log_entry`), to see whether a grant is actually used.

The cross-join of (model, host providers, teams, both permission tables, custom-permission keys) does not fit the existing per-team endpoints (`GET /api/admin/teams/{id}/model-permissions` / `provider-permissions` would mean N+1 fetches per team), so add a dedicated **admin-only webservice endpoint**, e.g. `GET /api/admin/models/{modelId}/access`, returning the matrix in one query.

For v1 the page can be **read-only**; granting/revoking stays where it is (team detail page). A "fix" affordance (one click to add the missing provider grant) would be a natural follow-up.

## Acceptance criteria

- [ ] Admin *Models* page rows link to a model detail page.
- [ ] Detail page lists all hosting providers of the model.
- [ ] Team matrix shows model grant, per-provider grants, and effective access; orphaned grants are visually flagged.
- [ ] Custom-permission keys with grants for the model are listed with the same flags.
- [ ] New webservice endpoint is admin-only (`@PreAuthorize` LOGOS_ADMIN) and serves the matrix in a single query.
- [ ] Running the SQL above against prod returns no orphaned grants while the page shows all of them (verification of parity).

wasnertobias:
## What

Adds an admin-only **Access tab** on the model details view that answers the question no existing view could: for one model, *which providers host it* and *which teams/keys actually have both the model grant and a grant for at least one host provider*.

This makes orphaned grants discoverable — the state behind the 7 prod teams found on 2026-09-15 holding a grant for `Qwen/Qwen3-Embedding-8B` (model 38) while lacking the provider grant for `Logos Mac1` (provider 24), its only host. The grant looked valid in every per-team list, but the model could never route.

Closes #975

## Screenshots

_Model management page (full page) — model rows open the details view on the new Access tab:_

models list

_Model Access tab for `Qwen/Qwen3-Embedding-8B` (full page — header, tab bar and every section): `alpha` holds the model grant but neither host-provider grant and is flagged **orphaned**; `beta` is **effective**. Custom-permission keys below show the same flags (orphaned `alpha-ci`, inactive `local admin key` / `beta-ci`):_

model access page

_Same tab for a multi-host model (`gpt-4o`, full page) — note the **Prices** tab, shown because the model has a cloud provider:_

model access page gpt-4o

_Mobile view (375px) of the **full** access page, top to bottom — header, tabs, hosting providers, team matrix and keys; the shared tables render as per-cell `data-label` cards labeled with the provider names:_

mobile team matrix

## How

**Webservice** — new `GET /admin/models/{modelId}/access` (`ModelAccessController`, `@PreAuthorize` LOGOS_ADMIN only), served by `ModelAccessService`:
- **Model facts** — name, aliases, tags, description, catalog price (in/out per 1M tokens), context window (`max_input_tokens`), last used. Reuses the existing pricing subquery shape (single-model twin of `findAllWithPricing`).
- **Pricing determinism** — the most-recently-routed-provider LATERAL in all three `ModelRepository` pricing queries now breaks ties on `log_entry.id` (serial PK), so the picked provider — and the price shown for it — is deterministic when requests share the newest `timestamp_request`.
- **Hosting providers** — all `model_provider` rows for the model, each with total request count and last request from `log_entry` (the nice-to-have #5, so a grant's real usage is visible).
- **Team access matrix** — every team crossed with every host provider in **one native query** over `teams × model_provider × team_model_permissions × team_provider_permissions`. Effective access = model grant ∧ ≥1 host-provider grant, computed server-side.
- **Custom-permission keys** — `use_custom_permissions = true` keys with a model or host-provider grant touching the model, same effective-access flags (team grants are ignored for these).

No N+1: the cross-join that the per-team `GET/PUT /admin/teams/{id}/…` endpoints can't express is one query.

**UI** — third **Access** tab on the existing model details view (`/models/:id/details?tab=access`), modeled on the team-detail layout:
- Hosting-providers table (with request counts / last request), team matrix, custom-permission-key matrix.
- **Orphaned grants** (model grant, no host provider) are row-tinted + badged — the broken state that motivated this issue. Provider-only grants (host provider, no model grant) are flagged as the secondary signal.
- Team-provider grants are **checkboxes** in the team matrix — one click grants, another revokes. Both directions are single atomic endpoints (`POST`/`DELETE /admin/teams/{id}/provider-permissions/{providerId}`) that never read or replace the team's full grant set, so a stale snapshot cannot clobber a concurrent admin edit. The grant is a pure upsert; the revoke is a single-row delete followed by the usual model-grant cascade in the same transaction, so revoking a team's last host provider for a model prunes the now-unreachable model grant instead of recreating the orphaned state the page diagnoses. The matrix reloads afterwards. All three team-provider-permission mutation paths (full-set PUT, atomic add, atomic remove) first take a per-team, transaction-scoped advisory lock, so a concurrent mutation of the same team cannot commit between another mutation's targeted row write and its model-grant cascade.
- Models-page rows now open the details view directly on the Access tab.

## Testing

- New `ModelAccessControllerTest` (9 cases, MockMvc + Testcontainers, same seed/cleanup as `PermissionControllerTest`): admin-only 403, unknown-model 404, effective team access, orphaned team grant, model-without-host-provider, custom-permission key with/without effective access, provider-only key.
- `PermissionControllerTest` extended for the atomic add **and remove** endpoints: role gates, append/remove keep the team's other grants, idempotency in both directions, the add path never re-runs the model-grant cascade, and the remove path prunes model grants orphaned by the revoke while keeping models still reachable through another provider. Plus a regression test pinning the advisory-lock contract: holding the per-team lock in one transaction blocks a fresh connection from locking the same team's key while another team's key stays free.
- Full webservice suite: **517 tests, 0 failures** (after merging latest `main`).
- UI `npm run build` passes (only pre-existing bundle-budget warnings).

## Acceptance criteria
- [x] Admin *Models* page rows link to a model detail page.
- [x] Detail page lists all hosting providers of the model.
- [x] Team matrix shows model grant, per-provider grants, and effective access; orphaned grants are visually flagged.
- [x] Custom-permission keys with grants for the model are listed with the same flags.
- [x] New webservice endpoint is admin-only (`@PreAuthorize` LOGOS_ADMIN) and serves the matrix in a single query.
- [ ] Running the prod SQL returns no orphaned grants while the page shows all of them — needs a prod DB, left as a manual verification step.

## Summary by CodeRabbit

- **New Features**
  - Added an admin-only **Access** tab to model details.
  - View model information, hosting providers, usage details, team permissions, and custom-permission API key access.
  - Access indicators show effective, orphaned, provider-only, and inactive-key states.
  - Selecting a model now opens its details directly on the Access tab.
  - Added loading and error states for access information.

- **Documentation**
  - Updated UI contribution guidance to require desktop and 375px mobile screenshots, including responsive table examples.


Claudia-Anthropica (approved):
@wasnertobias The admin boundary is enforced in both UI routing and the Spring controller, and the repository/service joins correctly handle team, host-provider, custom-key, and no-host cases. The integration coverage exercises authorization, missing models, effective access, and orphaned/provider-only grants; the captured UI export and lint checks passed, with some non-required checks still running.

wasnertobias:
please add a screenshot to the pr description. this is a hard requirement for all ui related prs. add this to your AGENTS.md as well in this PR.

logos/logos-ui/src/app/features/model-access/model-access.html
```diff
@@ -0,0 +1,224 @@
...
+            <div role="cell">
+              <i
+                class="pi {{ t.model_grant ? 'pi-check' : 'pi-times' }}"
+                [class.check--yes]="t.model_grant"
+                [class.check--no]="!t.model_grant"
+                [attr.aria-label]="'Model grant: ' + (t.model_grant ? 'yes' : 'no')"
+              ></i>
+            </div>
+            @for (g of t.provider_grants; track g.provider_id) {
+              <div role="cell">
```
Claudia-Anthropica: @wasnertobias [medium] Below 768px, the shared data table hides its header and displays labels only for cells carrying `data-label`; none of these new table cells set it. The team and key matrices therefore become unlabeled sequences of checkmarks, so users cannot tell which grant belongs to the model or each provider. Add labels to every fixed cell and bind each dynamic grant cell to its corresponding provider name.

In `logos/logos-ui/src/app/features/model-access/model-access.html`, the new table cells omit `data-label`, so the shared mobile layout hides the headers and leaves values and grant icons unlabeled. Add `data-label` attributes to every provider, team, and key cell, using the actual hosting-provider name for each dynamic grant column.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The endpoint authorization and permission-folding logic are consistent with the existing access rules. However, the new matrices lose their column labels at the shared mobile breakpoint, making the grant states unreadable. The captured `Test Webservice` check is red but provides only a generic exit code, so it cannot be attributed and is not a separate finding.

wasnertobias:
the screenshots shall be in the pr description only!

 I dislike what i see on the screenshots, right now the model details page have two tabs "Complete Logs" and "Performance" and I want you to add another tab instead of completely reworking it!

logos/logos-ui/src/app/features/models/models.ts
```diff
@@ -182,8 +182,9 @@ export class Models implements OnInit {
     return `Not used for ${daysSince(iso)} days`;
   }
 
-  openReport(model: Model): void {
-    this.router.navigate(['/models', model.id, 'details']);
+  /** Model row click: the access page (hosting providers, team/key grant matrix). */
+  openModel(model: Model): void {
+    this.router.navigate(['/models', model.id, 'access']);
```
Claudia-Anthropica: @wasnertobias [medium] This replaces the established model-details destination with a standalone access page, while the existing details component already owns the Complete Logs and Performance tabs and the requested design is a third tab there. Keep model-row navigation on the details view and integrate the access content into its tab bar.

In `logos/logos-ui/src/app/features/models/models.ts`, `openModel` routes model rows to a standalone access page instead of the requested third tab in the existing model-details view. Restore navigation to `/models/:id/details` and integrate model access as a third tab in the model-error-report component.

</details>

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ModelAccessService.java
```diff
@@ -0,0 +1,173 @@
...
+            boolean modelGrant = Boolean.TRUE.equals(key.getModelGrant());
+            return new KeyAccessDTO(
+                key.getKeyId(),
+                key.getKeyName(),
+                key.getIsActive(),
+                key.getTeamId(),
+                key.getTeamName(),
+                modelGrant,
+                grantsDTO,
+                effective(modelGrant, grantsDTO));
```
Claudia-Anthropica: @wasnertobias [medium] Custom-key `effectiveAccess` ignores `isActive`, so an inactive key holding both grants is returned and rendered as effective even though request authentication uses `findByKeyValueAndIsActiveTrue` and rejects it. Include active state in this calculation and cover the inactive-key case in the controller test.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ModelAccessService.java`, inactive custom-permission keys can be marked as having effective access. Require `isActive` when computing key effective access and add an integration assertion for an inactive key with both grants.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The prior [medium] mobile-matrix finding remains: the shared table hides headers below 768px, while the new cells still provide no `data-label` values. The model row also still opens a standalone access page instead of integrating access as the requested third model-details tab, and inactive custom keys can be labelled effective even though authentication rejects them. [low] The three screenshot PNGs remain committed under `logos/docs/screenshots/` despite the latest direction to keep screenshots only in the PR description. The captured CI snapshot has no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The latest commit fixes the mobile table labels, and the screenshot assets are no longer part of the diff. Two [medium] issues remain: model rows still navigate to the standalone `/access` page instead of adding Access as the requested third details tab, and custom-key `effectiveAccess` still ignores `isActive`, so disabled keys can be reported as usable. The captured CI snapshot has no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias Two unresolved [medium] findings remain on this head: `models.ts` still routes model rows to the standalone `/access` page instead of integrating Access into the existing details tabs, and `ModelAccessService` still computes key `effectiveAccess` without considering `isActive`. The earlier mobile-label issue is fixed, and the captured CI snapshot has no failed or pending checks.

logos/logos-ui/src/app/features/models/models.ts
wasnertobias: Fixed in 27feab626: model rows now navigate to the model details view (/models/:id/details) on a new third Access tab next to Complete Logs and Performance, deep-linkable via ?tab=access. The standalone route is gone and the embedded ModelAccess component dropped its own page header. Updated screenshots in the PR description show the tab layout (desktop + mobile).

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ModelAccessService.java
wasnertobias: Fixed in 27feab626: key effective access now requires is_active in ModelAccessService (an inactive key can never authenticate, so it never has effective access even with both grants). Added controller test getModelAccess_inactiveKeyIsNeverEffective covering the inactive key with both grants — suite is green (9/9).

wasnertobias:
The new description does not show any images. Fix it by using gh on commandline.

logos/logos-ui/src/app/features/model-access/model-access.ts
```diff
@@ -0,0 +1,117 @@
...
+      this.loadError.set(true);
+    } finally {
+      this.loading.set(false);
+    }
+  }
+
+  // ── Matrix helpers ───────────────────────────────────────────────────────
+  /** Model grant without any host-provider grant: can never route. */
+  isOrphaned(entry: TeamAccess | KeyAccess): boolean {
+    return entry.model_grant && !entry.effective_access;
```
Claudia-Anthropica: @wasnertobias [medium] After the backend fix makes every inactive key ineffective, this predicate marks an inactive key that has both model and provider grants as “orphaned.” The Access tab therefore reports a missing-provider configuration when the only blocker is key inactivity. Compute orphaned status from the grant columns (`model_grant && !provider_grants.some(g => g.granted)`) rather than from `effective_access`.

In `logos/logos-ui/src/app/features/model-access/model-access.ts`, `isOrphaned` misclassifies inactive keys that possess both required grants because it derives the label from `effective_access`. Define orphaned as a model grant with zero granted hosting providers, leaving inactivity to the existing active-state indicator.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The two previously blocking issues are fixed: Access is now the third model-details tab, and inactive keys cannot have effective access. One [medium] UI correctness issue remains: inactive keys with both grants are still classified as orphaned because the UI equates “not effective” with “no host-provider grant.” The captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The prior [medium] finding remains on this exact head: `isOrphaned` still derives orphan status from `effective_access`, so an inactive key holding both required grants is incorrectly reported as missing a host-provider grant. The captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The existing [medium] finding remains on this exact head: `isOrphaned` treats every model grant with `effective_access == false` as missing a host-provider grant, although the backend deliberately makes inactive keys ineffective even when both grants are present. This renders an inactive, fully granted key with the “orphaned” badge and row tint, so derive orphan status from `provider_grants` while leaving inactivity as the separate blocker; the captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The existing [medium] finding remains on this exact head: `isOrphaned` derives orphan status from `effective_access`, while the backend deliberately makes inactive keys ineffective even when both model and provider grants are present. This incorrectly badges and tints an inactive, fully granted key as orphaned rather than merely inactive; the captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The existing [medium] finding remains on this exact head: `isOrphaned` equates an ineffective model grant with a missing provider grant, although inactive keys are deliberately ineffective even when `provider_grants` contains a grant. This still badges and tints a fully granted inactive key as orphaned; derive orphan state from the grant columns and leave inactivity as the separate blocker. The captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The existing [medium] finding remains: `isOrphaned` derives orphan status from `effective_access`, while the backend deliberately makes inactive keys ineffective even when both model and provider grants are present. This still labels and tints a fully granted inactive key as orphaned; derive the state from `model_grant` and `provider_grants` instead, leaving inactivity as the separate blocker. The captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The existing [medium] finding remains: `isOrphaned` derives orphan status from `effective_access`, so an inactive key with both model and provider grants is still incorrectly labelled and tinted as orphaned. Derive orphan state from the grant columns instead; the captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The existing [medium] finding remains: `isOrphaned` derives orphan status from `effective_access`, so an inactive key holding both model and provider grants is incorrectly labelled and tinted as orphaned. Derive orphan state from the grant columns instead; the captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The existing [medium] finding remains on this exact head: `isOrphaned` derives orphan status from `effective_access`, so an inactive key holding both model and provider grants is incorrectly labelled and tinted as orphaned. Derive orphan state from `model_grant` and `provider_grants` instead, leaving inactivity as the separate blocker; the captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The existing [medium] finding remains: `isOrphaned` derives orphan status from `effective_access`, so an inactive key holding both model and provider grants is incorrectly labelled and tinted as orphaned. Derive orphan state from `model_grant` and `provider_grants`, leaving inactivity as the separate blocker; the captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The existing [medium] finding remains on this exact head: `isOrphaned` derives orphan status from `effective_access`, so an inactive key holding both model and provider grants is incorrectly labelled and tinted as orphaned; derive orphan status from the grant columns instead. The captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The unresolved [medium] finding remains on this exact head: `isOrphaned` still derives orphan status from `effective_access`, mislabeling inactive keys that hold both model and provider grants. The captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The existing [medium] thread remains valid on this exact head: `isOrphaned` derives orphan status from `effective_access`, while the backend deliberately makes inactive keys ineffective even when both model and provider grants are present, so the UI incorrectly labels and tints them as missing a host-provider grant. Derive orphan status from `model_grant` and `provider_grants`, leaving inactivity separate; the captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The existing [medium] finding remains on this exact head: `isOrphaned` still returns true for any model grant with `effective_access == false`, while the backend deliberately makes inactive, fully granted keys ineffective, so the Access tab incorrectly labels them as missing a provider grant. Derive orphan state from `model_grant` and the absence of any granted `provider_grants`; the captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The unresolved [medium] finding remains on this exact head: `isOrphaned` still uses `model_grant && !effective_access`, while the backend deliberately makes inactive keys ineffective even when they hold a provider grant. This falsely badges and tints an inactive, fully granted key as orphaned; derive orphan status from `model_grant` and the absence of any granted `provider_grants`. The captured CI snapshot is complete with no failed or pending checks.

Claudia-Anthropica (changes requested):
@wasnertobias The unresolved [medium] thread remains valid on this exact head: `isOrphaned` still returns true for every model grant with `effective_access == false`, while the backend deliberately makes an inactive key ineffective even when its `provider_grants` contains a granted host. This falsely labels and tints a fully granted inactive key as orphaned; derive orphan state from the model/provider grant columns and keep inactivity separate. The complete CI snapshot contains no failed or pending checks.

wasnertobias:
In the access tab there is a lot of clutter, you shall not concentrate on description/tags/aliases but on the problem you are solving solely. I want to be quickly able to add access to specific provider to a given team with one click. Remember what the exact problem was and how you did have to manually analyse it as there is no UI yet. That's what you are trying to solve, nothing more, nothing less.

logos/logos-ui/src/app/features/model-access/model-access.ts
wasnertobias: Fixed in 97db5df48: isOrphaned now derives orphan status from the grant columns directly (`model_grant && !provider_grants.some(g => g.granted)`) instead of from effective_access. An inactive key holding both grants is no longer badged 'orphaned' — inactivity stays with the existing Active column / 'inactive' chip. No screenshot change needed: the captured scenarios contain no inactive key with full grants, so the rendered matrices are unchanged (and the gist now serves the images as proper PNG binaries, fixing the rendering in the PR description).

logos/logos-ui/src/app/features/model-access/model-access.html
```diff
@@ -0,0 +1,209 @@
...
+            <div role="cell" data-label="Model">
+              <i
+                class="pi {{ t.model_grant ? 'pi-check' : 'pi-times' }}"
+                [class.check--yes]="t.model_grant"
+                [class.check--no]="!t.model_grant"
+                [attr.aria-label]="'Model grant: ' + (t.model_grant ? 'yes' : 'no')"
+              ></i>
+            </div>
+            @for (g of t.provider_grants; track g.provider_id; let gi = $index) {
+              <div role="cell" [attr.data-label]="a.providers[gi]?.name ?? 'Provider'">
```
Claudia-Anthropica, line 86: @wasnertobias [medium] The hosting-provider cells are still passive icons, and the hint explicitly sends admins to the team page. An orphaned team therefore cannot be repaired from this model view, so the requested one-click “grant this provider to this team” workflow is absent. Make each missing hosting-provider grant actionable, preserving the team's complete existing provider-permission set and reloading the matrix with loading/error feedback.

In `logos/logos-ui/src/app/features/model-access/model-access.html`, the team/provider matrix renders missing grants as passive icons, so an admin cannot grant the selected hosting provider to that team from this page. Turn each missing team-provider grant into a one-click action that adds that provider without replacing any other provider grants, then refreshes the matrix and reports failures.

</details>

logos/logos-ui/src/app/features/model-access/model-access.html
```diff
@@ -0,0 +1,209 @@
...
+      <i class="pi pi-spin pi-spinner"></i>
+      <span>Loading model access…</span>
+    </div>
+  } @else if (loadError()) {
+    <app-error-message variant="banner" message="Failed to load model access, please refresh." />
+  } @else if (access(); as a) {
+
+    <!-- ── Model facts ─────────────────────────────────────────────────── -->
+    <div class="facts">
+      <div class="fact">
```
Claudia-Anthropica: @wasnertobias [medium] The Access tab still opens with six generic model-fact cards, including the description, tags, and aliases that the latest feedback explicitly asked to remove. This pushes the provider/team remediation matrix below unrelated metadata and leaves the reported clutter intact; remove this block so the tab leads directly with hosting providers and access repair.

In `logos/logos-ui/src/app/features/model-access/model-access.html`, the Access tab still leads with description, tags, aliases, pricing, context, and usage cards despite the requirement to focus solely on diagnosing and repairing access. Remove the model-facts block and lead with the hosting-provider and team-access workflow.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The prior inactive-key orphan classification is fixed on this head. However, the latest requested workflow is still missing: admins cannot grant a hosting provider to an orphaned team from this page, and the tab still leads with the metadata cards requested for removal. The snapshot has no failed checks and no required checks, although several non-required jobs were still running.

Claudia-Anthropica (changes requested):
@wasnertobias The newest commit changes only `logos/AGENTS.md`, which is outside review scope, so both active [medium] findings remain unchanged on this exact head. `model-access.html` still renders the six-card model-facts block before the access workflow, and its team/provider cells remain passive icons while the hint sends admins to the team page, so the requested focused one-click grant workflow is absent. The captured snapshot has no failed or required checks, although several non-required jobs—including Test Webservice—were still running.

logos/logos-ui/src/app/features/model-access/model-access.html
wasnertobias: Fixed in 9e22e7247: the six model-fact cards (description, tags, aliases, catalog price, context window, last used) and their styles are gone, together with the now-unused formatPrice/formatTokens helpers — the tab leads directly with the hosting-providers section and the access-repair matrix. Updated screenshots in the PR description.

logos/logos-ui/src/app/features/model-access/model-access.html
wasnertobias: Fixed in 9e22e7247: every missing hosting-provider grant in the team matrix is now a 'grant' button. It re-reads the team's complete provider-permission set (GET /api/admin/teams/:id/provider-permissions), appends only the missing provider, writes the full set back (PUT), and reloads the matrix with the usual loading state — a single in-flight guard (granting) disables all grant buttons meanwhile, and a failure surfaces as an inline error notice. The hint text no longer sends admins to the team page. Desktop and mobile screenshots in the PR description show the buttons and the orphaned/effective rows.

Claudia-Anthropica (changes requested):
@wasnertobias The one-click provider-grant workflow and model-fact cleanup are fixed on this head. The retained [high] serialization finding remains valid: `JacksonConfig` supplies a bare `ObjectMapper`, so the new camelCase record fields are not converted to the snake_case properties consumed by the UI. The complete CI snapshot contains no failed or pending checks.

logos/logos-ui/src/app/features/model-access/model-access.html
Claudia-Anthropica: @wasnertobias Thanks, this is fixed now: missing provider grants are actionable while the team's existing provider permissions are preserved, and the matrix reloads with error feedback.

wasnertobias:
There are merge conflicts and before the PR the model details page have three tabs "Complete Logs" and "Performance" and "Price" and I want you to add another tab, not remove the existing ones

Claudia-Anthropica (dismissed):
@wasnertobias The current head correctly integrates the Access tab, preserves existing provider grants during one-click repair, and covers the model/team/key matrix with controller tests. The prior serialization concern is not supported: Spring Boot 4.1's Jackson 3 HTTP stack applies the configured snake_case strategy, while the bare `com.fasterxml` mapper is a separate legacy bean, and the endpoint tests assert the exact snake_case fields consumed by the UI. The complete captured CI snapshot has no failed or pending checks.

logos/logos-ui/src/app/features/model-access/model-access.ts
```diff
@@ -0,0 +1,142 @@
...
+   * with the usual loading/error feedback.
+   */
+  async grantProviderToTeam(teamId: number, providerId: number): Promise<void> {
+    if (this.granting()) return;
+    this.granting.set(true);
+    this.grantError.set(null);
+    try {
+      const current = await this.teamService.getTeamProviderPermissions(teamId);
+      const next = [...new Set([...current, providerId])];
+      await this.teamService.setTeamProviderPermissions(teamId, next);
```
Claudia-Anthropica: @wasnertobias [medium] This GET-then-PUT uses a potentially stale permission snapshot, while `setTeamProviderPermissions` deletes and recreates the team's entire provider set. If another admin adds or revokes a provider between these requests, this click can silently undo that update; losing a provider can also trigger cascading model-grant deletion. Use an atomic backend operation that adds only this team-provider grant, then reload the matrix.

In `logos/logos-ui/src/app/features/model-access/model-access.ts`, `grantProviderToTeam` performs a GET followed by a full-set PUT, so a concurrent permission edit can be overwritten and may cascade-delete model grants. Add and call an atomic backend operation that inserts only the requested team-provider grant without replacing other grants, then reload the matrix.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The merge preserves Complete Logs, Performance, Access, and Prices, and the prior serialization and inactive-key concerns are resolved. One [medium] data-integrity issue remains in the one-click grant path: its read/replace sequence can overwrite a concurrent provider-permission update; the snapshot has no failed or required checks, though several non-required jobs remain pending.

logos/logos-ui/src/app/features/model-access/model-access.ts
wasnertobias: Fixed in 6e787595f: the one-click repair no longer works from a client-side snapshot. The access tab now calls a new atomic endpoint — POST /admin/teams/{teamId}/provider-permissions/{providerId} (LOGOS_ADMIN) — which is a single INSERT ... ON CONFLICT DO NOTHING on the (team_id, provider_id) composite key: idempotent, adds only this one grant, leaves the team's other grants and the model-grant cascade untouched, so a concurrent admin edit can neither be clobbered nor trigger cascading model-grant deletion. The UI drops the GET+PUT round-trip and reloads the matrix after the call. New controller tests cover the role gate, append-keeps-other-grants, idempotency, and that the add path never re-runs the cascade (suite 23/23 green).

Claudia-Anthropica (dismissed):
@wasnertobias The concurrency finding is resolved on this exact head: the Access tab now calls an admin-only single-grant endpoint backed by an idempotent `INSERT ... ON CONFLICT DO NOTHING`, so it no longer replaces the team's full provider set or triggers the pruning cascade. The tests cover role enforcement, preservation of existing grants, idempotency, and retention of model grants. The captured snapshot has no failed or required checks; several non-required image-build jobs remain queued or in progress.

wasnertobias:
The current screenshots do not show the full page, which is not acceptable. add this to your agents.md and fix it!

wasnertobias:
Re-captured all PR screenshots as **full pages**: every shot now shows the page header, the tab bar and the complete scrollable content — no clipped viewports (desktop 1600px; mobile 375px top to bottom). The full-page requirement, including the scroll-container unlock technique for the app's inner scroll, is now documented in `logos/AGENTS.md` (1f46ff675) so future UI PRs don't ship clipped shots.

Claudia-Anthropica (dismissed):
@wasnertobias The exact head preserves all existing model-detail tabs and correctly implements the access matrix, inactive-key handling, responsive labels, and atomic single-provider grant workflow with focused controller coverage. No blocking findings remain; the captured CI has no failures and no required checks, while several non-required jobs are still unfinished.

wasnertobias:
1.) The screenshots are pixelated
2.) I only have ui interaction to grant access, but none to quickly remove it (agin). It does not have to be a spefic button, I would prefer to have a checkbox.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/PermissionService.java
```diff
@@ -79,4 +79,26 @@ public void setTeamProviderPermissions(int teamId, List<Integer> providerIds) {
...
+    }
+
+    /**
+     * Atomic counterpart of {@link #addTeamProviderPermission}: removes a
+     * single team-provider grant (one-click de-provisioning from the model
+     * access page) — idempotent, touches nothing else, no cascade.
+     */
+    @Transactional
+    public void removeTeamProviderPermission(int teamId, Integer providerId) {
+        teamProviderRepo.revoke(teamId, providerId);
```
Claudia-Anthropica, line 130: @wasnertobias [medium] Revoking a provider here bypasses the pruning that `setTeamProviderPermissions` performs via `deleteCascadeForTeam`. If this was the team's last granted host for a model, the checkbox removes effective access but leaves its model permission behind, immediately recreating the orphaned grant this page is intended to diagnose. Run the cascade in this transaction after the targeted delete; it already preserves models reachable through another granted provider.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/PermissionService.java`, `removeTeamProviderPermission` deletes the team-provider row without invoking `teamModelRepo.deleteCascadeForTeam(teamId)`, leaving orphaned model grants when the revoked provider was their last reachable host. Run the cascade after the atomic revoke in the same transaction and update the removal test to verify unreachable model grants are pruned while grants for models reachable through another provider remain.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The checkbox and atomic single-row removal implement the requested UI, but the removal path introduces one [medium] permission-integrity regression: it skips the existing model-grant cascade and can create orphaned grants. The captured CI snapshot has no failed or required checks; several non-required jobs were still in progress.

wasnertobias:
Both fixed in f539ff614:

1. **Pixelation** — all four screenshots were re-captured at `deviceScaleFactor: 2` (Retina: 3200px-wide desktop, 750px-wide mobile) with the same full-page technique, so text and icons are now crisp at native zoom. Same gist, same filenames/URLs, so the PR description links are unchanged.

2. **Quick removal via checkbox** — the team-matrix provider cell is now a checkbox that toggles in both directions: unchecked → grant, checked → revoke. The revoke path is a new atomic endpoint `DELETE /admin/teams/{teamId}/provider-permissions/{providerId}` (LOGOS_ADMIN) — a single-row `DELETE`, idempotent, no full-set PUT, so the team's other grants and the model-grant cascade are left exactly as with the atomic add: a concurrent admin edit can neither be clobbered nor trigger cascading model-grant deletion. Two new controller tests cover the role gate and that a removal keeps the team's other grants while never re-running the cascade (PermissionControllerTest 16/16, ModelAccessControllerTest 9/9).

Screenshots in the PR description show the checkboxes in all three states (orphaned team = unchecked, effective team = checked, no-grant team = unchecked).

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/PermissionService.java
wasnertobias: Fixed in 57c83be43: `removeTeamProviderPermission` now runs `teamModelRepo.deleteCascadeForTeam(teamId)` in the same transaction right after the targeted delete — the same pruning the full-set PUT performs. Revoking a team's last host provider for a model prunes that model grant immediately, so the checkbox can no longer recreate the orphaned state the page diagnoses; the cascade still only removes models unreachable through any of the team's remaining grants.

The targeted delete itself stays snapshot-safe (no read/replace of the team's other provider grants), so the add/remove split is: add = pure upsert, remove = single-row delete + cascade.

Tests: `removeTeamProviderPermission_removesOnlyThisGrant` now revokes the non-hosting grant and asserts the still-reachable model grant survives the cascade; new `removeTeamProviderPermission_prunesOrphanedModelGrants` seeds provider 6002 hosting model 5002, revokes the only host of 5001, and asserts 5001 is pruned while 5002 (reachable via 6002) is kept. PermissionControllerTest 17/17, ModelAccessControllerTest 9/9.

wasnertobias:
✅ Approved — I reviewed the latest access-matrix implementation and follow-up fixes; I found no remaining actionable correctness or authorization issue.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/PermissionService.java
```diff
@@ -79,4 +79,34 @@ public void setTeamProviderPermissions(int teamId, List<Integer> providerIds) {
...
+     * the model-grant cascade in the same transaction: the revoke itself is
+     * what can orphan model grants (revoking the last host provider of a
+     * model), and the cascade only prunes models no longer reachable through
+     * any of the team's remaining grants — exactly what the full-set PUT
+     * does.
+     */
+    @Transactional
+    public void removeTeamProviderPermission(int teamId, Integer providerId) {
+        teamProviderRepo.revoke(teamId, providerId);
+        teamModelRepo.deleteCascadeForTeam(teamId);
```
wasnertobias, line 131: This delete-plus-cascade can race with a concurrent provider-grant insertion under READ COMMITTED. Another admin can add a provider grant between these statements (or commit immediately afterward), while this cascade still removes the model grant. The insertion then commits with a provider grant that does not make its reachable model grant present. Please serialize the team’s provider-permission mutations or make the cascade atomic/serializable with retry, and add a concurrent revoke/grant regression test.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/repository/ModelRepository.java
```diff
@@ -14,6 +15,60 @@ public interface ModelRepository extends JpaRepository<Model, Integer> {
...
+               (SELECT MAX(le.timestamp_request)
+                FROM log_entry le
+                WHERE le.model_id = m.id
+               ) AS last_used_at
+        FROM models m
+        LEFT JOIN LATERAL (
+            SELECT le.provider_id
+            FROM log_entry le
+            WHERE le.model_id = m.id AND le.provider_id IS NOT NULL
+            ORDER BY le.timestamp_request DESC
```
wasnertobias: Please add a deterministic tie-breaker such as `le.id DESC` after `timestamp_request DESC`. Requests can share the same timestamp, so the selected provider and therefore the displayed pricing can otherwise vary nondeterministically between executions.

wasnertobias (commented):
❌ Rejected — the targeted revoke plus reachability cascade is not concurrency-safe under READ COMMITTED. A concurrent provider-grant insertion can commit after the cascade observes no remaining providers, leaving the new provider grant without the model grant it should make reachable. Please serialize a team’s provider-permission mutations or make the cascade atomic/serializable with retry; add a concurrent revoke/grant regression test.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/PermissionService.java
wasnertobias: Fixed in 10cc4a9a6. All three team-provider-permission mutation paths (full-set PUT, atomic add, atomic remove) now acquire a per-team, transaction-scoped `pg_advisory_xact_lock` first — key is `PermissionService.teamProviderPermsLockKey(teamId)` (stable "TEAMPERM" base + teamId, disjoint from other advisory-lock namespaces), acquired via `TeamProviderPermissionRepository.lockTeamProviderPermissions`. Under READ COMMITTED a concurrent mutation of the same team can no longer commit between another mutation's targeted row write and its model-grant cascade, since it blocks on the lock until the holder commits.

Added the concurrent revoke/grant regression test you asked for: `PermissionControllerTest.teamProviderPermissionMutations_areSerializedPerTeam` holds the very lock the mutation paths take, then proves from a physically fresh (autocommit) connection that `pg_try_advisory_xact_lock` on the same team's key blocks while another team's key stays free — pinning the per-team serialization contract all three paths rely on. Full webservice suite green (517 tests).

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/repository/ModelRepository.java
wasnertobias: Fixed in 10cc4a9a6: the LATERAL provider pick now orders by `le.timestamp_request DESC, le.id DESC` in all three pricing queries (`findWithPricingById`, `findAllWithPricing`, `findAllWithPricingForUser`). `log_entry.id` is the serial PK, so requests sharing the newest `timestamp_request` deterministically resolve to the highest log-entry id — the picked provider, and the price shown for it, no longer flip between reads. The index from migration 036, `(model_id, timestamp_request DESC) INCLUDE (provider_id) WHERE provider_id IS NOT NULL`, still answers the lateral; the `id` tiebreak only applies within rows sharing one timestamp (noted in the comment above `findAllWithPricing`). Full webservice suite green (517 tests).

PR #981, opened 2026-09-15
`Logos`: Harden the Python API surface

### LINKED ISSUE #978: `Logos`: Harden Python APIs
Python APIs are partly internal (INTERNAL_SECRET needed) but also partly reachable with any Logos API key. The second case shall be limited as much as possible. So either completely remove those and/or add generous rate limits (per IP) through ngnix proxy, which is put in front.

Security aspect is relevant. Other hardening ideas are relevant as well and shall be integrated as well.

wasnertobias:
Closes #978.

## What the audit found

The orchestrator's API surface is tiered — but the tiers leaked:

| Surface | Auth before | After |
|---|---|---|
| `/api/logosdb/scheduler_state` (publicly routed!) | **any Logos API key** | `LOGOS_INTERNAL_SECRET` (like `/internal/*`), public Traefik router removed |
| `POST /logosdb/connect_model_provider` | **none** — DB write + pipeline refresh | removed (webservice owns it: Keycloak JWT + `logos_admin`) |
| `POST /logosdb/update_provider_sdi_config` | **none** — DB write | removed (no callers anywhere) |
| `POST /logosdb/get_ollama_vram_stats` + `OPTIONS` (CORS `*`) | any Logos API key | removed (in prod the path is served by the webservice with proper auth; only dev/benchmark tooling used the copy, and it degrades gracefully) |
| `GET /forward_host` | **none** — echoed `X-Forwarded-Host` | removed (no callers) |
| `GET /docs`, `/redoc`, `/openapi.json` (routed on 443) | **none** — published the full endpoint map incl. all `/internal/*` paths | off by default; opt in with `LOGOS_DOCS_ENABLED` (dev compose enables it) |
| `/v1`, `/openai`, `/jobs`, `/health`, `/metrics`, `/logosdb/providers/logosnode/*`, `/internal/*` | (key / prometheus key / root key / secret — already fine) | unchanged, plus rate limits below |

Also audited: `logos-agent` is fully Keycloak-JWT + `logos_admin` gated (only `/health` is public) — no changes needed beyond the capacity-read switch below.

## Rate limits (in the compose files, per #978)

Both `docker-compose.yaml` and `docker-compose.dev.yaml` attach generous Traefik `rateLimit` middleware to the orchestrator routers (HTTP 429 when exceeded), env-tunable per deployment:

| Middleware | Routers | Prod default (avg rps / burst) | Dev default |
|---|---|---|---|
| `rl-model` | `/v1`, `/openai`, `/api` | 100 / 200 | 500 / 1000 |
| `rl-jobs` | `/jobs` | 50 / 100 | 250 / 500 |
| `rl-admin` | `/health`, `/docs`, `/metrics`, `/logosdb/providers/logosnode` | 20 / 40 | 100 / 200 |

Knobs: `LOGOS_RATE_LIMIT_{MODEL,JOBS,ADMIN}_{AVG,BURST}` in `.env`. Note Traefik counts per service, not per client IP — this is the in-stack backstop (on top of the orchestrator's existing per-API-key budgets for model traffic). **Per-IP limiting** belongs in the nginx in front of the stack; `logos/docs/deployment.md` now carries a ready-to-apply `limit_req_zone` snippet (30 r/s + burst for the model API, 5 r/s + burst for the control plane) that the operator can drop into the ingress config.

## Consequences / operational notes

- **`/logosdb/scheduler_state`** is now internal-only (secret-gated, no public route). The agent runner polls it and now presents `LOGOS_INTERNAL_SECRET` — the env var was already wired into the agent service and its `Settings`; the field was just unused. Deployments with the agent profile already ship the secret; no compose change needed there.
- Smoke/perf tooling: `test_capacity_planner_smoke.py` and `tests/performance/run_api_workload.py` take `--internal-secret` (defaults to the `LOGOS_INTERNAL_SECRET` env var).
- The benchmark's best-effort provider-name lookup (which already no-ops against prod, where the bare path was never routed) is removed with the endpoint.
- No UI changes: the UI's provider/VRAM calls go through the webservice, which is untouched.
- The removed endpoints had zero callers in this repo (verified by search: UI, webservice, agent, benchmarks, workers).

## Tests

- Orchestrator unit tests: **1860 passed** (incl. new `test_scheduler_state_endpoint.py` covering 403-no-secret / 401-wrong-secret / 401-plain-api-key / 503-not-initialized / 200-payload)
- Agent tests: **685 passed**
- `pre-commit` (autoflake/isort/black/flake8 + yaml checks) clean; both compose files validated with `docker compose config`

## Summary by CodeRabbit

* **New Features**
  * Added per-IP gateway rate limiting with separate budgets for model and control requests.
  * Added configurable trusted proxies, IP allowlists, thresholds, and upstream routing.
  * Added production and development support for the rate-limit gateway.

* **Security**
  * Scheduler status now requires an internal secret instead of a standard API key.
  * API documentation is disabled by default and configurable.
  * Public routing and administrative access are more tightly restricted.

* **Documentation**
  * Documented security tiers, rate limits, configuration options, and updated model connection procedures.


logos/docker-compose.yaml
```diff
@@ -149,6 +170,7 @@ services:
       - "traefik.http.routers.logos-orchestrator-api.tls=true"
       - "traefik.http.routers.logos-orchestrator-api.tls.certresolver=letsencrypt"
       - "traefik.http.routers.logos-orchestrator-api.service=logos-orchestrator"
+      - "traefik.http.routers.logos-orchestrator-api.middlewares=rl-model"
```
Claudia-Anthropica: @wasnertobias [medium] This limiter is attached only to the priority-100 orchestrator fallback, but real `/api` requests are selected by the priority-200/202 webservice, logosnode, and agent routers, whose middleware lists still contain only their strip middleware. Consequently `/api/logosdb/*`, `/api/admin/*`, identity, WebSocket, and agent traffic bypasses the advertised rate limit entirely.

In `logos/docker-compose.yaml`, the rate limit is attached to a lower-priority fallback and is bypassed by the higher-priority routers that actually serve `/api` requests. Add the appropriate `rl-*` middleware alongside each selected router's existing strip middleware, and mirror the correction in `logos/docker-compose.dev.yaml`.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias [medium] The configured `/api` rate limit does not protect the higher-priority routers that actually serve those requests; see the inline comment. The endpoint removals and internal-secret gate otherwise align with the existing Spring and agent integrations. [low] The current title `: Harden the Python API surface` does not match the enforced PR-title format; rename it to something such as `` `Logos`: Harden the Python API surface ``.

logos/docker-compose.yaml
wasnertobias: Fixed in d448f040d — every higher-priority router that serves /api now carries the limiter alongside its strip middleware (webservice identity/config/admin/WebSocket, the logosnode operator paths at 201/202, and agent-api), in both compose files. They all take rl-admin since they are control-plane traffic; model traffic still goes through the rl-model routers at /v1 and /openai. deployment.md documents the full router coverage now.

Claudia-Anthropica (dismissed):
@wasnertobias The earlier rate-limit gap is resolved: every higher-priority `/api` router in both compose files now carries `rl-admin` alongside its strip middleware, and both compose configurations validate. The internal-secret gate and agent caller wiring are consistent with the existing `/internal` authentication. The snapshot still shows optional build checks in progress, with no required checks configured.

logos/docker-compose.yaml
Claudia-Anthropica: @wasnertobias Thanks, this is fixed now.

wasnertobias:
It is important that the ratelimits are pre IP and there is also the possibility to whitelist certain IPs (to free them from the rate-limit) - for the latter one please add an example how to configure this, ideally through .env files. For the first one ensure this is the behaviour - if it is not then consider using nginx additionally/in exchange to traefik.

wasnertobias:
Done — both points are addressed in 5bb4650:

**Per-IP behaviour (verified, not just claimed).** You were right to ask: Traefik's `rateLimit` middleware counts requests *per service*, not per client, so the in-stack limits from before could not isolate one abusive IP. The PR now adds a small nginx container (`logos-rate-gateway`) inside the stack: the public entrypoints (443/8080) select only its catch-all router, which enforces two budgets **per client IP** (model class 30 r/s + burst 60; control-plane class 5 r/s + burst 20) and only then forwards to Traefik's internal entrypoint, where the previous per-service limits apply unchanged as a second layer. I verified the behaviour in a two-hop test rig: two different IPs are limited independently, and the real client IP is taken from `X-Forwarded-For` only for trusted proxy ranges (the Docker bridges by default, which no public client is — forged headers from the internet are stripped on the first hop).

**Whitelist with .env example.** Add this to the node's `.env` (it is picked up by the deploy workflow's auto-generated env, no workflow change needed):

```env
# Free the chair's benchmark rig and one office IP from all rate limits
LOGOS_RATE_LIMIT_WHITELISTED_IPS="129.79.32.0/20 1.2.3.4"
```

Space-separated IPv4 addresses and CIDRs; whitelisted IPs are exempt from both budget classes. Malformed entries are skipped with a log line instead of taking the gateway (and with it the stack) down. `docs/deployment.md` documents this plus the other new knobs (`LOGOS_IP_RATE_LIMIT_*`, `LOGOS_IP_CONTROL_RATE_LIMIT_*`, `LOGOS_GATEWAY_TRUSTED_PROXY_CIDRS` — the latter matters if the node sits behind the chair's nginx, so clients stay separated instead of all counting as that proxy's one IP).

logos/docker-compose.yaml
```diff
@@ -95,45 +180,58 @@ services:
...
+      # not per client, so they cap one service even when many IPs share
+      # the same key. The first layer — per-IP budgeting with an optional
+      # whitelist — is the logos-rate-gateway service in front of these
+      # routers (see docs/deployment.md).
+      # Raise via .env when a deployment needs more headroom.
+      - "traefik.http.middlewares.rl-model.ratelimit.average=${LOGOS_RATE_LIMIT_MODEL_AVG:-100}"
+      - "traefik.http.middlewares.rl-model.ratelimit.burst=${LOGOS_RATE_LIMIT_MODEL_BURST:-200}"
+      - "traefik.http.middlewares.rl-jobs.ratelimit.average=${LOGOS_RATE_LIMIT_JOBS_AVG:-50}"
+      - "traefik.http.middlewares.rl-jobs.ratelimit.burst=${LOGOS_RATE_LIMIT_JOBS_BURST:-100}"
+      - "traefik.http.middlewares.rl-admin.ratelimit.average=${LOGOS_RATE_LIMIT_ADMIN_AVG:-20}"
```
Claudia-Anthropica: @wasnertobias [medium] `rl-admin` is defined on the orchestrator container, but the webservice and agent routers also reference it. When the orchestrator is stopped or recreated, Traefik removes that container's dynamic middleware configuration, leaving those otherwise independent routes with a missing middleware and causing requests to fail—the same lifecycle problem the nearby `strip-api` labels were moved onto Traefik to avoid.

In `logos/docker-compose.yaml`, the shared `rl-*` middleware definitions are owned by the orchestrator even though routers on other services depend on `rl-admin`. Move the shared rate-limit middleware labels to the always-present Traefik service and mirror that ownership change in `logos/docker-compose.dev.yaml`.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias [medium] The shared rate-limit middleware is owned by the orchestrator even though independent webservice and agent routers reference it; see the inline comment. [medium] The existing entrypoint thread is confirmed: character-safe invalid CIDRs pass validation and can make nginx reject its generated configuration. [medium] Direct requests to the whitelist sentinel paths also bypass the public budget for upstream traffic. [low] `/redoc` remains unreachable when docs are enabled; the captured CI has no failed checks, with one non-required build still queued.

wasnertobias:
All four findings are addressed in bbcf70f00, verified in the two-hop test rig:

1. **Shared rate-limit middleware ownership** — the `rl-*` limiters are now defined on the Traefik container in both compose files (same lifecycle reason as `strip-api`). A restart of the orchestrator can no longer drop a limiter that the webservice and agent routers reference.
2. **Invalid CIDRs / IP entries** — the character allowlist in `entrypoint.sh` is replaced with a real parser (four dotted octets, no leading zeros, optional `/0-32` prefix) applied to both the whitelist and the trusted-proxy list. Verified: `999.999.999.999`, `1.2.3.4/33`, `1.2.3` and `01.2.3.4` are each skipped with a log line while the valid entries still apply, and nginx starts cleanly.
3. **Sentinel bypass** — both pass locations are now `internal`, so a direct `GET /__rl_pass_*` gets nginx's 404 instead of drawing on the 10000 r/s pass zone. Verified: direct requests (listed and unlisted IPs) all 404, while the whitelisted rewrite path still passes at full budget.
4. **`/redoc`** — the docs router now serves `/docs`, `/redoc` and `/openapi.json` in both compose files, and the gateway's control-plane class covers `/redoc` too.

Two CodeRabbit suggestions I deliberately did not take, with reasoning now in the code:

- **Encrypt the gateway→Traefik hop**: the `ratelimit` network is host-local and its only peers are this gateway and Traefik — the same trust domain as the `internal` network that already carries every other in-stack hop in cleartext (service APIs, database, Keycloak). Encrypting only this hop would not change what a bridge-level observer can read; the justification is documented in `upstream.conf`.
- **Split the proxy-trust list per entrypoint**: one shared list is sufficient because (a) the only source that can reach `internal8090` is the gateway itself (its peer list is that network and Traefik), so an outer-proxy entry matches no reachable source there, and (b) in the gateway, `real_ip_recursive` resolves the client as the *rightmost untrusted* address in the chain — a forged entry always sits to its left and is never reached. A two-variable split would also risk breaking the common "chair's nginx in front" deployment, where the public entrypoints must trust that proxy or all clients collapse to one IP.
- **`/api` fallback → `rl-admin`** (CodeRabbit, line 248) is taken: the gateway classifies `/api` as control-plane, so the second layer carries the control-plane budget there as well.

logos/rate-limit-gateway/upstream.conf
```diff
@@ -0,0 +1,31 @@
...
+# network that already carries every other in-stack hop in cleartext
+# (service APIs, database, Keycloak). Encrypting only this hop would not
+# change that trust domain — the attacker who could read bridge traffic
+# can already read all of it.
+
+# $request_uri keeps the original path+args, so the same include serves
+# both the normal locations and the named whitelist pass-throughs.
+proxy_pass http://${LOGOS_GATEWAY_UPSTREAM}$request_uri;
+proxy_http_version 1.1;
+proxy_set_header Host $host;
```
Claudia-Anthropica: @wasnertobias [medium] `$host` strips the explicit port from the incoming `Host` header. On the supported non-default endpoints—production `:8080` and development `:18081`—requests such as `POST /jobs` therefore reach Starlette without that port, so `request.url_for(...)` returns an unusable status URL on the default port. Forward the original host and port on this hop, for example by using `$http_host`.

In `logos/rate-limit-gateway/upstream.conf`, the gateway replaces the incoming Host value with `$host`, dropping explicit ports and breaking absolute URL generation on `:8080` and `:18081`. Forward the original host including its port, for example with `$http_host`.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The shared-middleware lifecycle, CIDR validation, sentinel bypass, and `/redoc` routing findings are fixed on this head. [medium] One proxy regression remains: the gateway drops explicit public ports from `Host`, causing application-generated absolute URLs to target the wrong endpoint; see the inline comment. The snapshot has no failed or required checks, although several non-required image-finalization jobs were still pending.

wasnertobias:
Fixed in d8f89ce4a. `$host` normalizes the incoming Host and drops an explicit port, so a client on a supported non-default endpoint (prod `:8080`, dev `:18081`) reached the services without it and `url_for` generated status URLs on the default port. The gateway now forwards `$http_host` (the original value including port); Traefik always sends the Host header on this hop, so it is never empty.

Verified in the two-hop rig: the upstream now receives the Host with its port, and the whitelist, per-IP limiting, and sentinel behaviour are all unchanged.

logos/logos-orchestrator/tests/performance/run_api_workload.py
```diff
@@ -301,11 +303,13 @@ async def collect_runtime_samples(
...
         sample: dict[str, object] = {"captured_at": isoformat_utc(datetime.now(timezone.utc))}
         try:
+            # scheduler_state is gated on the shared internal secret, not a
+            # user API key — the sample carries cluster internals.
             scheduler_payload = await _request_json(
                 client,
                 "GET",
                 f"{base_url.rstrip('/')}/logosdb/scheduler_state",
-                headers=headers,
+                headers={"Authorization": f"Bearer {internal_secret}"},
```
Claudia-Anthropica, line 321: @wasnertobias [medium] In remote/http-only mode, `base_url` is the public `--api-base`, but this PR removes `/logosdb/scheduler_state` from the public Traefik routers. Supplying the new secret therefore still reaches the UI catch-all or a 404, and `capture_once` swallows that failure into `runtime_samples.jsonl`, allowing a benchmark to finish while every documented scheduler/worker snapshot is missing. Add a separate internal telemetry base and wire it through `test_scheduling_performance.sh`, or otherwise provide a secret-gated reachable route without restoring user-key access.

In `logos/logos-orchestrator/tests/performance/run_api_workload.py`, remote benchmarks use the public API base for a scheduler endpoint this change makes internal-only, so every runtime sample becomes an error despite the new secret argument. Add and wire a separate internal telemetry base, including the remote wrapper, for scheduler-state collection.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The `$http_host` change fixes the previously reported non-default-port regression, and the earlier limiter, CIDR, sentinel, and documentation findings are resolved. [medium] The supported remote/http-only performance path now silently loses every scheduler runtime sample because it uses the public `--api-base` for an endpoint this PR removes from all public routers; see the inline comment. The captured CI has no failed or required checks, while the optional CodeRabbit review remains pending.

logos/logos-orchestrator/tests/performance/run_api_workload.py
wasnertobias: Fixed in 411930e2e with option 1 (separate internal telemetry base); no user-key access is restored.

- `run_api_workload.py` gains `--telemetry-base` (defaults to `--api-base`), used for the internal sampling endpoints (`/logosdb/scheduler_state`, `/logosdb/providers/logosnode/status`) and threaded through `run_workload` to `collect_runtime_samples`.
- `test_scheduling_performance.sh` now passes `--telemetry-base` and `--internal-secret` on both paths: local mode points it at the orchestrator's own `127.0.0.1:8080` (the secret is already in the container env); remote mode defaults to `--api-base` but lets the operator point it at a base that reaches the orchestrator internally.
- The failure is no longer silent: the first `scheduler_state` error prints a warning to stderr, and `run_meta.json` records `telemetry_base` + `telemetry_ok`, so a run whose `runtime_samples.jsonl` is empty is visible in the metadata.
- `/logosdb/scheduler_state` stays secret-gated and is intentionally not re-added to the public routers.

logos/logos-orchestrator/tests/performance/README.md
```diff
@@ -106,6 +106,29 @@ docker compose exec logos-orchestrator python /app/tests/performance/run_api_wor
   --workload /app/tests/performance/workloads/explicit/10m/workload_explicit_local5_skewed_bursty_10m.csv
 ```
 
+Remote run against a deployed environment (http-only, from a workstation):
+
+```bash
+./tests/performance/test_scheduling_performance.sh \
+  --logos-key "<ROOT_LOGOS_KEY>" \
+  --api-base "https://logos.example.org" \
+  --telemetry-base "http://<host-that-reaches-the-orchestrator>:8080" \
```
Claudia-Anthropica: @wasnertobias [high] This remote example sends `LOGOS_INTERNAL_SECRET` as a Bearer token to a plain-HTTP direct service port. Unlike the gateway's same-host bridge hop, this path crosses from a workstation to a deployment host or network, so anyone able to observe that link can recover the shared secret and authenticate to every `/internal/*` endpoint. Require a TLS-protected telemetry URL or an authenticated port-forward/tunnel, and make the example use that protected channel.

In `logos/logos-orchestrator/tests/performance/README.md`, the remote benchmark example sends the shared internal secret to a plain-HTTP telemetry endpoint, exposing the cluster credential on the workstation-to-host link. Require TLS or an authenticated tunnel/port-forward and update the example accordingly.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The separate `--telemetry-base` wiring resolves the prior missing-sample defect, and that thread can be closed. [high] The new remote-run example transmits `LOGOS_INTERNAL_SECRET` over plain HTTP to a host-reachable orchestrator port; see the inline comment. The captured CI snapshot has no failed or unfinished required checks.

logos/logos-orchestrator/tests/performance/run_api_workload.py
Claudia-Anthropica: @wasnertobias Thanks, the separate telemetry base plus the explicit warning and `telemetry_ok` metadata resolve the remote sampling issue.

logos/logos-orchestrator/tests/performance/README.md
wasnertobias: Fixed in 434fde9ea. The runner now refuses a plain-HTTP `--telemetry-base` for a non-local host (exits 2 with a message pointing at HTTPS or a local tunnel port); HTTPS and loopback bases are allowed. The README remote example no longer sends the secret over a cleartext direct port: it sets up an SSH tunnel to a local loopback port (`ssh -N -L 18443:127.0.0.1:8080 user@host`) and points `--telemetry-base` at `http://127.0.0.1:18443`, so the cleartext hop is confined to the workstation's own loopback. The requirement is also noted in the `--telemetry-base` help and the wrapper's remote-mode comment.

logos/logos-orchestrator/tests/performance/README.md
```diff
@@ -106,6 +106,40 @@ docker compose exec logos-orchestrator python /app/tests/performance/run_api_wor
...
 
+Remote run against a deployed environment (http-only, from a workstation):
+
+```bash
+# 1) Authenticated tunnel: local loopback 18443 -> the core node's orchestrator.
+#    The remote side (127.0.0.1:8080) is where /logosdb/scheduler_state is
+#    reachable on the node (its published admin port, or via `docker` on the
+#    host). SSH authenticates the link, so the cleartext hop stays on your
+#    own loopback.
+ssh -N -L 18443:127.0.0.1:8080 "$VM_USERNAME"@"$VM_HOST" &
```
Claudia-Anthropica: @wasnertobias [medium] The tunnel targets `127.0.0.1:8080` on the deployment host, but production `docker-compose.yaml` publishes that port for Traefik's TLS `secure8080` entrypoint; `logos-orchestrator` only has `expose: 8080` and no host binding. Thus the plain-HTTP telemetry request never reaches `/logosdb/scheduler_state` (and the benchmark merely warns and writes `telemetry_ok: false`), so the documented remote path remains unusable. Point the tunnel at an address that actually reaches `logos-orchestrator:8080`, such as its bridge IP, or add and document a loopback-only host binding for the orchestrator.

In `logos/logos-orchestrator/tests/performance/README.md`, the remote tunnel targets the host's Traefik port rather than the unpublished orchestrator port, so scheduler telemetry cannot be collected. Update the example to tunnel to an actual `logos-orchestrator:8080` endpoint, using a bridge address or a loopback-only host mapping.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The new URL validation prevents the shared secret from crossing a plain-HTTP remote link, but the replacement SSH example does not reach the orchestrator in the production Compose topology; see the inline finding. As written, remote benchmarks still complete without the promised scheduler samples. The snapshot has no failed or unfinished required checks; several optional jobs remain pending.

logos/logos-orchestrator/tests/performance/README.md
wasnertobias: Fixed in b5477ec37. You're right — the host's :8080 is Traefik's TLS secure8080 entrypoint, and logos-orchestrator only `expose`s 8080 on the compose network, so the old tunnel target never reached the orchestrator. The README now tunnels to the orchestrator's bridge IP (fetched via `docker inspect` on the host, which routes to compose-network IPs), keeping the cleartext hop on loopback. I went with the bridge-IP target rather than adding a loopback host binding to keep this security PR from growing the deployment surface; a stable ops port can be a follow-up if the team wants one.

I also addressed CodeRabbit's related note: the README now states the two telemetry endpoints use different credentials (scheduler_state = internal secret, provider status = a root logos_key), with a matching inline comment in collect_runtime_samples.

Claudia-Anthropica (changes requested):
@wasnertobias [medium] The existing trusted-proxy configuration thread remains blocking: the documented multi-CIDR value is space-delimited and passed unchanged to Traefik, whose CLI expects a comma-delimited list, while nginx separately expects whitespace-delimited entries. Normalize the value for both consumers so enabling the documented outer-proxy configuration does not leave it untrusted or invalidate Traefik's static configuration. The earlier telemetry fixes remain correct; captured CI has no failed or required checks, although optional checks remain unfinished.

wasnertobias (commented):
Fixed in 7562532dd. The value is now documented as **comma-delimited** (Traefik's `--forwardedHeaders.trustedIPs` is comma-separated, so a space-delimited value left the outer proxy untrusted / risked invalidating the static config), and the gateway entrypoint normalizes commas to spaces before its word-split loops — so nginx accepts comma *or* whitespace for both the trusted-proxy and whitelist lists. One documented value now serves both consumers: Traefik gets commas, nginx gets whitespace. Verified the normalization across comma / space / mixed / invalid / single inputs (the real `is_ip_entry` validator still rejects bad entries as before); compose comments and deployment.md updated to the comma form. Thanks for confirming the telemetry fixes were correct.

Claudia-Anthropica (dismissed):
@wasnertobias The blocking trusted-proxy issue is resolved: the documented comma-delimited value reaches Traefik unchanged and is normalized before nginx renders its per-range directives. The earlier secret-gating, telemetry, host-forwarding, and limiter-lifecycle fixes remain consistent with the current callers and Compose topology. Captured CI has no failed or required checks; optional jobs were still unfinished.

wasnertobias:
✅ Approved — I reviewed the latest head, including the gateway routing and scheduler-state changes, and found no remaining actionable blocker.

logos/logos-orchestrator/tests/performance/run_api_workload.py
```diff
@@ -375,13 +404,10 @@ def fetch_runtime_artifacts(
...
-    provider_vram = _request_json_sync(
-        "POST",
-        f"{base_url.rstrip('/')}/logosdb/get_ollama_vram_stats",
-        logos_key=logos_key,
-        json_body={"day": vram_day},
-    )
+    # The orchestrator's /logosdb/get_ollama_vram_stats endpoint was removed
+    # (the webservice serves the dashboard with proper auth); VRAM history is
+    # covered by the per-sample provider_status payloads instead.
+    provider_vram = None
```
wasnertobias, line 410: Every benchmark now writes `provider_vram.json` as `null`, while the performance README still describes this as the provider VRAM history artifact. Existing downstream analysis will silently lose that payload. Please preserve the schema using an equivalent authenticated source, or remove/update the artifact and add a regression test documenting the replacement.

logos/rate-limit-gateway/nginx.conf
```diff
@@ -0,0 +1,112 @@
...
+    default upgrade;
+    '' close;
+}
+
+# The geo block rendered by entrypoint.sh (whitelist).
+include /etc/nginx/rate-gateway/whitelist.conf;
+
+server {
+    listen 80;
+    client_max_body_size 100m;
```
wasnertobias, line 65: The gateway hard-caps uploads at 100 MB while the application batch/audio limits remain environment-configurable. A deployment configured above 100 MB will be rejected unexpectedly at the gateway. Make `client_max_body_size` configurable and align it with the application limits, or document/enforce the 100 MB maximum consistently.

wasnertobias (commented):
✅ Approved with non-blocking follow-ups — the main hardening changes are sound, but these compatibility/documentation gaps should be addressed before relying on the affected tooling in all supported deployments.

wasnertobias:
Follow-up: `logos/e2e/harness/clients/admin.py:96-100` still sends the admin Logos key to `/logosdb/scheduler_state`, but that route now requires `LOGOS_INTERNAL_SECRET`; the E2E helper will receive 401. Update it to send the internal secret (or remove the stale helper) and add coverage for the credential distinction.

PR #988, opened 2026-09-16
`Athena`: Fix high severity Dependabot alerts

krusche:
### Motivation and Context

First of six PRs clearing the open **high** severity [Dependabot alerts](https://github.com/ls1intum/edutelligence/security/dependabot?q=is%3Aopen+severity%3Ahigh). All criticals are now closed; this one covers Athena (~133 of the 334 alerts). The remaining subprojects follow in their own PRs so each stays reviewable.

Scope is safe in-range updates plus genuinely low-risk majors. Deep-testing majors (transformers 5, starlette/FastAPI, pyarrow, cryptography) are deliberately deferred.

### Description

Most alerts are transitive and fit inside existing ranges, so they are lockfile-only. The four manifest changes are each forced by something specific:

| Change | Why it was necessary |
| --- | --- |
| `axios` 1.13.1 → 1.20.0, `mysql2` 3.9.8 → 3.24.4 | Both are semver **minor**. They only appeared breaking because the manifest pins exact versions. |
| `postcss` override `^8.5.23` | `next` 15.5.25 pins postcss to exactly 8.4.31, so npm proposed upgrading to **next 16**. The override stays inside the API-compatible 8.x line and avoids a second major Next upgrade. |
| `serialize-javascript` override `^7.1.1` | Dependents still declare `^6`, and npm proposed *downgrading* `@docusaurus/core` to 3.5.2. |
| `webpackbar` override `^7.0.0` | See below. |
| `react` + `react-dom` both floored `^19.3.0` | See below. |

**Two real build breakages were introduced by the naive update and fixed here.** Worth reviewing, since neither is visible from the lockfile diff alone:

1. **webpack ≥ 5.106.0 breaks the Docusaurus build.** It tightened the `ProgressPlugin` option schema, which now rejects the `name` / `color` / `reporters` / `reporter` options that `webpackbar` 6.0.1 passes — `ValidationError: options has an unknown property 'name'`. I bisected the boundary: **5.105.4 builds, 5.106.0 does not**. webpack is not itself in any alert, so pinning it back was an option, but `webpackbar` 7.0.0 is compatible with current webpack — overriding that instead keeps webpack receiving future security fixes.

2. **react / react-dom drifted apart.** The update moved `react` to 19.3.0 but left `react-dom` at 19.2.0; React rejects this at build time (`Incompatible React versions ... must have the exact same version`). Both floors are now `^19.3.0` so they can't separate again.

`protobuf` 6.31.1 → 6.33.5 in `module_text_cofee` is an exact pin whose comment warns the Dockerfile base image must be kept in step. The comment still referenced `golang:alpine3.22` while the Dockerfile is already on `alpine3.24`, so that stale reference is corrected. The protobuf runtime is forward-compatible with the existing generated `cofee_pb2.py`; the Docker build in CI exercises this.

### Not fixed here

- **`nltk`** (6 alerts — `athena/tests` and four modules) has **no patched release upstream**. `first_patched_version` is null, so there is nothing to upgrade to. These alerts will stay open until upstream ships a fix.
- **`starlette`** in `athena/tests` needs 1.3.1, a major bump pinned through FastAPI. Deferred per the agreed scope.

### Steps for Testing

Playground:
1. `cd athena/playground && npm ci`
2. `npm run build` — compiles all 20 routes
3. `npx tsc --noEmit` — exits 0; `npm run lint` — 0 errors
4. `npm audit` reports **0 high, 0 critical** (2 moderate remain)

Docs:
1. `cd athena/docs && npm ci && npm run build` — must reach `[SUCCESS] Generated static files`
2. `npm audit` reports **0 high, 0 critical** (18 moderate remain)

Python:
- `cd athena/evaluation && poetry check --lock`
- `cd athena/modules/text/module_text_cofee && poetry check --lock`; the module's Docker build covers the protobuf bump.

All of the above were run locally and pass.

## Summary by CodeRabbit

* **Chores**
  * Updated underlying application components and build tooling to newer versions.
  * Improved compatibility with current runtime and build environments.
  * Applied maintenance updates across the text-processing and playground components.


athena/docs/package.json
```diff
@@ -20,8 +20,8 @@
     "@easyops-cn/docusaurus-search-local": "^0.52.1",
     "@mdx-js/react": "^3.0.0",
     "prism-react-renderer": "^2.3.0",
-    "react": "^19.0.0",
-    "react-dom": "^19.0.0"
+    "react": "^19.3.0",
```
Claudia-Anthropica: @krusche [low] Using the same caret floor does not keep React and React DOM on identical versions: each range resolves independently, so a staggered 19.x release can recreate the mismatch this change is intended to prevent. Pin both packages to `19.3.0` (or otherwise enforce one exact shared version) so future lockfile refreshes cannot diverge.

In `athena/docs/package.json`, the independent `^19.3.0` ranges can resolve React and React DOM to different 19.x releases even though their runtime versions must match. Pin both dependencies to the same exact version or enforce one exact shared override, then regenerate the lockfile.

</details>

Claudia-Anthropica (dismissed):
@krusche The manifests and lockfiles are internally aligned, and the CoFee protobuf runtime matches the Alpine 3.24 generator image. Athena tests, lint, and the CoFee image build passed in the captured CI snapshot; several non-required image builds, including the playground image, were still running. I found one low-severity durability issue in the React constraints.

athena/docs/package.json
krusche: Good catch — you are right, and I have fixed it in 87846dca6 (and the same change on Atlas #989, which carried the identical defect).

The reasoning holds: npm resolves each range independently, so matching `^19.3.0` floors guarantee nothing. A staggered 19.x release — react 19.4.0 landing before react-dom 19.4.0 — would let a lockfile refresh pick react 19.4.0 against react-dom 19.3.x and reproduce exactly the `Incompatible React versions` build failure the floors were added to prevent. The floors only fixed the symptom I had already hit, not the mechanism.

Both are now pinned to exact `19.3.0`. That also makes them consistent with every `@docusaurus/*` dependency in the same file, which are all pinned exact for the same class of reason. Verified the docs build still reaches `[SUCCESS] Generated static files` on both sites after the change.

PR #989, opened 2026-09-16
`Atlas`: Fix high severity Dependabot alerts

krusche:
### Motivation and Context

Second of six PRs clearing the open **high** severity [Dependabot alerts](https://github.com/ls1intum/edutelligence/security/dependabot?q=is%3Aopen+severity%3Ahigh). Covers Atlas (78 alerts: 45 in `atlas/docs`, 33 in `atlas/AtlasMl`). Athena is #988, Iris #990, Logos #994.

Scope is safe in-range updates plus low-risk majors; deep-testing majors are deferred and listed below.

#### `atlas/docs`

Same Docusaurus pattern as #988. Three of these fix build breakages that the naive `npm audit fix` introduced:

| Change | Why |
| --- | --- |
| `webpackbar` override `^7.0.0` | webpack ≥ 5.106.0 tightened the `ProgressPlugin` option schema and now rejects the `name`/`color`/`reporter` options webpackbar 6.0.1 passes — this **broke the build**. I bisected the boundary: 5.105.4 builds, 5.106.0 does not. webpackbar 7.0.0 works with current webpack, so overriding it keeps webpack receiving future security fixes rather than pinning webpack back. |
| `react` + `react-dom` floored `^19.3.0` | The update moved react to 19.3.0 but left react-dom at 19.2.0; React rejects the mismatch at build time (`Incompatible React versions`). |
| `@docusaurus/theme-mermaid` pinned `3.9.1` | It was the **only** `@docusaurus/*` entry using a caret, so it drifted to 3.10.2 while core stayed at 3.9.1 — Docusaurus then failed with `Invalid name=docusaurus-theme-mermaid version number=3.10.2`. Pinning it matches how every sibling `@docusaurus/*` dep is already declared. |
| `serialize-javascript` override `^7.1.1` | Dependents still declare `^6`; npm proposed downgrading `@docusaurus/core` to 3.5.2 instead. |

#### `atlas/AtlasMl`

`pillow` 12.3.0, `aiohttp` 3.14.3, `soupsieve` 2.9.2, `urllib3` 2.8.0 — lockfile-only.

`protobuf` 5.29.4 → **6.33.6**. The alert only required 5.29.6, but poetry resolves to the newest allowed version. I checked this is safe rather than assuming: `weaviate-client` declares `protobuf >=4.21.6,<7.0.0`, and AtlasMl contains **no** `.proto` files or generated `_pb2.py`, so nothing depends on a specific codegen version. (Contrast `module_text_cofee` in #988, which *does* have generated code and is therefore held to a patch bump.)

### Note on the commit history

This branch has three commits. The second added a `black` 26 bump plus the 18-file reformatting it forces, and the third **reverts both** — so the net diff touches no AtlasMl source at all, only lockfiles and `atlas/docs/package.json`. Black 26 makes its new multiline-string handling part of the *stable* style with no opt-out, which is a team style decision rather than something to fold into a security fix. It moves to its own cross-cutting PR covering AtlasMl, logos-agent, memiris and shared.

Confirm with `git diff main --stat`: only `atlas/AtlasMl/poetry.lock`, `atlas/docs/package-lock.json` and `atlas/docs/package.json`.

### Not fixed here

Deferred, each needing its own testing:

- `cryptography` → 50 (major)
- `pyarrow` → 23 (major)
- `starlette` → 1.3.1 (major, pinned through FastAPI)
- `transformers` → 5 (major)
- `black` → 26 — moved to the dedicated style PR; this alert stays open until that lands.

### Steps for Testing

1. `cd atlas/docs && npm ci && npm run build` — must reach `[SUCCESS] Generated static files`. `npm audit` reports **0 high, 0 critical** (18 moderate remain).
2. `cd atlas/AtlasMl && poetry check --lock`, then `poetry install` and `poetry run pytest`.

The docs build, the audit and `poetry check --lock` were run locally and pass. AtlasMl's pytest suite was **not** run locally (it pulls torch/sentence-transformers); CI covers it.

## Summary by CodeRabbit

* **Chores**
  * Updated documentation tooling and runtime configuration for improved consistency and compatibility.
  * Refined documentation build settings to support more reliable results.


Claudia-Anthropica (dismissed):
@krusche The dependency manifests and locks are aligned, React and ReactDOM resolve together at 19.3.0, and the captured documentation and Atlas image builds succeeded. The Python reformat preserves the code structure, apart from removing trailing whitespace in two docstrings. The non-required `test` check is red, but its captured evidence contains only a generic exit code 2, so I could not identify a failing step or attribute the failure to this change.

PR #993, opened 2026-09-16
`Logos`: Centralize the HuggingFace token on the orchestrator

### LINKED ISSUE #984: `Logos`: Huggingfacetoken shall be centralized
Right now the hf_token is decentralized (every workernode has it configured in .env). But it could also be centrally configured on the orchestrator side, and sent on bridge connect to every workernode.

Maybe other .env variables are to be centralized as well.

bensofficial:
## Summary
- The orchestrator can now configure `HF_TOKEN` once and send it to every worker node in the `/logosdb/providers/logosnode/auth` response, instead of requiring `HF_TOKEN` in each worker's own `.env`.
- A worker adopts the centrally-sent token as its own `HF_TOKEN` on every successful auth; if the orchestrator has none configured, the worker keeps using its own local value (backwards compatible).

Fixes #984

## Test plan
- [x] `logos-orchestrator`: `PYTHONPATH=src pytest tests/unit/` — 1864 passed
- [x] `logos-workernode`: `PYTHONPATH=. pytest tests/test_logos_bridge.py` — 89 passed
- [x] `pre-commit run --config logos/.pre-commit-config.yaml` on all changed files — passed

## Summary by CodeRabbit

* **New Features**
  * Added optional centralized Hugging Face token configuration for worker authentication.
  * Workers now receive the centrally configured token during authentication and startup, enabling access to gated models.
  * Local worker tokens remain available as a fallback when centralized credentials are unavailable.
  * Startup continues gracefully with locally configured credentials if the central service cannot be reached.

* **Documentation**
  * Updated configuration examples with setup guidance, token permissions, and centralized-token behavior.


logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -292,6 +292,10 @@ async def _authenticate(self) -> dict[str, Any]:
         if "worker_id" in data:
             self._resolved_worker_id = str(data["worker_id"])
 
+        central_hf_token = str(data.get("hf_token", "")).strip()
```
Claudia-Anthropica, line 309: @bensofficial [medium] The centralized credential is installed only during `_authenticate()`, but `lifespan()` schedules capability prefetches and awaits all static/restored `apply_lanes()` calls before starting the bridge. A fresh worker relying solely on the orchestrator token can therefore download or launch a gated model without credentials; a static gated lane can fail startup before authentication ever occurs.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, the central HF token is applied only after startup prefetch and lane initialization may already have used it. Complete an initial authenticated token bootstrap before scheduling capability prefetches or starting static/restored lanes, and add a cold-start ordering test.

</details>

logos/logos-workernode/logos_worker_node/logos_bridge.py
```diff
@@ -292,6 +292,10 @@ async def _authenticate(self) -> dict[str, Any]:
...
+        if central_hf_token:
```
Claudia-Anthropica, line 310: @bensofficial [medium] An empty token response leaves `os.environ` untouched. After this process accepts central token A, removing that token from the orchestrator and reconnecting leaves A usable indefinitely, while the original local fallback cannot be restored because it was overwritten; revocation therefore requires restarting every worker.

In `logos/logos-workernode/logos_worker_node/logos_bridge.py`, an empty central-token response retains a previously injected central credential instead of revoking it. Preserve the worker's initial local fallback and, after every successful auth, either install the nonempty central token or restore/unset the local value; test a nonempty response followed by an empty one.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The centralized token is currently acquired too late for startup model operations, and a subsequently removed central token remains active in the worker process. The captured checks passed, but these credential-lifecycle defects need correction before the feature is reliable.

bensofficial:
@Claudia-Anthropica Both fixed in 72684964d:

1. **Acquired too late for startup model operations** — moved the central `HF_TOKEN` fetch into a new `LogosBridgeClient.bootstrap_hf_token()`, called from `lifespan()` right after config load, before the missing-model prefetch and static/dynamic lane `apply_lanes()` calls that spawn vLLM/download weights. It's a best-effort one-off `/auth` call (short-circuits if the bridge is disabled, swallows network errors and falls back to the local `.env` value).
2. **Stale central token after removal** — `_authenticate()` now captures the worker's original local `HF_TOKEN` at construction time and re-derives `os.environ["HF_TOKEN"]` on every auth: central value if the server sends one, else the local value, else clears it. So a removed central token is no longer stuck forever — it clears on the worker's next auth/reconnect cycle.

Added regression tests for both (lifespan ordering test in `test_main.py`, and revert-on-removal + bootstrap tests in `test_logos_bridge.py`). Full worker + orchestrator suites and pre-commit are green.

Claudia-Anthropica (approved):
@bensofficial The current code resolves both previous findings: authentication now precedes cache prefetch and lane startup, and every successful auth restores or removes the captured local fallback when the central token disappears. Regression coverage exercises both paths, and the captured Logos test and lint checks pass; only non-required checks remain queued.

PR #990, opened 2026-09-16
`Iris`: Fix high severity Dependabot alerts

krusche:
### Motivation and Context

Third of six PRs clearing the open **high** severity [Dependabot alerts](https://github.com/ls1intum/edutelligence/security/dependabot?q=is%3Aopen+severity%3Ahigh). Covers Iris (59 alerts: 38 in `iris/docs`, 21 in `iris`). Athena is #988, Atlas is #989.

#### `iris/docs`

- `serialize-javascript` override `^7.1.1` — its dependents still declare `^6`, and npm proposed *downgrading* `@docusaurus/core` instead.
- `webpackbar` override `^7.0.0` — webpack ≥ 5.106.0 tightened the `ProgressPlugin` option schema and rejects the options webpackbar 6.0.1 passes. **`iris/docs` resolves webpack 5.105.4 and is not yet affected**, so this is preventative rather than a fix: it stops the next webpack bump from breaking the build, and keeps all three docs sites on the same shape.

Notably, iris/docs needed **neither** of the realignments the other two docs sites did — `react` and `react-dom` already resolve to a matching 19.2.4, and every `@docusaurus/*` entry is already pinned exact so nothing drifted out of step with core.

#### `iris`

`aiohttp` 3.14.3, `langchain-classic` 1.0.8, `langsmith` 0.12.5, `orjson` 3.12.0, `urllib3` 2.8.0, `protobuf` 6.33.6 — all lockfile-only, no manifest changes. `protobuf` 6.31.1 → 6.33.6 stays within the same major.

### Not fixed here

Deferred as agreed, each needing its own testing:

- `cryptography` → 50 (major)
- `starlette` → 1.3.1 (major, pinned through FastAPI)
- `transformers` → 5 (major)

### Steps for Testing

1. `cd iris/docs && npm ci && npm run build` — must reach `[SUCCESS] Generated static files`. `npm audit` reports **0 high, 0 critical** (19 moderate remain).
2. `cd iris && poetry check --lock` — passes.
3. `poetry install && poetry run pytest`, plus the existing Lint job.

The docs build, the audit and `poetry check --lock` were run locally and pass. The Iris pytest suite was **not** run locally; CI covers it.

PR #994, opened 2026-09-16
`Logos`: Fix high severity Dependabot alerts

krusche:
### Motivation and Context

Fourth of six PRs clearing the open **high** severity [Dependabot alerts](https://github.com/ls1intum/edutelligence/security/dependabot?q=is%3Aopen+severity%3Ahigh). Covers Logos (37 alerts). Athena is #988, Atlas #989, Iris #990.

#### `logos-ui`

Lockfile-only. The in-range updates clear **every** alert — `npm audit` now reports no vulnerabilities at all for this project, at any severity.

#### `logos-orchestrator`

`pillow` 12.3.0, `aiohttp` 3.14.3, `langchain-core` 1.6.3, `langsmith` 0.12.5, `python-multipart` 0.0.32, `urllib3` 2.8.0. Lockfile-only, no `pyproject.toml` change.

One thing worth knowing for anyone reproducing this: regenerating the orchestrator lock requires the `shared` path dependency to resolve, and `logos/logos-orchestrator/shared` **does not exist in the repo** — the Dockerfile synthesises it by copying the repository-root `shared/` into place. `uv lock` therefore fails locally with `Distribution not found`. I regenerated against a temporary symlink and removed it afterwards; confirm nothing stray was committed with `git status`. The lock keeps its existing format (`version = 1`, `revision = 3`), so the pinned uv 0.12.5 in the Dockerfile still reads it.

### Not fixed here

- `starlette` → 1.3.1 and `transformers` → 5 are major bumps needing their own testing.
- **`black`** is deliberately excluded. Bumping it to 26 forces a style migration (its new stable multiline-string handling), which rewrites much of `logos-agent/app/db.py`. That belongs in its own PR rather than buried in a security change, so `logos-agent` is untouched here. Its `black` and `starlette` alerts stay open for now.

### Steps for Testing

1. `cd logos/logos-ui && npm ci && npm run build`; `npm audit` reports **zero** vulnerabilities.
2. `cd logos/logos-orchestrator && uv sync` and run the orchestrator's tests.
3. The Logos Docker builds in CI exercise the orchestrator lock end to end.

The `logos-ui` audit and the orchestrator lock regeneration were verified locally. The orchestrator test suite was **not** run locally (the `shared` path dependency is not resolvable outside the Docker context); CI covers it.


logos/logos-orchestrator/uv.lock
```diff
@@ -32,7 +32,7 @@ wheels = [
 
 [[package]]
 name = "aiohttp"
-version = "3.13.5"
+version = "3.14.3"
```
Claudia-Anthropica, line 35: @krusche [high] These patched versions never become an input to the orchestrator image or its test installation: both run `uv pip install .`, the Dockerfile does not copy `uv.lock`, and `pyproject.toml` is unchanged. The cached Docker dependency layer can therefore retain the vulnerable versions, while a clean installation merely re-resolves unconstrained transitive versions and ignores these pins. Add the patched minimum versions represented by this lock update to `logos/logos-orchestrator/pyproject.toml` (declaring explicit constraints for transitive packages), then regenerate the lock so the existing installation path and Docker cache are forced to use them.

In `logos/logos-orchestrator/uv.lock`, the security upgrades are ignored by the orchestrator's `uv pip install .` build and test paths because they consume only `pyproject.toml`. Add every patched minimum to `logos/logos-orchestrator/pyproject.toml`, including explicit constraints for affected transitive packages, and regenerate the lock.

</details>

Claudia-Anthropica (changes requested):
@krusche The UI lockfile is consumed by both `npm ci` paths, and the captured UI build passed. The orchestrator security update is not consumed by its installation or image-build path, so changes are required.

logos/logos-orchestrator/uv.lock
krusche: Confirmed and fixed in a8dae5e09 — you were right, and the finding was more serious than a durability concern: the fix was entirely inert.

I verified the mechanism rather than taking it on trust. `logos/logos-orchestrator/Dockerfile` copies `pyproject.toml` (line 29) and the `shared/` package, and there is no `COPY` of `uv.lock` anywhere in the file. Combined with `uv pip install .`, that means the image resolved transitive dependencies afresh from the manifest and never saw a single version I had pinned. The cached dependency layer would have kept serving the vulnerable set on top of that.

The patched minimums are now constraints in `[tool.poetry.dependencies]`, which the poetry-core backend turns into the metadata `uv pip install .` actually consumes:

- `aiohttp` `^3.9.0` → `>=3.14.3,<4.0.0`
- `python-multipart` `>=0.0.20` → `>=0.0.30`
- `langchain-core >=1.3.3`, `langsmith >=0.8.18`, `pillow >=12.3.0`, `urllib3 >=2.7.0` added — these were transitive only, so nothing in the manifest kept a rebuild off a vulnerable version.

I used the advisories' patched minimums rather than the resolved versions, so the constraints express the security requirement without freezing the tree.

Verified along the path the image actually takes, not by reading the lock — `uv pip compile pyproject.toml --no-cache`, which ignores `uv.lock` exactly as the Dockerfile does:

```
aiohttp==3.14.3
langchain-core==1.6.3
langsmith==0.12.5
pillow==12.3.0
python-multipart==0.0.32
urllib3==2.8.0
```

I am applying the same treatment to the other PRs in this batch, since the same lockfile-only weakness applies to them even where the build path does read the lock.

Claudia-Anthropica (approved):
@krusche The prior blocker is resolved: `pyproject.toml` now carries the patched minimums consumed by both Docker and CI installation paths, and the regenerated lock resolves those floors. The UI lock remains consistent, and the captured orchestrator tests, UI build, lint, and Docker builds all succeeded at this exact head.

PR #995, opened 2026-09-16
`Memiris`: Fix high severity Dependabot alerts

krusche:
### Motivation and Context

Fifth of six PRs clearing the open **high** severity [Dependabot alerts](https://github.com/ls1intum/edutelligence/security/dependabot?q=is%3Aopen+severity%3Ahigh). Covers Memiris (19 alerts). Athena is #988, Atlas #989, Iris #990, Logos #994.

### Description

Lockfile-only — no `pyproject.toml` change:

`aiohttp` 3.14.3, `langchain-classic` 1.0.8, `langchain-core` 1.6.3, `langgraph-checkpoint` 4.2.0, `langsmith` 0.12.5, `orjson` 3.12.0, `urllib3` 2.8.0, `protobuf` 6.33.6.

**On `protobuf` crossing a major boundary:** the alert only required 5.29.6, but poetry resolves to the newest allowed version, landing on 6.33.6. I verified this rather than assuming — every declared constraint in the lock admits it (`weaviate-client >=4.21.6,<7.0.0`, `opentelemetry-proto >=5.0,<8.0`, `googleapis-common-protos <7.0.0`), and memiris contains **no** `.proto` files or generated `_pb2.py`, so nothing depends on a specific codegen version.

### Not fixed here

- `cryptography` → 50 (major) needs its own testing.
- **`black`** → 26 is excluded on purpose. It makes its new multiline-string handling part of the *stable* style with no opt-out, so upgrading forces a reformat. That is a team style decision and belongs in its own cross-cutting PR (AtlasMl, logos-agent, memiris, shared) rather than inside a security change. Notably, memiris is the one project where black 26 would reformat **zero** files — but it still waits for the shared PR so the bump lands consistently everywhere.

### Steps for Testing

1. `cd memiris && poetry check --lock` — passes.
2. `poetry install && poetry run pytest`.

`poetry check --lock` was verified locally. The pytest suite was **not** run locally — its Weaviate tests need Docker testcontainers; CI covers it.


Claudia-Anthropica (dismissed):
@krusche The updated lockfile remains consistent with Memiris's declared constraints, including the protobuf/OpenTelemetry and LangSmith transitive dependency changes. The exact-head Memiris test and lint checks completed successfully.

PR #996, opened 2026-09-16
`General`: Fix high severity Dependabot alerts in shared

krusche:
### Motivation and Context

Last of six PRs clearing the open **high** severity [Dependabot alerts](https://github.com/ls1intum/edutelligence/security/dependabot?q=is%3Aopen+severity%3Ahigh). Covers `shared` (7 alerts). Athena is #988, Atlas #989, Iris #990, Logos #994, Memiris #995.

### Description

`python-multipart` 0.0.20 → 0.0.32, closing **all three** of its alerts.

**This required raising the Python floor first, which is the only interesting part of this PR.** Every `python-multipart` release that fixes the advisories requires Python **≥ 3.10**, while `shared` declared `>=3.9`. Poetry therefore could not move past 0.0.20 no matter what it was asked to update — `poetry update --lock python-multipart` silently resolved back to the same version. The same floor blocks the starlette fixes, which also need ≥3.10.

The floor is now `>=3.11`, which is what the repository actually uses:

| Project | requires-python |
| --- | --- |
| Athena (all modules) | `3.11.*` ← lowest |
| athena/evaluation | `3.12.*` |
| AtlasMl | `^3.13` |
| iris, memiris, logos-* | `>=3.13` |

Nothing in the repo could ever have used `shared` on 3.9 or 3.10. I chose `>=3.11` over the strictly-minimal `>=3.10` because 3.10 is a version no consumer uses, and leaving it would just re-block the next dependency that drops it.

`shared` is **not published** — `edutelligence-shared` does not exist on PyPI, and consumers depend on it by local path (`shared = { path = "./shared", develop = true }`), so no external consumer is affected. (The `shared` package that *is* on PyPI is an unrelated project.)

### Not fixed here

- **`starlette`** — needs 1.3.1 to clear all three of its alerts. `fastapi` 0.115.11 pins `starlette >=0.40.0,<0.47.0`, so this needs a FastAPI bump to a release with an open-ended starlette range (0.141.1 requires just `starlette>=0.46.0`). Deferred as agreed. For whoever picks it up: `shared` is 328 LOC and touches only `BaseHTTPMiddleware` from starlette, so its own surface is small — but the bump propagates to every consumer of `shared`, which is what makes it worth its own PR.
- **`black`** — moved to the dedicated style PR.

### Steps for Testing

1. `cd shared && poetry check --lock` — passes (`All set!`).
2. `poetry install` on Python 3.11+ and import `shared.health` / `shared.security`.
3. The Logos Docker builds consume `shared` as a path dependency and exercise it in CI.

`poetry check --lock` and the version resolution were verified locally.


Claudia-Anthropica (dismissed):
@krusche The project metadata and lockfile consistently raise the Python floor to 3.11 and pin `python-multipart` 0.0.32. The repository consumer requires Python 3.13, the exact-head CI checks succeeded, and I found no blocking compatibility or correctness issue.

PR #998, opened 2026-09-16
`General`: Bump black to 26 and reformat

krusche:
### Motivation and Context

Closes the `black` alerts ([GHSA-3936-cmfr-pm3m](https://github.com/advisories/GHSA-3936-cmfr-pm3m) — arbitrary file writes from unsanitized input in the cache file name) for AtlasMl, logos-agent and memiris.

This is deliberately **separate** from the six per-subproject security PRs (#988, #989, #990, #994, #995, #996), because bumping black is not really a dependency update — it is a style migration, and it should be your team's decision on its own terms rather than something buried in a security diff.

### Description

**Black 26 promotes its new multiline-string handling into the *stable* style, with no flag to opt out.** I checked for an escape hatch — `--preview` / `--enable-unstable-feature` only *add* behaviour; there is no way to keep the old output. So taking the security fix means taking the reformat.

The characteristic change:

```python
# before
text(
    """
    SELECT ...
    """
)

# after
text("""
    SELECT ...
    """)
```

It is denser, and in places arguably less readable — `logos-agent/app/db.py` is where you will notice it most. Worth looking at that file specifically before approving.

**Scale of the reformat:**

| Project | Files reformatted |
| --- | --- |
| atlas/AtlasMl | 18 |
| logos/logos-agent | 1 (`app/db.py`) |
| memiris | **0** |

`memiris` already declared `^26.1.0`, so only its lock moves — it gets the fix for free.

**Safety:** formatting only, no behaviour change. Black verifies AST equivalence in its default safe mode, and I separately confirmed all 19 changed Python files still parse.

### `shared` is not included

Its black alert **cannot** be fixed yet: black 26 requires Python ≥ 3.10, while `shared` still declares `>=3.9`, so poetry silently resolves back to 25.1.0. The floor is raised to `>=3.11` in #996 — once that merges, `shared` can take the same bump. Its black alert stays open until then.

### Steps for Testing

1. `cd atlas/AtlasMl && poetry install && poetry run black --check .` — passes.
2. `cd logos/logos-agent && uv sync && uv run black --check .` — passes.
3. `cd memiris && poetry install && poetry run black --check .` — passes.
4. Run each project's test suite to confirm the reformat is inert.

All three `black --check` runs and the parse check were verified locally. The test suites were not run locally; CI covers them.

### Review tip

`git show --stat` on the single commit separates the intent cleanly: the four manifest/lock files are the actual change, the 19 `.py` files are mechanical output.

## Summary by CodeRabbit

- **Chores**
  - Updated Black formatting tools and pre-commit hooks across several projects.
  - Standardized code formatting, whitespace, line breaks, and quote styles throughout the codebase.
  - Reformatted database queries and configuration declarations without changing behavior.

- **Tests**
  - Reformatted test fixtures, assertions, and setup code while preserving existing coverage and expectations.

- **Generated Files**
  - Cleaned up whitespace in generated service integration files.


Claudia-Anthropica (approved):
@krusche The Python changes preserve parsed behavior apart from removing trailing spaces in two docstrings, and the lockfile deltas are confined to Black-related dependencies and regenerated artifact metadata. The captured test, lint, and build checks all succeeded; the remaining Black 25.1.0 pin in logos-orchestrator is already covered by the existing non-blocking CodeRabbit thread.

PR #1001, opened 2026-09-16
`Logos`: Add a GPU-less E2E suite, and fix the node registration it found broken

bensofficial:
## Motivation

Logos' expensive failures live at seams nothing currently tests: GPU/lane startup (compute-capability → attention-backend selection, nvcc preflight, fatal-CUDA classification, compile-cache poisoning, rejected pre-sharded checkpoints, stuck VRAM, watchdog reboot), the orchestrator↔node link, and the client contract. Today those are found in production — see `logos-workernode/research/bench_results/vllm_worker_crash_postmortem_20260310.md`.

The question this answers: **can that be tested without GPUs?** Mostly yes, because every GPU decision in the worker is mediated by a text boundary we can own.

| Decision | Read from | Simulated by |
|---|---|---|
| Attention backend, `TORCH_CUDA_ARCH_LIST` | `nvidia-smi --query-gpu=compute_cap` | shim binary |
| VRAM, degraded/`ERR!` telemetry | `nvidia-smi --query-gpu=…` | state-file-driven shim |
| Stuck CUDA contexts | `nvidia-smi --query-compute-apps` | the state file's VRAM ledger |
| nvcc / C-compiler preflight | `shutil.which`, `CUDA_HOME` | which shims are on `PATH` |
| Fatal CUDA, cache poisoning, broken shards | the vLLM log blob | replayed corpus |
| Lane liveness, sleep/wake, metrics | HTTP to the lane | fake vLLM server |

## ⚠️ It found a production bug on its first stack run

**`/logosdb/providers/logosnode/register` returned 400 for every request** — it called `add_provider` without `privacy_level`, which that function rejects outright. No worker node could bootstrap through it. Broken since the Spring migration (#597) introduced the requirement.

Nothing caught it: the unit tests mock `DBManager`, and a node registered through the UI takes a different path entirely. Worker nodes are self-hosted GPU hardware, so they now register as `LOCAL`; `test_registration_mints_a_usable_provider` pins it. All 1862 orchestrator unit tests still pass.

That is the case for this suite, made by the suite.

## What this adds

`logos/e2e/` — four tiers. **No worker or orchestrator code is stubbed**; only the layer below the worker is simulated.

| Tier | Drives | Needs | Runtime |
|------|--------|-------|---------|
| 1 — `tests/gpu` | the real `LaneManager` / `VllmProcessHandle` against simulated GPUs | nothing | ~75 s |
| 2 — `tests/node` | orchestrator + Postgres + worker nodes on simulated hardware | Docker | minutes |
| 3 — `tests/client` | the same stack through the real `openai` SDK | Docker | minutes |
| 4 — `logos-ui/e2e` | the UI in a real browser (Playwright) | Docker + Node | minutes |

The VRAM ledger is what makes Tier 1 more than canned output: the fake vLLM debits its `--gpu-memory-utilization` share on startup and credits it back on a clean exit, so the capacity planner watches memory actually move — and a scripted `leak_vram` reproduces a stuck CUDA context exactly as `_verify_vram_released` detects one.

Tier 2's nodes run the **real** `logos_worker_node` with the shims first on `PATH`, and self-register through the production endpoint rather than a seeded provider row. Simulated telemetry is asserted end to end — GPU simulator answers `nvidia-smi`, the worker parses it, the bridge ships it, the orchestrator stores it — down to per-card memory, identity, temperature, and the per-sensor health block the watchdog escalates on.

Tier 4 drives the real Keycloak redirect rather than injecting a token, because the redirect, the token exchange, and the role mapping the route guards read are all part of what breaks. It walks every admin route — each is a lazy `loadComponent`, so a broken import is a blank page that the unit tests and the static export both stay green through — and asserts the simulated GPUs actually render. It sits behind a `ui` compose profile so the inter-node tier stays fast.

## The corpus is the oracle

`harness/corpus/manifest.yaml` pairs each captured failure log with the verdict the worker must reach **and** the recovery it must perform:

- fatal CUDA error → no cache purge, no retry
- poisoned compile cache → purge and retry **exactly once**, `modelinfos/` survives
- rejected pre-sharded checkpoint → discarded, retried from the full checkpoint
- `corpus/benign/` → none of the above

That last group is the expensive direction: widening a fingerprint to catch a gated-repo 401 or a missing `ninja` would wipe every model's compile cache on the node. Two tests keep the corpus and `vllm_process.py` in step — every fingerprint needs an entry, every log file needs a manifest block.

**The growth rule: every production GPU incident lands a corpus file in the same PR as its fix.** That is what keeps a GPU-less suite tracking reality.

## What it cannot catch

Stated plainly in `README.md`: kernel-level and numerical errors, the real OOM threshold at a given GMU, actual NCCL behaviour at TP>1, driver Xid/GSP events (symptoms simulated, occurrence not), and anything timing-dependent. No GPU runner is available, so the mitigation is calibration-by-artifact — the corpus — not hardware.

## CI

New `logos_e2e.yml`, PR-blocking on `logos/**`:
- `gpu-sim` — Tier 1, no Docker. **Passing.**
- `stack` — compose up, Tiers 2 + 3, all service logs uploaded on failure
- `browser` — Tier 4 with the `ui` profile; traces, screenshots and video on failure

Base images are pre-pulled with retries: the stack job first failed on a Docker Hub `connection reset by peer`, and a registry flake reported as a test failure is the kind of CI noise that trains people to ignore a red E2E.

## Notes for review

- The stack is started separately from the tests on purpose; a suite that owns the lifecycle rebuilds images every run and tears down the evidence at the moment a failure needs inspecting. With the stack down, Tiers 2–4 skip with the command to start it.
- Node liveness reads `/internal/provider_status` (the worker registry), not `scheduler_state` — that facade only lists a provider once it has deployments attached, so a connected idle node reads as offline there. This was my own bug, caught the same way.
- `harness/stack/seed.sql` seeds a known admin key and a plain developer key; the client tier uses the developer key so a privilege leak shows up as a test failure.
- The corpus `.log` files needed a `.gitignore` negation — the repo-root `*.log*` rule was silently dropping all fifteen. The suite's own integrity check caught it.
- `harness/faults/` is scaffolded but empty: the WS frame-level fault proxy (dropped acks, duplicated `cmd_id`, replayed event ids) is the next milestone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed worker-node registration failures and ensured privacy levels are validated and applied correctly.

* **Tests**
  * Added comprehensive end-to-end coverage for GPU compatibility, telemetry, lifecycle management, watchdog recovery, registration, API authentication, and error handling.
  * Added browser-based coverage for navigation, access control, simulated hardware, statistics, and provider views.
  * Added simulated GPU profiles and failure scenarios for broader hardware coverage.

* **Documentation**
  * Documented testing tiers, coverage, and known limitations.


logos/e2e/harness/stack/docker-compose.e2e.yaml
```diff
@@ -0,0 +1,282 @@
...
+    depends_on:
+      logos-orchestrator:
+        condition: service_healthy
+    environment:
+      LOGOS_SIM_PROFILE: l40s
+      LOGOS_SIM_GPU_COUNT: "2"
+      LOGOS_URL: "http://logos-orchestrator:8080"
+      LOGOS_ALLOW_INSECURE_HTTP: "true"
+    healthcheck:
+      test: ["CMD-SHELL", "curl -fsS http://localhost:80/health > /dev/null || exit 1"]
```
Claudia-Anthropica: @bensofficial [high] This healthcheck can never succeed: the worker FastAPI app defines `/` but no `/health` route, so `curl -f` receives 404 and `docker compose up --wait` marks both simulated nodes unhealthy. This blocks both the stack and browser tiers before their tests run; point both node healthchecks at the existing `/` endpoint.

In `logos/e2e/harness/stack/docker-compose.e2e.yaml`, both simulated-node healthchecks request a nonexistent `/health` endpoint and therefore keep the compose stack unhealthy. Point both healthchecks at the worker's existing `/` endpoint.

</details>

logos/logos-orchestrator/src/logos/routers/logosnode.py
```diff
@@ -274,6 +275,12 @@ async def logosnode_register(data: LogosNodeRegisterRequest):
             auth_name="",
             auth_format="{}",
             provider_type="logosnode",
+            # add_provider rejects a missing privacy_level outright, so omitting
+            # it made this endpoint return 400 for every request. A worker node
+            # is self-hosted GPU hardware, which is what LOCAL means; operators
+            # renting third-party GPUs can move it to THIRD_PARTY_HARDWARE
+            # afterwards, the same as for any other provider.
+            privacy_level=ThresholdLevel.LOCAL.value,
```
Claudia-Anthropica: @bensofficial [high] This silently classifies every self-registered worker as `LOCAL`, the most trusted privacy tier, even though the comment explicitly acknowledges rented third-party workers. Such a worker can become connected and eligible for sensitive traffic before an operator manually downgrades it; require and validate the intended privacy level in `LogosNodeRegisterRequest` and pass that value to `add_provider` instead of defaulting to `LOCAL`.

In `logos/logos-orchestrator/src/logos/routers/logosnode.py`, node registration assigns every worker the most trusted `LOCAL` privacy level, which can expose sensitive requests to third-party hardware. Make `LogosNodeRegisterRequest` require a valid privacy level and pass that supplied value to `add_provider`.

</details>

Claudia-Anthropica (changes requested):
@bensofficial Both captured Logos E2E stack checks are failing; although their annotations only expose exit code 1, the first finding identifies a source-level startup failure shared by both jobs. The registration fix also assigns the highest privacy trust level to every newly registered worker, including third-party hardware.

Claudia-Anthropica (changes requested):
@bensofficial [high] The two existing blockers remain on this exact head. Both simulated nodes probe a nonexistent `/health` route, causing `docker compose up --wait` to fail before the stack and browser tests run, while self-registration still hard-codes `ThresholdLevel.LOCAL`, allowing third-party workers to receive traffic restricted to operator-controlled hardware.

logos/e2e/harness/nodesim/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+# protocol, but `nvidia-smi` and `vllm` on PATH are the shims from
+# harness/gpusim/bin.
+#
+# That keeps the whole inter-node tier runnable on an ordinary CI runner, and
+# keeps the thing being tested honest — no worker module is stubbed or patched.
+#
+# Build context is the `logos/` directory:
+#   docker build -f e2e/harness/nodesim/Dockerfile logos/
+
+FROM python:3.13-slim
```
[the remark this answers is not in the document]
bensofficial: Not taking this one. This image exists only to run the simulated worker in CI: it installs no CUDA stack (the `vllm` binary is a shim), it is never published, and it is layer-cached between runs, so the multi-stage split would trade a slimmer image for a longer Dockerfile without changing CI time in any way we could measure. Single-stage also keeps `docker run --rm  python ...` usable for poking at a failing node, which is how the registration bug on this branch actually got diagnosed.

Happy to revisit if this image ever gets published or grows a real dependency tree.

logos/e2e/pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+
+[project.optional-dependencies]
+# Tier 1 imports the worker package directly. Installed editable from the
+# sibling directory rather than vendored, so the tests always run against the
+# working tree.
+worker = ["pydantic>=2.9"]
+
+[build-system]
+requires = ["setuptools>=75"]
+build-backend = "setuptools.build_meta"
```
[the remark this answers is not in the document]
bensofficial: Declining for now, though the consistency point is fair — logos-orchestrator does build with poetry-core.

Two reasons. This package is a test harness installed with `uv pip install -e .`, never built or published, so the build backend has no consumer beyond that one command; and the GPU-simulation job is currently the only green tier on this PR, so swapping its build system is the one change here I cannot verify without running it (local test runs are off for this branch at the author's request). Trading a working install for convention, unverified, is not a trade worth making mid-review.

The lockfile half is a separate question: a pinned lockfile would make E2E runs reproducible, which is worth having. Better as its own change, with a CI run to prove it.

logos/e2e/harness/gpusim/bin/vllm
```diff
@@ -0,0 +1,174 @@
...
+    raw_gmu = _flag(argv, "--gpu-memory-utilization")
+    gmu = float(raw_gmu) if raw_gmu else 0.9
+    try:
+        sim = gpustate.load()
+    except gpustate.GpuSimStateError:
+        return 0.0
+    if not devices:
+        return 0.0
+    device = sim.device(devices[0])
+    return (device.memory_total_mb * gmu) if device else 0.0
```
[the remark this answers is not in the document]
bensofficial: Fixed in 928b6c7. `_resolve_vram_mb` returns a per-device mapping now, and `run_server` books each card its own amount. You are right that nothing exercises the mixed-profile TP path yet, but the ledger is what this simulator exists to get right — a capacity assertion is only as good as the numbers under it.

logos/e2e/tests/client/test_openai_contract.py
```diff
@@ -0,0 +1,71 @@
...
+        client.chat.completions.create(
+            model="definitely-not-a-real-model",
+            messages=[{"role": "user", "content": "hello"}],
+        )
+
+    assert caught.value.status_code in (
+        400,
+        404,
+        422,
+    ), f"unknown model returned HTTP {caught.value.status_code} — a client cannot act on that"
```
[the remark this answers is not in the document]
bensofficial: Fixed in 928b6c7. The test now asserts the body parses as an object and carries a message a client could surface. Worth catching: the test name promised a structured error while the assertion would have accepted an HTML error page with the right status.

logos/e2e/tests/gpu/test_gpu_compatibility_matrix.py
```diff
@@ -0,0 +1,132 @@
...
+
+    ``_auto_attention_backend`` only overrides when *every* card is pre-Ampere.
+    Forcing TRITON_ATTN on the mixed node would needlessly slow the newer card;
+    the sm75 card is handled by not scheduling FlashInfer work onto it.
+    """
+    env, handle, ctx = await spawned(GpuScenario(profiles=["rtx2080ti", "l40s"]))
+    try:
+        await handle.spawn(lane_harness.lane_config())
+        assert env.arg_value(ATTENTION_BACKEND_FLAG) is None
+        assert env.last_env().get("TORCH_CUDA_ARCH_LIST") == "7.5;8.9"
```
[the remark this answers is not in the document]
bensofficial: Fixed in 928b6c7, and this one caught a false claim. The test comment said the sm75 card is handled by not scheduling FlashInfer onto it; nothing in the worker does that, exactly as you describe — `gpu_devices` unset falls back to the worker-wide `all`, `CUDA_VISIBLE_DEVICES` is never set, and the pre-Ampere card stays visible. Rather than constrain the scenario and hide it, the test now asserts that exposure, and a companion test pins the mechanism that removes it (pinning the lane to the Ampere card). The underlying production question — whether a mixed node should auto-pin — is real and outside this PR.

logos/logos-ui/e2e/navigation.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    test(`${route.path} renders`, async ({ page }) => {
+      const failures: string[] = [];
+      page.on('pageerror', error => failures.push(error.message));
+
+      await page.goto(route.path);
+
+      await expect(page).toHaveTitle(route.title);
+      // A lazy chunk that fails to load leaves the shell up and the outlet
+      // empty, so the title alone is not enough — assert something rendered.
+      await expect(page.locator('router-outlet + *, main, [role="main"]').first()).toBeVisible();
```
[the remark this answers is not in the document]
bensofficial: Fixed in 928b6c7. Verified against `shell.html`: `<main class="main-content">` wraps the outlet, so the `main` fallback matched on precisely the blank-page failure the test exists to catch. Now requires `router-outlet + *`.

logos/logos-ui/e2e/navigation.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+test.describe('unauthenticated access', () => {
+  test.use({ storageState: { cookies: [], origins: [] } });
+
+  test('a protected route sends an anonymous visitor to the login page', async ({ page }) => {
+    await page.goto('/statistics');
+
+    // The guard must not let the shell paint first — a flash of admin chrome
+    // before the redirect is a real leak, not a cosmetic one.
+    await expect(page.getByRole('button', { name: /sign in with tum/i })).toBeVisible({ timeout: 30_000 });
```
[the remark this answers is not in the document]
bensofficial: Fixed in 928b6c7 along the lines you suggested — an `addInitScript` observer armed before `goto` records any shell render, and the test asserts none occurred after the login page appears.

logos/e2e/harness/stack/docker-compose.e2e.yaml
bensofficial: Fixed in 25b2e46. Confirmed independently before your review landed: every `/health` in logos_worker_node is the worker probing *vLLM*, and the app itself defines only `/`. Both node healthchecks now use `/`. This was the root cause of both red jobs — the containers were marked unhealthy and `compose up --wait` aborted before any test ran.

logos/logos-orchestrator/src/logos/routers/logosnode.py
bensofficial: Fixed in 25b2e46, and I took your stronger version. I had made it an overridable field defaulting to LOCAL; your point that a default still silently grants the most-trusted tier to any self-registering worker is the right one. It is now required and validated on `LogosNodeRegisterRequest` with no default.

One argument in favour that neither of us made: requiring it is costless *right now* precisely because the endpoint has been returning 400 for every request since #597, so it has no working callers to break. That stops being true the moment this merges.

logos/e2e/pyproject.toml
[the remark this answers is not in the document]
bensofficial: Following up with a correction, since I checked the guideline this cites.

`logos/AGENTS.md` documents the opposite of the lockfile half:

> **Dependency Management**: `uv` (lockfile: `logos-orchestrator/uv.lock`, committed). `pyproject.toml` still declares a Poetry-format `[tool.poetry]` packaging section, but there is no `poetry.lock` and installs go through `uv pip install .` / `uv sync`, not `poetry install`.

and for CI:

> The CI workflow installs dependencies via `uv` (`astral-sh/setup-uv`), not Poetry — there is no Poetry cache step anymore.

So "generate and commit `poetry.lock`, update CI to install from that lockfile" would move this package *away* from the documented standard, not toward it. The repo deliberately migrated off Poetry installs.

The part that does hold is the build backend: logos-orchestrator builds with `poetry-core`, and matching that is reasonable. The reproducibility goal is also right, but the artifact is `uv.lock`, not `poetry.lock`.

So the accurate follow-up is: switch the backend to `poetry-core` and commit a `uv.lock`. Worth doing, still worth its own change with a CI run behind it rather than blind mid-review. No issue needed — I will carry it.

logos/e2e/harness/nodesim/Dockerfile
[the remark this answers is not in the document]
bensofficial: Revisiting this one — you were right that a guideline backs it, and my earlier reply did not engage with that. `logos/AGENTS.md` does document the pattern:

> **Containerization**: Docker multi-stage build with `uv` (pinned)
> **Docker build**: Uses multi-stage build with `uv` (pinned version) ... Runtime stage uses slim Python image with `VIRTUAL_ENV=/opt/venv`.

So "CI-only image, not worth the complexity" was the wrong frame: there is a stated convention, and this Dockerfile does not follow it.

Not changing it in this PR, for a narrower reason than before: the stack tier went green for the first time on this exact image a few minutes ago, after several red runs, and local test runs are off for this branch at the author's request — so a rewrite of the working build is the one change here I could not verify before pushing. Converting it to the orchestrator's builder/runtime pattern is mechanical and I will do it once the tiers are stably green, or immediately if the author prefers.

Keeping the thread resolved to reflect that it is decided rather than ignored; the follow-up is mine to carry.

logos/e2e/harness/stack/docker-compose.e2e.yaml
```diff
@@ -0,0 +1,290 @@
...
+      interval: 2s
+      timeout: 5s
+      retries: 45
+    # Only consumed when the `ui` profile brings Traefik up; inert otherwise.
+    # The worker-node admin routes must outrank the webservice's /api/logosdb
+    # catch-all (200) — the orchestrator holds the node registry, and the
+    # webservice would 404 them.
+    labels:
+      - "traefik.enable=true"
+      - "traefik.http.routers.orch-api-logosnode.rule=PathPrefix(`/api/logosdb/providers/logosnode`)"
```
Claudia-Anthropica, line 103: @bensofficial [medium] This priority-201 prefix also captures the six operator-facing lane/calibration paths that production deliberately routes through Spring at priority 202. The UI sends a JWT and `{provider_id, ...}`, whereas the orchestrator endpoints require `logos_key`, so these actions return validation/auth errors and Tier 4 does not exercise the production route. Add the same exact-path webservice router used by the production compose file, with priority above this catch-all.

In `logos/e2e/harness/stack/docker-compose.e2e.yaml`, the orchestrator's priority-201 logosnode prefix steals the operator lane and calibration routes from the JWT-authenticated webservice, causing browser requests to hit incompatible orchestrator request models. Add a priority-202 webservice router for the production exact-path list so those actions pass through Spring before reaching the orchestrator.

</details>

logos/e2e/harness/gpusim/bin/vllm
```diff
@@ -0,0 +1,181 @@
...
+from harness.gpusim.server import run as run_server  # noqa: E402
+
+CORPUS_DIR = E2E_ROOT / "harness" / "corpus"
+
+#: Environment prefixes worth recording. These are the ones the worker sets
+#: deliberately; the rest is inherited noise.
+ENV_PREFIXES = ("VLLM_", "TORCH", "CUDA_", "NCCL_", "HF_", "FLASHINFER_", "CC", "LOGOS_")
+
+#: Recorded as a presence flag only — never the value.
+SECRET_ENV = {"HF_TOKEN"}
```
Claudia-Anthropica: @bensofficial [medium] `VllmProcessHandle` copies its parent environment into the vLLM subprocess, and the broad `LOGOS_` capture therefore writes values such as `LOGOS_API_KEY`, `LOGOS_INTERNAL_SECRET`, and `LOGOS_ADMIN_KEY` verbatim into `gpusim.json`; only `HF_TOKEN` is redacted. Redact every credential-bearing variable before recording, or replace the prefix capture with an allowlist of non-secret Logos settings.

In `logos/e2e/harness/gpusim/bin/vllm`, `_captured_env` persists inherited Logos credentials because `LOGOS_` is captured wholesale while only `HF_TOKEN` is redacted. Redact all token, key, secret, and password variables—or allowlist only non-sensitive Logos variables—before writing the invocation state.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The previous healthcheck and explicit privacy-level blockers are fixed on this head. Two medium issues remain: the browser stack misroutes operator lane actions, and the GPU simulator records inherited Logos credentials without redaction.

logos/e2e/harness/stack/docker-compose.e2e.yaml
bensofficial: Fixed in 62d7f8f. Added the same exact-path webservice router at priority 202, mirroring docker-compose.dev.yaml, so the six lane/calibration paths reach Spring rather than the orchestrator catch-all at 201.

You were right about the consequence as well as the cause: I had copied the orchestrator half of production's routing and not the webservice half, so Tier 4 would have been exercising a route production does not use — passing while telling us nothing about the real one.

logos/e2e/harness/gpusim/bin/vllm
bensofficial: Fixed in 62d7f8f. Confirmed the leak was live, not theoretical: nodesim's entrypoint sets `LOGOS_API_KEY` to the shared key minted at registration, and the worker copies its parent environment into the subprocess, so the real key was being written verbatim into `gpusim.json` — a file the tests read and CI collects on failure.

Redaction is now by name fragment (TOKEN/KEY/SECRET/PASSWORD/CREDENTIAL) rather than the allowlist option, so a credential added later is redacted by default instead of leaking until someone extends a list. Presence is still recorded as `<set>`. A test pins both directions: no secret value may appear anywhere in the serialised state, and a non-secret setting must still be captured — a redaction that swallowed everything would satisfy the first check alone.

logos/e2e/harness/stack/compose.py
```diff
@@ -0,0 +1,212 @@
...
+    while time.monotonic() < deadline:
+        if is_up():
+            return
+        time.sleep(1.0)
+    raise TimeoutError(f"orchestrator did not answer at {ORCHESTRATOR_URL} within {timeout_s:.0f}s")
+
+
+def ui_is_up() -> bool:
+    try:
+        return httpx.get(UI_URL, timeout=2.0, follow_redirects=True).status_code < 500
```
Claudia-Anthropica: @bensofficial [medium] This only proves that the static UI container is serving `/`. `logos-webservice` has no healthcheck and `logos-ui` depends only on `service_started`, while Angular's application initializer immediately fetches `/api/info` and throws on a non-2xx response. Consequently `up --ui` can return while Spring is still booting, leaving the first Playwright page load permanently blank and relying on the configured retry to recover. Probe `/api/info` through `UI_URL` and validate its Keycloak configuration before declaring the browser stack ready.

In `logos/e2e/harness/stack/compose.py`, `wait_for_ui` considers the stack ready when only the static UI root responds, even though Angular startup requires the unhealthchecked webservice's `/api/info` endpoint. Change the readiness probe to wait for a successful `/api/info` response through `UI_URL` with the expected Keycloak configuration before returning.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The previous lane-routing and credential-redaction blockers are fixed on this head. One medium issue remains: the browser stack declares readiness before the webservice endpoint required by Angular is available. Optional E2E and build checks were still running in the captured snapshot.

logos/e2e/harness/stack/compose.py
bensofficial: Valid — fixed in 217c5f4f1. Verified each part before changing it: `app.config.ts` registers `provideAppInitializer(() => initKeycloak())`, and `core/auth/keycloak.ts` fetches `/api/info` and throws unless the response carries `keycloak.issuer` and `keycloak.client_id`; the webservice serving it has no healthcheck and logos-ui only waits on `condition: service_started`.

It also explains a failure I had already misdiagnosed. A Playwright run spent 60s looking for the sign-in button and found nothing — I read that as a slow cold start and raised the timeout. The button was never going to appear: Angular had not bootstrapped, because `/api/info` was not being served yet. Treating the symptom left the race in place.

`ui_is_up()` now makes the same call the app makes and requires the same fields, and the timeout message prints the webservice logs since that is what a failure here is about. The comment in auth.setup.ts is corrected to say what actually happened rather than repeat my wrong explanation.

logos/logos-ui/e2e/navigation.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    // admin chrome painted first and was then replaced — and a flash of another
+    // tenant's navigation is a real leak, not a cosmetic one.
+    await page.addInitScript(() => {
+      (window as unknown as { __shellSeen: boolean }).__shellSeen = false;
+      const check = () => {
+        if (document.querySelector('app-shell, main.main-content')) {
+          (window as unknown as { __shellSeen: boolean }).__shellSeen = true;
+        }
+      };
+      new MutationObserver(check).observe(document.documentElement, { childList: true, subtree: true });
```
Claudia-Anthropica: @bensofficial [medium] `addInitScript` executes before the page markup is guaranteed to provide a `documentElement`. If it is still `null`, `observe()` throws after `__shellSeen` was initialized to `false`; the later assertion then reads that false value and passes even if the protected shell subsequently renders. Observe the `document` node itself, or defer attachment until a root exists while making setup failure visible to the test.

In `logos/logos-ui/e2e/navigation.spec.ts`, the initialization script can call `MutationObserver.observe` with a null `documentElement`, leaving `__shellSeen` false without installing the observer and allowing a protected-shell flash to go undetected. Attach the observer to `document` or reliably wait for a root node, and ensure observer setup failures cause the test to fail.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The previous routing, credential-redaction, and UI-readiness blockers are fixed, and all captured checks passed on this head. One medium issue remains: the anonymous-navigation observer can fail before attaching and let the protected-shell assertion pass falsely.

Claudia-Anthropica (changes requested):
@bensofficial The earlier routing, credential-redaction, and UI-readiness issues remain fixed, and all captured checks passed on this exact head. The unresolved medium issue in `navigation.spec.ts` remains: the init script calls `MutationObserver.observe(document.documentElement, ...)` before a root is guaranteed, so setup can throw while leaving `__shellSeen` false and let the assertion pass without an observer. Attach the observer to `document`, or defer attachment until a root exists while making setup failure visible to the test.

Claudia-Anthropica (changes requested):
@bensofficial All captured checks passed, and the prior routing, credential-redaction, and UI-readiness issues remain fixed. The unresolved medium issue in `navigation.spec.ts` remains on this head: the init script still calls `MutationObserver.observe(document.documentElement, ...)` without guaranteeing a root exists, so setup can throw after leaving `__shellSeen` false and the test can pass without an observer.

Claudia-Anthropica (changes requested):
@bensofficial All 52 captured checks passed, and the prior fixes remain present. One medium issue is still concrete on this exact head: `navigation.spec.ts` observes `document.documentElement` before guaranteeing that it exists, so observer setup can throw after leaving `__shellSeen` false and allow the protected-shell assertion to pass without an active observer. The existing unresolved thread covers this defect, so I have not duplicated its inline comment.

Claudia-Anthropica (changes requested):
@bensofficial All 52 captured checks passed, and the prior fixes remain present. One medium issue remains on this exact head: `navigation.spec.ts` observes `document.documentElement` before guaranteeing it exists, so setup can throw after initializing `__shellSeen` to false and let the protected-shell assertion pass without an active observer. The existing unresolved thread covers this defect, so I have not duplicated its inline comment.

Claudia-Anthropica (changes requested):
@bensofficial All 52 captured checks passed, and the prior fixes remain present. One unresolved medium issue remains on this exact head: `navigation.spec.ts` observes `document.documentElement` before guaranteeing it exists, so setup can throw after initializing `__shellSeen` to false and allow the protected-shell assertion to pass without an active observer; the existing unresolved thread already covers this defect.

logos/e2e/harness/gpusim/bin/vllm
```diff
@@ -0,0 +1,196 @@
...
+        sim = gpustate.load()
+    except gpustate.GpuSimStateError:
+        return []
+    visible = [d.index for d in sim.visible_devices()]
+    pinned = os.environ.get("CUDA_VISIBLE_DEVICES")
+    if pinned is not None:
+        if not pinned.strip():
+            return []
+        wanted = [int(p) for p in pinned.split(",") if p.strip().isdigit()]
+        visible = [i for i in visible if i in wanted]
```
Claudia-Anthropica: @bensofficial [medium] `CUDA_VISIBLE_DEVICES` defines both visibility and logical-device order, but this membership filter restores simulator order. With `CUDA_VISIBLE_DEVICES=2,0` and TP=1, real vLLM uses physical GPU 2 while the simulator returns GPU 0, so the ledger can certify placement and capacity against the wrong card. Preserve the pinned order when selecting the first `tensor_parallel_size` devices.

In `logos/e2e/harness/gpusim/bin/vllm`, `_target_devices` reorders `CUDA_VISIBLE_DEVICES` into simulator enumeration order, causing TP lanes to book VRAM against the wrong physical GPU. Preserve the environment variable's device order when filtering available devices and add a reversed-order TP=1 regression test.

</details>

Claudia-Anthropica (changes requested):
@bensofficial All 52 captured checks passed, and the prior routing, credential-redaction, and UI-readiness fixes remain present. Two medium issues remain on this exact head: the unresolved init-script observer can fail before attachment and falsely pass, while the GPU simulator reorders `CUDA_VISIBLE_DEVICES` before selecting TP devices and can book VRAM on the wrong physical card. The existing thread covers the observer issue; the inline comment below covers the device-order defect.

Claudia-Anthropica (changes requested):
@bensofficial All 52 captured checks passed, but two medium test-fidelity issues remain on this exact head. The anonymous-navigation observer can throw before attaching when `document.documentElement` is unavailable, leaving the false sentinel to pass, and the GPU shim filters `CUDA_VISIBLE_DEVICES` through simulator order, so a reversed pin such as `2,0` allocates the wrong physical card. The existing unresolved threads cover both findings, so I have not duplicated their inline comments.

Claudia-Anthropica (changes requested):
@bensofficial All 52 captured checks passed, but two medium test-fidelity issues remain on this exact head. The anonymous-navigation observer can throw before attaching when `document.documentElement` is unavailable, and the GPU shim restores simulator order instead of preserving `CUDA_VISIBLE_DEVICES` order; the existing unresolved threads cover both defects.

logos/logos-ui/e2e/navigation.spec.ts
bensofficial: Fixed in 911570c3. You were right that it could not fail: `__shellSeen` was set to false before `observe()`, so a throw on a null root left the flag false, the init-script exception did not fail the test, and the assertion passed whether or not the shell had flashed.

It now observes `document`, and it records any setup failure so the test asserts the watcher actually attached before trusting what it reports — the second half of your suggestion, which is what turns a silent pass into a visible one.

This is the second check in this PR that reported success because it never ran (the earlier one matched `main`, which shell.html always renders). That pattern is why I also made the specs typecheck in this commit: `e2e/tsconfig.json` asked for `types: ["node"]` while `@types/node` was never installed, so `tsc` could not run at all.

logos/e2e/harness/gpusim/bin/vllm
bensofficial: Fixed in 911570c3. Confirmed the semantics before changing it: the filter was `[i for i in visible if i in wanted]`, which rebuilds the simulator's own order, so `CUDA_VISIBLE_DEVICES=2,0` with TP=1 selected GPU 0 where real vLLM uses physical GPU 2.

The pinned list is now iterated in order, dropping indices that name no present device as CUDA does. The regression test pins it on a heterogeneous node on purpose — with two identical cards the wrong pick is invisible, because both allocations look the same.

Worth saying that this one mattered more than its severity suggests: VRAM bookkeeping is the thing the simulator exists to get right, so a lane recorded against a card it never touched would have let capacity and placement assertions certify the wrong hardware.

PR #1002, opened 2026-09-16
`Logos`: Build the orchestrator image from uv.lock

krusche:
### Motivation and Context

While reviewing #994, [a comment](https://github.com/ls1intum/edutelligence/pull/994#discussion_r4026625294) pointed out that the orchestrator's lockfile updates never reach the image. Digging in turned up three compounding problems: **the orchestrator's `uv.lock` was decorative, and the image re-resolved its dependencies on every build.** This fixes all three.

### 1. `uv.lock` recorded a machine-specific absolute path

On `main` the lock contained:

```
directory = "/Users/timgrobecker/TUM/S26/[CIT000000] BT/edutelligence/logos/logos-orchestrator/shared"
```

It churned whenever a different person ran `uv lock`, and could never be installed from on another machine. Pre-existing, not introduced by #994.

**Cause:** `shared` was declared *only* in `[tool.poetry.dependencies]`. uv sees that through poetry-core's built metadata, which hands back an **absolute** directory. I confirmed the mechanism by testing the alternatives — neither a real subdirectory nor a symlink avoids it.

**Fix:** declare it to uv directly, beside the existing `torch` entry:

```toml
[tool.uv.sources]
shared = { path = "../../shared" }
```

uv now records `directory = "../../shared"` — relative, portable, resolving in-repo with no symlink.

### 2. `.dockerignore` did not whitelist the lock

The context is a whitelist (`*` then `!path`), so `uv.lock` was absent and copying it failed with `failed to compute cache key … not found`. `iris/poetry.lock` was already whitelisted this way.

### 3. The Dockerfile ignored the lock

`uv pip install .` re-resolves from the manifest. It now runs `uv sync --locked`, with:

- **builder `WORKDIR` mirroring the repo layout** (`/build/logos/logos-orchestrator`) so `../../shared` resolves;
- **`UV_PROJECT_ENVIRONMENT=/opt/venv`** so uv syncs into the venv the runtime stage copies instead of creating `.venv`;
- **`--no-install-project`** so only dependencies land in the cached layer — the service's own source is copied into the runtime stage and imported via `PYTHONPATH`. The existing layer-caching design is preserved.

`--locked` rather than `--frozen` deliberately: a lock that has drifted from `pyproject.toml` should fail the build, not be silently worked around.

### Verification

Built the real image locally rather than reasoning about it:

- **Build succeeds**, and the image reports exactly what the lock pins — `aiohttp 3.14.3`, `pillow 12.3.0`, `langchain-core 1.6.3`, `langsmith 0.12.5`, `python-multipart 0.0.32`, `urllib3 2.8.0`, `shared 2.7`.
- **Imports work** in the runtime image: `fastapi`, `aiohttp`, `sqlalchemy`, `langchain_core`, and `shared.health` / `shared.security` from the path dependency.
- **The guard actually guards**, twice over. Raising a constraint in `pyproject.toml` without relocking fails the build with `The lockfile at uv.lock needs to be updated`. It then caught a *real* case unprompted: after #996 merged and changed `shared/pyproject.toml`, the orchestrator's lock was correctly reported stale, because it depends on `shared` by path. That cross-project drift was previously invisible.

### Note on the diff size

`uv.lock` shows a large diff because it was regenerated from scratch to drop the absolute path. **Dependency versions are unchanged** from #994 — the meaningful change is the two `directory =` lines.

### Follow-up worth considering

`logos/logos-agent` has the same shape (`uv pip install .`, no lock copied). Left untouched here to keep this reviewable, but the same three-part fix applies.

### Steps for Testing

1. `docker build -f logos/logos-orchestrator/Dockerfile -t orch .` from the repo root.
2. `docker run --rm orch python -c "from importlib.metadata import version; print(version('urllib3'))"` → matches `uv.lock`.
3. Edit any constraint in `logos/logos-orchestrator/pyproject.toml`, rebuild without running `uv lock` → build must fail.
4. `cd logos/logos-orchestrator && uv lock --check` passes, and running `uv lock` on another machine must not rewrite the `shared` path.


Claudia-Anthropica (approved):
@krusche The Docker build now consumes the portable lockfile, preserves the repository layout required by the `shared` path dependency, and installs into the runtime venv with locked resolution. The orchestrator image build and Logos tests succeeded; the only unfinished snapshot check was the unrelated, non-required vLLM worker build.

PR #1004, opened 2026-09-16
`General`: Fix the remaining high severity Dependabot alerts

krusche:
### Motivation and Context

Clears every remaining **high** severity [Dependabot alert](https://github.com/ls1intum/edutelligence/security/dependabot?q=is%3Aopen+severity%3Ahigh) that has an upstream fix — including the majors deferred from #988–#996, and the `black` work from #998 (which had gone stale after those merges; **#998 can be closed as superseded**).

High alerts: **334 → 53** after the six merges → **7 after this**. All 7 remaining are blocked upstream, not deferred; details at the bottom. Critical is already 0.

Every patched minimum is declared in the **build file**, not just the lock, so a later relock cannot resolve back onto a vulnerable release.

### Description

**Review the two commits separately** — the second is pure `black` reformatting and carries no logic.

#### `starlette` ≥ 1.3.1 — 17 alerts

`athena/tests`, `atlas/AtlasMl`, `iris`, `logos-agent`, `logos-orchestrator`, `shared`.

starlette only reaches 1.x through FastAPI, which capped it below 0.47 until 0.141.1. So FastAPI moves alongside it:

| Project | fastapi |
| --- | --- |
| atlas/AtlasMl | `^0.115.5` → `^0.141.1` |
| logos-agent | `==0.115.9` → `>=0.141.1,<1.0.0` |
| logos-orchestrator | `0.115.9` → `>=0.141.1,<1.0.0` |
| shared | `>=0.115.9` → `>=0.141.1` |

`shared` is the only place with code written directly against starlette (`BaseHTTPMiddleware`). Verified it still imports and still resolves that symbol on starlette 1.6.0 + fastapi 0.141.1.

#### `transformers` ≥ 5.10.0 — 6 of 7 alerts

`atlas/AtlasMl` additionally needed `sentence-transformers` `^4.0.2` → `^5.0.0`, which was capping transformers below 5. `logos-orchestrator` was already on the 5.x line. Reassuringly, logos already ran sentence-transformers 5.x, so the major is proven in this repo.

Verified the APIs actually used — `transformers.pipeline`, `transformers.AutoTokenizer`, `sentence_transformers.SentenceTransformer` — against transformers 5.17.0 / sentence-transformers 5.7.0 in a clean env. These are the only three entry points across AtlasMl and iris.

#### `cryptography` ≥ 50.0.0 (9), `pyarrow` ≥ 23.0.1 (1), `httpx2` ≥ 2.12.0 + `httpcore2` ≥ 2.10.0 (8)

All transitive. The `httpx2`/`httpcore2` advisories are new — they appeared while the earlier PRs were in flight, as `httpx` is superseded upstream.

#### `black` ≥ 26.3.1 — 5 alerts

`atlas/AtlasMl`, `logos-agent`, `memiris`, `shared`. `shared` was blocked until #996 raised its Python floor to ≥3.11 (black 26 needs ≥3.10); that has landed, so it's included now.

**black is pinned in two independent places per project.** Bumping `pyproject.toml` does nothing for the pre-commit hook, so the `iris`, `memiris` and `logos` configs move to `rev: 26.5.1` too — they were running a black *inside* the advisory range.

### Not fixed — blocked upstream, not deferred

- **`nltk`** (6 alerts). [GHSA-8mgp-746c-j5xp](https://github.com/advisories/GHSA-8mgp-746c-j5xp) covers `<= 3.10.3`, and **3.10.3 is the newest release on PyPI**. There is no version to move to. I re-checked rather than assuming it hadn't changed.
- **`transformers` in `iris`** (1 alert). `optimum[onnxruntime]` pulls in `optimum-onnx`, whose newest release (0.1.0) still requires `transformers <4.58.0`. iris cannot leave the 4.x line while it depends on optimum. The constraint now carries a comment recording why it is pinned there, so the next person doesn't repeat the attempt.

### Steps for Testing

1. `poetry check --lock` in `athena/evaluation`, `athena/tests`, `atlas/AtlasMl`, `iris`, `memiris`, `shared` — all pass.
2. `uv lock --check` in `logos-orchestrator` and `logos-agent` — both pass. (The orchestrator needs its `shared` path dependency to resolve; see #1002.)
3. `poetry run pytest` per project — **AtlasMl and iris matter most here**, since they carry the transformers/sentence-transformers major.
4. `black --check` across AtlasMl, logos, memiris, iris, shared — all clean.
5. `pre-commit run --config logos/.pre-commit-config.yaml --all-files` — black passes.

Everything except the per-project pytest suites was verified locally (they pull torch/Weaviate testcontainers); CI covers those.

### Note on #1002

#1002 also touches `logos/logos-orchestrator/pyproject.toml` and `uv.lock`, so whichever merges second will need a rebase. They're independent changes — #1002 makes the image build *from* the lock; this one changes what the lock contains.

## Summary by CodeRabbit

- **New Features**
  - Competency records now support an indexed course identifier, improving course-based filtering and lookup.

- **Security and Reliability**
  - Updated web framework, machine-learning, and supporting packages to newer compatible versions.
  - Added minimum-version safeguards for several transitive dependencies to help prevent vulnerable versions from being installed.

- **Maintenance**
  - Updated development tooling and standardized code formatting across multiple services and tests.


Claudia-Anthropica (dismissed):
@krusche The dependency floors and regenerated locks are internally consistent, and the formatting commit preserves Python behavior apart from removing trailing spaces in two docstrings. The red `build_image (athena/modules/modeling/module_modeling_llm)` check failed on an unrelated registry `unknown blob` error. The red Logos `Test` check remains unattributed because the captured evidence contains only a generic exit code and identifies no failing step or test.

krusche:
CI is green — 71 checks pass, 0 failures. Notes on the two reviews:

**@Claudia-Anthropica** — both red checks you flagged are resolved.

The Logos `Test` failure that the captured evidence left unattributed was real, and I have fixed it in 254d124bd. It was a behaviour change from the starlette 1.x / fastapi 0.141 bump in this PR: `app.include_router()` no longer flattens routes into `app.routes`, it appends a single `_IncludedRouter`, so three tests iterating `app.routes` and reading `.path` failed with `AttributeError: '_IncludedRouter' object has no attribute 'path'`. I reproduced that outside the suite before fixing it, and switched the tests to the public `fastapi.routing.iter_route_contexts`, which yields every effective route with the prefix applied and `.methods` intact, in registration order — order matters, because those tests exist to prove concrete routes are registered ahead of the `/v1/{path:path}` catch-alls. Production code is unaffected: the only non-test use of `app.routes` is `get_openapi(routes=app.routes)`, which FastAPI handles itself.

Verified by running the suite inside the built image: 1849 passed, with the only failures being 13 wrapper tests that need `/logos-ui/public/claude-logos.sh`, which the runtime image does not ship. Test counts reconcile exactly with the CI run (1862 both).

Your read on `build_image (athena/modules/modeling/module_modeling_llm)` was right — `unknown blob`, and this PR touches nothing under `athena/modules`. Re-run, now green.

**@coderabbitai** — the `course_id` finding is correct, and I checked it against the code rather than taking it at face value. `atlas/AtlasMl/atlasml/clients/weaviate.py:235` does pass the `indexFilterable` value into `index_searchable`, and `Property` exposes `index_filterable`, `index_searchable` and `index_range_filters` as distinct parameters, so the numeric `course_id` does not get the filterable index it asks for.

I am deliberately **not** fixing it here. It is pre-existing on `main` — this branch only reformatted the neighbouring property dicts under black 26, not that line — and it is a data-layer behaviour change that wants its own PR and its own migration testing, rather than being folded into a dependency-security change where it would be easy to miss. Raising it separately.

Claudia-Anthropica (dismissed):
@krusche The dependency floors and regenerated locks are consistent, and the FastAPI route-enumeration follow-up correctly traverses included routers while preserving registration-order assertions. The reported `course_id` mapping defect is unchanged from the merge base rather than introduced here; the snapshot contains 71 passing checks and no failures.

wasnertobias (dismissed):
build green, changes in Logos make sense. Ready to go out of my view.

PR #1016, opened 2026-09-16
`Iris`: Move to transformers 5, and run the mypy hook serially

krusche:
### Motivation and Context

Two things, both needed to get the high-severity queue and `main` clean.

**1. `main`'s `Lint` is still red.** #1012 merged before this commit reached its branch, so main got the testcontainers fix but not the mypy one. That commit is included here.

**2. Closes the 3 high severity `transformers` alerts on iris** (`< 5.3.0`, `< 5.5.0`, `< 5.10.0`).

### The mypy crash — root cause, not a pin

`mypy INTERNAL ERROR` (exit 2) reproduced on CI every single time and never locally. That inconsistency was the clue: **pre-commit splits the file list across concurrent mypy processes that share one `.mypy_cache`.** The run logs show repeated `Success: no issues found in 4 source files` rather than one pass over 59. On a cold cache they race to populate it and crash; a warm cache needs no writes, so local runs passed while CI — always cold — failed every time.

`require_serial: true` gives one process and one writer. Verified by deleting `.mypy_cache` and running CI's exact command (`pre-commit run --all-files --hook-stage manual memiris`): it crashes reliably without the flag and is clean with it, across repeated cold runs. **mypy stays on 2.x** — this was a hook configuration bug, not a version problem.

### The transformers alerts — two blockers

**`optimum` capped the whole lock.** `optimum-onnx` (newest release 0.1.0) still requires `transformers <4.58.0`, pinning even the transformers that *ships in the runtime image* — while `optimum` itself is imported only by `models/global_search_intent/train.py`. Nothing installs it: neither CI nor the Dockerfile passes `--with train`, and the image downloads the pre-built ONNX model rather than exporting it. It is no longer declared, with a comment recording why and how to install it for the export step.

**iris could not resolve its lock at all:**

```
Because memiris depends on langfuse (^4.15.2) and iris depends on langfuse (^3.15.0),
memiris is forbidden.
```

memiris moved to langfuse `^4` in #1006; iris pins `^3` and consumes memiris as a path dependency. **That break would have blocked any future iris dependency change**, not just this one. memiris returns to `^3.15.0`, which also reverts the `start_as_current_observation` call the v4 bump required.

*Why revert memiris rather than move iris forward:* nothing in memiris needed langfuse 4 — no advisory asks for it, it was a routine update. Moving **iris** to 4 is a real migration: v4 drops `update_current_trace`, and trace attributes (`user_id`, `session_id`, `tags`) move to `propagate_attributes()`, a **context manager that must wrap pipeline execution** rather than being applied after it, as iris does today at `abstract_agent_pipeline.py:206`. That restructures the tracing layer and needs verification against a langfuse backend, so it belongs in its own change.

### Verification

- `pre-commit run --all-files --hook-stage manual memiris` on a **deleted** `.mypy_cache` — passes; reproducibly crashes without `require_serial`
- memiris: **68 tests pass** on langfuse 3.15.0
- iris resolves to **transformers 5.12.1**, `optimum`/`optimum-onnx` absent, `poetry check --lock` clean
- The only transformers API iris uses is `AutoTokenizer` (`global_search_intent_classifier.py:20`), checked against the 5.x line

### Alert status

High severity went **10 → 4** and lands at **1** once this merges:

- 3 × `transformers` (iris) — closed by this PR
- 6 × `nltk` — dismissed separately: the advisory covers `<= 3.10.3` and 3.10.3 is the newest release on PyPI, and nltk is genuinely used (`word_tokenize`, `sent_tokenize`, `ngrams`) across five Athena modules, so removal would be a functional rewrite
- 1 × `black` (logos-agent) — already fixed on main (`black==26.5.1` in manifest and lock); stale since 2026-09-02, closes on the next rescan

### Steps for Testing

1. `cd memiris && rm -rf .mypy_cache && poetry install --with dev`, then from the repo root with that venv on PATH: `pre-commit run --all-files --hook-stage manual memiris`
2. `cd memiris && poetry run pytest tests` — 68 passed
3. `cd iris && poetry check --lock && poetry install && poetry run pytest`
4. The ONNX export script now needs `poetry run pip install "optimum[onnxruntime]"` alongside the train group

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of type-checking workflows by preventing concurrent cache conflicts.
  - Updated tracing behavior to maintain compatibility with the supported Langfuse integration.

- **Compatibility**
  - Updated language-model tooling to support newer transformer versions.
  - Aligned supporting library versions across components to preserve model training, export, and tracing workflows.


Claudia-Anthropica (approved):
@krusche The dependency set is internally consistent: Iris resolves Transformers 5.12.1 without Optimum, the Langfuse rollback restores the prior 3.x-compatible API, and the serialized mypy hook addresses the shared-cache race. All captured Iris and Memiris test, lint, and build checks succeeded. The existing CodeRabbit thread already captures the non-blocking training-script installation-message mismatch, so I am not duplicating it.

PR #1018, opened 2026-09-16
`Logos`: Rate-limit publicly reachable and unauthenticated endpoints

### LINKED ISSUE #839: `Logos`: Rate-limit publicly reachable and unauthenticated endpoints
## Problem

The webservice and the orchestrator expose endpoints that can be called without a valid credential, and none of them is rate limited. A search across `logos-webservice` finds no `RateLimiter`, no `bucket4j`, and no filter that counts requests per client.

Two categories are affected.

**Publicly reachable without any credential** (verified against logos-dev):

| Endpoint | Method | Status without key |
| --- | --- | --- |
| `/health` (orchestrator) | GET | 200 |
| `/info` (webservice) | GET | 200 |

Both are intentionally open — `/info` tells the UI which Keycloak to talk to, `/health` is the liveness signal. But every call to `/health` runs a database query (the deployment inventory), so an unauthenticated caller can generate database load at whatever rate they like.

**Authenticated endpoints whose rejection path is unmetered.** `/v1/models`, `/v1/chat/completions` and `/logosdb/get_model_health` correctly answer 401 without a key, but there is no limit on how often that 401 can be provoked. A leaked API key can be tested for validity indefinitely, and each attempt costs a key lookup in the database. Brute forcing a key is not realistic given their entropy — confirming a key someone already holds is.

## Proposal

Add a rate limit in front of both categories:

- A per-IP limit on the unauthenticated endpoints (`/health`, `/info`), generous enough for monitoring probes but bounded.
- A per-IP limit on authentication *failures* across the API-key paths, so repeated 401s from one source back off. Successful authenticated traffic should stay governed by the existing per-key limits rather than by this.

Worth deciding at the same time: whether `/health` should keep serving a database-backed payload at all, or answer from a cached snapshot refreshed on a timer. That removes the amplification factor independently of any rate limit.

## Context

Came up while reviewing the model-level health check, but it is not specific to that change — the gap predates it and applies to every endpoint listed above.

bensofficial:
## Summary

Closes #839.

`/health` (orchestrator) and `/info` (webservice) take no credential and each run backend work on every call (a DB query, in `/health`'s case), so an unauthenticated caller could drive unbounded backend load. `/v1/models`, `/v1/chat/completions` and `/logosdb/get_model_health` correctly answer 401 without a key, but the rejection path was unmetered, so a leaked API key could be tested for validity indefinitely.

This adds two kinds of per-IP rate limiting, in-process on both services (neither runs more than one instance, so no shared/distributed store is needed):

- **Public endpoints** (`/health`, `/info`): every request is bounded — generous enough for monitoring probes, bounded against anyone else. Orchestrator: `LOGOS_PUBLIC_ENDPOINT_RPM` (default 60/min). Webservice: `logos.rate-limiting.public-endpoint-requests-per-minute` / `LOGOS_RATE_LIMIT_PUBLIC_RPM` (default 60/min).
- **Auth failures** on the API-key paths (`/v1/models`, `/v1/chat/completions`, `/logosdb/get_model_health`): only a 401 spends budget, so a client that keeps authenticating successfully never approaches the limit — successful traffic stays governed entirely by the existing per-key RPM/TPM limits. Orchestrator: `LOGOS_AUTH_FAILURE_RPM` (default 20/min). Webservice: `logos.rate-limiting.auth-failure-requests-per-minute` / `LOGOS_RATE_LIMIT_AUTH_FAILURE_RPM` (default 20/min).

The orchestrator (Python) reuses its existing `InMemoryRateLimiter` (extended with a peek/consume pair for the "only failures spend budget" pattern). The webservice (Java) gets a new lightweight `IpRateLimiterService` in the same spirit as `RateLimitService`/`RateLimitingProperties`/`RateLimitType` in ls1intum/Artemis, scaled down to this project's needs (no Bucket4j/Redis/Hazelcast, since logos-webservice is a single instance) — a 429 with `Retry-After` on both services either way.

Left out of scope, per the issue: whether `/health` should serve from a cached snapshot instead of hitting the DB on every call — worth a separate follow-up.

## Test plan

- [x] Orchestrator: `pytest tests/unit` — 1874 passed
- [x] Webservice: `mvn test` — 467 passed, including new `IpRateLimiterServiceTest`, `PublicConfigControllerRateLimitTest`, `ModelHealthControllerRateLimitTest`
- [x] New unit tests for the failure-only-spends-budget behavior on both sides (repeated bad-key calls trip 429; repeated successful calls never do)

## Summary by CodeRabbit

- **New Features**
  - Added a per-IP rate-limiting gateway for model, jobs, administrative, health, and service-information requests.
  - Added configurable protection against repeated failed API-key authentication attempts.
  - Added `/redoc` support and configurable API documentation visibility.
  - Scheduler state now requires the internal secret.
- **Bug Fixes**
  - Rate-limit responses return HTTP 429 with retry timing.
  - Successful authentication does not consume failed-authentication quota.
- **Security**
  - Restricted internal traffic and trusted proxy handling to configured networks.


logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/common/IpRateLimiterService.java
```diff
@@ -0,0 +1,125 @@
...
+     * Whether {@code clientIp} still has an unspent authentication-failure slot, without spending one. Call before
+     * attempting authentication; pair with {@link #consumeAuthFailure} on the failure path only.
+     */
+    public boolean hasAuthFailureBudget(String clientIp) {
+        if (!properties.enabled() || properties.authFailureRequestsPerMinute() <= 0 || clientIp == null) {
+            return true;
+        }
+        return count(key(clientIp, "auth_fail")) < properties.authFailureRequestsPerMinute();
+    }
+
```
Claudia-Anthropica, line 118: @bensofficial [medium] `hasAuthFailureBudget` releases the deque lock before `ModelController` performs authentication and later calls `consumeAuthFailure`. Concurrent requests can therefore all observe the same available slot, execute their key lookups, and record failures afterward, exceeding the configured RPM by up to the request concurrency. Reserve an in-flight slot atomically before authentication, then remove that reservation on success or convert it to a failure timestamp on failure, and cover this with a concurrent test.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/common/IpRateLimiterService.java`, authentication-budget checking and failure consumption are not atomic, so concurrent invalid requests can exceed the configured limit. Implement atomic per-IP in-flight reservation that is released on successful authentication and converted to a failure timestamp on failed authentication, and add a concurrent regression test.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The sequential rate-limit paths are well covered, but the webservice does not enforce the configured authentication-failure threshold under concurrent requests. The budget check and failure recording are separate critical sections around the database lookup, allowing a burst of invalid-key probes to exceed the limit.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/common/IpRateLimiterService.java
bensofficial: Fixed in ebc1f03a: replaced the check-then-record pair with `tryReserveAuthFailureSlot` (atomic check-and-reserve, called before authentication) plus `releaseAuthFailureSlot` (called on successful auth to give the slot back). Added a concurrency regression test (`concurrentAuthFailures_neverExceedTheConfiguredLimit`) that fires 50 concurrent callers against a limit of 5 and asserts exactly 5 get through.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/controller/ModelController.java
```diff
@@ -66,15 +71,30 @@ public ResponseEntity<?> getModels(@RequestAttribute("authContext") AuthContext
...
+            throw new RateLimitExceededException(60);
+        }
+
         String apiKey = extractApiKey(request);
         if (apiKey == null) {
             return ResponseEntity.status(401).body(Map.of("detail", "Invalid or missing API key"));
         }
         return modelService.getModelHealth(apiKey)
-            .map(ResponseEntity::ok)
+            .map(result -> {
```
Claudia-Anthropica: @bensofficial [medium] The reservation is released only after `modelService.getModelHealth()` finishes, but that method performs the orchestrator health call after validating the key. With the default limit, 20 concurrent requests using valid keys can therefore occupy every failure slot while their downstream calls are in flight, causing the next valid request to receive 429; a downstream exception also retains its slot until expiry. Separate key validation from health retrieval and release the reservation immediately once the key is known valid, including on non-401 exceptions, and cover concurrent valid requests with a regression test.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/controller/ModelController.java`, successful authentication retains its failure-budget reservation throughout downstream model-health work and may never release it on an exception. Refactor the flow to release immediately after successful key validation and add a concurrent valid-request regression test.

</details>

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/common/IpRateLimiterService.java
```diff
@@ -0,0 +1,127 @@
...
+ * if that ever changes, this is the place to swap in a shared store (e.g. Redis) behind the same interface.
+ */
+@Service
+public class IpRateLimiterService {
+
+    private static final Duration WINDOW = Duration.ofMinutes(1);
+
+    private final RateLimitingProperties properties;
+
+    private final ConcurrentHashMap<String, ArrayDeque<Long>> requestWindows = new ConcurrentHashMap<>();
```
Claudia-Anthropica, line 57: @bensofficial [medium] `requestWindows` never removes an IP key; `prune` only removes timestamps when that same key is used again. Because `/info` is unauthenticated, traffic from continually new source addresses accumulates map entries for the process lifetime. The Python limiter now has the same issue through `_request_windows.setdefault`, so both services need bounded or expiry-based key eviction rather than timestamp pruning alone.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/common/IpRateLimiterService.java`, `requestWindows` permanently retains every per-IP key after its timestamps expire. Implement safe bounded or expiry-based key eviction here and in the equivalent Python per-IP store, and add a high-cardinality churn test.

</details>

logos/logos-webservice/src/main/resources/application.properties
```diff
@@ -54,3 +54,11 @@ logos.auth.sync.admin-base-url=${KEYCLOAK_ADMIN_BASE_URL:http://localhost:8085}
 logos.auth.sync.realm=${KEYCLOAK_REALM:tum}
 logos.auth.sync.client-id=${KEYCLOAK_SYNC_CLIENT_ID:logos-sync}
 logos.auth.sync.client-secret=${KEYCLOAK_SYNC_CLIENT_SECRET:}
+
+# --- Per-IP rate limiting for endpoints that take no valid credential ---
+# Covers public endpoints (/info) and the failure path of API-key auth
+# (/logosdb/get_model_health): the orchestrator keeps the matching limits for
+# its own /health, /v1/models and /v1/chat/completions (see logos.rate_limiter).
+logos.rate-limiting.enabled=${LOGOS_RATE_LIMITING_ENABLED:true}
```
Claudia-Anthropica, line 62: @bensofficial [medium] Neither `docker-compose.yaml` nor `docker-compose.dev.yaml` passes these new webservice variables—or the orchestrator's `LOGOS_PUBLIC_ENDPOINT_RPM` and `LOGOS_AUTH_FAILURE_RPM`—into the containers. Consequently, setting them in the normal host environment or Compose `.env` file has no effect and deployed users are locked to the defaults, including the 60-RPM public limit shared by clients behind one NAT. Wire all five variables into the appropriate service environments in both Compose files.

In `logos/logos-webservice/src/main/resources/application.properties`, the advertised rate-limit environment overrides are not forwarded by the repository's production or development Compose services. Add explicit environment mappings for the three webservice variables and the two orchestrator variables to both Compose configurations.

</details>

Claudia-Anthropica (changes requested):
@bensofficial The prior Java check/record race is fixed by the atomic reservation, so the resolved thread can stay closed. Three current-head issues remain: valid concurrent health requests retain failure-budget reservations through downstream work, both per-IP stores retain identities indefinitely, and the advertised environment overrides are not passed into the shipped Compose services. These can reject valid traffic or allow unauthenticated traffic to grow memory without bound.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/controller/ModelController.java
bensofficial: Fixed in e114b564: split ModelService.getModelHealth into resolveAccessibleModelsForApiKey (DB-only key/permission check) and getModelHealthForAccessibleModels (the orchestrator call). ModelController now calls releaseAuthFailureSlot immediately after the former succeeds, before the orchestrator call runs — so the reservation is held only for the fast DB lookup, never for the slow/exception-prone downstream call. Added ModelHealthControllerReleaseTimingTest, a deterministic regression test (latches force the ordering rather than relying on real concurrency/timing) that fails with a clear 429 assertion error if release is moved back to after the downstream call — I verified this manually by temporarily reverting the fix.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/common/IpRateLimiterService.java
bensofficial: Fixed in e114b564: replaced the plain ConcurrentHashMap.computeIfAbsent + synchronized(window) pattern with ConcurrentHashMap.compute (atomic remap per key, so no separate lock is needed), and added an @Scheduled evictStaleWindows sweep (every 5 minutes, longer than the 1-minute rate-limit window so a window with no new activity is guaranteed to have nothing left to prune by the time it runs). Made the window duration injectable via a package-private constructor so IpRateLimiterServiceTest can use a short window and observe real-time eviction directly (evictStaleWindows_reclaimsWindowsOnceTheyHaveNothingLeftAfterPruning), plus a companion test confirming a window with recent activity is left alone. Applied the same sweep to the Python orchestrator's InMemoryRateLimiter (inline, lock-guarded, triggered from the existing entry points rather than a new background task) — added a matching high-cardinality churn test there too.

logos/logos-webservice/src/main/resources/application.properties
bensofficial: Fixed in e114b564: wired all five variables into both docker-compose.yaml and docker-compose.dev.yaml — LOGOS_RATE_LIMITING_ENABLED, LOGOS_RATE_LIMIT_PUBLIC_RPM, LOGOS_RATE_LIMIT_AUTH_FAILURE_RPM on logos-webservice, and LOGOS_PUBLIC_ENDPOINT_RPM, LOGOS_AUTH_FAILURE_RPM on logos-orchestrator, each defaulted to match the values already documented in application.properties / rate_limiter.py. Validated both compose files with 'docker compose config -q'.

logos/logos-orchestrator/src/logos/rate_limiter.py
```diff
@@ -74,9 +119,34 @@ def record_tokens(self, key: str, token_count: int) -> None:
...
+            self._prune_requests(dq, cutoff)
+            return len(dq) < limit
+
+    def consume(self, key: str) -> None:
+        """Spend one request slot for `key`, unconditionally."""
+        now = time.monotonic()
+        with self._lock:
+            self._sweep_locked(now)
+            dq = self._request_windows.setdefault(key, deque())
+            dq.append(now)
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 974fac87: authentication now tracks in-flight lookups separately, records only completed failures atomically, and releases reservations on successful authentication or errors. The focused orchestrator auth/rate-limiter tests pass.

logos/logos-orchestrator/src/logos/routers/monitoring.py
```diff
@@ -36,7 +37,17 @@ async def health():
...
+
+    It is also reachable without any credential, and every call runs the
+    database query below, so it is rate limited per source IP — generous
+    enough for monitoring probes, bounded against anyone else.
     """
+    # Local import: see the matching comment in logos.auth.authenticate_api_key
+    # for why logos.rate_limiter must not be imported at module scope here.
+    from logos.rate_limiter import PUBLIC_ENDPOINT_RPM, enforce_ip_rate_limit
+
+    enforce_ip_rate_limit(get_client_ip(request), "health", PUBLIC_ENDPOINT_RPM)
```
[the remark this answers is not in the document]
wasnertobias: Fixed in ee75c2a3: the development orchestrator mapping is now 127.0.0.1:18080:8080, so the direct debug port is not exposed on host interfaces.

logos/logos-webservice/src/test/java/de/tum/cit/aet/logos/logoswebservice/common/IpRateLimiterServiceTest.java
```diff
@@ -0,0 +1,217 @@
...
+                    }
+                    catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
+                        return;
+                    }
+                    if (limiter.tryReserveAuthFailureSlot("9.9.9.9")) {
+                        admitted.incrementAndGet();
+                        // Simulates a "key turned out valid" caller: release right away, as
+                        // ModelController now does, rather than holding the slot through downstream work.
+                        limiter.releaseAuthFailureSlot("9.9.9.9");
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 974fac87: the concurrent valid-request test now exercises the separate pending-reservation path, where every reservation is admitted and immediately released independently of the failure RPM. The Java focused suite passes.

logos/logos-webservice/src/test/java/de/tum/cit/aet/logos/logoswebservice/configuration/ModelHealthControllerRateLimitTest.java
```diff
@@ -0,0 +1,93 @@
...
+        "INSERT INTO team_model_permissions (team_id, model_id) VALUES (2001, 5001)",
+        "INSERT INTO team_provider_permissions (team_id, provider_id) VALUES (2001, 6001)"
+    }, executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD)
+    void successfulAuthReleasesTheReservedSlot() throws Exception {
+        when(modelHealthClient.getModelHealth()).thenReturn(List.of(Map.of("name", "gpt-4", "status", "UP")));
+
+        // Each successful call reserves-then-releases its slot, so many more
+        // than the configured limit must succeed without ever tripping 429.
+        for (int i = 0; i < 5; i++) {
+            mvc.perform(post("/logosdb/get_model_health")
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 974fac87: the two shared-context tests use distinct explicit remote addresses, 192.0.2.10 and 192.0.2.20, so in-memory limiter state cannot leak between them.

logos/logos-webservice/src/test/java/de/tum/cit/aet/logos/logoswebservice/configuration/ModelHealthControllerReleaseTimingTest.java
```diff
@@ -0,0 +1,102 @@
...
+    @MockitoBean OrchestratorModelHealthClient modelHealthClient;
+
+    @Test
+    void aSecondValidRequestSucceedsWhileTheFirstIsStillBlockedDownstream() throws Exception {
+        CountDownLatch reachedDownstream = new CountDownLatch(1);
+        CountDownLatch releaseDownstream = new CountDownLatch(1);
+        when(modelHealthClient.getModelHealth()).thenAnswer(invocation -> {
+            reachedDownstream.countDown();
+            releaseDownstream.await(5, TimeUnit.SECONDS);
+            return List.of(Map.of("name", "gpt-4", "status", "UP"));
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 974fac87: an AtomicInteger gates only the first downstream mock invocation, and its latch is released in finally; later calls return immediately.

PR #1019, opened 2026-09-16
`Logos`: Add a refresh button to the provider page

### LINKED ISSUE #997: `Logos`: Provider page shall have refresh button
Refresh button shall trigger the fetching of all available models instantly.

wasnertobias:
Fixes #997

The cloud model sync — the orchestrator re-reading every cloud provider's `/v1/models` listing — only ran on its 15-minute interval or as a side effect of adding or editing a provider. An operator looking at stale or missing models had no way to ask for the pass to run now.

## Changes

**Webservice**
- New admin endpoint `POST /api/logosdb/refresh_models` (`ProviderController` + `ProviderService.refreshModels()`) that announces a cloud model sync to the orchestrator via the existing `notifyRefresh(false, true)` machinery. The pass is scheduled, not awaited — the endpoint returns as soon as it has been handed off.

**UI**
- Refresh button in the provider page toolbar (admin-only page, matching the admin-only endpoint).
- After the trigger, the page refetches the model catalogue and the expanded provider's connections every 2s until two consecutive snapshots match (max ~16s), so newly synced models appear while the operator watches. The button spins while running and a banner shows if the trigger fails.

## Tests

- `ProviderControllerTest`: endpoint requires `logos_admin`; admin call verifies `notifyRefresh(false, true)` (a plain pipeline refresh would leave the upstream listings unread until the next interval).
- `providers.spec.ts`: settle behavior (stops once two snapshots match, keeps polling while the list moves), refetch of an expanded provider's connections, no double-trigger while running, error surfacing.
- Local runs: webservice full suite 460/460, providers spec 13/13, `npm run build` (static export gate) green.


logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ProviderService.java
```diff
@@ -187,6 +187,19 @@ public Map<String, Object> disconnectModelProvider(DisconnectModelProviderReques
...
+    /**
+     * Trigger an immediate cloud model sync in the orchestrator: every cloud
+     * provider's {@code /v1/models} listing is re-read now instead of at the
+     * next 15-minute interval. The pass is scheduled on the orchestrator and
+     * writes the resulting models and links itself, so nothing is stored here.
+     * Not transactional — there is no write to defer the notification past,
+     * which is exactly what lets the pass start at once.
+     */
+    public Map<String, Object> refreshModels() {
+        orchestratorNotificationService.notifyRefresh(false, true);
```
Claudia-Anthropica: @wasnertobias [medium] The explicit refresh always returns 200 before `sendRefresh` runs: `notifyRefresh` invokes an `@Async void` method whose missing-configuration and HTTP-failure paths return or catch internally. Consequently an unavailable orchestrator is reported as “triggered,” and the UI cannot surface the failed handoff. Use an acknowledgement-returning notification path here that waits only for the orchestrator to accept/schedule the refresh and propagates delivery failures.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ProviderService.java`, the manual refresh reports success without knowing whether the asynchronous notification was delivered. Add an acknowledgement-returning notification method for this endpoint and propagate missing-configuration and HTTP failures as a non-2xx response.

</details>

logos/logos-ui/src/app/features/providers/providers.ts
```diff
@@ -289,6 +294,67 @@ export class Providers implements OnInit {
...
+  }
+
+  private async waitForSyncToSettle(): Promise<void> {
+    let previous = await this.snapshotModelLists();
+    for (let round = 0; round < Providers.REFRESH_SETTLE_MAX_ROUNDS; round++) {
+      await new Promise<void>((resolve) => {
+        this.refreshSettleTimer = setTimeout(resolve, Providers.REFRESH_SETTLE_INTERVAL_MS);
+      });
+      const snapshot = await this.snapshotModelLists();
+      if (snapshot === previous) return;
```
Claudia-Anthropica: @wasnertobias [medium] One unchanged two-second interval is not evidence that this asynchronous sync has settled. The orchestrator schedules the pass after responding and then contacts providers serially, so these snapshots can match before the first database update—or before a later provider is reached—causing the UI to stop while its catalogue is still stale. Require an observed update followed by stability, or poll an explicit completion status, before returning.

In `logos/logos-ui/src/app/features/providers/providers.ts`, an unchanged snapshot before any observed sync update terminates polling even though the asynchronous refresh can still be running. Require an observed update followed by stability, or poll an explicit backend completion status, before returning.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The admin authorization and cloud-sync flag are covered, and the captured checks completed successfully. Two asynchronous-boundary defects can make the button silently finish before the refresh is delivered or observed, so changes are required.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/orchestrator/OrchestratorModelSyncClient.java
```diff
@@ -0,0 +1,64 @@
...
+            headers.set("Authorization", "Bearer " + internalSecret);
+            var response = restTemplate.exchange(
+                orchestratorUrl + "/internal/cloud_model_sync_status",
+                HttpMethod.GET,
+                new HttpEntity<Void>(headers),
+                Map.class);
+            return response.getBody() instanceof Map<?, ?> body && Boolean.TRUE.equals(body.get("running"));
+        } catch (Exception e) {
+            log.warn("Failed to fetch cloud model sync status from orchestrator: {}", e.getMessage());
+            return false;
```
Claudia-Anthropica: @wasnertobias [medium] `isSyncRunning()` maps every status failure—including a transient timeout or rolling-deploy 404—to `false`. The UI interprets that value as confirmed completion and stops polling even though the previously accepted sync may still be writing, leaving the catalogue stale. Propagate an unknown/error result and let the UI retry within its existing cap (or surface the failure); only report `false` after a successful orchestrator response explicitly says `running: false`.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/orchestrator/OrchestratorModelSyncClient.java`, status lookup failures are returned as “not running,” causing an accepted refresh to be treated as complete prematurely. Propagate status lookup failures and make the UI retry them within its polling cap; only return “not running” from a successful orchestrator response with `running: false`.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The acknowledged trigger and in-flight counter address the two earlier failure modes, and the captured Logos tests, lint, and relevant UI/webservice builds succeeded. However, a failed status lookup is still interpreted as completed, which can stop polling while an accepted sync remains in progress.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/configuration/service/ProviderService.java
wasnertobias: Fixed in de66a6fc6. The webservice now has an acknowledgement-returning path: `OrchestratorNotificationService.sendRefreshSync` sends the same `POST /internal/refresh_pipeline` request synchronously and returns true only when the orchestrator accepts and schedules the refresh (false on unconfigured/unreachable). `ProviderService.refreshModels` propagates a failed handoff as `IllegalStateException`, and `ProviderController` maps that to 503 with an error body — so the UI can surface a failed trigger instead of assuming success. Covered by `refreshModels_reports503WhenTheOrchestratorCouldNotBeReached` and the new `OrchestratorNotificationServiceTest` cases.

logos/logos-ui/src/app/features/providers/providers.ts
wasnertobias: Fixed in de66a6fc6. Instead of inferring completion from snapshot stability, the UI now polls an explicit completion status: `POST /api/logosdb/model_sync_status` → orchestrator `GET /internal/cloud_model_sync_status`, which reports whether a sync pass is running or queued via a pass-in-flight counter in `CloudModelSyncService` (`is_busy()`). The counter accounts for the interval pass (counted only while the pass runs) and out-of-band refreshes (counted from task creation until the loop exits, so a queued follow-up keeps the sync busy). The providers page polls every 2s and refetches the model lists until the pass reports done, capped at ~90s; an unreachable status endpoint reads as done so a rolling deploy cannot wedge the button. Covered by the new `providers.spec.ts` polling tests, `test_internal_sync_status_endpoint.py`, and the busy-counter tests in `test_cloud_model_sync.py`.

Claudia-Anthropica (changes requested):
@wasnertobias The synchronous handoff and 503 mapping corroborate the triggering reply, while the in-flight status resolves the earlier snapshot-settling issue. However, `OrchestratorModelSyncClient.isSyncRunning()` still catches every status lookup failure and returns `false`; the controller consequently reports confirmed completion, and the UI stops polling while an accepted sync may still be running. The captured checks succeeded, but this existing medium-severity thread remains unresolved.

Claudia-Anthropica (changes requested):
@wasnertobias The current head still catches every status lookup failure in `OrchestratorModelSyncClient.isSyncRunning()` and returns `false`; `modelSyncStatus()` consequently reports completion and the UI exits polling after one refetch even if the accepted sync is still running. The captured checks completed successfully, but this existing medium-severity thread remains unresolved.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/orchestrator/OrchestratorModelSyncClient.java
wasnertobias: Fixed in 1ac6d78b3. `OrchestratorModelSyncClient.isSyncRunning()` now returns a nullable `Boolean`: `null` when the state could not be read (unconfigured, any transport failure, or an answer without a usable `running` field), and `false` only on a successful orchestrator response that explicitly says `running: false`. `model_sync_status` passes that `null` through (`{"running": null}`), and the providers page treats an unreadable status as unknown, not done — it retries it on the next poll round within the existing 45-round cap instead of settling the wait. Covered by the new `OrchestratorModelSyncClientTest` (true / explicit false / unreachable / unconfigured / malformed body), the controller's unknown-state case, and two UI spec tests (retry-until-cap on persistent failures, stop only on an explicit idle after mixed unreadable rounds).

Claudia-Anthropica (approved):
@wasnertobias The nullable status path now preserves unreadable states as unknown and retries them, while only an explicit `false` terminates polling; this resolves the remaining review concern without regressing the acknowledged trigger or busy-counter behavior. The captured webservice, UI, orchestrator test, lint, and title checks succeeded, with only non-required image-build jobs still pending.

PR #1021, opened 2026-09-16
`Logos`: Priority queues — team priority with role tiebreak (#986)

### LINKED ISSUE #986: `Logos`: Priority Queues
Priority Queues are needed so most important traffic is threated with the appropriate attention.

Right now in the webUI there is no possibility to easily set Priorities for teams, applications or members.

Per default the priority within a team shall always be:
application > app admin > developer

They shall be configurable individually, though.

The overall teams priorities shall be defined by Logos Admins only.

wasnertobias:
Closes #986.

Important traffic now gets the right treatment in the local model queue.

**What**

- **Team priority** — new nullable `teams.priority` (1–10, `NULL` = unset, i.e. today's policy behavior is unchanged for all untouched teams). Logos admins set it via `PATCH /api/teams/{id}/priority` and a new **PRIORITY** column in the team management table exposing every value 1–10 (read-only for other roles).
- **Role tiebreak** — within equal effective priority, callers are ordered by role rank: application keys (2) before app/logos admins (1) before developer/service traffic (0), then FIFO. This gives the default intra-team ordering *application > app admin > developer*.
- **Key priority "unset" state** — `0` is now consistently the persisted marker for "no explicit key priority": `ApiKeyFactory` writes `0` for newly auto-provisioned developer keys, so team priorities apply to their traffic out of the box, and changeset 036 re-points the column default to `0` for future raw inserts. Existing rows are deliberately untouched: a stored `1` on a developer key is ambiguous (legacy default vs. deliberately pinned LOW) with no reliable provenance, so pinned overrides survive; the changeset documents the per-key / per-team opt-in reset for teams that want the legacy behavior gone.

**Design**

Effective queue priority per request (relevant for local queueing; cloud requests are accepted immediately):

```
key override (default_priority != 0)  →  team priority (!= NULL)  →  policy priority  →  0
```

The full-precision value (1–10) survives into the queue: the heap orders by `(-raw_priority, -role_rank, ts)`, so a team at 7 dequeues before a plain 5 inside the NORMAL bucket, while buckets (HIGH > NORMAL > LOW) still dominate. Per-key overrides (`api_keys.default_priority`, #805) remain the individual escape hatch.

Note: log/monitoring consumers report the bucket name, so a raw 7 is logged as "normal" (documented in the queue README).

**Deploy**

Roll the **webservice** (Liquibase 035 + 036) out **before** the orchestrator — the orchestrator selects `teams.priority` directly from the database.

**Tests**

- orchestrator: `pytest tests/unit` — 1876 passed (new: raw-priority + role-rank ordering, precedence matrix, escalation keeps the role rank)
- webservice: `mvn test` — 465 passed (new: `TeamControllerTest` coverage for 200/403/400/404 and the NULL reset; `LiquibaseBaselineTest` coverage that 036 preserves existing pinned priorities and unsets only future defaults)
- ui: `npm run build` passes; `team-management.spec.ts` green, including the full 1–10 option set and a non-bucket selection

## Summary by CodeRabbit

* **New Features**
  * Logos admins can set, update, or clear team queue priorities from Team Management.
  * Team priority values range from 1–10, with labels for Low, Normal, and High.
  * Queue ordering now considers API key priority, team priority, caller role, and FIFO order.
  * Queue priority details are included in team information and authentication context.

* **Documentation**
  * Updated authentication, pipeline, and queue documentation to describe priority resolution and ordering rules.


logos/logos-orchestrator/src/logos/pipeline/pipeline.py
```diff
@@ -55,35 +55,65 @@ def _privacy_ok(threshold: str, level: str) -> bool:
...
+        team_priority: The Logos admin's priority for the key's team, or
+            0/None when unset.
         policy_priority: The policy's ``priority`` value (may be 0/None).
 
     Returns:
         The effective integer priority for the request's queue entry.
     """
     if default_priority:
         return int(default_priority)
+    if team_priority:
```
Claudia-Anthropica, line 94: @wasnertobias [medium] This fallback is unreachable for normal developer keys: `ApiKeyFactory.createDeveloperKey()` and the database schema initialize `default_priority` to `1`, so line 87 returns before any configured team priority is considered. Consequently, setting a team to 2–10 moves its application-key traffic but leaves its auto-provisioned app-admin/developer keys in LOW; the new tests only exercise synthetic `default_priority=0` requests.

In `logos/logos-orchestrator/src/logos/pipeline/pipeline.py`, auto-provisioned developer keys persist `default_priority=1`, making the team-priority fallback unreachable for their traffic. Introduce a persisted unset state for key priority, use it for new and legacy auto-provisioned developer keys while preserving explicit overrides, and test team-priority resolution using that production default.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The new team-priority precedence is unreachable for auto-provisioned developer keys, so team settings do not apply to normal app-admin/developer traffic; see the inline finding. The already-open CodeRabbit selector finding is independently confirmed by the current template and remains unresolved. The non-required `pre-commit` check is red, but the captured evidence contains only exit code 1, so I could not attribute it to a specific hook or this change.

logos/logos-ui/src/app/features/team-management/team-management.html
```diff
@@ -72,6 +76,29 @@ <h1 class="page-title">Team Management</h1>
...
+              title="Queue priority of this team's traffic (applies to all its keys without an explicit key priority)"
+              [value]="team.priority ?? ''"
+              [disabled]="prioritySavingId() === team.id"
+              (click)="$event.stopPropagation()"
+              (change)="$event.stopPropagation(); changeTeamPriority(team, $any($event.target).value)"
+            >
+              <option value="">Default</option>
+              <option [value]="1">Low</option>
+              <option [value]="5">Normal</option>
+              <option [value]="10">High</option>
```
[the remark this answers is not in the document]
wasnertobias: Fixed in a1a7d7a32: the select now renders Default plus every value 1-10, keeping Low/Normal/High as annotations for the bucket values (e.g. Normal (5)). Added DOM tests for the complete option set and for selecting the non-bucket value 7 via a change event.

Claudia-Anthropica (changes requested):
@wasnertobias The selector and new-key fallback are now fixed. However, migration 036 rewrites every developer key at priority 1 to unset—including explicitly pinned LOW overrides—so those keys can unexpectedly inherit a higher team or policy priority; the existing thread remains blocking. The non-required `pre-commit` check is red, but its captured evidence contains only exit code 1, so I could not attribute it to a hook or this change.

logos/logos-orchestrator/src/logos/pipeline/pipeline.py
Claudia-Anthropica: @wasnertobias [medium] The fallback is reachable for newly provisioned keys now, but the explicit-override part remains unresolved: migration 036 rewrites every developer key with `default_priority = 1`, and its own comment acknowledges that deliberately pinned LOW overrides are erased. Those keys can then inherit a team or policy priority as high as 10 without owner intent. Keep 0 for new auto-provisioned keys, but preserve ambiguous existing priority-1 values unless reliable provenance or an opt-in migration distinguishes defaults from explicit overrides; add a regression test for an existing developer key explicitly pinned to LOW.

wasnertobias: Agreed - fixed in a1a7d7a32 on the webservice side: 0 was already the persisted 'not explicitly set' marker (the admin key API defaults a missing value to 0), so ApiKeyFactory.createDeveloperKey now writes 0, and new changeset 036 resets the legacy auto-provisioned developer keys (key_type='developer' AND default_priority=1) to 0 and re-points the column default. Explicit 2-10 values and all non-developer keys are untouched; a developer key explicitly pinned to 1 cannot be told apart from the legacy default, so it is reset as well (re-pinnable via the key update API) - documented in the changeset. Tests: LiquibaseBaselineTest.migration036_resetsAutoProvisionedDeveloperKeysToUnsetPriority covers the reset and the preserved values, and the (0, team, policy) resolution test now documents 0 as the production default for auto-provisioned developer keys.

Claudia-Anthropica (changes requested):
@wasnertobias The selector and new-key default are fixed, but migration 036 still unconditionally converts every developer key at priority 1 to 0. Because the current UI and API permit explicitly pinning developer keys to 1, those keys will silently begin inheriting team or policy priorities after deployment; I staged a follow-up on the existing thread. The non-required `pre-commit` check remains red, but the snapshot provides only exit code 1, so I could not attribute it to a specific hook or this change.

logos/logos-orchestrator/src/logos/pipeline/pipeline.py
Claudia-Anthropica: @wasnertobias Thanks for confirming the ambiguity. This remains blocking because the current key-settings UI accepts priority 1 and `updateKey` persists it, while migration 036 unconditionally changes every such developer key to 0; after deployment, an explicitly LOW key therefore inherits its team or policy priority until someone discovers and re-pins it. Please preserve ambiguous existing values unless provenance or an opt-in transition distinguishes legacy defaults, and add a regression covering an existing developer key explicitly pinned to LOW.

Claudia-Anthropica (approved):
@wasnertobias The team-priority path, complete 1–10 selector, and raw-priority/role queue ordering are consistent on this head. Migration 036 now preserves existing priority-1 keys while changing only the default for future inserts, resolving the remaining migration concern. The non-required `pre-commit` check is red, but its captured evidence reports only exit code 1, so I could not attribute it to this change.

logos/logos-orchestrator/src/logos/pipeline/pipeline.py
Claudia-Anthropica: @wasnertobias Thanks, the current migration preserves existing priority-1 values and the regression test covers the pinned-LOW case, so this is fixed now.

wasnertobias: Agreed - the bulk reset from a1a7d7a32 is gone. Changeset 036 now only re-points the column default to the 'unset' marker 0 (the factory keeps writing 0 for newly provisioned keys); existing rows are left untouched, so a developer key deliberately pinned to LOW keeps its explicit override and can no longer be silently lifted to team/policy priority. The changeset comment documents the opt-in for teams that want the legacy reset: per key via the key update API (an explicit default_priority of 0) or per team via the documented one-liner after checking that no key is pinned LOW. New regression test: LiquibaseBaselineTest.migration036_preservesExistingPrioritiesAndUnsetsFutureDefaults asserts that a pinned-LOW developer key (1) and the other explicit values survive, and that only a new row omitting the column picks up the default 0.

logos/logos-ui/src/app/features/team-management/team-management.ts
```diff
@@ -148,6 +152,41 @@ export class TeamManagement implements OnInit {
...
+    this.priorityError.set('');
+    this.teams.update((list) => list.map((t) => (t.id === team.id ? { ...t, priority } : t)));
+    this.prioritySavingId.set(team.id);
+    try {
+      await this.teamService.updateTeamPriority(team.id, priority);
+    } catch {
+      this.teams.update((list) => list.map((t) => (t.id === team.id ? { ...t, priority: previous } : t)));
+      this.priorityError.set(`Failed to update the queue priority of '${team.name}'.`);
+    } finally {
+      this.prioritySavingId.set(null);
```
Claudia-Anthropica: @wasnertobias [medium] `prioritySavingId` represents only one row, so starting a save for team B replaces team A's ID while A is still pending, and completion of either request clears the saving state for both. This re-enables an in-flight selector; another edit can then race the first PATCH, allowing an older response or rollback to leave the persisted priority and optimistic UI inconsistent. Track pending saves per team and add a test covering overlapping updates on different rows followed by another edit to an in-flight row.

In `logos/logos-ui/src/app/features/team-management/team-management.ts`, the single shared `prioritySavingId` loses track of concurrent per-team saves and permits racing updates with stale rollback state. Track pending saves independently by team ID, keep each team's selector disabled until its own request completes, and add an overlapping-save regression test.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The priority propagation, migration preservation, selector, and queue ordering are consistent on this head, and the resolved migration thread remains resolved. One UI race remains: the shared saving ID allows overlapping updates to re-enable an in-flight team selector, so later edits can leave the displayed and persisted priorities inconsistent. The non-required `pre-commit` check is red, but its captured evidence contains only exit code 1, so I could not attribute it to this change.

logos/logos-ui/src/app/features/team-management/team-management.ts
wasnertobias: Fixed in 8ae155f7c: pending saves are now tracked per team (prioritySaving is a Set of team ids instead of a single id). Each team's selector stays disabled until its own PATCH settles, so starting a save for team B no longer unlocks team A's in-flight selector and a second edit can no longer race the first request or its rollback. New regression test: overlapping saves on two rows, then completion of only the first - asserts that only the finished row unlocks. Also: the red pre-commit check was a black formatting violation in priority_queue.py; the file is reformatted and pre-commit is green locally across all branch files.

wasnertobias:
✅ Approved — the latest head covers priority ordering, role tie-breaking, migration behavior, and overlapping UI saves without a remaining actionable blocker.

logos/logos-orchestrator/src/logos/pipeline/pipeline.py
```diff
@@ -55,35 +55,65 @@ def _privacy_ok(threshold: str, level: str) -> bool:
...
+        return int(team_priority)
     if policy_priority:
         return int(policy_priority)
     return 0
```
wasnertobias: `0` is normalized to the NORMAL bucket downstream, but it is passed as `raw_priority=0`; this makes default requests rank below explicit NORMAL traffic with raw priority 5 within the same bucket. It also prevents the role-rank tiebreak from applying between default requests and explicit NORMAL requests. Return NORMAL's raw value (5), or normalize before enqueueing, and add a regression covering the all-unset/default case.

wasnertobias (commented):
❌ Rejected — follow-up review found a blocking priority-normalization bug. Default/unset traffic returns raw priority 0, but that value is placed in the NORMAL bucket and ordered below explicit NORMAL raw priority 5, so role tie-breaking and historical default ordering are bypassed. Please normalize the raw value before enqueueing and add a regression test.

logos/logos-orchestrator/src/logos/pipeline/pipeline.py
wasnertobias: Fixed in ab60480b6. resolve_queue_priority now returns Priority.NORMAL (5) instead of 0 when key, team and policy are all unset, so the entry's raw_priority matches the bucket Priority.from_int already chooses for it. Default traffic now ranks level with explicit NORMAL (5) in the same bucket and the role-rank tiebreak applies between them. Added two regressions: test_all_unset_resolves_to_normal_raw_not_zero (resolver returns 5 and raw matches its bucket) and test_default_priority_ranks_with_explicit_normal_by_role_rank (a default application-key request dequeues ahead of an explicit NORMAL developer request, proving the tiebreak is reached). Full unit suite green (1878 passed, 1 xfailed); pre-commit hooks pass.

PR #1023, opened 2026-09-17
`Logos`: Fix the request volume chart and cut the statistics page load time

### LINKED ISSUE #1022: Logos: Request volume is broken
On stats page.

Split it into two pages/tabs at this section. Also remove VRAM remaining. 

Explore the use of MVs for stats page to decrease load time. Test based on prod db snapshot. Same for models page and other pages, overall the load times shall be xplored and improved.

wasnertobias:
## Summary

The issue carried four things. All four are here, plus the two other pages the
issue asked to look at — both of which turned out to be worse than the statistics
page.

Every number below was measured against production data: either by running the
page's own query on the production host (`ssh logos`, read-only), or against a
full snapshot of it restored locally. The two differ by roughly 4x, so each
figure says which it is.

| Page | Query | Before (prod) | After |
|---|---|---|---|
| Statistics | 8 aggregates per load | 6,460 ms | ~50 ms¹ |
| Statistics | enqueue event payload | 25 MB / load | gone |
| Models | `findAllWithPricing` | 9,648 ms | ~2 ms¹ |
| Billing | `findTeamBudgetHistory` | 18,264 ms | ~2,400 ms² |

¹ snapshot figure; the production host runs ~4x slower in absolute terms.
² extrapolated from the snapshot's 4,746 ms → 693 ms at the same factor.

## 1. Request volume was broken

Not slow — **wrong**. The chart was built in the browser from the raw enqueue
event of every request in the range, and the server capped that list at 200,000
rows ordered **oldest first**. The default 30-day window holds 416,722 requests
on production, so the chart silently lost everything past the cut: the most
recent **12 days 22 hours** were a flat line while the KPI card directly above it
counted all 416,722 requests.

It now renders `stats.timeSeries`, which is aggregated in the database over the
whole range, arrives already bucketed, and costs a few kilobytes instead of
**25 MB per page load**. The event payload, its delta stream, the client-side
re-bucketing and the three repository queries behind them are gone.

## 2. Split into two tabs

Per the discussion, split **thematically** rather than literally at that section,
because the page was answering two unrelated questions at two unrelated refresh
rates:

- **Requests** — KPI strip, request volume, distribution, recent requests. Under
  the time range and the scope filters.
- **Infrastructure** — provider selector, VRAM utilization, lane health, workers
  & GPUs. Always the latest sample; the time range never applied to it, which was
  part of what made the combined page misleading.

The active tab is mirrored into `?tab=` so links and the back button work. The
scope/time controls now render only on the tab they actually drive.

## 3. VRAM remaining removed

Panel, component and its range slider deleted. It was the only consumer of the
VRAM *history* — every other panel reads just the latest sample per provider.

## 4. Materialized views for the statistics page

**An expression index is not the answer, and this is worth stating because it was
the obvious first move.** Every aggregate filtered `log_entry` on
`COALESCE(timestamp_forwarding, timestamp_request, timestamp_response)`, which no
index covered. I built that index and measured it: it does not help the default
view, because 30 days is ~60% of the table and at that selectivity a sequential
scan is the cheaper plan — the planner correctly keeps choosing it, and several
queries got *slower* with the index present. Counting 400k rows into 120 buckets
is O(n); only pre-aggregation removes it.

`log_entry_hourly_stats` (migration 035) collapses the whole history to **12,591
rows / 3.4 MB** — a factor of 55.

- **Only whole hours**, so `max(bucket_hour) + 1 hour` is a watermark the reader
  can trust.
- **Ids, not derived values** — `provider_id`/`model_id` stay ids and the reader
  joins live, so a renamed model or changed privacy level needs no refresh.
- **Sums and counts, never averages** — averaging pre-averaged hours would weight
  a quiet hour like a busy one.

The reader unions the rollup for hours below the watermark with `log_entry` for
the partial head hour and everything newer, through one shared split point
(`logos_stats_rollup_window`). Sharing it is deliberate: if two aggregates
disagreed by an hour about where the rollup ends, one would double-count the
overlap and another would drop a gap, and nothing would look wrong.

**Correctness never depends on the refresh.** Stale, lagging or failed moves work
back to `log_entry`, never numbers. Sub-hour buckets bypass the rollup entirely
and use the expression index, which *is* right at that selectivity.

The expression index is still created — it is what makes the live tail and the
short presets cheap.

### Three queries nothing rendered

`findQueueDepth`, `findRuntimeByColdStart` and `findLastEventTs` ran on every
aggregate push and are read by **no component**. ~700 ms per push on production
for data nobody displayed. Removing `findQueueDepth` also removed the one
aggregate a rollup cannot serve (`PERCENTILE_CONT(0.95)`).

## 5. Models page — 9.6 s on production

9,500 ms of it in one `LEFT JOIN LATERAL` resolving each model's most recently
used provider: 195 ms × 49 models. A comment claimed migration 014's
`(model_id, timestamp_request)` made this a single index seek. It did not — that
index does not carry `provider_id`, and the lateral filters on
`provider_id IS NOT NULL`, so the planner chose `idx_log_entry_performance_window`
instead, which is covering but has `model_id` **third**. Equality on a non-leading
column is not seekable, so every model walked the index until one of its rows
appeared.

Migration 036 adds `(model_id, timestamp_request DESC) INCLUDE (provider_id)
WHERE provider_id IS NOT NULL`. Snapshot: **252 ms → 1.7 ms**.

## 6. Billing page — 18.3 s on production

`log_entry_cost` joined its token rollup as `LEFT JOIN LATERAL ... ON NOT
le.cost_finalized AND ...`. A lateral is evaluated for every row and only *then*
filtered by its `ON` clause, so every settled row paid for a full `usage_tokens`
aggregation before the join discarded it: the plan runs **415,233 aggregations to
keep 22**.

Moving the rollup into the `CASE` branches that use it as a scalar subquery makes
it evaluate only when that branch is taken. **Verified equivalent over the full
snapshot: all 689,682 rows identical, identical totals.** Snapshot 4,746 ms →
693 ms. The changeset is `runOnChange`, so it is an edit in place rather than a
duplicated definition.

## Database changes

| Migration | What |
|---|---|
| `023_token_cost_decomposition.xml` | `log_entry_cost` restructured (`runOnChange`) |
| `035_log_entry_hourly_stats.xml` | rollup + expression index + `logos_stats_rollup_window`; pins `search_path` on the `logos_*` functions |
| `036_log_entry_model_provider_recent_index.xml` | models-page seek index |

The `search_path` pin is required, not incidental: `CREATE`/`REFRESH MATERIALIZED
VIEW` run with a restricted `search_path`, and `logos_price_usage` referenced
`providers` unqualified, so any rollup over `log_entry_cost` failed with
*"relation providers does not exist"* until the functions carried their own.

`RequestLogStatsRefreshService` refreshes hourly (`REFRESH ... CONCURRENTLY`,
advisory-locked so instances do not pile up rebuilds). No pg_cron dependency.

## Testing

- **465 backend tests pass**, including the 50 `TokenCostFunctionTest` cases
  covering the restructured pricing view.
- New `RequestLogStatsRollupTest` (7 tests) asks each question **twice** — once
  with the rollup empty so everything comes from `log_entry`, once after a
  refresh — and demands the same numbers. Covers totals, status counts, model
  breakdown, both time series, the team scope reaching both branches, a range
  that does not start on the hour, a sub-hour bucket bypassing the rollup, and a
  double refresh not double-counting.
- Migration verified end-to-end against the production snapshot via Liquibase.
- `ng build` passes (template type-checking included).

**Not verified:** the UI was not opened in a browser — the tab split is a
structural move of existing blocks plus one deletion, checked by the Angular
compiler only. `ng test` fails on this repo for unrelated reasons (`.scss?raw`
imports have no configured loader); it fails identically on an unmodified tree
and CI does not run it.

## Notes

Measurements, plans, and what is still open (the billing nested-loop join,
incremental rollup refresh, `provider_snapshots` retention, per-tab
subscriptions): `logos/docs/page-load-times.md`.

https://claude.ai/code/session_011Qe89ipnyvL8sDupkFCDjC

## Summary by CodeRabbit

- **New Features**
  - Statistics are now organized into Requests and Infrastructure tabs.
  - Request-volume data is aggregated server-side for faster loading.
  - Historical statistics use hourly rollups, while recent data remains live.
  - Statistics rollups refresh automatically each hour.

- **Performance**
  - Statistics, model, and billing queries load significantly faster and transfer less data.

- **Changes**
  - Removed the VRAM remaining chart and VRAM range controls.
  - Removed timeline event and delta updates.
  - Removed queue-depth, cold-start runtime, and average VRAM statistics.

- **Documentation**
  - Added documentation describing page-load performance improvements and remaining optimization opportunities.


logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/operations/service/RequestLogStatsRefreshService.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Five past the hour: the view's newest bucket is the hour that just closed,
+     * and the offset leaves room for requests that were still in flight when it
+     * did.
+     */
+    @Scheduled(cron = "${logos.stats.rollup.refresh-cron:0 5 * * * *}")
+    public void refreshHourlyStats() {
+        if (!refreshEnabled) return;
+        refreshNow();
```
Claudia-Anthropica, line 40: @wasnertobias [high] This is a self-invocation, so Spring does not apply the `@Transactional(REQUIRES_NEW)` advice on `refreshNow()` when the scheduler enters through `refreshHourlyStats()`. The scheduled path therefore reaches the native `executeUpdate()` without an active JPA transaction, while the transaction-scoped advisory lock is released with the preceding statement; the integration tests only call `refreshNow()` through the bean proxy and do not exercise this path. Move the transactional refresh into a separately injected bean (or put the transaction on the scheduled entry point) and cover scheduled invocation.

In `logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/operations/service/RequestLogStatsRefreshService.java`, the scheduled method self-invokes the transactional refresh, bypassing Spring's transaction proxy. Move the transactional refresh into a separate injected Spring bean and add a test that invokes the scheduled entry point.

</details>

logos/logos-webservice/src/main/resources/liquibase/changelog/035_log_entry_hourly_stats.xml
```diff
@@ -0,0 +1,239 @@
...
+            FROM log_entry le
+            LEFT JOIN LATERAL (
+                SELECT SUM(ut.token_count) AS total_tokens
+                FROM usage_tokens ut
+                JOIN token_types tt ON tt.id = ut.type_id
+                WHERE ut.log_entry_id = le.id
+                  AND tt.name = 'total_tokens'
+            ) tok ON TRUE
+            LEFT JOIN log_entry_cost lec ON lec.log_entry_id = le.id
+            WHERE COALESCE(le.timestamp_forwarding, le.timestamp_request, le.timestamp_response)
```
Claudia-Anthropica: @wasnertobias [medium] Ending an hour does not make all of its `log_entry` rows immutable. For example, a request forwarded at 04:59 and still running during the 05:05 refresh is rolled up with a null status and no response duration, tokens, or finalized cost; when it completes afterward, every reader excludes its 04:59 base row below `mv_hi`, so the dashboard remains stale until another refresh. Keep mutable/not-yet-consumed rows in the live branch until a later refresh captures their finalized state, and add a test that finalizes a pre-watermark request after refresh.

In `logos/logos-webservice/src/main/resources/liquibase/changelog/035_log_entry_hourly_stats.xml`, the hour-based cutoff rolls up requests that can still receive lifecycle, usage, and cost updates, after which the live queries exclude them. Change the rollup/live boundary so rows remain live until their finalized state has been captured, and test finalization after a refresh.

</details>

logos/logos-ui/src/app/features/statistics/statistics.ts
```diff
@@ -111,6 +100,38 @@ const TIMELINE_EVENT_CAP = 200_000;
...
+  setActiveTab(tab: StatsTab): void {
+    if (this.activeTab() === tab) return;
+    this.activeTab.set(tab);
+    this.router.navigate([], {
+      relativeTo: this.route,
+      // 'requests' is the default, so it stays out of the URL rather than
+      // decorating every link with a parameter that changes nothing.
+      queryParams: { tab: tab === 'requests' ? null : tab },
+      queryParamsHandling: 'merge',
+      replaceUrl: true,
```
Claudia-Anthropica: @wasnertobias [medium] The promised browser history behavior cannot work here: `replaceUrl: true` removes each tab switch from history, and `ngOnInit` reads only a query-parameter snapshot, so back/forward changes would not update `activeTab` even if entries existed. Push tab changes into history and derive the signal from `queryParamMap` updates.

In `logos/logos-ui/src/app/features/statistics/statistics.ts`, tab navigation replaces history and only reads the query parameter once, so browser back/forward cannot switch tabs. Remove replacement navigation and subscribe the active tab to query-parameter changes.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The rollup direction and equivalence coverage are strong, and the snapshot contains no failed checks or unfinished required checks. However, the scheduled refresh bypasses its transaction, completed-hour rows can still mutate after being rolled up, and tab history is not connected to browser navigation.

wasnertobias:
Recent requests shall not be at the end of Requests tabs but at the beginning

Default tab shall be Infrastructure, rename this tab to "Local Providers" and set it first.

Can you show in the Local Providers tab the RAM utilization just like the VRAM utilization? So per model and as a pie-chart?

Claudia-Anthropica (changes requested):
@wasnertobias The separate refresher bean and scheduled-entry test resolve the transaction finding, and the query-parameter subscription now makes tab history work. The rollup still uses a fixed six-hour cutoff: `LOGOS_TIMEOUT_S` explicitly supports day-long requests, while response and usage persistence can mutate a row after its forwarding hour has entered the view, leaving status, timings, tokens, and cost stale until another successful refresh. Captured CI has no failed or unfinished required checks, although several non-required jobs were still running.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/operations/service/RequestLogStatsRefreshService.java
Claudia-Anthropica: @wasnertobias Thanks, moving the transactional work into `RequestLogStatsRollupRefresher` makes the scheduled invocation cross Spring's proxy, and the scheduled-entry test now covers that path.

logos/logos-webservice/src/main/resources/liquibase/changelog/035_log_entry_hourly_stats.xml
Claudia-Anthropica: @wasnertobias This remains on the current head. The six-hour delay reduces the likelihood but is not a correctness boundary: `LOGOS_TIMEOUT_S` supports requests lasting a day, and the new test mutates only a two-hour-old row that is deliberately outside the materialized view. Please use mutation-aware rollup state or a change watermark and cover finalizing a row that has already crossed the rollup cutoff.

wasnertobias: Fixed in cac8cf3e5 — though deliberately not the way suggested, and the difference matters.

The cutoff now stops **six hours short of now** rather than at the last closed hour, so a request forwarded at 04:59 is still on the live side when the 05:05 refresh lands, and stays there until 11:05 — by which time it has long finished (p99.9 request duration on production is 20 minutes). This was real rather than theoretical: production had 1,050 rows sitting in closed hours with a null status.

What I did **not** do is base the boundary on finalization, because that reintroduces the same class of bug in a harder-to-see form. The rollup would hold what `cost_finalized` said at refresh time while the reader tests what it says *now*, so a row that settles between the two is in neither branch and vanishes from the totals entirely. A boundary that is a function of time alone partitions the range exactly once, whatever any row does afterwards.

A row that still changes after six hours (late settlement — the oldest unsettled row on production is 9.5 days old) is picked up by the next rebuild rather than never: `REFRESH ... CONCURRENTLY` recomputes the whole view, so such an error is transient, not permanent.

Covered by `a_request_still_running_when_the_refresh_lands_is_not_frozen_by_it`, which finalizes a pre-watermark request after a refresh and asserts the aggregates follow it.

logos/logos-webservice/src/main/resources/liquibase/changelog/035_log_entry_hourly_stats.xml
```diff
@@ -0,0 +1,239 @@
...
+                  < date_trunc('hour', now())
+            GROUP BY 1, 2, 3, 4, 5, 6, 7;
```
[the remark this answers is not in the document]
wasnertobias: Fixed in cac8cf3e5 — though deliberately not the way suggested, and the difference matters.

The cutoff now stops **six hours short of now** rather than at the last closed hour, so a request forwarded at 04:59 is still on the live side when the 05:05 refresh lands, and stays there until 11:05 — by which time it has long finished (p99.9 request duration on production is 20 minutes). This was real rather than theoretical: production had 1,050 rows sitting in closed hours with a null status.

What I did **not** do is base the boundary on finalization, because that reintroduces the same class of bug in a harder-to-see form. The rollup would hold what `cost_finalized` said at refresh time while the reader tests what it says *now*, so a row that settles between the two is in neither branch and vanishes from the totals entirely. A boundary that is a function of time alone partitions the range exactly once, whatever any row does afterwards.

A row that still changes after six hours (late settlement — the oldest unsettled row on production is 9.5 days old) is picked up by the next rebuild rather than never: `REFRESH ... CONCURRENTLY` recomputes the whole view, so such an error is transient, not permanent.

Covered by `a_request_still_running_when_the_refresh_lands_is_not_frozen_by_it`, which finalizes a pre-watermark request after a refresh and asserts the aggregates follow it.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/operations/repository/LogEntryRepository.java
```diff
@@ -676,70 +608,90 @@ List<RequestLogProjection> findRequestLogsByUser(
...
+                SELECT SUM(ut.token_count) AS total_tokens
+                FROM usage_tokens ut
+                JOIN token_types tt ON tt.id = ut.type_id
+                WHERE ut.log_entry_id = le.id AND tt.name = 'total_tokens'
+            ) tok ON TRUE
+            LEFT JOIN log_entry_cost lec ON lec.log_entry_id = le.id
+            CROSS JOIN w
+            WHERE COALESCE(le.timestamp_forwarding, le.timestamp_request, le.timestamp_response) BETWEEN :start AND :end
+              AND (COALESCE(le.timestamp_forwarding, le.timestamp_request, le.timestamp_response) <  w.mv_lo
+                OR COALESCE(le.timestamp_forwarding, le.timestamp_request, le.timestamp_response) >= w.mv_hi)
```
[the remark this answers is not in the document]
wasnertobias: Fixed in cac8cf3e5 — though deliberately not the way suggested, and the difference matters.

The cutoff now stops **six hours short of now** rather than at the last closed hour, so a request forwarded at 04:59 is still on the live side when the 05:05 refresh lands, and stays there until 11:05 — by which time it has long finished (p99.9 request duration on production is 20 minutes). This was real rather than theoretical: production had 1,050 rows sitting in closed hours with a null status.

What I did **not** do is base the boundary on finalization, because that reintroduces the same class of bug in a harder-to-see form. The rollup would hold what `cost_finalized` said at refresh time while the reader tests what it says *now*, so a row that settles between the two is in neither branch and vanishes from the totals entirely. A boundary that is a function of time alone partitions the range exactly once, whatever any row does afterwards.

A row that still changes after six hours (late settlement — the oldest unsettled row on production is 9.5 days old) is picked up by the next rebuild rather than never: `REFRESH ... CONCURRENTLY` recomputes the whole view, so such an error is transient, not permanent.

Covered by `a_request_still_running_when_the_refresh_lands_is_not_frozen_by_it`, which finalizes a pre-watermark request after a refresh and asserts the aggregates follow it.

logos/logos-webservice/src/main/java/de/tum/cit/aet/logos/logoswebservice/operations/service/RequestLogStatsRefreshService.java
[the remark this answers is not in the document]
wasnertobias: Correct — and it was worse than "may": the scheduled path had no transaction at all, so `pg_try_advisory_xact_lock` was committed and released by the very statement that took it. The lock guarded nothing.

Fixed in cac8cf3e5: the transactional refresh moved into its own bean (`RequestLogStatsRollupRefresher`), reached through injection so the proxy actually applies. `the_scheduled_entry_point_refreshes_through_the_transaction_proxy` now drives `refreshHourlyStats()` rather than the inner method — the path that was untested is exactly why this slipped through.

logos/logos-ui/src/app/features/statistics/statistics.ts
```diff
@@ -111,6 +100,38 @@ const TIMELINE_EVENT_CAP = 200_000;
 export class Statistics implements OnInit, OnDestroy {
   private statsWs = inject(StatsWebsocketService);
   private statisticsService = inject(StatisticsService);
+  private readonly router = inject(Router);
+  private readonly route = inject(ActivatedRoute);
```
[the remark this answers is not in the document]
wasnertobias: Right — the fixture would have thrown `NullInjectorError` during construction, before any test body ran. `provideRouter([])` added to the spec in cac8cf3e5.

Worth recording that CI would not have caught it either way: `ng test` is not part of the pipeline, and it currently fails on this repository for an unrelated reason (`.scss?raw` imports have no configured loader) — identically on an unmodified tree.

logos/logos-ui/src/app/features/statistics/statistics.ts
wasnertobias: Fixed in cac8cf3e5. `replaceUrl` is gone, so each tab switch is its own history entry, and `activeTab` is driven by a `queryParamMap` subscription (unsubscribed in `ngOnDestroy`) instead of a one-shot snapshot.

`setActiveTab` now only navigates and lets that subscription own the signal, so a click, a pasted link and Back all reach the state the same way and cannot disagree.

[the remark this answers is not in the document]
wasnertobias: Fixed in cac8cf3e5. `replaceUrl` is gone, so each tab switch is its own history entry, and `activeTab` is driven by a `queryParamMap` subscription (unsubscribed in `ngOnDestroy`) instead of a one-shot snapshot.

`setActiveTab` now only navigates and lets that subscription own the signal, so a click, a pasted link and Back all reach the state the same way and cannot disagree.

logos/logos-ui/src/app/features/statistics/statistics.ts
```diff
@@ -915,20 +891,21 @@ export class Statistics implements OnInit, OnDestroy {
   // ── Lifecycle ─────────────────────────────────────────────────────────────────
 
   ngOnInit(): void {
+    const tabParam = this.route.snapshot.queryParamMap.get('tab');
+    if (tabParam && this.TABS.some((t) => t.id === tabParam)) {
+      this.activeTab.set(tabParam as StatsTab);
```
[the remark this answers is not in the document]
wasnertobias: Fixed in cac8cf3e5. `replaceUrl` is gone, so each tab switch is its own history entry, and `activeTab` is driven by a `queryParamMap` subscription (unsubscribed in `ngOnDestroy`) instead of a one-shot snapshot.

`setActiveTab` now only navigates and lets that subscription own the signal, so a click, a pasted link and Back all reach the state the same way and cannot disagree.

logos/logos-ui/src/app/features/statistics/statistics.html
```diff
@@ -59,447 +87,431 @@ <h1 class="page-title">Statistics</h1>
...
+        } @else {
+          <app-stats-skeleton variant="kpi" />
+        }
+
+        <!-- Active lanes -->
+        @if (vramReady()) {
+          <app-stats-kpi-card
+            label="Active lanes"
+            [accent]="getLaneStateColor('loaded')"
+            [value]="derivedActiveLanes() + ' active'"
```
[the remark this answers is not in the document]
wasnertobias: Agreed — and it quietly undercut the reason for splitting the tabs at all. Moved in cac8cf3e5: the Active-lanes card is out of the period-scoped totals.

The tabs were reordered in the same commit, with Local Providers first and as the default.

logos/logos-webservice/src/test/java/de/tum/cit/aet/logos/logoswebservice/operations/RequestLogStatsRollupTest.java
```diff
@@ -0,0 +1,222 @@
...
+class RequestLogStatsRollupTest {
+
+    @Autowired RequestLogStatsService statsService;
+    @Autowired RequestLogStatsRefreshService refreshService;
+    @Autowired JdbcTemplate jdbc;
+    @MockitoBean JwtDecoder jwtDecoder;
+
+    /** Wide enough to cover the whole seed, ending now so the range stays live. */
+    private static final String START = Instant.now().minus(24, ChronoUnit.HOURS).toString();
+    private static final String END   = Instant.now().plus(1, ChronoUnit.MINUTES).toString();
```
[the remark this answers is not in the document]
wasnertobias: Fixed in cac8cf3e5. Both bounds are derived per call and anchored to the hour (`start()` / `end()`), so the range always covers the seed's `hour + 1min` and `hour + 2min` rows no matter which minute of the hour the suite starts in.

logos/logos-webservice/src/main/resources/liquibase/changelog/035_log_entry_hourly_stats.xml
wasnertobias: You were right on both counts, and my earlier reply here was written before I had read this follow-up — apologies for resolving the thread over it.

**The test was vacuous.** It seeded at two hours back, which the six-hour cutoff keeps on the *live* side, so the row was never in the rollup and the assertion passed for the wrong reason. Replaced in d3009378d by two tests that say what they mean:

- `a_row_that_changes_after_the_cutoff_is_corrected_by_the_next_refresh` seeds **nine hours** back, asserts the row really is in `log_entry_hourly_stats` before touching it, then changes it and pins both halves of the contract — stale while the rollup is stale, correct after a rebuild.
- `a_request_still_running_at_the_refresh_stays_on_the_live_side` pins the case the cutoff exists for: a row inside the gap is asserted absent from the rollup, and its completion shows with no refresh at all.

**On the timeout.** The day-long figure is what `LOGOS_TIMEOUT_S` would allow if set; it is unset in production, where the effective bound is `LOGOSNODE_INFER_TIMEOUT_SECONDS=600`. So six hours is ~36× the longest a request can still be running. The comment now derives the interval from that variable and says to raise it if the timeout ever goes past an hour, rather than resting on a percentile.

**On mutation-aware state.** I did not adopt it, and I want to be explicit about why rather than leave it looking like an oversight. A materialized view rebuilt whole has no way to retract a row's stale contribution. So:

- flag-based exclusion (`roll up only settled rows`) drops rows that settle between build and read — they are in neither branch;
- a change watermark (`take rows updated since the build live as well`) double-counts rows that were already in the rollup when they changed.

Both are worse than a time boundary, which partitions the range exactly once whatever a row does afterwards. Getting it right means the rollup must be able to *retract*, i.e. incremental maintenance over a real table upserted per changed hour, keyed off a change marker on `log_entry` (a trigger-maintained `updated_at`). That is a different design, not a tweak to this one, and it is recorded under *Still open* in `docs/page-load-times.md`.

**What that leaves.** Not "permanently wrong": `REFRESH` recomputes the whole view, so a late settlement is stale for at most one refresh interval. The residual is a cost figure lagging by up to an hour on the 22 completed-but-unsettled rows production currently has. If that window is unacceptable I will do the incremental table — say the word and I will take it on rather than argue the point.

Claudia-Anthropica (changes requested):
@wasnertobias The revised tests now accurately exercise both sides of the cutoff and confirm the refresher and tab-history fixes. However, `a_row_that_changes_after_the_cutoff_is_corrected_by_the_next_refresh` also proves the dashboard serves stale status—and likewise late token or cost changes—until a successful rebuild; a failed or disabled refresh extends that inconsistency beyond the claimed hour. The rollup still needs mutation-aware incremental maintenance, or mutable aggregates must remain on a live path.

logos/logos-webservice/src/main/resources/liquibase/changelog/035_log_entry_hourly_stats.xml
Claudia-Anthropica: @wasnertobias Thanks for correcting the test; it now confirms rather than resolves the finding. After the update, the dashboard deliberately still reports the old status until `refreshNow()` succeeds, and the same snapshotting affects late tokens and costs, so an unsuccessful or disabled refresh extends the stale result beyond one hour. Please implement the incremental mutation-aware table/change-marker design you outlined, or keep mutable aggregates on a live path.

wasnertobias: Implemented in 32bec8779. The rollup is a table maintained incrementally, not a view rebuilt whole, and `log_entry` carries an `updated_at` stamped by a trigger — a trigger rather than application writes because log_entry is updated from several places in the orchestrator and a marker that can be forgotten in one of them is worse than none.

Each pass recomputes exactly the hours whose rows changed. Measured against a production snapshot (689k rows):

| | |
|---|---|
| `ALTER TABLE ADD COLUMN updated_at` | **4 ms** — no default, so no rewrite of the 13 GB table |
| initial backfill | 3.7 s |
| idle pass | **2.9 ms** |
| pass after a late settlement on a ten-month-old row | **110 ms** (1 hour, 2 rows) |

That is what makes a one-minute cadence affordable, and it is what the six-hour cutoff was standing in for. The cutoff is gone.

Two bounds replace it, and both are yours:

- **The write lag.** A pass holds its cutoff a minute behind `now()`. Without it a transaction that began before the cutoff but commits after it would carry an `updated_at` below a watermark the pass never saw, and no later pass would look at it again. It is a parameter, not a constant, because tests need zero.
- **The freshness guard.** Past fifteen minutes with no completed pass, `logos_stats_rollup_window` collapses and every aggregate reads `log_entry`. This is the direct answer to your point about an unsuccessful or disabled refresh: a rollup nobody maintains stops being trusted, so the failure mode is a slow page, never a wrong one.

Three tests pin the behaviour rather than describing it: a row corrected by the next pass, a row whose `timestamp_forwarding` arrives late and moves it to another hour leaving no count behind in the one it left, and an aged state row sending every aggregate live with identical numbers.

**Measuring this caught a mistake of my own**, and it is the same one this PR fixes in `log_entry_cost`. The live branch took its bounds from a joined CTE, which the planner cannot turn into an index condition, so it materialised all 416,721 rows of the range — token `LATERAL` and cost join included — and filtered down to 694 afterwards. Read as scalar subqueries instead: **1837 ms → 52 ms** for the same answer. I would have shipped that without your push for the incremental design, so thank you for it.

The remaining invariant, stated in the migration: every closed hour below the watermark has been through a pass. It holds because rows only ever arrive stamped `now()` — the orchestrator never back-dates — so a row cannot appear inside an hour already rolled up. Fixtures that do back-date have to age the state row, which the operations seed now does.

wasnertobias:
✅ Approved — the latest rollup and statistics changes address the earlier freshness, scope, and test-coverage concerns; I found no remaining actionable blocker.

PR #1042, opened 2026-09-17
`Iris`: Accept course exclusions in lecture search

Nayer-kotry:
### Motivation

Artemis lets a user hide courses from a global search. It can apply that itself for a normal user, by subtracting the excluded courses from the scope it forwards, because that scope is a concrete list of the user's courses. An administrator carries no such list: the search runs without a course ceiling, so there is nothing to subtract from and the exclusion cannot be expressed on the Artemis side.

Iris therefore has to be able to receive the exclusion and apply it to the query.

### Description

`LectureSearchRequestDTO` gains an optional `excludeCourseIds`. The lecture search and its vector-override variant take the same list as an optional argument and pass it into both collection queries.

The exclusion is resolved as late as possible:

- A caller with a course ceiling has the excluded ids subtracted from the effective scope, so the query stays a plain positive filter. An exclusion that empties the scope skips the search, exactly like an empty access context does today.
- An unrestricted caller has no scope to subtract from, so the ids travel into the query as one not-equal clause per course, on the segment and the transcription collection alike.

Not-equal clauses are used rather than a negated `contains_any` because that form is understood by every Weaviate version the deployments run. The list stays short, since Artemis bounds the number of course ids it forwards.

The field is optional and additive: a request without it produces the same filters, the same queries and the same results as before, which is what the current Artemis release sends.

### Testing

Prerequisites: none beyond the repository test setup.

1. `poetry run pytest tests/test_lecture_global_search_exclusions.py tests/test_lecture_global_search_course_scope.py tests/test_lecture_visibility_retrieval_filters.py`
2. Confirm a request without `excludeCourseIds` parses with `exclude_course_ids` as `None` and queries both collections unfiltered.
3. Confirm an unrestricted caller with `excludeCourseIds` queries both collections with a filter.
4. Confirm an exclusion that empties a bounded scope returns no results without computing an embedding.

The paired Artemis change, which forwards the field for administrators and subtracts it for everyone else, is part of the search palette filters pull request.

## Summary by CodeRabbit

- **New Features**
  - Lecture search now supports excluding specific courses from results.
  - Course exclusions can be provided using the `excludeCourseIds` request field.
  - Searches automatically avoid excluded courses, including unrestricted searches.
  - Searches return no results when all eligible courses are excluded.

- **Tests**
  - Added coverage for course exclusion behavior and backward compatibility when no exclusions are provided.


Claudia-Anthropica (approved):
@Nayer-kotry The exclusion field is correctly propagated through bounded and unrestricted searches, including the vector-override path, while preserving the existing behavior when omitted. The captured Iris tests and lint checks succeeded, and I found no blocking correctness or security issues.

PR #1043, opened 2026-09-17
`Iris`: Fix global search intent classifier on transformers 5.x

Nayer-kotry:
## Summary

Fixes a regression that was already live on `main` (introduced by the transformers `4.57.6` → `5.10.0` security-floor bump): the global search intent classifier fails to load on startup, silently disabling intent filtering.

**Root cause:** the exported ONNX graph declares `token_type_ids` as a required input. Transformers 5's tokenizer rewrite (`AutoTokenizer` now resolves to a new internal `TokenizersBackend` class) stopped including `token_type_ids` in the default encoding — even though the tokenizer's pinned revision is unchanged. Confirmed directly: loading the pinned revision under transformers 5.12.1 gives `model_input_names == ['input_ids', 'attention_mask']`, with no `token_type_ids` key produced at all. The classifier's code only filtered the tokenizer's output down to keys the ONNX model wanted (`{k: v for k, v in enc.items() if k in self._input_names}`) — it never checked the reverse direction (a required model input the tokenizer didn't produce) — so the missing key reached `onnxruntime`'s `InferenceSession.run()`, which raised `ValueError: Required inputs (['token_type_ids']) are missing from input feed`. `_get_classifier()` catches that as a permanent load failure and `classify()` silently falls back to always returning `TRIGGER_AI`, so the app never crashes but intent filtering is effectively disabled — no error surfaces past a startup log line.

**Fix:** supply `token_type_ids` locally (all zeros — correct for this single-sentence classifier, confirmed against what the tokenizer itself produces when explicitly asked via `return_token_type_ids=True`) whenever the ONNX model requires it and the tokenizer didn't provide it, instead of depending on tokenizer output shape staying stable across major library versions.

## Test plan

- [x] New regression test (`tests/test_global_search_intent_classifier.py`) loads the real classifier against the real pinned tokenizer and the real ONNX model and asserts it loads and classifies successfully. Verified the test actually reproduces the bug: reverted the fix locally, confirmed the test fails with the exact same `ValueError`; restored the fix, confirmed it passes.
- [x] The compiled `.onnx` weights are gitignored (dev-machine/CI-local artifact, not checked into the repo), so the test is `skipif`-guarded on the artifact being present — verified it skips cleanly when absent and passes when present, so it won't spuriously fail in an environment without the model file.
- [x] Full Iris suite: 413 passed (412 + 1 new), 0 failed.
- [x] `black --check` and `pylint` (10.00/10) clean on both changed files.
- [x] Confirmed via `grep` that this classifier has zero callers outside the unrelated global-search feature (`global_search_pipeline.py`, `main.py` startup warm-up, `web/routers/pipelines.py`) — no interaction with lecture ingestion.


iris/tests/test_global_search_intent_classifier.py
```diff
@@ -0,0 +1,39 @@
...
+import iris.pipeline.shared.global_search_intent_classifier as intent_classifier
+from iris.domain.search.search_intent_dto import SearchIntent
+
+# The compiled .onnx weights are gitignored (fetched/built locally, not
+# checked into the repo — see models/global_search_intent/.gitignore), so this
+# test only has a real model to exercise wherever that artifact is present.
+_HAS_ONNX_MODEL = any(intent_classifier._model_dir().glob("*.onnx"))
+
+
+@pytest.mark.skipif(
```
Claudia-Anthropica, line 24: @Nayer-kotry [medium] This guard skips the only regression test in Iris CI: `actions/checkout` cannot provide the gitignored ONNX binary, and the test workflow never downloads it. Consequently, the new input-synthesis behavior can regress while CI remains green; keep this optional integration check if useful, but add a non-skipped unit test using a fake tokenizer and session.

In `iris/tests/test_global_search_intent_classifier.py`, the sole regression test is skipped in CI because its ONNX prerequisite is absent. Add a CI-executed unit test using a fake tokenizer and ONNX session that asserts missing `token_type_ids` are synthesized as all zeros with the same shape and dtype as `input_ids`.

</details>

Claudia-Anthropica (changes requested):
@Nayer-kotry The runtime fix correctly supplies an all-zero `token_type_ids` tensor for this single-segment model, and the captured Iris test and lint checks passed. However, the only added regression test is always skipped in Iris CI because the workflow never downloads the gitignored ONNX model, leaving this behavior without automated protection.

PR #1070, opened 2026-09-19
`Logos`: Persist requested model on queued stats rows again

wasnertobias:
## Summary
- After #1035, enqueue/schedule metrics were buffered until completion, so `log_entry.model_id` stayed null while a request queued and the statistics recent-requests feed showed no requested model.
- Write `model_id` / `provider_id` at enqueue and those plus `scheduled_ts` at schedule (live-visible columns only); keep other lifecycle fields buffered for the single completion UPDATE.
- Extend monitoring recorder unit tests to pin the eager live writes and the still-buffered remainder.

## Test plan
- [x] `pytest tests/unit/monitoring/test_recorder.py tests/unit/monitoring/test_recorder_buffering.py tests/unit/monitoring/test_in_flight_gauge.py`
- [ ] On a running stack, open Statistics → Recent requests, submit a request that waits in queue, and confirm the queued row shows the requested model name (not blank) before it starts running.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Live monitoring now displays the requested model and provider as soon as requests are queued or scheduled.
  * Scheduled requests show their queued-to-running transition in the recent requests feed before completion.

* **Bug Fixes**
  * Monitoring data now remains consistent across request completion, draining, and discarded requests.
  * Rate-limit admission status is preserved for both admitted and non-admitted requests.

* **Documentation**
  * Updated monitoring documentation to clarify which lifecycle details appear immediately and which are recorded at completion.


logos/logos-orchestrator/src/logos/monitoring/recorder.py
```diff
@@ -147,6 +147,10 @@ def record_enqueue(
             "timeout_s": timeout_s,
         }
         self._buffer(request_id, **payload)
+        # The statistics recent-requests feed reads model_id off the log row
+        # while the request still queues. Buffering until completion left that
+        # column null for the whole flight, so queued rows showed no model.
+        self._write_live(request_id, model_id=model_id, provider_id=provider_id)
```
Claudia-Anthropica, line 154: @wasnertobias [medium] This UPDATE silently affects zero rows on the warm non-streaming path: `auth_parse_log` returns `_PendingLog` when deployments are cached, so the `log_entry` is not created until terminal persistence. Those requests therefore still cannot expose their requested model while queued, and the mock-based recorder test cannot detect the missing row. Materialize the pending log before the pipeline records enqueue so this write targets an existing row.

In `logos/logos-orchestrator/src/logos/monitoring/recorder.py`, the enqueue-time live UPDATE is a no-op when creation of the matching `log_entry` was deferred. Materialize deferred logs before `record_enqueue` so the model/provider update targets an existing row.

</details>

logos/logos-orchestrator/src/logos/monitoring/recorder.py
```diff
@@ -196,6 +201,15 @@ def record_scheduled(
                 ]:
                     payload[key] = value
         self._buffer(request_id, **payload)
+        # Stage (queued → executing) and the serving model come from these
+        # columns; without an eager write the feed kept showing Queued with
+        # no model until the completion flush.
+        self._write_live(
```
Claudia-Anthropica, line 208: @wasnertobias [medium] `record_scheduled` runs on the async pipeline's event-loop thread, but this call—and the matching enqueue call—opens a synchronous `DBManager`, performs database queries, and commits inline. That adds two blocking pool round trips to every monitored request, reversing the write-behind design used elsewhere specifically to keep synchronous database work off the event loop. Enqueue these live writes on the existing write-behind worker after ordering the log insertion ahead of them.

In `logos/logos-orchestrator/src/logos/monitoring/recorder.py`, both eager lifecycle writes perform synchronous database I/O on the async request event loop. Dispatch the live writes through the write-behind queue while preserving insert-before-update ordering for each request.

</details>

Claudia-Anthropica (changes requested):
@wasnertobias The eager fields work for already-materialized rows, and the captured CI has no failed checks. However, the deferred-log path turns the enqueue update into a no-op, while both eager updates introduce blocking database work on the async request path.

wasnertobias:
Addressed both review notes in 9a50d4625:

1. **Deferred `_PendingLog` path** — before `_pipeline.process`, enqueue an idempotent `ensure_log_usage` INSERT on the write-behind worker so the warm non-streaming path has a row for the live identity UPDATE. Terminal `_materialize_log_id` now goes through the same idempotent helper (lookup-by-`request_id` / insert / unique-race retry).

2. **Blocking DB on the event loop** — `_write_live` now only `write_queue.enqueue`s the UPDATE; the FIFO worker runs the deferred INSERT first, then the enqueue/schedule identity writes.

logos/logos-orchestrator/src/logos/monitoring/recorder.py
wasnertobias: Fixed in 9a50d4625: deferred `_PendingLog` rows are inserted on the write-behind queue (FIFO) before this live identity UPDATE, and `_materialize_log_id` is idempotent via `ensure_log_usage`.

logos/logos-orchestrator/src/logos/monitoring/recorder.py
wasnertobias: Fixed in 9a50d4625: `_write_live` now only enqueues onto the write-behind worker — no synchronous `DBManager` on the async request path.

logos/logos-orchestrator/src/logos/dbutils/dbmanager.py
```diff
@@ -4052,6 +4062,51 @@ def log_usage(
...
+        except sqlalchemy.exc.IntegrityError:
+            # Another writer won the unique request_id race.
+            self.session.rollback()
+            if request_id:
+                return self.get_log_id_by_request_id(request_id)
+            raise
+
     def set_time_at_first_token(self, log_id: int):
         sql = text("""
                    UPDATE log_entry
```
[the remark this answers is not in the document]
wasnertobias: Fixed in 216a265b0: only `idx_log_entry_request_id_unique` races return the existing row; other IntegrityErrors are re-raised after rollback.

